PR #26, opened 2016-01-15
FEATURE: Support Chrome Redux Dev Tools Extension

skurfuerst:
https://github.com/zalmoxisus/redux-devtools-extension

Note: for meaningful debugging, you need to:
- go to chrome://extensions/
- find "Redux DevTools" there and click "Options"
- Enable "Serialization" in these options.

(taken from https://github.com/zalmoxisus/redux-devtools-extension/issues/35)

### DIFF
--- a/Resources/Private/JavaScript/Host/index.js
+++ b/Resources/Private/JavaScript/Host/index.js
@@ -1,3 +1,4 @@
+import compose from 'lodash.compose';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import {Provider} from 'react-redux';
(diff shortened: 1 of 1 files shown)


Resources/Private/JavaScript/Host/index.js
```diff
@@ -44,7 +44,14 @@ document.addEventListener('DOMContentLoaded', () => {
     const translations = JSON.parse(appContainer.querySelector('[data-json="translations"]').innerHTML);
     const nodeTypeSchema = JSON.parse(appContainer.querySelector('[data-json="nodeTypeSchema"]').innerHTML);
     const reducers = reducerFactory(initialState);
-    const store = createStore(reducers);
+
+    const finalCreateStore = compose(
```
grebaldi: Seems like compose is missing at this point. We have included `lodash.compose` in our dependencies already, so it needs to be imported here.

Anyway, great idea to include the dev tools :)

skurfuerst:
Fixed compose here... forgot to commit it

grebaldi:
Works like a charm, thanks!

PR #27, opened 2016-01-15
Loading indicator in tree

skurfuerst:


Resources/Private/JavaScript/Host/Service/TYPO3CR/NodeTreeService.js
```diff
@@ -88,7 +91,11 @@ class NodeTreeService {
                 })
             })
             .then(response => response.json())
-            .then(json => this.store.dispatch(actions.UI.PageTree.setSubTree(storePath, json)));
+            .then(json => {
+                const nodeWhichIsNotLoadingAnymore = $set(updatedRootInStore, 'isLoading', false);
+                this.updateNode(nodeWhichIsNotLoadingAnymore);
+                this.store.dispatch(actions.UI.PageTree.setSubTree(storePath, json));
+            });
```
skurfuerst: for this part, I'd be curious about a code review, specifically:

- is this code idiomatic?
- is the name "nodeWhichIsNotLoadingAnymore" good? (I think not). Better suggestions?

grebaldi: Hey Sebastian,

thanks for this PR :). To your questions:

> is this code idiomatic?

Looks good to me. But I have to admit, that I'm about to refactor the entire server-communication with redux-thunk to get rid of those lengthy Promise chains and have loading state handled more implicitly. I will also refactor the service layer into a purely functional form, since the current architecture invites to introduce side effects.

> is the name "nodeWhichIsNotLoadingAnymore" good? (I think not). Better suggestions?

I agree there, it seems a bit off^^. Since you have two isomorphic operations going on before and after the remote call, I would suggest to describe that with `preLoadingRootState` and `postLoadingRootState` or something in that fashion. In this particular case, I think it's even okay to drop the constants alltogether and just pass the `$set` call to the update method. The resulting statement seems pretty descriptive anyway and since the updated node state objects are not reused anywhere else, it might be okay to omit the assignments.

dimaip: @grebaldi has written what I was about to write, and what I had already suggested to Tyll. Great we're on the same page here!

Inkdpixels:
Looks good, as @grebaldi said, we will refactor the promise based actions later on with redux-thunk :) :+1:

grebaldi:
I merged this for now, since it works with the current implementation and thus will be carried over for the refactoring.

PR #37, opened 2016-01-25
FEATURE: FullScreen mode toggle

dimaip:


Resources/Private/JavaScript/Host/Containers/FooterBar/style.css
```diff
@@ -0,0 +1,4 @@
+.footerBar {}
+.footerBar--hidden {
```
Inkdpixels: Hey @dimaip - Thanks for the work! :)
For consistency reasons, this should be named `.footerBar--isHidden`

Resources/Private/JavaScript/Host/Containers/LeftSideBar/index.js
```diff
@@ -10,17 +10,20 @@ import style from './style.css';
...
-        isHidden: PropTypes.bool.isRequired
+        isHidden: PropTypes.bool.isRequired,
+        isFullScreen: PropTypes.bool.isRequired
     };
 
     render() {
         const classNames = mergeClassNames({
             [style.leftSideBar]: true,
-            [style['leftSideBar--isHidden']]: this.props.isHidden
+            [style['leftSideBar--isHidden']]: this.props.isHidden,
```
Inkdpixels: The `isFullScreen` class could be removed, not only would the UX benefit from this, since the transform is animated, the browser could also improve the performance of the rendering, since mutating the `display`property would cause a reflow, and `transform` only a repaint.

I would just merge both props into one, f.e.
```
[style['leftSideBar--isHidden']]: this.props.isHidden || this.props.isFullScreen
```

grebaldi: Agreed. Quick question: Wouldn't it make sense to even put that into @connect in https://github.com/PackageFactory/PackageFactory.Guevara/pull/37/files#diff-7f5ed69c466f20c5f6463003e7164ed8L12 - or would it be better to let the container handle the fullscreen state separately?

Resources/Private/JavaScript/Host/Containers/FooterBar/index.js
```diff
@@ -4,24 +4,31 @@ import Immutable from 'immutable';
...
+import mergeClassNames from 'classnames';
+import style from './style.css';
 import {immutableOperations} from 'Shared/Utilities/';
 
 const {$get} = immutableOperations;
 
 @connect(state => ({
-    tabs: $get(state, 'ui.tabs')
+    tabs: $get(state, 'ui.tabs'),
+    isFullScreen: $get(state, 'ui.fullScreen.isFullScreen')
```
grebaldi: I would rename this prop to `isHidden`, just for contextual consistency

Resources/Private/JavaScript/Host/Containers/FullScreen/index.js
```diff
@@ -0,0 +1,26 @@
...
+import {IconButton} from 'Host/Components/';
+import style from './style.css';
+import {immutableOperations} from 'Shared/Utilities/';
+
+const {$get} = immutableOperations;
+
+@connect(state => ({
+    isFullScreen: $get(state, 'ui.fullScreen.isFullScreen')
+}))
+export default class ContextBar extends Component {
```
grebaldi: Looks like `ContextBar` got copied accidentally ;) - I would name the component `FullscreenToggler` though... But `Fullscreen` might be also okay... matter of taste I think :)

Resources/Private/JavaScript/Host/Containers/TopBar/index.js
```diff
@@ -7,12 +7,25 @@ import MenuToggler from './MenuToggler/';
...
 import EditModePanelToggler from './EditModePanelToggler/';
 import style from './style.css';
+import mergeClassNames from 'classnames';
+import {immutableOperations} from 'Shared/Utilities/';
 
-@connect()
+const {$get} = immutableOperations;
+
+@connect(state => ({
+    isFullScreen: $get(state, 'ui.fullScreen.isFullScreen')
```
grebaldi: Same here, I would rename the prop to `isHidden`

Resources/Private/JavaScript/Host/Containers/LeftSideBar/index.js
Inkdpixels: No you are right, that one should be put into the decorator itself to clean up the component! :+1:

dimaip: That's exactly how I had that originally implemented: in the connect. But it didn't work with the RightSideBar, as when it is hidden it still has that chevron toggle shown. Thanks for the hint about repaint, didn't think about it.
So do I just add some extra CSS to hide the toggle button for the rightSideBar or what?

dimaip: Also, if we leave the sidebars animated, should I also add animation to top and bottom bars? I wasn't sure if the animation is necessary for going fullscreen at all, but it's not hard to add it.

Resources/Private/JavaScript/Host/Containers/FullScreen/index.js
dimaip: I wanted to call it just `FullScreen` because in the future there may be more stuff than just a
 toggle button.

skurfuerst:
Hey,

a short question: Feeling-wise, I'd rather do more logic inside the Reducers, than instead of inside the @connect...

So currently, we have:
1 Action (toggle Full Screen) changes 1 part of the State, which in turn is read by *many* components.

I'd rather propose: "1 Action (toggle Full Screen) changes many parts of the state (collapseLeftSidebar, collapseRightSidebar, collapseTopBar, ..."   -> and every component only reads its local state.

What do you think?
All the best,
Sebastian

dimaip:
My reply from slack:
In my experience with Redux, it's best to keep writes (commands/actions whatever you call them) more simple than reads (query/select), for this reason I didn't want to disperse the state of being fullScreen throughout multiple reducers.
At least that's the perception I got from Redux best practices of trying to keep the state normalized as much as possible and doing all hard work with reselect selectors in connect.
Complex selecors should be extracted into separate functions and tested.

And this is just how we did it in our chat app: https://github.com/shri-2015-org/shrimp/tree/master/app/selectors

skurfuerst:
OK, curious what @Inkdpixels  and @grebaldi  say :) It's good to have this discussion.

Inkdpixels:
I agree with @dimaip - I think such a tiny but of logic inside a connect decorator wouldn't cause much harm.

dimaip:
I tried to make additional point: if the query logic gets complex it's not a problem too, just take it out into a selector function, possibly with reselect. The collection of such selectors becomes something like Q from CQRS.

grebaldi:
I agree with @dimaip. Imho action creators and reducers shouldn't actually be aware of components interpreting the state. `@connect` is the correct point, where the application state should be translated to whatever the component is interested in. This seems more semantically correct and leads to much more flexibility, since components then display not only a view on the state but on everything that can be concluded from it.

I also like the idea of having reselect in there - it's a pretty powerful concept. And since those selectors memoize things and recalculate their results based on referential equality, there is actually little to no risk of having to much computation going on in `@connect`.

dimaip:
So back to the change itself, any impediments to merge it?

Inkdpixels:
I would advise 2 changes still: 
* Rename the `FullScreen ` component as proposed from @grebaldi to `FullScreenToggler``
* Re-Evaluate each `isFullscreen`prop that you've added to the components, and see if you can also simplify it via the or operator `false || true` in the `@connect` decorator.

Otherwise I would say good to go!

dimaip:
@Inkdpixels 
1. I left a comment that I didn't want to rename it to `FullScreenToggler` because I wanted it to be a container for the whole layer of FullScreen UIs, e.g. we could add something else there in the future.

2. I've left `ifFullScreen` only in RightSideBar (needed to hide the toggle icon) and in ContentView(needed to adjust padding, but can't call it `isHidden` there because it doesn't acutally hide that component, just adjusts its padding to fullscreen mode).

Inkdpixels:
Ahh, sorry, missed that one out. Will merge it. Thanks for the work @dimaip! :) :+1:

dimaip:
Great, thanks for review!

skurfuerst:
Thanks for all your explanations :) Makes perfect sense; :+1: for using Reselect for more complex expressions there!

PR #38, opened 2016-01-31
Add initial functionality to the add node dialog

dimaip:
TODO:

- [x] use real nodetypes data from schema;
- [ ] apply restrictions
- [ ] prompt for node title before
creation

### DIFF
--- a/Classes/PackageFactory/Guevara/Domain/Model/Changes/Create.php
+++ b/Classes/PackageFactory/Guevara/Domain/Model/Changes/Create.php
@@ -7,6 +7,7 @@ use TYPO3\TYPO3CR\Domain\Service\NodeTypeManager;
 class Create extends AbstractCreate
 {
     /**
+     * TODO: find out why everything breaks if I remove this dependency
      * @var NodeTypeManager
      * @Flow\Inject
      */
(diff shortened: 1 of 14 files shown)


Classes/PackageFactory/Guevara/Domain/Model/Changes/Create.php
```diff
@@ -1,9 +1,19 @@
 <?php
 namespace PackageFactory\Guevara\Domain\Model\Changes;
 
+use TYPO3\Flow\Annotations as Flow;
+use TYPO3\TYPO3CR\Domain\Service\NodeTypeManager;
+
 class Create extends AbstractCreate
 {
     /**
+     * TODO: find out why everything breaks if I remove this dependency
```
dimaip: Hey people who are more into Flow, please help me understand why $this->nodeTypeManager in AbstractCreate.php becomes null (i.e. not injected), if I remove it from this file???

skurfuerst: Well because AbstractCreate does not have a nodeTypeManager property which is injected. You can/should move this property up to AbstractCreate...

dimaip: Hey, but why, it IS there, AbstractCreate has property nodeTypeManager...

dimaip: Though @grebaldi wasn't able to verify it, I'll try to debug it once again now

dimaip:
I will create smaller PRs, here's the first one: #73

PR #40, opened 2016-02-08
FEATURE: Allow installation alongside classical Neos UI

skurfuerst:


dimaip:
Hey, great job there! Do I still have to delete the default flowpack Neos login package?
Also, what stops us from adding Composer support, so it would be possible to deploy it to production for further testing?

skurfuerst:
@dimaip +2 for adding composer support; I don't think you need to delete the default Flowpack Neos Login package.

PR #78, opened 2016-03-09
Basic Inspector Implementation

skurfuerst:
- [x] implement basic inspector editor structure
- [x] rebase on immutable.js work
- [x] Load image metadata
- [x] optimize performance (still slow, even after immutable.js. Dig into this!)
- [ ] Work on image editor
  - [x] Display uploaded images in preview
  - [x] Upload new images
  - [x] Crop images
  - [x] Resize images
  - [x] Choose images in media module
  - [x] Loading Indicator
  - [x] Media Module -> display image details
  - [ ] details in image cropping (fixed aspect ratio)
  - [x] Persist uploaded images
  - [ ] Persist image changes
- [x] Persist data changes from inspector
- [ ] remove node "blur" action and ensure always one node is selected. (action-wise)
- [x] localization of buttons etc in image editor
- [x] make editors configurable using public API (NodeTypes etc)
- [x] add textarea editor
- [x] Make sure tests run through locally
- [ ] Make sure tests run through on Travis

### DIFF
--- a/Classes/PackageFactory/Guevara/Aspects/FlowpackFrontendLoginCompatibilityAspect.php
+++ b/Classes/PackageFactory/Guevara/Aspects/FlowpackFrontendLoginCompatibilityAspect.php
@@ -1,0 +1,51 @@
+<?php
+namespace PackageFactory\Guevara\Aspects;
+
+/*                                                                        *
+ * This script belongs to the TYPO3 Flow package "PackageFactory.Guevara".*
+ *                                                                        *
+ *                                                                        */
+
+use TYPO3\Flow\Annotations as Flow;
+use TYPO3\FLOW\AOP\JoinPointInterface;
+use Flowpack\Neos\FrontendLogin\Security\NeosRequestPattern;
+
+/**
+ * @TODO !!! Remove this, when there's a better way to achieve compatibility
+ *
+ * This Aspect helps us to get around compatibility issues with the Flowpack.Neos.FrontendLogin package. The background
+ * is a RequestPattern that is applied via that package.
+ *
+ * This request pattern tries to distinguish between frontend and backend routes, but only matches backend routes that
+ * start with '/neos'.
+ *
+ * With this aspect the behavior is enhanced to consider routes starting with '/che!' as well.
+ *
+ * It is important to find a better way to support Flowpack.Neos.FrontendLogin, since PackageFactory.Guevara should not
+ * have a dependency to it. Therefore this aspect avoids a direct dependency, it is still implicit though.
+ *
+ * @Flow\Scope("singleton")
+ * @Flow\Aspect
+ */
+class FlowpackFrontendLoginCompatibilityAspect
+{
+
+    /**
+     * @Flow\Around("method(Flowpack\Neos\FrontendLogin\Security\NeosRequestPattern->matchRequest())")
+     * @param JoinPointInterface $joinPoint
+     * @return boolean
+     */
+    public function enhanceMatchRequest(JoinPointInterface $joinPoint)
+    {
+        $request = $joinPoint->getMethodArgument('request');
+        $requestPath = $request->getHttpRequest()->getUri()->getPath();
+
+        if ($joinPoint->getProxy()->getPattern() === NeosRequestPattern::PATTERN_BACKEND) {
+            if (strpos($requestPath, '/che!') === 0) {
+                return true;
+            }
+        }
+
+        return $joinPoint->getAdviceChain()->proceed($joinPoint);
+    }
+}
--- a/README.md
+++ b/README.md
@@ -20,14 +20,7 @@
 ## Installation
 *Full composer support coming soon...*
 
-1. Since our package currently conflicts with the default FrontendLogin package, we need to
-   remove it.  
-   **Remove the following line** from your `composer.json` file:  
-   ```
-   "flowpack/neos-frontendlogin": "@dev"
-   ```
-
-2. Add the following lines to your `composer.json` (our package is not published to composer
+1. Add the following lines to your `composer.json` (our package is not published to composer
    repository yet, so we need to give composer hints where to find it):
    ```
     "repositories": [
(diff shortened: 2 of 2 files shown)


Inkdpixels:
As far as I can tell it looks really good so far! :+1: :)
(Though I cannot review the php part)

Resources/Private/JavaScript/Host/Containers/RightSideBar/Inspector/SingleInspectorEditor/index.js
```diff
@@ -0,0 +1,65 @@
...
+    }
+};
+
+@connect($transform({
+    focusedNode: CR.Nodes.focusedSelector,
+    currentInspectorValue: UI.Inspector.currentValue
+}), {
+    writeValue: actions.UI.Inspector.writeValue
+})
+export default class SingleInspectorEditor extends Component {
```
dimaip: I could think about refactoring this one to a reusable component, so it could be used e.g. in the wizard. But I guess it's OK to leave it here for now, not hard to refactor it later.

Resources/Private/JavaScript/Host/Redux/Sagas/UI/Inspector/index.js
```diff
@@ -0,0 +1,29 @@
...
+
+export function* applyInspectorState(getState) {
+    yield* takeEvery(actionTypes.UI.Inspector.APPLY, function* applyAllChanges(action) {
+        const state = getState();
+        const {nodeContextPath} = action.payload;
+
+        const transientNodeInspectorValues = $get([nodeContextPath], getTransientInspectorValues(state));
+
+        for (const key of Object.keys(transientNodeInspectorValues)) {
+            yield put(actions.Changes.add({
```
dimaip: Maybe we can gather all changes into an array here, and add persist them in one go? That would be a huge performance win.

dimaip: Besides, there's no `actions.Changes.add` anymore, I renamed it to `persistChange`. I would add another action, e.g. `persistChanges` or something, to take an array of changes. Our API takes arrays of changes in any case, so it should be very easy.

Resources/Private/JavaScript/Host/Redux/UI/Inspector/index.js
```diff
@@ -0,0 +1,46 @@
...
+
+//
+// Export the reducer
+//
+export const reducer = {
+    [WRITE_VALUE]: ({nodeContextPath, propertyId, value}) => $set(['ui', 'inspector', 'valuesByNodePath', nodeContextPath, propertyId], value),
+    // APPLY is handled by a saga.
+    // APPLY_FINISHED is only triggered by the saga which listens on APPLY.
+    [APPLY_FINISHED]: ({nodeContextPath}) => $drop(['ui', 'inspector', 'valuesByNodePath', nodeContextPath]),
+    // TODO: APPLY_FINISHED needs to UPDATE the Node properties on the client. Or should the "Changes" API do this? I guess yes.
```
dimaip: Yeah, it should work via feedback mechanism, the way we now update workspaces information on publish? I think @grebaldi has some idea for this feedback stuff.

dimaip: Also it would be great to apply updates to the document without reloading the page, like we do currently in Ember UI: as feedback from the server, we could get the re-rendered markup for the updated node. I hate page reloads.

aertmann:
Hey did you guys decide anything regarding custom editors? I remember a discussion about Webpack not being able to handle runtime dependencies. What I wanted to add was that one downside of the existing solution is that all editors are loaded that way, meaning that the built in ones are always loaded and not optimized as part of the build.

skurfuerst:
Making some great progress; Image cropping / Resizing now works; missing is still persistence, aspect ratio support, styling fixes.

dimaip:
@skurfuerst Congrats on the progress! By the way, how do you load custom editors with webpack? As @aertmann said, I didn't find a way to do it a while ago.

grebaldi:
Regarding the inspector persistence:

@dimaip and I had a small conversation about that yesterday and came up with the following idea:

1. Keep the "changes" mechanism that is already used for inline changes
2. Get rid of "Apply" and "Discard" and persist inspector changes on the fly
3. Utilize the feedback stream of the "changes API" to signal a required page reload and then provide the user with a graceful notification and a button to optionally reload the page.

This way, we could get rid of the "Throw away"-dialog that most editors I spoke to find annoying and confusing. Besides that it would be possible for editors to perform lots of changes without numerous page reloads (which are also perceived as being quite distracting...).

So, what do you think @skurfuerst?

PS: Just realized, that we should prevent inline editing on inconsistent guest frame states in this scenario. But since in the current implementation, the content is overlayed with that "Throw away"-dialog trigger, we could do something similar as soon as the system recognizes a required page reload. We could then display a transparent overlay that on click reloads the content view.

skurfuerst:
I don't understand why the travis build fails, but it works locally... any ideas?

skurfuerst:
Hey everybody,

I think the pull request could take a review now :+1: -- not all features are implemented yet (but most are; see above) and the code is more-or-less cleaned up; so it would be really helpful to already gain some feedback!

I would not merge this yet; as I'd like to implement the missing features.

All the best,
Sebastian

grebaldi:
ReactCrop.css is missing in the sources. Doesn't seem to be built along with the other styles. If it's third-party I suggest to put it in Public/Plugins for now.

Also I saw several elements overlapping scrollbars:

1. The close button in the expanded media and/or cropping tool
2. Apply/Revert at the bottom of the inspector

The media package integration works good.

grebaldi:
Meh... I made a mistake during rebase, so all commits were gone in here...

Obviously, github decides to just close this then and does not allow me to reopen it, after I fixed the mess.

So, I'm sorry. And this PR continues here: #138

PR #86, opened 2016-03-14
BUGFIX: reduce log output in Webpack console

skurfuerst:
See https://github.com/webpack/webpack/blob/master/lib/Stats.js#L26-L40
for possible options.


grebaldi:
Awesome! That's a relief :)

PR #89, opened 2016-03-15
Replace exposers with a redux-saga driven mechanism

grebaldi:
This changes the handling of node state observation as exposed via the UI API plugin. We had a mechanism called `exposers` for this, that on every state change would inform their subscribers.

Now this is handled through `redux-saga` resulting in much less function calls and a better error handling.


dimaip:
Cool, looks more consistent with the rest of the app. :+1:

Inkdpixels:
:+1: as well

PR #90, opened 2016-03-15
Make Preview button functional

### LINKED ISSUE #9: Add functionality to the OpenInNewTab Button in the ContextBar

grebaldi:
The current live preview url is now rendered as part of the guest frame document information. I therefore have changed the way this information is composed to fit more with what we do elsewhere.

This seems to significantly slow down the server side rendering, but I need confirmation for that since I don't trust my machine here :) If anybody experiences the same issues, we may have to change the entire TypoScript driven state rendering.

This closes #9


dimaip:
Can't test performance atm, but I always prefer TS solutions to PHP, so code-wise this looks much better.

Resources/Private/JavaScript/Host/Containers/ContextBar/index.js
```diff
@@ -25,27 +27,34 @@ export default class ContextBar extends Component {
...
+            [style['contextBar__buttonLink--isDisabled']]: !previewUrl
+        });
 
         return (
             <div className={classNames}>
                 <DimensionSwitcher />
 
                 <div className={style.contextBar__rightHandedActions}>
-                    <IconButton icon="external-link" onClick={this.onClickOpenInNewTab.bind(this)} />
+                    {previewUrl ?
```
grebaldi: any thoughts on this @Inkdpixels?

Inkdpixels: I would personally just add an anchor without an href attribute if the previewUrl was not passed. Also note https://mathiasbynens.github.io/rel-noopener/

dimaip: perhaps you can change it to smth like `{previewUrl} ? previewUrl : #`?

dimaip: ah @Inkdpixels you bet me to it

grebaldi:
Thanks guys :) I updated the component according to the comments here and will merge now.

PR #94, opened 2016-03-20
TASK: Create Redux store for the guest frame

grebaldi:
### Yet another guest frame refactoring PR

Yeah, it won't be the last one, I'm afraid :) To get this right is crucial to the entire application, so it might take a number of attempts to get a really stable solution. I think, this one's a huge step closer to what we need.

This change introduces a second Redux store for the guest frame.

### Why on earth?

The first idea was not to implement the guest frame with a React/Redux combination, since this kind of contradicts our principle to decouple host and guest as far as possible. As long as we had a veeery thin guest frame in mind, this worked out pretty well. Now the time has come, where complexity has brought our guest frame implementation to its limits and a React/Redux stack is far more efficient to address that.

We need to stick to the decoupling principle, though. The methods of communication between host and guest remain controlled and well defined - it is what will become our public JS API.

So, why a second redux store? Our Host application is not entirely aware of what is going on in the guest frame. There are just certain parts of the guest state that can safely be assumed to be there (like `focusedNode` or `hoveredNode`). And there it is already: the guest frame has its own state - it actually is an application of its own. After having parts of the guest frame UI implemented in React, the next logical step would be to manage the guest frame state with Redux as well. Until realisation, I was afraid that creating a second store would violate the core principles of Redux - but looking at the situation as two separate applications, it actually makes a lot of sense.

The problem now is, how to synchronize those parts of the state that host and guest are sharing. Another problem are those parts of the guest frame UI that are not implemented in React.

### Attempt to draw a communication scheme

guest-host communication scheme

**Observer API** this is a part of the public API, that exposes certain parts of the host store, whenever relevant changes occur

**UI API** A couple of public API methods that in most cases dispatch actions in the host frame

**Node Component** All logic, that is attached to dom elements in the guest frame, that represent CR nodes. This is the only place where we have to address nodes with their `contextPath` *and* their `typoScriptPath`. The component does nothing more than handling the hovered and focused state of nodes. And it certainly should not do anything more than that. This is the unavoidable leak in the implementation alongside with any inline editor, since both of these require direct DOM access and we can make almost no assumptions about the DOM that gets loaded.

**Central store synchronization** With Redux in the guest frame, we now have the opportunity to centralize all state and therefore also to centralize synchronization between host and guest. (All of which is happening in `Guest/Process/Synchonization.js`)

I hope this drawing helps a little^^ 

### TODOs

- [x] Initialize Redux store for the guest frame
- [x] Move `handleActions` from `Host` to `Shared`
- [x] Connect guest frame components with state 
- [x] Remove `ui` and `connection` props from guest frame components
- [x] Harmonize naming of `NodeToolbar` component


dimaip:
Looks great! I think you should just merge things like this, and PRs here are mainly informative, for everyone to stay up to date with what you do there. Don't be worried that we all have to rebase our work all the time, it's very important to get this thing right, so you can experiment as much as needed.

PR #98, opened 2016-03-24
TASK: make initial UI state configurable using YAML

skurfuerst:
As discussed, this makes the UI state configurable using YAML. So far, this does not modify the state generation in any way. This will follow as separate commits.


grebaldi:
The FlowQuery endpoint is broken now, but I guess that's not a surprise :) Besides that this looks good to me and can be merged. :+1:

dimaip:
Hey, from a quick glance it works as I would imagine it.
Btw, I wonder how does it compare performance-wise with TS rendering? Is it a lot quicker?

PR #183, opened 2016-06-27
TASK: rename and move the package under official neos organization

dimaip:
- [x] Transfer the project to Neos Github org
- [x] Rename package and namespaces
- [x] Adjust Travis and other integrations


dimaip:
I have tested the package locally, and the tests are passing, so I'm merging it as is, to get the project out of stale state. Some other little fixes may be necessary, but they may follow in new PRs.

PR #195, opened 2016-07-05
refactor(Icon): stop printing errors/warnings

sryzycki:
Currently per each legacy/incorrect icon name the `Icon` component console logs 3 lines of warnings/errors. Clean console is important for each developer to quickly spot any warnings/errors they have just caused.

I suggest coming back to logging some stuff once we begin integration of `neos-ui`.

### DIFF
--- a/Resources/Private/JavaScript/Components/Icon/index.js
+++ b/Resources/Private/JavaScript/Components/Icon/index.js
@@ -1,11 +1,7 @@
 import React, {PropTypes} from 'react';
 import mergeClassNames from 'classnames';
-import {service} from 'Shared/index';
 import style from './style.css';
 import {fontAwesome} from 'Shared/Utilities/index';
-const {logger} = service;
-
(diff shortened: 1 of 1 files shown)


Inkdpixels:
Hey @sryzycki - Thanks for this PR, while I understand the reasoning, I would not like to remove the code completely, instead we should look for a solution on how to disable this logging-mess just in development :D

F.e. a possible solution would be to print the env variable `FLOW_CONTEXT` into the initial state of the application, to have it accessible within. From there on we can easily disable these loggings for development instances. Wdyt? :)

Inkdpixels:
Hey @sryzycki - Sorry for not working on your fork, but I had no permissions to do so and I just wanted to get this PR of the table. You can find a new PR with the additions from the comment here: #220 

I will close this PR and merge the #220 one if you are ok with my changes? Your commit is of course still left intact and will be merged as the origin of the idea :) :+1:

PR #203, opened 2016-07-09
Various stability bugfixes

skurfuerst:
I have fixed a few bugs leading to an unstable UI after re-installing it. The most severe ones were the one about publishableNodes.count not found, as it was no Immutable object anymore; and I needed to downgrade Redux Saga a little as I could not find the root cause of the error yet.

All the best,
Sebastian


skurfuerst:
Note: merging this one to continue with progress

PR #207, opened 2016-07-11
FEATURE: collapsible AddNodeModal nodetype groups and current mode display

dimaip:
Review on commit-by-commit basis, I have slipped in a few styling refactorings and fixes.

TODO: put state into redux, and persist in in browser local storage.


skurfuerst:
:+1: by reading

PR #215, opened 2016-07-13
FEATURE: inline creation of nodes

dimaip:
TODO: 

- [ ] reflect allowed insert positions in AddNode button, both inline and in tree
- [ ] refactor toolbar into a presentational component and use it from both tree and inline.
- [ ] how do we lazily append a newly created node, without making a full reload of the guest frame?
- [ ] learn to use `neos.ui` api properly (see comment)

### DIFF
--- a/Resources/Private/JavaScript/Guest/Containers/NodeToolbar/Buttons/AddNode/index.js
+++ b/Resources/Private/JavaScript/Guest/Containers/NodeToolbar/Buttons/AddNode/index.js
@@ -4,8 +4,12 @@ import {
     Icon
 } from 'Components/index';
 
+// TODO: hackish! What is the proper way to call `ui` methods?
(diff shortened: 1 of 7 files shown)


Resources/Private/JavaScript/Guest/Containers/NodeToolbar/Buttons/AddNode/index.js
```diff
@@ -4,8 +4,12 @@ import {
     Icon
 } from 'Components/index';
 
+// TODO: hackish! What is the proper way to call `ui` methods?
+const {ui} = window.neos;
```
dimaip: How to do this properly? What are signals and should they be used instead? Or should I fire a guest redux action and call `ui.openAddNodeModal` from Saga? Help, I have not clue!

PR #219, opened 2016-07-15
Radically Simplified guest frame (no React inside anymore, just single redux store)

skurfuerst:


Resources/Private/JavaScript/Components/Frame/index.js
```diff
@@ -0,0 +1,109 @@
...
+    getDefaultProps: function() {
+        return {
+            initialContent: '<!DOCTYPE html><html><head></head><body><div></div></body></html>',
+            contentDidMount: function() {},
+            contentDidUpdate: function() {}
+        };
+    },
+    render: function() {
+        // The iframe isn't ready so we drop children from props here. #12, #17
+        return React.createElement('iframe', assign({}, this.props, {children: undefined}));
```
dimaip: Just wondering, why aren't you using JSX here and use `createElement` by hand?

Resources/Private/JavaScript/Components/Frame/index.js
```diff
@@ -0,0 +1,109 @@
+// 98997e98ea6d5ececc49db0de528151e40e272b7 from https://github.com/ryanseddon/react-frame-component/commit/98997e98ea6d5ececc49db0de528151e40e272b7
```
dimaip: Ah, so you just copy pasted the whole component from somewhere else? Why not include it via npm?

skurfuerst: Hey,

I heavily patched it; we'll eventually can submit PRs upstream - but for now this was the best way to evaluate whether the idea worked :)

All the best,
Sebastian

Resources/Private/JavaScript/Host/Containers/ContentCanvas/InlineUI/MarkActiveNodeAsFocused/index.js
```diff
@@ -0,0 +1,32 @@
...
+import {connect} from 'react-redux';
+import {CR} from 'Host/Selectors/index';
+import {$transform} from 'plow-js';
+import style from './style.css';
+import {node as NodePropType} from 'Shared/PropTypes/CR/Node/index';
+
+@connect($transform({
+    focusedNode: CR.Nodes.focusedSelector
+}))
+export default class MarkActiveNodeAsFocused extends Component {
```
dimaip: Why does it need to be a react component at all, and not just plain JS module, if it's not rendering anything, but rather directly manipulating DOM?

Resources/Private/JavaScript/Host/Containers/ContentCanvas/index.js
```diff
@@ -31,17 +58,82 @@ export default class ContentCanvas extends Component {
...
+                    this.props.setActiveFormatting(activeFormatting);
+                }
+            };
+
+            _iframeWindow.NeosCKEditorApi.initialize(editorConfig);
+
+            //
+            // Initialize inline editors
+            //
+            [].slice.call(iframeDocument.querySelectorAll('.neos-inline-editable')).forEach(dom => {
```
dimaip: Just a sidenote, I think this is a better way to iterate over NodeList:

```
Array.prototype.forEach.call(elements, (el, i) => {});
```

Resources/Private/Templates/Backend/Guest.html
```diff
@@ -1,6 +1,6 @@
 <f:section name="guestFrameApplication">
     <script>window.neos = window.parent.neos;</script>
     <script src="{f:uri.resource(path: 'resource://Neos.Neos.Ui/Public/Plugins/ckeditor/ckeditor.js')}"></script>
-    <script src="{f:uri.resource(path: 'resource://Neos.Neos.Ui/Public/JavaScript/Guest.js')}"></script>
-    <link rel="stylesheet" href="{f:uri.resource(path: 'resource://Neos.Neos.Ui/Public/Styles/Guest.css')}">
+    <script src="{f:uri.resource(path: 'resource://Neos.Neos.Ui/Public/Javascript/Guest.js')}"></script>
+    <link rel="stylesheet" href="{f:uri.resource(path: 'resource://Neos.Neos.Ui/Public/Styles/Host.css')}">
```
dimaip: Maybe call them Host.css, Shared.css and Guest.css? Current naming is confusing, as I would expect `Host.css` to be only loaded for Host, based on its name.

Resources/Private/JavaScript/Host/Containers/ContentCanvas/InlineUI/MarkActiveNodeAsFocused/index.js
skurfuerst: well, we need to listen to changes in the Store -- so if we did not use a React component, we'd need to manually hook into the store similar to what is done in https://github.com/reactjs/react-redux/blob/master/src/components/connect.js#L80 -- thus I thought it was better/easier to wrap it in a component.

Resources/Private/JavaScript/Host/Containers/ContentCanvas/index.js
skurfuerst: fine, so I'll change this ;)

Resources/Private/Templates/Backend/Guest.html
skurfuerst: Hey, I agree but I do not know how we could do that, as the file is currently called "Host.js" and "Host.css". I don't know how we can change the naming for CSS, but not for JS...

skurfuerst:
Hey everybody,

This is *finally* ready for merging I'd say, after I've just rebased it again.

@Inkdpixels do you wanna have a final look before I press the button? :-)

All the best,
Sebastian

Resources/Private/JavaScript/Host/Containers/ContentCanvas/InlineUI/MarkActiveNodeAsFocused/index.js
dimaip: It's perfectly fine to subscribe to store updates in redux, it's the core API. Wrapping some random js code into React component just for the sake of subscribing to Redux is not a very good practice I think.

Resources/Private/Templates/Backend/Guest.html
dimaip: Just rename the CSS file? They don't have to match.

Resources/Private/JavaScript/Host/Containers/ContentCanvas/InlineUI/MarkActiveNodeAsFocused/index.js
dimaip: On the other hand, it's not even necessary to subscribe to changes manually, you can just take the code from MarkActiveNodeAsFocused, but it into plain JS module, and call it from the render method of InlineUi. Makes sense?

skurfuerst: yeah, fine :)

PR #220, opened 2016-07-17
TASK: Refactor deprecation warnings

Inkdpixels:
Originated from #195 but with the addition to keep the code and only throw warnings outside of development. In future we might adjust the condition to throw warnings not in `Production` instead.

cc @sryzycki


Inkdpixels:
FYI @sryzycki I just moved to a more sophisticated approach regarding deprecation warnings with it's own method in the logger and disabled deprecation warnings not only for the icons but in the whole code base while serving the UI in the `FLOW_CONTEXT=Development`. I hope you are fine with me merging this so we can move on fast. :)

PR #320, opened 2016-09-05
Deep CKEditor integration

skurfuerst:
Fully-fledged CKEditor configuration, including:

- all styles supported by Aloha
- extensible 
- tables


hlubek:
I have some issues getting this working with the Neos demo site:

bildschirmfoto 2016-09-10 um 15 00 35

Is this caused by the Aloha configuration or am I doing something wrong? (current master)

hlubek:
Something seems to be wrong with the active formatting actions, too. When clicking around in the text it doesn't always update correctly but lags behind until the correct button is active.

PR #324, opened 2016-09-09
BUGFIX: Switch order of dispatched actions in content canvas

hlubek:
This fixes an issue where switching dimensions when navigating
causes an error because the context path was set before nodes were added
from the document information.

This needs a refactoring (e.g. single action) but fixes the issue for now.


grebaldi:
Looks fine to me so far by reading and testing. But I couldn't reproduce any error on master - so, somebody else should have a look at it.

skurfuerst:
@grebaldi I think the problem exists together with the Dimension selector Christopher is currently working on!

hlubek:
I could reproduce it on master by switching e.g. to French in the language menu on the website (without my changes). Anyway, thx. for merging ;)

PR #325, opened 2016-09-09
FEATURE: Drawer (main) menu implemention

grebaldi:
## Implementation of the main navigation menu 

* Menu items are an additional json structure that lives outside the application state and configuration
  * potentially to be fetched asynchronously later on
* Menu items are rendered via TypoScript
* Menu items can either target the window or the content canvas

TODOS:

- [x] Render content module
- [x] Render backend modules
- [x] Handle multiple targets
  - [x] Window (default)
  - [x] ContentCanvas
- [x]  Improve mouseleave threshold

solves: #6

### DIFF
--- a/Resources/Private/JavaScript/Host/Containers/Drawer/index.js
+++ b/Resources/Private/JavaScript/Host/Containers/Drawer/index.js
@@ -24,20 +24,20 @@ export default class Drawer extends Component {
     static propTypes = {
         isHidden: PropTypes.bool.isRequired,
         hideDrawer: PropTypes.func.isRequired,
-		menuData: PropTypes.objectOf(
-			PropTypes.shape({
-				icon: PropTypes.string,
-				label: PropTypes.string.isRequired,
-				children: PropTypes.arrayOf(
-					PropTypes.shape({
-						icon: PropTypes.string,
-						label: PropTypes.string.isRequired,
-						uri: PropTypes.string.isRequired,
-						isActive: PropTypes.bool.isReqired
-					})
-				)
-			})
-		).isRequired
+        menuData: PropTypes.objectOf(
+            PropTypes.shape({
+                icon: PropTypes.string,
+                label: PropTypes.string.isRequired,
+                children: PropTypes.arrayOf(
+                    PropTypes.shape({
+                        icon: PropTypes.string,
+                        label: PropTypes.string.isRequired,
+                        uri: PropTypes.string.isRequired,
+                        isActive: PropTypes.bool.isReqired
+                    })
+                )
+            })
+        ).isRequired
     };
 
     shouldComponentUpdate(...args) {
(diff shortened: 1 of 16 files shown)


Resources/Private/JavaScript/Host/Containers/Drawer/index.js
```diff
@@ -15,23 +15,83 @@ import style from './style.css';
...
+            this.setState({
+                mouseLeaveTimeout: null
+            });
+        }
     }
 
     render() {
-        const {isHidden, hideDrawer} = this.props;
+        const {isHidden} = this.props;
+        const clearMouseLeaveTimeout = ::this.clearMouseLeaveTimeout;
```
Inkdpixels: Hey Wilhelm!

Doesn't the stage 0 `::` operator create a new function, just as `.bind(this)` does? If that is the case I would be careful with this due to https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md

What is your opnion on that?

Resources/Private/JavaScript/Host/Containers/Drawer/index.js
```diff
@@ -49,51 +110,46 @@ export default class Drawer extends Component {
...
-            }, {
-                icon: 'calendar',
-                title: 'History'
-            }]
-        }];
-
-        return staticMenuData.map((item, index) => this.renderMenuItem(item, index));
+        const {menuData} = this.props;
+
+        return Object.keys(menuData).map(k => menuData[k]).map((item, index) => this.renderMenuItem(item, index));
```
Inkdpixels: While I think that every experienced developer knows that `k` is the iterated `key`, I would still prefer fully written variables names to be honest :smile: 

Nobody knows if we get some less experienced contributors on board to the project, and we should do those little things to make code reading a bit easier I guess, what do you think?

Resources/Private/JavaScript/Host/Containers/Drawer/index.js
```diff
@@ -49,51 +110,46 @@ export default class Drawer extends Component {
...
     }
 
     renderMenuItem(item, key) {
-        const {title, icon, children} = item;
+        const {setContentCanvasSrc, hideDrawer} = this.props;
+        const {label, icon, children, uri, target} = item;
         const onClick = () => {
-            console.log(`change to menu page "${title}"`);
+            switch (target) {
```
Inkdpixels: Same as the `::` operator, functions created in render loops yo! :)

Resources/Private/JavaScript/Host/Containers/Drawer/index.js
grebaldi: You're absolutely right. I have never thought about that issue tbh, but makes a lot of sense. Probably it would be best to move those bindings to the constructor then. wdyt?

Resources/Private/JavaScript/Host/Containers/Drawer/index.js
grebaldi: You're right of course. But I think we should even go one step further and use `Object.values`+polyfill or some other abstraction to avoid this pattern alltogether, since it is pretty counter-intuitive in and for itself.

grebaldi:
hey @Inkdpixels,

thanks for the comments. I'll fix the code tomorrow. Feels really good to work on this again :tada:

PR #326, opened 2016-09-09
FEATURE: basic reference editor

skurfuerst:


skurfuerst:
FYI: I am not sure whether we break React/Redux principles here; that's why I'd like a review ;)

Resources/Private/JavaScript/InspectorEditors/Reference/index.js
```diff
@@ -0,0 +1,65 @@
...
+    siteNodePath: $get('cr.nodes.siteNode')
+}))
+export default class ReferenceEditor extends Component {
+
+    static propTypes = {
+        commit: PropTypes.func.isRequired,
+        value: PropTypes.any.isRequired,
+
+        options: PropTypes.shape({
+            nodeTypes: PropTypes.arrayOf(PropTypes.string)
```
dimaip: What about the starting point option?

Resources/Private/JavaScript/InspectorEditors/Reference/index.js
```diff
@@ -0,0 +1,65 @@
...
+    };
+
+    render() {
+        const {commit} = this.props;
+        const valueSoFar = this.props.value;
+
+        const options = ({value, callback}) => {
+            const searchQuery = {
+                // TODO: dimensons!
+                workspaceName: this.props.activeWorkspace,
```
dimaip: Does the editor need to be aware of all this low-level stuff like dimensions and workspaces? Maybe move it to `searchNodes` API?

Resources/Private/JavaScript/InspectorEditors/Reference/index.js
```diff
@@ -0,0 +1,65 @@
...
+            } else if (!value || value.length < 2) {
+                return;
+            } else {
+                searchQuery.searchTerm = value;
+            }
+
+            searchNodes(searchQuery)
+            .then(result => result.text())
+            .then(result => {
+                const d = document.createElement('div');
```
dimaip: Oh wow, I didn't see it at first that you manipulate the DOM directly here. Why? Why not move it to component state and do it the React way?

dimaip: uff sorry it's late here, of course you are not manipulating the DOM directly, just use it to extract nodes... Still I don't quite figure out how it works and where the state is stored, but I'll try to get it now.

Resources/Private/JavaScript/InspectorEditors/Reference/index.js
```diff
@@ -0,0 +1,65 @@
...
+                d.innerHTML = result;
+                const nodes = d.querySelector('.nodes');
+
+                return Array.prototype.map.call(nodes.querySelectorAll('.node'), node => ({
+                    label: node.querySelector('.node-label').innerText,
+                    value: node.querySelector('.node-identifier').innerText
+                }));
+            })
+            .then(optionsForForm => {
```
dimaip: nitpick: I guess this could as well be written as `optionsForForm => callback(optionsForForm)`

Resources/Private/JavaScript/InspectorEditors/Reference/index.js
```diff
@@ -0,0 +1,65 @@
...
+            } else {
+                searchQuery.searchTerm = value;
+            }
+
+            searchNodes(searchQuery)
+            .then(result => result.text())
+            .then(result => {
+                const d = document.createElement('div');
+                d.innerHTML = result;
+                const nodes = d.querySelector('.nodes');
```
dimaip: This somehow feels so hackish. Is there a way to either make the server return json or at least hide this data extraction behind the API?

grebaldi: I'm afraid this pattern must be applied in a lot of places, due to the design of the current backend APIs.

I'd prefer a cleaner JSON approach as well, but it would be okay to do that at a later point. 

In the meantime, :+1: for moving this part to the API.

grebaldi:
After reading the code, I would like to suggest to make those options part of the editors state and move the API call to fetch them out of the `render` method.  This would be a little closer to what we do elsewhere. 

IMHO, the SelectBox component should not handle async operations at all. How come it does :)?

skurfuerst:
obsolete. merged in a newer version.

PR #330, opened 2016-09-21
BUGFIX: I18n fallback chain

grebaldi:
This PR will include trans-unit IDs to the I18n components fallback chain, so that literal labels (those that are not marked with `i18n` in the NodeType configuration) are no longer skipped.

Also, the no longer used I18n service gets removed with this.


Inkdpixels:
:+1:

PR #331, opened 2016-09-21
BUGFIX: Apply fixed height to toolbar buttons

grebaldi:
This fixes the page tree toolbar breaking, due to varying button sizes within.


Inkdpixels:
:+1:

PR #345, opened 2016-10-06
TASK: refactor root react component out from View saga

dimaip:
Currently it's very confusing, that the root component of our React
rendering is placed in as weird a place as a Saga. This may be difficult
for React developers to grasp, so I refactor it to be in a more expected
place.


dimaip:
I'll also extract the loader into a separate component, once we'll have a single repo again, too lazy to open two PRs for it.

PR #352, opened 2016-10-07
BUGFIX: Implement workaround to make CK editor work for inline elements

grebaldi:


Inkdpixels:
Related to https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8988348/ - One word, wow.

PR #360, opened 2016-10-24
Fix content editing

skurfuerst:


packages/neos-ui/src/Containers/ContentCanvas/index.js
```diff
@@ -229,7 +229,7 @@ export default class ContentCanvas extends Component {
                 return;
             }
 
-            const nodeFormattingRules = calculateEnabledFormattingRulesForNodeType(node.nodeType);
+            const nodeFormattingRules = this.calculateEnabledFormattingRulesForNodeType(node.nodeType);
```
skurfuerst: actually, this is the fix; the rest is code cleanup

PR #362, opened 2016-10-25
FEATURE: Close secondary inspector when applying or cancelling the inspector

skurfuerst:
This is done by introducing a new callback function which can be used to render the secondary
inspector.


grebaldi:
Works :+1:  But linting fails. Can be merged as soon as travis is green.

PR #413, opened 2016-11-23
BUGFIX: Let the watch task work again

mstruebing:
This fixes the "watch" task for all other packages than Neos-UI


grebaldi:
Sorry for the downvote. But `onchange` combined with `npm run build:ui` will trigger the entire webpack build, including all of its bootstrapping, which would be a massive waste of resources.

Let's find a solution that works natively with webpack.

mstruebing:
Now it is checked via RegExp.
The webpack config excludes ALL node_modules except @neos-project.

PR #414, opened 2016-11-23
FEATURE: Reload all opened node children in the PageTree when clickin…

Inkdpixels:
…g on the reload ToolBar button #405

### DIFF
--- a/packages/neos-ui-redux-store/src/CR/Nodes/selectors.js
+++ b/packages/neos-ui-redux-store/src/CR/Nodes/selectors.js
@@ -53,8 +53,8 @@ export const focusedNodePathSelector = createSelector(
         focused,
         getCurrentContentCanvasContextPath
     ],
-    (focused, getCurrentContentCanvasContextPath) => {
-        return focused || getCurrentContentCanvasContextPath;
+    (focused, currentContentCanvasContextPath) => {
+        return focused || currentContentCanvasContextPath;
     }
 );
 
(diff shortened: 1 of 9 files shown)


packages/neos-ui-redux-store/src/CR/Nodes/selectors.js
```diff
@@ -25,13 +36,25 @@ export const nodeByContextPath = state => contextPath => {
     return node;
 };
 
+export const makeGetDocumentNodes = nodeTypesRegistry => createSelector(
+    [
+        nodes
+    ],
+    nodesMap => {
+        const documentSubNodeTypes = nodeTypesRegistry.getSubTypesOf('TYPO3.Neos:Document');
+        const nodes = nodesMap.toJS();
```
grebaldi: meh: we found out, that `toJS` at this point leads to performance issues, since it always results in different object references, while Immutable itself leaves references intact. So we should try to avoid `toJS` it selectors.

packages/neos-ui-redux-store/src/CR/Nodes/selectors.js
```diff
@@ -25,13 +36,25 @@ export const nodeByContextPath = state => contextPath => {
...
+
 export const focusedNodePathSelector = createSelector(
     [
         focused,
-        currentDocumentNode
+        getCurrentContentCanvasContextPath
     ],
-    (focused, currentDocumentNode) => {
-        return focused || currentDocumentNode;
+    (focused, getCurrentContentCanvasContextPath) => {
```
grebaldi: nitpick: just for readability, I would remove the `get` prefix for the parameter.

packages/neos-ui-redux-store/src/CR/Nodes/selectors.js
Inkdpixels: Yep, but so does the following filter operation on the next line, wdyt?

PR #415, opened 2016-11-23
React Performance improvements

skurfuerst:


skurfuerst:
https://github.com/neos/neos-ui/issues/406

PR #416, opened 2016-11-23
FEATURE: Add node inline operation lifecycle

grebaldi:
This pull request will contain the entire lifecycle of adding a node in the guest frame.

TODOS:

- [x] Extend add node mechanism to pass relevant fusion paths to the server
- [x] Render the created node on server side and give feedback
- [x] Insert the rendered node at the right place in the guest frame DOM
- [x] Add new node to the store

### DIFF
--- a/Classes/Neos/Neos/Ui/Domain/Model/Changes/AbstractCreate.php
+++ b/Classes/Neos/Neos/Ui/Domain/Model/Changes/AbstractCreate.php
@@ -10,6 +10,7 @@ use TYPO3\TYPO3CR\Domain\Service\NodeTypeManager;
 use Neos\Neos\Ui\Exception\InvalidNodeCreationHandlerException;
 use Neos\Neos\Ui\Domain\Model\AbstractChange;
 use Neos\Neos\Ui\Domain\Model\ChangeInterface;
+use Neos\Neos\Ui\Domain\Model\Feedback\Operations\RenderNode;
 
 abstract class AbstractCreate extends AbstractChange
 {
@@ -19,6 +20,13 @@ abstract class AbstractCreate extends AbstractChange
      * @var NodeType
      */
     protected $nodeType;
+
+    /**
+     * The referenceData for the rendered node
+     *
+     * @var array
+     */
+    protected $referenceData = [];
 
     /**
      * @var NodeTypeManager
@@ -47,6 +55,13 @@ abstract class AbstractCreate extends AbstractChange
     protected $nodeService;
 
     /**
+     * Get the insertion mode (before|after|into) that is represented by this change
+     *
+     * @return string
+     */
+    abstract public function getMode();
+
+    /**
      * Set the node type
      *
      * @param string|NodeType $nodeType
@@ -72,6 +87,28 @@ abstract class AbstractCreate extends AbstractChange
     public function getNodeType()
     {
         return $this->nodeType;
+    }
+
+
+    /**
+     * Set the referenceData
+     *
+     * @param array $referenceData
+     * @return void
+     */
+    public function setReferenceData(array $referenceData)
+    {
+        $this->referenceData = $referenceData;
+    }
+
+    /**
+     * Get the referenceData
+     *
+     * @return array
+     */
+    public function getReferenceData()
+    {
+        return $this->referenceData;
     }
 
     /**
@@ -160,6 +197,15 @@ abstract class AbstractCreate extends AbstractChange
         $node = $parent->createNode($name, $nodeType);
 
         $this->applyNodeCreationHandlers($node);
+
+        if ($nodeType->isOfType('TYPO3.Neos:Content') && $this->getReferenceData()['subject']['fusionPath']) {
+            $renderNode = new RenderNode();
+            $renderNode->setNode($node);
+            $renderNode->setReferenceData($this->getReferenceData());
+            $renderNode->setMode($this->getMode());
+
+            $this->feedbackCollection->add($renderNode);
+        }
 
         return $node;
     }
--- a/Classes/Neos/Neos/Ui/Domain/Model/Changes/AbstractCreate.php
+++ b/Classes/Neos/Neos/Ui/Domain/Model/Changes/AbstractCreate.php
@@ -3,6 +3,7 @@ namespace Neos\Neos\Ui\Domain\Model\Changes;
 
 use Neos\Neos\Ui\NodeCreationHandler\NodeCreationHandlerInterface;
 use TYPO3\Flow\Annotations as Flow;
+use TYPO3\Eel\FlowQuery\FlowQuery;
 use TYPO3\TYPO3CR\Domain\Model\NodeType;
 use TYPO3\TYPO3CR\Domain\Model\NodeInterface;
 use TYPO3\TYPO3CR\Domain\Service\NodeServiceInterface;
@@ -12,6 +13,7 @@ use Neos\Neos\Ui\Domain\Model\RenderedNodeDomAddress;
 use Neos\Neos\Ui\Domain\Model\AbstractChange;
 use Neos\Neos\Ui\Domain\Model\ChangeInterface;
 use Neos\Neos\Ui\Domain\Model\Feedback\Operations\RenderContentOutOfBand;
+use Neos\Neos\Ui\Domain\Model\Feedback\Operations\ReloadDocument;
 use Neos\Neos\Ui\Domain\Model\Feedback\Operations\UpdateNodeInfo;
 
 abstract class AbstractCreate extends AbstractChange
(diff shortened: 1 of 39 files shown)


Classes/Neos/Neos/Ui/Domain/Model/Feedback/Operations/RenderNode.php
```diff
@@ -0,0 +1,157 @@
...
+        return $this->node;
+    }
+
+    /**
+     * Set the referenceData
+     *
+     * @param string $referenceData
+     * @return void
+     */
+    public function setReferenceData(array $referenceData)
```
skurfuerst: instead of get/setReferenceData, I'd rather like to use "typoScriptPathToBeRendered" or so

Classes/Neos/Neos/Ui/Domain/Model/Changes/AbstractCreate.php
```diff
@@ -20,6 +22,13 @@
     protected $nodeType;
 
     /**
+     * The referenceData for the rendered node
```
skurfuerst: Actually, I'd even more prefer if we could not use an array here, but more specifically set the Fusion path as string explicitely

packages/neos-ui-redux-store/src/CR/Nodes/index.js
```diff
@@ -95,11 +94,10 @@ export const reducer = handleActions({
             siteNode: $get('cr.nodes.siteNode', state) || '',
             focused: new Map({
                 contextPath: '',
-                typoscriptPath: ''
+                fusionPath: ''
```
skurfuerst: add a comment here why fusionPath is needed (e.g. only for out-of-band-rendering when adding new nodes)

packages/neos-ui-redux-store/src/UI/AddNodeModal/index.js
```diff
@@ -23,7 +23,7 @@ export const actionTypes = {
 /**
  * Opens the add node modal.
  */
-const open = createAction(OPEN, (contextPath, mode) => ({contextPath, mode}));
+const open = createAction(OPEN, (reference, mode) => ({reference, mode}));
```
skurfuerst: please add docs what "reference" is. String? Node Context Path? Object?

packages/neos-ui-redux-store/src/UI/AddNodeModal/index.js
```diff
@@ -59,25 +60,32 @@ export const reducer = handleActions({
     [system.INIT]: () => $set(
         'ui.addNodeModal',
         new Map({
-            referenceNode: '',
+            reference: null,
```
skurfuerst: hm, what is "reference"? I find it's hard to understand, because it is unclear what is in here and why we need it.

packages/neos-ui/src/Containers/ContentCanvas/InlineUI/NodeToolbar/Buttons/AddNode/index.js
```diff
@@ -37,9 +39,23 @@ export default class AddNode extends PureComponent {
...
         const {
             openAddNodeModal,
-            node
+            fusionPath,
+            contextPath
         } = this.props;
+        const closestCollectionElement = dom.closestNode(
+            dom.findNode(contextPath, fusionPath)
+        );
+        const collection = {
```
skurfuerst: collection -> "contentCollection". becuase that's what it is.

packages/neos-ui/src/Containers/ContentCanvas/InlineUI/NodeToolbar/Buttons/AddNode/index.js
```diff
@@ -37,9 +39,23 @@ export default class AddNode extends PureComponent {
     handleOpenModalBtnClick() {
         const {
             openAddNodeModal,
-            node
+            fusionPath,
+            contextPath
         } = this.props;
+        const closestCollectionElement = dom.closestNode(
```
skurfuerst: add a TODO here, that we always assume that closestNode is a collection; and that is kinda fragile I think...

packages/neos-ui/src/Containers/ContentCanvas/InlineUI/index.js
```diff
@@ -12,26 +12,29 @@ import style from './style.css';
 
 @connect($transform({
     hoveredNode: selectors.CR.Nodes.hoveredSelector,
-    focusedNode: selectors.CR.Nodes.focusedSelector
+    focused: $get('cr.nodes.focused')
```
skurfuerst: ? why is this change actually needed?

packages/neos-ui/src/Containers/ContentCanvas/InlineUI/index.js
```diff
@@ -12,26 +12,29 @@ import style from './style.css';
...
+            fusionPath: PropTypes.string
+        })
     };
 
     render() {
-        const {focusedNode, hoveredNode} = this.props;
+        const {focused, hoveredNode} = this.props;
 
-        if (!focusedNode && !hoveredNode) {
+        if (!focused.get('contextPath') && !hoveredNode) {
```
skurfuerst: use $get instead of assuming this is an immutable Object

packages/neos-ui/src/Containers/ContentCanvas/InlineUI/index.js
```diff
@@ -12,26 +12,29 @@ import style from './style.css';
...
 @connect($transform({
     hoveredNode: selectors.CR.Nodes.hoveredSelector,
-    focusedNode: selectors.CR.Nodes.focusedSelector
+    focused: $get('cr.nodes.focused')
 }))
 export default class InlineUI extends PureComponent {
     static propTypes = {
         hoveredNode: PropTypes.object,
-        focusedNode: PropTypes.object
+        focused: PropTypes.shape({
```
skurfuerst: probably this fails because the object is an immutable object?

packages/neos-ui/src/Containers/Modals/AddNode/index.js
```diff
@@ -53,6 +54,16 @@ const calculateActiveMode = (currentMode, allowedNodeTypesByMode) => {
 }))
 export default class AddNodeModal extends PureComponent {
     static propTypes = {
+        reference: PropTypes.shape({
+            subject: PropTypes.shape({
+                contextPath: PropTypes.string.isRequired,
+                fusionPath: PropTypes.string
+            }).isRequired,
+            collection: PropTypes.shape({
```
skurfuerst: naming: what is "collection"? "enclosingContentCollection"?

Classes/Neos/Neos/Ui/Domain/Model/Changes/AbstractCreate.php
```diff
@@ -176,6 +213,20 @@ protected function createNode(NodeInterface $parent)
             $node->setProperty($key, $value);
         }
 
+        if ($nodeType->isOfType('TYPO3.Neos:Content') && $this->getReferenceData()['subject']['fusionPath']) {
+            $renderNode = new RenderNode();
```
skurfuerst: ...AddToContentCollectionOutOfBand is maybe even better; as the code in JS makes assumptions this is only working in Content Collections

Classes/Neos/Neos/Ui/Domain/Model/Changes/AbstractCreate.php
```diff
@@ -176,6 +213,20 @@ protected function createNode(NodeInterface $parent)
             $node->setProperty($key, $value);
         }
 
+        if ($nodeType->isOfType('TYPO3.Neos:Content') && $this->getReferenceData()['subject']['fusionPath']) {
```
skurfuerst: actually, I think this condition is not fully correct:

1) additionally, the parent node must *always* be a ContentCollection for having this feedback work.

2) if the parent is NO contentCollection, we should reload the page using the feedback we already have.

Classes/Neos/Neos/Ui/Domain/Model/Feedback/Operations/RenderNode.php
dimaip: yes, but let's use the name fusion please

Classes/Neos/Neos/Ui/Domain/Model/Changes/AbstractCreate.php
grebaldi: This doesn't suffice. For inline node operations, we need to distribute fusion path for both, the reference node and the closest collection node, so the Feedback Handler knows exactly where to put the rendered node.

Classes/Neos/Neos/Ui/Domain/Model/Changes/AbstractCreate.php
grebaldi: Is there actually any scenario, in which we create content nodes outside of content collections?

packages/neos-ui/src/Containers/ContentCanvas/InlineUI/NodeToolbar/Buttons/AddNode/index.js
grebaldi: See my comment above. I think this actually pretty implicit.

packages/neos-ui/src/Containers/ContentCanvas/InlineUI/index.js
grebaldi: `focusedSelector` returns a node, `$get('cr.nodes.focused')` returns a contextPath and a fusionPath, which are the two values needed to safely identify a rendered node in the DOM.

packages/neos-ui/src/Containers/Modals/AddNode/index.js
grebaldi: I'd prefer `closestContentCollection`, but I don't like bloating this, since it can't be anything else in this context. I'd rather change the name of `reference` to make more clear what it is about. Maybe something like `creationContext`.

PR #438, opened 2016-11-24
FEATURE: improve node focus in the tree

mgoldbeck:
Now it is possible to focus a node in the tree by clicking it in the same line.

So there is a active node and a focused one.

neos_cms


mstruebing:
https://github.com/neos/neos-ui/pull/439

This let the build work :)

PR #440, opened 2016-11-25
FEATURE: validation

dimaip:
TODO:

- [ ] add more validators
- [ ] improve styling for tooltips
- [x] block apply in inspector on failed validation. Same in creation dialog


grebaldi:
Looks to me, as if we could split out the remaining todos (except for blocking "apply" maybe). Writing validators could be done in parallel by multiple people.

dimaip:
@grebaldi I will work on this change during the flight, but I would like to fix all remaining issues separately, as I just spend considerable amount of time again rebasing after your recent changes, and it was not so obvious how to resolve those merge conflicts. Could you check especially your part of code (`calculateDomAddressesFromMode` and stuff), if I merged it correctly?

packages/neos-ui-i18n/src/index.js
```diff
@@ -51,8 +51,44 @@ export default class I18n extends Component {
         return [packageKey, sourceName, id];
     }
 
+    substitutePlaceholders(textWithPlaceholders, parameters) {
+        let startOfPlaceholder;
+        while ((startOfPlaceholder = textWithPlaceholders.indexOf('{')) !== -1) {
```
Inkdpixels: Not sure about this `while` loop in the render call, can we solve this in another way maybe? Also it would be nice to explain/document this magic a bit if a newcomer stumbles upon it

packages/neos-ui-validators/src/NotEmpty/index.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import I18n from '@neos-project/neos-ui-i18n';
+
+const NotEmpty = value => {
+    if (value === null || value === '' || value === []) {
+        return <I18n id="content.inspector.validators.notEmptyValidator.isEmpty" fallback="The value can't be empty"/>;
+    }
```
Inkdpixels: I would personally always return at least `null` as a fallback, `undefined`  is dangerous since it can be overwritten

packages/neos-ui/src/Containers/Modals/AddNode/index.js
```diff
@@ -132,183 +127,64 @@ export default class AddNodeModal extends PureComponent {
...
-        if (!this.props.referenceNode) {
+        const {referenceNode, handleClose, nodeTypesRegistry, getAllowedNodeTypesByModeGenerator} = this.props;
+        if (!referenceNode) {
             return null;
         }
 
         if (this.state.step === 1) {
-            return this.renderStep1();
+            return (
+                <Step1
```
Inkdpixels: Is `step1` / `step2` the best name possible here? Maybe something more descriptive would be nice :)

packages/neos-ui-validators/src/StringLength/index.js
```diff
@@ -0,0 +1,28 @@
...
+    const stringLength = value.toString().length;
+    if (stringLength < minimum || stringLength > maximum) {
+        if (minimum > 0 && maximum < 10000) {
+            return <I18n id="content.inspector.validators.stringLength.outOfBounds" params={{minimum, maximum}}/>;
+        } else if (minimum > 0) {
+            return <I18n id="content.inspector.validators.stringLength.smallerThanMinimum" params={{minimum}}/>;
+        }
+        return <I18n id="content.inspector.validators.stringLength.greaterThanMaximum" params={{maximum}}/>;
+    }
+};
```
Inkdpixels: Also make sure to return `null` here as well :)

packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
```diff
@@ -96,14 +104,28 @@ export default class Inspector extends PureComponent {
...
+            Object.keys(transientValues.toJS()).forEach(key => {
+                transientProps[key] = $get([key, 'value'], transientValues);
+            });
+            validationErrors = validate(transientProps, nodeType.properties, validatorRegistry);
+        }
+
         const viewConfiguration = nodeTypesRegistry.getInspectorViewConfigurationFor(focusedNode.nodeType);
-        const isDisabled = transientValues === undefined;
+        const isDiscardDisabled = transientValues === undefined;
+        const isApplyDisabled = transientValues === undefined || validationErrors !== null;
```
Inkdpixels: You could re-use the `isDiscardDisabled` constant here

packages/neos-ui-i18n/src/index.js
dimaip: This is a copy-paste from existing Ember version of this thing. I didn't not want to tamper with its logic, as it works and tested by its age :)

packages/neos-ui-validators/src/NotEmpty/index.js
dimaip: Same as above, but OK, this makes sense fixing.

packages/neos-ui/src/Containers/Modals/AddNode/index.js
dimaip: I thought about it, but since here we have only to steps and we don't plan to expand this logic in the future, I left it the way @skurfuerst coded it. If you think it's really important, please suggest better names.

packages/neos-ui-validators/src/StringLength/index.js
dimaip: +

packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
dimaip: Not sure it's a good idea. That would be D.R.Y. for wrong reasons. The logic for `isDiscardDisabled` may change anytime, there's no semantical dependency between the states of these two buttons.

packages/neos-ui-validators/src/index.js
```diff
@@ -0,0 +1,46 @@
...
+                return checkValidator(elementValue, validatorName, validatorConfiguration);
+            });
+            return validationResults.filter(result => result);
+        }
+    };
+
+    const errors = {};
+    let hasErrors = false;
+    Object.keys(values).forEach(elementName => {
+        if ((elementName in values) && (elementName in elementConfigurations)) {
```
dimaip: Fair point, I've thought about it. In that case we would need to merge transient and node properties, and run validation on it. I'd do it. :+1:

dimaip:
@skurfuerst @Inkdpixels thanks for the review! I'll fix your comments and merge it later today.

dimaip:
@skurfuerst @Inkdpixels guys, I think I fixed your comments. Let's merge this PR if there are no serious blockers left.

PR #445, opened 2016-11-25
BUGFIX: Fix the selection of a new media asset.

Inkdpixels:
Previously to this commit plow-js threw an error stating that it couldnt set the  property on a string, which is the current value for new media nodes. Due to the error the image wasnt being applied correctly to the store, and therefore noe change from the initial placeholder image was possible

### DIFF
--- a/packages/neos-ui-editors/src/Image/index.js
+++ b/packages/neos-ui-editors/src/Image/index.js
@@ -162,19 +162,26 @@ export default class ImageEditor extends Component {
         this.props.renderSecondaryInspector(undefined, undefined);
     }
 
+    getValue() {
+        return this.props.value ? this.props.value : {};
+    }
+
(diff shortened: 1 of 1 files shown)


packages/neos-ui-editors/src/Image/index.js
```diff
@@ -162,19 +162,26 @@ export default class ImageEditor extends Component {
         this.props.renderSecondaryInspector(undefined, undefined);
     }
 
+    getValue() {
+        return this.props.value ? this.props.value : {};
```
Inkdpixels: Question, should I use an immutable Map here instead?

grebaldi: Nope, outside the store Immutable should be avoided to be used explicitly

PR #459, opened 2016-12-14
FEATURE: Delete node functionality in tree and content view

grebaldi:
Backend:
- [x] Change class
- [x] Feedback class

Frontend:
- [x] Extend redux store
- [x] Saga
- [x] FeedbackHandler
- [x] Confirmation dialog
- [x] Implement node removal workflow for page tree
- [x] Implement node removal workflow for content canvas

### DIFF
--- a/Classes/Neos/Neos/Ui/Domain/Model/Feedback/Operations/RemoveNode.php
+++ b/Classes/Neos/Neos/Ui/Domain/Model/Feedback/Operations/RemoveNode.php
@@ -1,0 +1,83 @@
+<?php
+namespace Neos\Neos\Ui\Domain\Model\Feedback\Operations;
+
+use Neos\Neos\Ui\Domain\Model\FeedbackInterface;
+use Neos\ContentRepository\Domain\Model\NodeInterface;
+use Neos\Flow\Mvc\Controller\ControllerContext;
+
+class RemoveNode implements FeedbackInterface
+{
+    /**
+     * @var NodeInterface
+     */
+    protected $node;
+
+    /**
+     * Set the node
+     *
+     * @param NodeInterface $node
+     * @return void
+     */
+    public function setNode(NodeInterface $node)
+    {
+        $this->node = $node;
+    }
+
+    /**
+     * Get the node
+     *
+     * @return NodeInterface
+     */
+    public function getNode()
+    {
+        return $this->node;
+    }
+
+    /**
+     * Get the type identifier
+     *
+     * @return string
+     */
+    public function getType()
+    {
+        return 'Neos.Neos.Ui:RemoveNode';
+    }
+
+    /**
+     * Get the description
+     *
+     * @return string
+     */
+    public function getDescription()
+    {
+        return sprintf('Node "%s" has been removed.', $this->getNode()->getLabel());
+    }
+
+    /**
+     * Checks whether this feedback is similar to another
+     *
+     * @param FeedbackInterface $feedback
+     * @return boolean
+     */
+    public function isSimilarTo(FeedbackInterface $feedback)
+    {
+        if (!$feedback instanceof RemoveNode) {
+            return false;
+        }
+
+        return $this->getNode()->getContextPath() === $feedback->getNode()->getContextPath();
+    }
+
+    /**
+     * Serialize the payload for this feedback
+     *
+     * @return mixed
+     */
+    public function serializePayload(ControllerContext $controllerContext)
+    {
+        return [
+            'contextPath' => $this->getNode()->getContextPath(),
+            'parentContextPath' => $this->getNode()->getParent()->getContextPath()
+        ];
+    }
+}
--- a/packages/neos-ui-redux-store/src/CR/Nodes/index.js
+++ b/packages/neos-ui-redux-store/src/CR/Nodes/index.js
@@ -1,6 +1,6 @@
 import {createAction} from 'redux-actions';
 import Immutable, {Map} from 'immutable';
-import {$all, $set, $get} from 'plow-js';
+import {$all, $set, $drop, $get} from 'plow-js';
 
 import {handleActions} from '@neos-project/utils-redux';
 import {actionTypes as system} from '../../System/index';
@@ -10,6 +10,10 @@ import * as selectors from './selectors';
 const ADD = '@neos/neos-ui/Transient/Nodes/ADD';
 const FOCUS = '@neos/neos-ui/Transient/Nodes/FOCUS';
 const UNFOCUS = '@neos/neos-ui/Transient/Nodes/UNFOCUS';
+const COMMENCE_REMOVAL = '@neos/neos-ui/Transient/Nodes/COMMENCE_REMOVAL';
+const REMOVAL_ABORTED = '@neos/neos-ui/Transient/Nodes/REMOVAL_ABORTED';
+const REMOVAL_CONFIRMED = '@neos/neos-ui/Transient/Nodes/REMOVAL_CONFIRMED';
+const REMOVE = '@neos/neos-ui/Transient/Nodes/REMOVE';
 
 //
 // Export the action types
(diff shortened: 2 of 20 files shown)


packages/neos-ui/src/Sagas/CR/NodeOperations/index.js
```diff
@@ -0,0 +1,33 @@
...
+        const nextAction = Object.keys(waitForNextAction).map(k => waitForNextAction[k])[0];
+
+        if (nextAction.type === actionTypes.CR.Nodes.REMOVAL_ABORTED) {
+            return;
+        }
+
+        if (nextAction.type === actionTypes.CR.Nodes.REMOVAL_CONFIRMED) {
+            const nodeToBeRemovedContextPath = $get('cr.nodes.toBeRemoved', state);
+
+            yield put(actions.Changes.persistChange({
```
dimaip: All the stuff to this point is syncronous. Do we need a saga for that? We handle feedback elsewhere anyways.

packages/neos-ui/src/manifest.js
```diff
@@ -428,4 +429,31 @@ manifest('main', {}, globalRegistry => {
...
+            const parentNodeUri = $get(['cr', 'nodes', 'byContextPath', parentContextPath, 'uri'], state);
+
+            store.dispatch(actions.UI.ContentCanvas.setSrc(parentNodeUri));
+            store.dispatch(actions.UI.ContentCanvas.setContextPath(parentContextPath));
+        } else {
+            dom.findAll(`[data-__neos-node-contextpath="${contextPath}"]`)
+                .forEach(el => el.remove());
+        }
+
+        store.dispatch(actions.CR.Nodes.remove(contextPath));
```
dimaip: No optimistic update? Stuff to think about later how to better handle them.

packages/neos-ui/src/Containers/Modals/DeleteNode/index.js
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        return (
+            <Dialog
+                actions={[this.renderAbort(), this.renderConfirm()]}
+                title={this.renderTitle()}
+                onRequestClose={this.handleAbort}
+                isOpen
+                >
+                <div style={{padding: '1em'}}>
```
dimaip: same

Classes/Neos/Neos/Ui/Domain/Model/Feedback/Operations/RemoveNode.php
```diff
@@ -0,0 +1,83 @@
+<?php
+namespace Neos\Neos\Ui\Domain\Model\Feedback\Operations;
+
+use Neos\Neos\Ui\Domain\Model\FeedbackInterface;
+use Neos\ContentRepository\Domain\Model\NodeInterface;
+use Neos\Flow\Mvc\Controller\ControllerContext;
+
+class RemoveNode implements FeedbackInterface
```
grebaldi: Actually, `DocumentNodeCreated` is currently the exception among the feedback classes. We have stuff there like `UpdateWorkspaceInfo` or `RenderContentOutOfBand` and conceptually, the idea was, that those "feedbacks" are commands rather than events, so actually `DocumentNodeCreated` should be renamed to `AddDocumentNode`.

The idea was, that the backend should indeed have a mechanism, to tell the UI what to do (in a sense). The backend merely tells the UI about side effects, that happened during the operation.

But I get, that this might be confusing and we should discuss, whether this is still a good idea.

packages/neos-ui-redux-store/src/CR/Nodes/index.js
```diff
@@ -41,13 +49,41 @@ const focus = createAction(FOCUS, (contextPath, fusionPath) => ({contextPath, fu
  */
 const unFocus = createAction(UNFOCUS);
 
+/**
+ * Start node removal workflow
```
grebaldi: I get your point, but I disagree here. Imho we should not split this core workflow into multiple reducers.

From the pov of a caller (e.g. a "delete" button), it is much more natural in terms of workflow to dispatch a `CR.Nodes.commenceRemoval` action, than a `UI.DeleteModal.show` action, simply because the caller is interested in a node being removed, not a modal being shown.

packages/neos-ui/src/Containers/LeftSideBar/NodeTreeToolBar/Buttons/DeleteSelectedNode/index.js
```diff
@@ -1,42 +1,57 @@
...
-            className
-        } = this.props;
+        const {className, focusedNodeContextPath, siteNodeContextPath, getNodeByContextPath} = this.props;
+        const node = getNodeByContextPath(focusedNodeContextPath);
+
+        //
+        // Do not allow deletion, when there's no focused node, the focused node is auto created or the focused node
+        // is the site node
+        //
+        const isDisabled = !node || $get('isAutoCreated', node) || siteNodeContextPath === focusedNodeContextPath;
```
grebaldi: I agree. Imho this should be moved to `selectors.CR.Nodes`, but I'd postpone that, since this is not such big of a concern for the alpha.

packages/neos-ui/src/Containers/Modals/DeleteNode/index.js
```diff
@@ -0,0 +1,121 @@
...
+    renderTitle() {
+        const {nodeToBeDeletedContextPath, getNodeByContextPath, nodeTypesRegistry} = this.props;
+        const node = getNodeByContextPath(nodeToBeDeletedContextPath);
+        const nodeType = $get('nodeType', node);
+        const nodeTypeLabel = $get('ui.label', nodeTypesRegistry.get(nodeType)) || 'Neos.Neos:Main:node';
+
+        return (
+            <div>
+                <Icon icon="exclamation-triangle"/>
+                <span style={{marginLeft: '1em'}}>
```
grebaldi: You're absolutely right. I overlooked these :) Will fix

packages/neos-ui/src/Sagas/CR/NodeOperations/index.js
grebaldi: The way sagas are used here, allows to express the ongoing workflow decoupled from what is happening in the UI components. 

I actually think, we should use this pattern more often, to keep business logic out of our components, which are currently quite bloaty.

packages/neos-ui/src/manifest.js
grebaldi: Well, yeah... :) I agree we should do optimistic updates, but we still need a way to revert them in case of error. The DOM update from above is a good example for a side effect, that currently could not be reverted if anything goes wrong.

So for now, we need to stay pessimistic.

Classes/Neos/Neos/Ui/Domain/Model/Feedback/Operations/RemoveNode.php
skurfuerst: Agreeing to Wilhelms thoughts here :) Instead of "AddDocumentNode", I'd rather use more UI-specific namings like "LoadDocumentNodeToStore" or so...

packages/neos-ui/src/Containers/LeftSideBar/NodeTreeToolBar/Buttons/DeleteSelectedNode/index.js
skurfuerst: yes, selectors would be a natural place :)

skurfuerst:
Thanks for your work :) I'm merging the code now 👍

packages/neos-ui-redux-store/src/CR/Nodes/index.js
dimaip: Maybe Sagas could be a thing to orchestrate these smaller events together for certain workflows?

packages/neos-ui/src/Containers/LeftSideBar/NodeTreeToolBar/Buttons/DeleteSelectedNode/index.js
dimaip: :+1:

packages/neos-ui/src/manifest.js
dimaip: Yeah that immersive Dom mangling is nasty but hard to avoid. Something to keep in mind what one of the benefits of staying in declarative immutable paradise is :)

dimaip: s/immersive/imperative can't edit, typing from my phone

dimaip:
Thanks for your comments, important to talk through these things sometimes to stay on the same page :)

PR #460, opened 2016-12-14
PROPOSAL: Refactor the editor toolbar as an overlay

grebaldi:
This PR can be seen as an experiment. I wanted to know, how it feels, if the editor toolbar does not replace the dimensionswitcher, but instead appears below it. So here's, what I've implemented:

- The editor toolbar only appears, when there are matching formatting rules
- It appears below the dimension switcher
- It "overlays" the content, that means, the content area does not shrink
- It stays, even when in fullscreen mode, thus allowing a richer editing experience while viewing the site at full size

This is just a proposal that addresses an issue reported by a number of editors. It seems to be hard to mentally project the state change between dimension switcher and editor toolbar (hence the question: "Where's the dimension switcher gone?").

After having tried this, I think, that this might need some polishing, but actually feels fine. So, I'd like to hear your feedback :)

@Inkdpixels @hlubek @robertlemke @skurfuerst @dfeyer @dimaip @aertmann


dimaip:
I like how it sounds from the description. Would check it out later. I think adding a few screenshots could help to provide quick feedback without checking out and recompiling.

hlubek:
>  It seems to be hard to mentally project the state change between dimension switcher and editor toolbar (hence the question: "Where's the dimension switcher gone?").

Full ack. That's always hard to explain to editors. ("Click on the page - no here - yes now you see it again")

dfeyer:
Yes it's a good move

We need to take care of the tablet usability also, but first thing first ;)

dimaip:
What do we do about this one?

aertmann:
@dimaip As far as I can tell no one objects to the change, so by all means go ahead with it. However consider if it's worth the effort doing before a stable release or not of course as with everything else at this state.

dfeyer:
Hi, let's dont touch this one currently, more research I do on the UX ... more splitted am i between this fixed position editor toolbar ... or some floating contextual bar ... near the mouse cursor, on at least attached to the currently selected node ... but need more research.

grebaldi:
I'm closing this since it's much too outdated.

PR #465, opened 2016-12-18
FEATURE: Hide nodes

grebaldi:
TODOs:

- [x] Implement hide/show workflow for page tree
- [x] Implement hide/show workflow for content canvas

Description:

This comes with a different idea for styling the "hidden" state in the tree. Formerly, we had a small red icon indicating, that the node was hidden.

I replaced that with a strikethrough style on the node label, since I always felt, that the little red icon is a bit misleading.

So please, take this decision as a suggestion. I will gladly restore the known behavior if anything speaks against such a change.

### DIFF
--- a/Classes/Neos/Neos/Ui/Domain/Model/Changes/Property.php
+++ b/Classes/Neos/Neos/Ui/Domain/Model/Changes/Property.php
@@ -28,7 +28,7 @@ class Property extends AbstractChange
     /**
      * The value, the property will be set to
      *
-     * @var mixed
+     * @var string
      */
     protected $value;
 
@@ -56,7 +56,7 @@ class Property extends AbstractChange
     /**
      * Set the value
      *
-     * @param mixed $value
+     * @param string $value
      */
     public function setValue($value)
     {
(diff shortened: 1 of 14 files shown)


Classes/Neos/Neos/Ui/Domain/Model/Changes/Property.php
```diff
@@ -137,7 +137,11 @@ public function apply()
             $value = $this->nodePropertyConversionService->convert(
                 $node->getNodeType(), $this->getPropertyName(), $this->getValue(), $node->getContext());
 
-            $node->setProperty($this->getPropertyName(), $value);
+            if ($this->getPropertyName() === '_hidden') {
```
skurfuerst: this should be generalized to using ObjectAccess::setProperty when the first character starts with "_"

grebaldi: Agreed, will fix

dimaip:
Looked through the change and don't see any other issues. Great job! :+1:

PR #470, opened 2016-12-25
TASK: refactor I18n and render placeholders for editors

dimaip:
This allows to use i18n logic without rendering a react element.


skurfuerst:
Looks great for me :+1:

Issue #418, opened 2016-11-23
Fix the returned contextPath of the DocumentNodeCreated feedback operation


Inkdpixels:
Currently the `DocumentNodeCreated` feedback operation does not return the newly crearted contextType (https://github.com/neos/neos-ui/blob/master/Classes/Neos/Neos/Ui/Domain/Model/AbstractChange.php#L87), but the parent nodes contextType, this is not a major bug, but should be resolved soon

Issue #407, opened 2016-11-23
After creating a new document node it needs to be set as the active one in the ContentCanvas and NodeTree


Inkdpixels:
continuing #395

PR #472, opened 2016-12-27
TASK: load page after creation

dimaip:


Classes/Neos/Neos/Ui/Domain/Model/AbstractChange.php
```diff
@@ -95,13 +95,13 @@ protected function reloadDocument()
...
-     *
+     * @param NodeInterface $subject
      * @return void
      */
-    protected function addDocumentNodeCreatedFeedback()
+    protected function addDocumentNodeCreatedFeedback($subject = null)
     {
+        $node = $subject ?: $this->getSubject();
         $nodeService = new NodeService();
-        $node = $nodeService->getClosestDocument($this->getSubject());
```
dimaip: @grebaldi why was this `getClosestDocument` necessary? The method is called `addDocumentNodeCreatedFeedback`, so I assume it should only create feedback for document nodes, no?

dimaip:
Fixes #418

dimaip:
Fixes #407

Classes/Neos/Neos/Ui/Domain/Model/AbstractChange.php
skurfuerst: correct!

PR #476, opened 2016-12-29
TASK: use ButtonGroup component in InsertModeSelector

dimaip:
I propose to use a button group instead of SelectBox for selecting new element's position.
Why? -- It saves one click, which is quite a lot!

The final styling could be improved with the help of a designer.

What are your thoughts?

New behaviour:

image

### DIFF
--- a/Resources/Private/Translations/en/Main.xlf
+++ b/Resources/Private/Translations/en/Main.xlf
@@ -11,6 +11,9 @@
 			<trans-unit id="insert" xml:space="preserve">
 				<source>Insert</source>
 			</trans-unit>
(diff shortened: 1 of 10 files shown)


dimaip:
Rebase when this is merged: #478

grebaldi:
I like it, but I think it should go with an explanation, like:

```
Insert mode: [Button] [Button] [Button]
```

dimaip:
@grebaldi agreed, will do it

dimaip:
@grebaldi like this?

image

skurfuerst:
I like!

PR #481, opened 2017-01-02
TASK: various AddNode modal improvements

dimaip:
- fix close button
- autofocus on first input
- better styling

image


codev0:
It works.

Issue #13, opened 2016-01-04
Save a subset of the store in session storage


grebaldi:
e.g. `ui.tabs` and `user.settings`

dimaip:
How do we do it? By subscribing to store? Create a store enhancer from it? Then we'd need to debounce writes, right?

PR #482, opened 2017-01-02
TASK: highlight transient inspector fields

dimaip:
image

PR #483, opened 2017-01-03
TASK: persist ui state to local storage

dimaip:
Saves the `ui` part of the state to local storage on every action that
matches the `persistentActionsPattern`.

TODO:

- [x] Init other parts of the state, not only left and right sidebars

Closes: #13

### DIFF
--- a/packages/neos-ui-redux-store/src/UI/EditModePanel/index.js
+++ b/packages/neos-ui-redux-store/src/UI/EditModePanel/index.js
@@ -1,6 +1,6 @@
 import {createAction} from 'redux-actions';
-import {Map} from 'immutable';
-import {$set, $toggle} from 'plow-js';
+import Immutable from 'immutable';
+import {$get, $set, $toggle} from 'plow-js';
 
 import {handleActions} from '@neos-project/utils-redux';
 import {actionTypes as system} from '../../System/index';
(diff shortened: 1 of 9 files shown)


dimaip:
Merge after #481 and #482 or rebase

packages/neos-ui-redux-store/src/UI/EditModePanel/index.js
```diff
@@ -30,11 +30,9 @@ export const actions = {
 // Export the reducer
 //
 export const reducer = handleActions({
-    [system.INIT]: () => $set(
+    [system.INIT]: payload => $set(
         'ui.editModePanel',
-        new Map({
-            isHidden: true
```
dimaip: btw, can we just throw away fallback state from here and rely on default state defined in Settings? https://github.com/neos/neos-ui/blob/master/Configuration/Settings.yaml#L87

skurfuerst: we could; although I quite like that you see the state structure in the reducers.

dimaip: I think we should do either or. It's really hard to trace the source of a particular part of state now, good source for bugs.

packages/neos-ui/src/localStorageMiddleware.js
```diff
@@ -0,0 +1,41 @@
...
+//
+// Local storage middleware.
+//
+// Saves "ui" part of the state to local storage on every action
+// that matches the "persistentActionsPattern"
+//
+const localStorageMiddleware = ({getState}) => {
+    let timer = null;
+    const debounceLocalStorageTimeout = 1000;
+    const persistentActionsPatterns = ['@neos/neos-ui/UI', '@neos/neos-ui/User/Settings'];
```
dimaip: This feels a bit wacky to me now, I think we need a more expressive convention of how we decide what to store in the local storage and what not.

PR #491, opened 2017-02-01
TASK: Only cover own controllers in security provider request patterns

aertmann:


dimaip:
Wow @aertmann really excited to see PRs from you here!

Issue #328, opened 2016-09-21
RFC: Serialization of nodes (pre-loading) when loading documents


hlubek:
# Current implementation (UI rewrite)

To pre-load nodes into the store when loading a document, the nodes are injected into the metadata as `documentNode.nodes` configured in _Settings_ with the following expression:

```
Neos.Ui.NodeInfo.renderDocumentNodeAndChildContent(documentNode, controllerContext)
```

This will collect the document node itself and all non-document descendants recursively.

While this basically works for simple cases where content is directly placed into a content collection that is a child or descendant of the document it is not working for every case and has problems with caching.
## Issue 1: Rendering of arbitrary nodes

In many cases nodes outside of the current document node will be rendered (e.g. a shared footer, a referenced and editable node or a listing of nodes). These are only determined by the actual rendering and cannot be supplied in a single expression.

We should extract all needed nodes from the actual rendering (i.e. every node that is passed into the content element wrapping) instead of pre-determining an approximation.
## Issue 2: Caching...

In regards to caching there are multiple problems of centralizing the node metadata outside of the actual cache segments containing the rendered nodes with wrapping.

When injecting the metadata for all nodes into the cache segment for the document it will happen that flushing (by editing or cache lifetime) invalidates an inner cache segment for a particular content node or content collection. In that case the inner segment and the outer segment containing the metadata need to be updated. Currently the outer segment will be re-rendered in cases where an inner segment is missing, but that's not a strict necessity (because TypoScript rendering should allow to render the inner segment out of band). Another issue is the inefficient invalidation that occurs for _every_ editing under a document node and needs to re-serialize all nodes (if we want to do it correct like stated in issue 1 that would mean a full re-rendering of the document).
# Proposal: Split serialized nodes between cache segments

To solve the issue of a correct list of nodes and to be compliant with caching we should emit serialized nodes for each cache segment while rendering the document. This means we collect rendered nodes through a processor (e.g. content element wrapping - but also others depending on editing UI needs) and emit the serialized nodes whenever a cache segment is written.
## Proposed structure

Each cache segment that has associated metadata (not limited to nodes) should include a _single_ script tag that appends the serialized data to a global variable (naming to be decided, could be a random token).

```
Cache Segment

<div ...>
</div>
<script>
var Neos_UI_Data = window.Neos_UI_Data || {};
Neos_UI_Data['nodes']['{node_identifier_X}'] = { ... serialized node ... };
</script>
```

[...]

PR #492, opened 2017-02-02
FEATURE: Node serialization

hlubek:
This implements a basic implementation of #328

A custom content element wrapping Fusion implementation will be implemented in a follow-up.

A more sophisticated implementation (serialized data per cache segment) needs more work and could be part of another follow-up.


hlubek:
@dimaip Yes, I want to get rid of the aspect, too. That would be the next thing I'd implement.

skurfuerst:
looks awesome, thanks <3

PR #494, opened 2017-02-02
WIP TASK: return immutable data from selectors

dimaip:


hlubek:
@dimaip Do we have unit tests for the selectors that could check the assertion that only immutable values are returned?

dimaip:
@hlubek so far we only used instanceof to check like this: https://github.com/neos/neos-ui/blob/master/packages/neos-ui-redux-store/src/CR/Nodes/index.spec.js#L77
There are libs like this for chai, but we don't use chai for assertions: https://github.com/astorije/chai-immutable

skurfuerst:
@dimaip I've incorporated your changes into https://github.com/neos/neos-ui/pull/524 :) -- so they are part of there.

PR #524, opened 2017-02-22
TASK: Fix performance

skurfuerst:


dimaip:
Looks awesome! Can't do a detailed review right now, let's just merge. Would test drive it later.

PR #528, opened 2017-02-22
FEATURE: styled ck-editor table dialog to look like neos backend

crazydrecksack:
Used the minimalistic skin and did some heavy restyling
tailored to make the table dialog look nice. Others
might still look strange.

TODO: create a react component as soon as possible.


skurfuerst:
looks great :)

PR #532, opened 2017-02-22
BUGFIX: only store very specific state in LocalStorage

skurfuerst:
... we should use a Saga for storing and restoring!

### DIFF
--- a/packages/neos-ui/src/localStorageMiddleware.js
+++ b/packages/neos-ui/src/localStorageMiddleware.js
@@ -9,14 +9,17 @@ import {$get} from 'plow-js';
 const localStorageMiddleware = ({getState}) => {
     let timer = null;
     const debounceLocalStorageTimeout = 1000;
-    const persistentActionsPatterns = ['@neos/neos-ui/UI', '@neos/neos-ui/User/Settings'];
+    const persistentActionsPatterns = [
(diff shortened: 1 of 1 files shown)


packages/neos-ui/src/localStorageMiddleware.js
```diff
@@ -9,24 +9,31 @@ import {$get} from 'plow-js';
 const localStorageMiddleware = ({getState}) => {
     let timer = null;
     const debounceLocalStorageTimeout = 1000;
-    const persistentActionsPatterns = ['@neos/neos-ui/UI', '@neos/neos-ui/User/Settings'];
+    const persistentActionsPatterns = [
+        '@neos/neos-ui/UI/LeftSideBar/TOGGLE',
```
dimaip: there's a lot of other stuff that should go in: collapsing node creation dialog groups, full screen, probably a lot more

skurfuerst: yes; but we need to be very careful what we put in there; we had various bugs due to too much being in there :-) (e.g. refreshing the current page did not work as another page was in local storage for Content Canvas).

We'll change the approach to be saga-based; and then we can start putting more data in there :)

All the best,
Sebastian

PR #534, opened 2017-02-23
FEATURE: update and clean code

samuelhauser:


samuelhauser:
Unfortunately no...

packages/neos-ui-editors/src/Reference/index.js
```diff
@@ -1,65 +1,63 @@
...
 @connect($transform({
-    activeWorkspace: $get('cr.workspaces.personalWorkspace.name'),
-    siteNodePath: $get('cr.nodes.siteNode')
+    contextForNodeLinking: selectors.UI.NodeLinking.contextForNodeLinking
 }))
 @neos(globalRegistry => ({
     i18nRegistry: globalRegistry.get('@neos-project/neos-ui-i18n')
 }))
-export default class ReferenceEditor extends PureComponent {
+export default class ReferenceEditor extends Component {
```
grebaldi: Is it necessary to let the Editor inherit from `Component`? We should definitely stick to `PureComponent` otherwise

PR #537, opened 2017-02-23
TASK: validator tests

mstruebing:


mstruebing:
BTW: I have only 90% branch coverage for the `Count`-Validator.
Had tested every scenario which I can think of and modified the code of the validator multiple times.

If anyone knows why I would be very curious why this is the case.
All other coverages are 100%

grebaldi:
Hey @mstruebing, great PR!

As far as I can see, there's no coverage for the case `length > maximum`, thus the missing 10% in coverage.

otherwise great work!

PR #545, opened 2017-02-24
FEATURE: views for the inspector

samuelhauser:
An approach for views in the inspector. Not finished yet. Suggestions and changes are very welcome


skurfuerst:
I like the change 👍 ❤️

PR #551, opened 2017-02-27
FEATURE: Implement Content Tree

grebaldi:


skurfuerst:
@grebaldi is this already mergeable?

grebaldi:
I'd say so. The entire tree implementation needs more polishing and contains a few glitches here and there, but I'd prefer to handle those in a separate PR.

dimaip:
Wow, really elegant, much better then current Ember implementation with duplication and inheritance...

PR #554, opened 2017-03-05
WIP: fix "insert into" constraints for auto-created child nodes

skurfuerst:
## How to reproduce the bug

* Ensure you have a content constraint for Auto Created Child nodes (e.g. "Main") set up.
* you will see ALL node types are displayed, not just the correct ones in the "add node" dialog.

I am not sure if the solution above is actually correct or has weird side-effects. @grebaldi would be great to have some more docs on the selectors :)

All the best,
Sebastian


grebaldi:
This is definitely more correct than before :)

I admittedly could have done a better job on naming that `elevator` function. It is supposed to be either `identity` or `parentNodeContextPath` as seen at the top of the file.

This way the logic for constraints could be reused for all insert modes.

PR #604, opened 2017-03-28
TASK: Unsupported props for Button components

dfeyer:


dimaip:
@dfeyer can you fix the test and let's merge? Those errors are really annoying...

dimaip:
I fixed this at a higher leve. Basically we should not spread at such components as buttons...

PR #608, opened 2017-03-28
FEATURE: use `baseNodeType` setting to filter the document tree

dimaip:


Classes/Neos/Neos/Ui/FlowQueryOperations/FilteredChildrenOperation.php
```diff
@@ -0,0 +1,78 @@
...
+ * the filter expression specified as optional argument.
+ */
+class FilteredChildrenOperation extends AbstractOperation
+{
+    /**
+     * {@inheritdoc}
+     *
+     * @var string
+     */
+    protected static $shortName = 'filteredChildren';
```
skurfuerst: I'd call this "neos_ui_filteredChildren", to ensure this is not accidentally called by other people

packages/neos-ui/src/index.js
```diff
@@ -139,6 +134,11 @@ function * application() {
     const configuration = yield system.getConfiguration;
 
     //
+    // Bootstrap the saga middleware with initial sagas
+    //
+    allSagas.forEach(saga => sagaMiddleWare.run(saga, {store, globalRegistry, configuration}));
```
skurfuerst: that's I think too far below; as we cannot listen to actions.System.ready() anymore in sagas f.e.

PR #610, opened 2017-03-28
BUGFIX: fix adding of nodes if document node properties are edited on the page

skurfuerst:


gerhard-boden:
Tested on Demo Page, inserting document nodes is now possible again

Issue #614, opened 2017-03-28
ImageEditor: Image not showing in preview pane 


gerhard-boden:
![image](https://cloud.githubusercontent.com/assets/10533739/24415796/f29c9cd6-13e2-11e7-9995-c6f9b5a98616.png)

PR #616, opened 2017-03-28
BUGFIX: Show image preview panel on click

gerhard-boden:
Show image preview when clicking the image in the ImageEditor within the inspector.

Fixes: #614


skurfuerst:
:+1: once travis says +1

PR #627, opened 2017-03-29
BUGFIX: include react-ui-components directly

skurfuerst:
this removes the extra-compile step for react UI components,
streaming the dev experience


dimaip:
Probably that's OK to do it like this for now :+1:

PR #632, opened 2017-03-29
TASK: Refactor Content initialization and move to separate package

grebaldi:

- [x] Create @neos-project/neos-ui-guest-frame package
- [x] Move Content initialization to @neos-project/neos-ui-guest-frame
- [x] Move focused node styling to @neos-project/neos-ui-guest-frame
- [x] Move empty content collection placeholders to @neos-project/neos-ui-guest-frame
- [x] Implement correct behavior for empty content collection placeholders when inserting/removing dynamic content
- [x] Move InlineUI component to @neos-project/neos-ui-guest-frame
- [x] Move CKEditor initialization completely to @neos-project/neos-ui-ckeditor-bindings
- [x] Add indirection for inline editables

Proof of Concepts:

- [x] Can a new inline editor be created?
- [x] Can the initialization of content be manipulated?


skurfuerst:
Hey @grebaldi - I'd not like to merge this today, but after neos con. Fine with you? :)

dimaip:
Had a chance to test this PR on a website with 500 editables. So now it takes ~2s to boot up the UI, compared to 20s normally and 30s in the old UI. Nice job!

Within my website I have a scrollable area (fixed height; overflow: scroll), where most of the editables are. The first 10 editables in it are init'ed correctly, if I scroll further the remaining editables are not init'ed.

Also when I scroll the page fast it feels sluggish, while it's initing the editables in the background.

Maybe it's still somehow possible to init on click? E.g. you click on it, and it sets a timer to do the click on it again in 100ms, if it's not ready yet it resets the timer, and so on, until it's ready?

skurfuerst:
@dimaip I think initializing on click won't work, as AFAIK you cannot "capture" the full event including mouse position etc etc; and then "re-trigger" it so that a newly created ContentEditable would set the cursor position. At least that's what I fear.

@grebaldi would you be up for fixing the merge conflicts, so we can get this one in; now that beta1 is out?

packages/neos-ui-contentrepository/src/registry/NodeTypesRegistry.js
```diff
@@ -163,4 +163,17 @@ export default class NodeTypesRegistry extends SynchronousRegistry {
 
         return viewConfiguration;
     }
+
+    getInlineEditorForProperty(nodeTypeName, propertyName) {
+        const nodeType = this.get(nodeTypeName);
+
+        return $get(['properties', propertyName, 'ui', 'inline', 'editor'], nodeType) || 'ckeditor';
```
skurfuerst: this additional configuration option is quite "ad-hoc"; we need to document these in a central place. Maybe for now "README.md"?

packages/neos-ui-contentrepository/src/registry/NodeTypesRegistry.js
```diff
@@ -163,4 +163,17 @@ export default class NodeTypesRegistry extends SynchronousRegistry {
...
+    }
+
+    getInlineEditorOptionsForProperty(nodeTypeName, propertyName) {
+        const nodeType = this.get(nodeTypeName);
+
+        return $get(['properties', propertyName, 'ui', 'inline', 'editorOptions'], nodeType) ||
```
skurfuerst: this additional configuration option is quite "ad-hoc"; we need to document these in a central place. Maybe for now "README.md"?

packages/neos-ui-guest-frame/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@neos-project/neos-ui-guest-frame",
+  "version": "4.4.8",
+  "description": "Guest frame initialization bindings",
+  "private": true,
+  "main": "./src/manifest.js",
+  "scripts": {
+    "prebuild": "check-dependencies && npm run clean",
+    "test": "npm run lint && nyc --lines 95 --functions 95 --branches 95 npm run ava",
+    "test:watch": "npm run ava -- --watch",
```
skurfuerst: ava -> jest as with the other packages

packages/neos-ui-guest-frame/package.json
```diff
@@ -0,0 +1,49 @@
...
+    ],
+    "include": [
+      "src/**"
+    ]
+  },
+  "devDependencies": {
+    "@neos-project/build-essentials": "^4.4.8"
+  },
+  "dependencies": {
+    "@neos-project/neos-ui-extensibility": "^4.4.8",
```
skurfuerst: extensibility version has to be reduced to 1.0.0-beta1

packages/neos-ui-guest-frame/src/initializeGuestFrame.js
```diff
@@ -0,0 +1,108 @@
...
+                        nodeTypesRegistry,
+                        inlineEditorRegistry,
+                        nodes
+                    })
+                );
+                guestFrameWindow.removeEventListener('scroll', handle);
+            }
+        };
+
+        guestFrameWindow.addEventListener('scroll', handle);
```
skurfuerst: I think we should debounce the "handle", so that it is called a little more lazy than on "every" scroll.

packages/neos-ui-guest-frame/src/initializePropertyDomNode.js
```diff
@@ -0,0 +1,61 @@
...
+
+import {getGuestFrameWindow, closestContextPathInGuestFrame} from './dom';
+
+export default ({store, globalRegistry, nodeTypesRegistry, inlineEditorRegistry, nodes}) => propertyDomNode => {
+    const guestFrameWindow = getGuestFrameWindow();
+    if (!guestFrameWindow['@Neos.Neos.Ui:InitializedInlineEditors']) {
+        guestFrameWindow['@Neos.Neos.Ui:InitializedInlineEditors'] = {};
+    }
+
+    const initializedInlindeEditorApis = guestFrameWindow['@Neos.Neos.Ui:InitializedInlineEditors'];
```
skurfuerst: Inlin*D*eEditor

packages/neos-ui-guest-frame/src/initializePropertyDomNode.js
```diff
@@ -0,0 +1,61 @@
...
+    const propertyName = propertyDomNode.getAttribute('data-__neos-property');
+    const contextPath = closestContextPathInGuestFrame(propertyDomNode);
+    const nodeTypeName = $get([contextPath, 'nodeType'], nodes);
+    const nodeType = nodeTypesRegistry.get(nodeTypeName);
+    const isInlineEditable = $get(['properties', propertyName, 'ui', 'inlineEditable'], nodeType) !== false;
+
+    if (isInlineEditable) {
+        const editorIdentifier = nodeTypesRegistry.getInlineEditorForProperty(nodeTypeName, propertyName);
```
skurfuerst: the method name `getInlineEditorForProperty` doesn't make clear that an "editor identifier" is returned. Suggestion: `getInlineEditorIdentifierForProperty`

packages/neos-ui-guest-frame/src/initializePropertyDomNode.js
```diff
@@ -0,0 +1,61 @@
...
+
+                bootstrap({
+                    setFormattingUnderCursor:
+                        (...args) => store.dispatch(setFormattingUnderCursor(...args)),
+                    setCurrentlyEditedPropertyName:
+                        (...args) => store.dispatch(setCurrentlyEditedPropertyName(...args))
+                });
+
+                initializedInlindeEditorApis[editorIdentifier] = true;
+            } catch (err) {
```
skurfuerst: when does this try/catch fail? Is it just "safe programming" or is it there for a reason? Please add a short comment if you know 😄

packages/neos-ui-guest-frame/src/initializePropertyDomNode.js
```diff
@@ -0,0 +1,61 @@
...
+                contextPath,
+                nodeType,
+                editorOptions,
+                globalRegistry,
+                persistChange: change => store.dispatch(
+                    actions.Changes.persistChange(change)
+                )
+            });
+        } catch (err) {
+            console.error(err);
```
skurfuerst: when does this try/catch fail? Is it just "safe programming" or is it there for a reason? Please add a short comment if you know 😄

packages/neos-ui-guest-frame/src/manifest.js
```diff
@@ -0,0 +1,98 @@
...
+    findAllOccurrencesOfNodeInGuestFrame,
+    createEmptyContentCollectionPlaceholderIfMissing
+} from './dom';
+import InlineUI from './InlineUI';
+
+import style from './style.css';
+
+manifest('@neos-project/neos-ui-guestframe', {}, globalRegistry => {
+    const guestFrameRegistry = new SynchronousRegistry(`
+        # Registry for guest-frame specific functionalities
```
skurfuerst: Please document the structure for makeInitializeGuestFrame and InlineUIComponent; with a short sentence what these do, and what their API is...

packages/neos-ui-redux-store/src/CR/Nodes/index.js
```diff
@@ -192,7 +192,9 @@ export const reducer = handleActions({
             )
         ))
     ),
-    [FOCUS]: ({contextPath, fusionPath}) => $set('cr.nodes.focused', new Map({contextPath, fusionPath})),
+    [FOCUS]: ({contextPath, fusionPath}) => $all(
+        $set('cr.nodes.focused', new Map({contextPath, fusionPath}))
+    ),
```
skurfuerst: why wrap the $set in $all, if it is just 1 statement? I think the old code was more concise; and they are functionally equivalent, aren't they?

packages/neos-ui/src/Containers/ContentCanvas/index.js
```diff
@@ -117,105 +100,10 @@ export default class ContentCanvas extends PureComponent {
...
-        });
-
-        const editorConfig = {
-            formattingRules: formattingRulesRegistry.getAllAsObject(),
-            setFormattingUnderCursor: formattingUnderCursor,
-            setCurrentlyEditedPropertyName
-        };
-
-        iframeWindow.NeosCKEditorApi.initialize(editorConfig);
+        iframeDocument.__isInitialized = true;
```
skurfuerst: is this propery used anywhere? I don't find it :)

packages/neos-ui/src/manifest.js
```diff
@@ -21,6 +15,13 @@ manifest('main', {}, globalRegistry => {
     `));
 
     //
+    // Create inline editor registry
+    //
+    globalRegistry.add('inlineEditors', new SynchronousRegistry(`
+        # Registry for inline editors
```
skurfuerst: documentation here is required; explaining the basic API here and interacting registries

packages/neos-ui-contentrepository/src/registry/NodeTypesRegistry.js
grebaldi: I think it's best to leave a `TODO` there and not document it at all for now, since I believe this deserves further discussion before becoming official.

Documenting it as publicly as in the README might suggest it likely to become public API. But since it cannot be considered stable until there is a certain level of confidence within the team, I'd prefer to avoid confusion of that kind.

However, I think it's justifiable to create an example within https://github.com/neos/neos-ui-extensibility-examples to showcase the general concept for people to give it a try. wdyt?

packages/neos-ui-guest-frame/src/initializePropertyDomNode.js
grebaldi: The entire function is executed in a saga-context. Since we're fiddeling with the guest frame at that point, there might be plenty of exceptions completely unknown to us, that would become invisible or at least hard to read, if we do not display them explicitly like this.

I'll add an inline comment, to make that more transparent :)

packages/neos-ui/src/Containers/ContentCanvas/index.js
grebaldi: it's used about six lines above to prevent the initialization from occurring multiple times ;)

grebaldi:
@dimaip as @skurfuerst said, clicks cannot be captured and re-triggered unfortunately (I actually tried to build something like this a couple of months ago - browsers are getting pretty angry if you try it :D)

What might work, is to initialize it `onmouseenter`, since this reliably triggers before the click and should - in the vast majority of cases - leave enough time for the initialization to take place before any click is registered. In a smaller number of cases, the user would have to click twice (but users might notice over time, that this happens only, if their action is to fast, which might not even be surprising to them).

However, if we rely on mouse actions, this has some considerable implications for accessibility. The current solution isn't perfect either, but is likely a more promising foundation and should work in most contexts.

Another problem would be, that mouse actions require a particular design for inline editors. So far, we only have a rich text editor, that ideally doesn't change the overall look and feel of the layout it is integrated in. But I'd imagine that there's good reason to break this paradigm, once it is generally possible. In that case, initializing on mouseenter or on click would mean, that there's unexpected stuff popping up on the page, while hovering over an element or clicking on an innocent-looking headline.

So, my suggestion would be to stick with the lazy initialization on scroll, while trying to improve the performance of that (I'm sure, there's plenty of room for improvement :D).

packages/neos-ui-contentrepository/src/registry/NodeTypesRegistry.js
skurfuerst: OK, let's do it then this way <3

packages/neos-ui-guest-frame/src/initializePropertyDomNode.js
skurfuerst: Wonderful 👍 Thanks for the explanation!

packages/neos-ui/src/Containers/ContentCanvas/index.js
skurfuerst: makes sense 💃

skurfuerst:
@grebaldi totally makes sense :)

Do you need help merging the change and resolve the conflicts, or will you do this yourself?

All the best,
Sebastian

grebaldi:
@skurfuerst I've updated the PR according to your comments, it's ready for review now :tada: 

@dimaip, @skurfuerst I changed the behavior of the lazy initialization. It doesn't happen on scroll anymore. Instead I utilize `requestIdleCallback` to initialize whenever no keystrokes, click events or scroll events, etc. are happening. Each node gets its own idle callback after its predecessor is done initializing.

Sometimes this is still blocking the UI thread, but most of the time the initialization feels faster than before. We might have to polyfill `requestIdleCallback` for IE, but I first wanted hear your opinions on this approach :) (especially your case @dimaip)

dimaip:
Hey @grebaldi your idea sounds really awesome! I'm on vacation since tomorrow, so I'll have a chance to give it a spin! Needs rebasing though.

skurfuerst:
will rebase/merge now :)

PR #643, opened 2017-03-30
BUGFIX: Apply new images in the inspector without throwing exceptions

gerhard-boden:
Before this change an type conversion exception was thrown when trying to change the image inside the image inspector.


gerhard-boden:
We also removed the `merge` and `canMerge` methods, since there is no use for them anymore. We tested the PR in this combination, this is why we decided to bundle this PR with that commit as well.

PR #656, opened 2017-03-30
BUGFIX: Set removed image to empty string

gerhard-boden:
Before an exception was thrown, now empty string will be used for removed images within the inspector (the empty string is consistent with how removed images are handled in the current UI).


dimaip:
@skurfuerst @grebaldi can you have a look at this, I'm not quite sure why it didn't work before.

PR #657, opened 2017-03-30
BUGFIX: Show placeholder image if no image is selected in ImageEditor

gerhard-boden:


gerhard-boden:
Before:
image

After:
image

Issue #587, opened 2017-03-27
wrong coloring of content dimension menu leading to bad contrast for dropdown


crazydrecksack:
old UI
![screen shot 2017-03-27 at 19 07 08](https://cloud.githubusercontent.com/assets/683773/24368493/e6c99566-1320-11e7-97da-f4133c19323c.png)

vs.

new UI
![screen shot 2017-03-27 at 19 06 55](https://cloud.githubusercontent.com/assets/683773/24368506/f778af0a-1320-11e7-8517-4131fff9db85.png)

PR #660, opened 2017-03-30
TASK: Fix color contrast issue in DimensionSwitcher

dfeyer:
Now the DropDown component accept a new style attributes. The default rendering is not touched. You can use ```darker``` to have the color styling used in the DimensionSwitcher. The ```padded``` attributes is added to add padding around the content area of the DropDown.

The change on the DropDown make it more self contains and allow us to remove most of the styling from the DimensionSwitcher. 

Lerna story is updated so you can test the different rendering.

Fixes: #587


dimaip:
@dfeyer thanks for taking the time to do it the right way, love it.

PR #661, opened 2017-03-30
FEATURE: set property isSearchable for selectbox component

samuelhauser:


dimaip:
What's a motivation for this change? To make isSearchable logic more explicit or any other reasons?

samuelhauser:
@dimaip Yes. I think it's not the right way to decide, if the selectBox is searchable, by saying the onDelete has to be a function. that because for example the multiselectBox has no onDelete function.

dimaip:
Ok, merged it for now, but @skurfuerst eventually wants to rewrite the whole thing

PR #662, opened 2017-03-30
FEATURE: Add references editor

samuelhauser:
Needs to be refactored with multiple select box


samuelhauser:
This version should work now and is ready for review ;)

PR #670, opened 2017-04-05
TASK: Strip html tags from nodename labels

### LINKED ISSUE #669: Strip tags from label titles in the tree
image

mstruebing:
Closes #669

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "semantic-release": "^6.3.2",
     "sinon": "^2.1.0",
     "skin-deep": "^0.16.0",
+    "striptags": "^3.0.1",
     "style-loader": "^0.16.0",
     "stylelint": "^7.5.0",
     "watch": "^1.0.2",
--- a/packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js
+++ b/packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js
@@ -131,7 +131,7 @@ export default class Node extends PureComponent {
                     isHidden={$get('properties._hidden', node)}
                     isHiddenInIndex={$get('properties._hiddenInIndex', node)}
                     hasError={this.hasError()}
-                    label={$get('label', node)}
+                    label={$get('label', node).replace(/<\/?[^>]+(>|$)/g, '')}
                     icon={this.getIcon()}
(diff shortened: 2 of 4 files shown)


packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js
```diff
@@ -131,7 +131,7 @@ export default class Node extends PureComponent {
                     isHidden={$get('properties._hidden', node)}
                     isHiddenInIndex={$get('properties._hiddenInIndex', node)}
                     hasError={this.hasError()}
-                    label={$get('label', node)}
+                    label={$get('label', node).replace(/<\/?[^>]+(>|$)/g, '')}
```
dimaip: Would be nice to put this strip tags logic to some utility class, in case we would want to reuse it again. Or even use some npm lib for this...

mstruebing: Then I would use a npm lib for this, am I right that I have to put this dependency at the  rootlevel AND the package which uses it?

dimaip: I think not, as it's not a dev dep. It's enough to just put it as a dep to something that uses it, in this case it's neos-ui.
@grebaldi might say if I'm wrong.

package.json
```diff
@@ -78,6 +78,7 @@
     "semantic-release": "^6.3.2",
     "sinon": "^2.1.0",
     "skin-deep": "^0.16.0",
+    "striptags": "^3.0.1",
```
dimaip: Why do you add it as a dev dep to the root package.json?

mstruebing: Because I'm getting this error without striptags in the root package.json:

    ERROR in JavaScript/Host.js from UglifyJs
    SyntaxError: Unexpected token: name (context) [./packages/neos-ui/~/striptags/src/striptags.js:17,0]

dimaip: @grebaldi is this normal? Can anyone explain why it happens?

dimaip:
Merging this for now. @mstruebing can you ask Wilhelm why it was necessary to include that dep as a dev dep to the root? Ideally it would be nice to have a short post in our wiki explaining how to work with dependencies, as most of us have no clue :)

PR #672, opened 2017-04-06
BUGFIX: Change the first letter of Production to uppercase

mstruebing:
This made the weird errors when `yarn build` go away.


package.json
```diff
@@ -9,7 +9,7 @@
     "build:ui:watch": "npm run build:ui -- --watch",
     "build:components": "echo 'Not needed anymore, just use build:ui'",
     "build:components:watch": "echo 'Not needed anymore, just use build:ui:watch'",
-    "build": "cross-env NODE_ENV=production npm run build:ui",
+    "build": "cross-env NODE_ENV=Production npm run build:ui",
```
dimaip: eh, but it should be `production` with lower case. What you are doing is basically disabling the production build...

mstruebing: argh okay, thanks!

PR #684, opened 2017-04-10
BUGFIX: increase height of codemirror editor

### LINKED ISSUE #671: HTML editor does not take full height
image

Also can't click the close button in top right corner

mstruebing:
closes #671

### DIFF
--- a/packages/neos-ui-editors/src/CodeMirror/CodeMirrorWrap/codemirror.vanilla-css
+++ b/packages/neos-ui-editors/src/CodeMirror/CodeMirrorWrap/codemirror.vanilla-css
@@ -3,7 +3,7 @@
 .CodeMirror {
   /* Set height, width, borders, and global font properties here */
   font-family: monospace;
-  height: 300px;
+  height: 85vh;
   color: black;
 }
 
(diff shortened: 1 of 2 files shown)


packages/neos-ui-editors/src/CodeMirror/CodeMirrorWrap/codemirror.vanilla-css
```diff
@@ -3,7 +3,7 @@
 .CodeMirror {
   /* Set height, width, borders, and global font properties here */
   font-family: monospace;
-  height: 300px;
+  height: 85vh;
```
dimaip: Why 85vh? Isn't it possible to make it take the full available height?

mstruebing: That would be a little bit more complicated than just a style property but I can try :1st_place_medal:

mstruebing: I hope this is better now

Issue #680, opened 2017-04-10
Default order of properties in the inspector is different from the old UI


dimaip:
Properties used to be ordered according to .yaml source order, now they are not.

mstruebing:
This can be closed

PR #709, opened 2017-05-03
BUGFIX: sorting messes up order for not positioned properties

dimaip:
Currently the sorter doesn't sort unpositioned elements correctly. They should retain the order defined in source code.
Also I extracted the logic to make it more clear and easier to test, and actually written a test for this sorting behavior.


dimaip:
Fixes #680

Issue #691, opened 2017-04-14
References / Select Editor Redone


skurfuerst:
# References / Select Editor

**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

**the current Pull Request (WIP) can be found at #726**

This document is divided into three parts:

* *Use Cases*, detailing the requirements for the Select/References editor (Problem Space)
* *Problems with current implementation*, reasoning what we learned from building the current implementation
* *Implementation Idea* - the new implementation idea

## Use Cases

The Select component is used in a variety of ways in the UI:

* as a "simple select" to choose from predefined options in various UI parts
* in the inline link editor, where the user starts typing and gets an *async* list of link suggestions
* in the reference+link editors, where the user starts typing and gets an *async* list of reference suggestions
* in the select editor:
  * choose between predefined options
  * choose between predefined options, but with search
  * choose between async loaded options, which are all loaded on startup (dataProvider)
  * [would be nice, but not possible now] choose between async loaded options, which are all loaded when the user types

* in the MultiSelect editor:
  * choose between predefined options
  * choose between predefined options, but with search
  * choose between async loaded options, which are all loaded on startup (dataProvider)
  * [would be nice, but not possible now] choose between async loaded options, which are all loaded when the user types
  * allow to sort multiple selections 

In all cases, the select value bound from the outside is *not* the label shown in the select box; but often some kind of identifier (i.e. node identifier). Sometimes, the select box includes the "empty" value, sometimes not (depending on the configuration).

In all cases, if an already-selected option value is not existing anymore, show [invalid value: XY].

In the async cases, we need to show a loading indicator.

### Asynchronous loading

It turns out the synchronous cases are all rather easy; it's the async cases which are difficult. We need to distinguish between the following cases:

* Pre-populate the list options on load; and if something is selected, also fetch the label for the selected ID. Pseudo API:
  `getOptions() returns Promise`
* Load the options lazily on typing (autocomplete). Here, the initial case needs to be handled separately, as we need to resolve by ID and not by label:
  `getSelectedValuesInitial(IDlist) returns Promise`
  `getOptions(SearchString) returns Promise`

## Problems with the current implementation

The current version of the Select Box is already the second bigger iteration. In the first iteration, we had just a single stateful component dealing with all aspects of the implementation.

In the second iteration (which is currently in master), we tore apart different components, leading to the structure shown in the linked diagram:

[...]

PR #726, opened 2017-05-18
FEATURE: rewritten Select box component

skurfuerst:
## This is a PR for #691

- [x] write basic Select Box component
- [x] write backend loading code (Sagas etc)
- [x] use Select Box editor in LinkEditor
  - [x] allow Node links
  - [x] allow URL links
  - [ ] allow asset links
- [x] use SelectBox component in Reference Editor
- [x] use SelectBox component in inline link editor
  - [x] allow Node links
  - [x] allow URL links
  - [ ] allow asset links
- [ ] use SelectBox component in Asset editor
- [x] write basic MultiSelect Box component
- [ ] ... (some more TODOs)


skurfuerst:
rework the editor / data loader concept after our discussions

I was able to greatly simplify most of the code as we discussed
in the meeting last week:

- the Sagas were removed
- the special Registry implementation was removed. It was a bad idea in the
  first place; especially because special Registry implementations should only
  be used if we need them for the WRITE side (in manifest.js); however, we used
  them on the READ side (in connected components).
- the Redux parts of the data loaders were removed
- the DataLoaders now return simple promises instead of being generator functions
- the cache is an implementation detail of the data loader.
- DropDown is now available in stateful and stateless fashion; I decided to
  not use recompose as it was not needed.
- The DataLoaders are now directly triggered from the relevant component
  (like LinkEditor).

By removing all layers between the components and the data loaders, this allows
to implement data loaders with DIFFERENT APIs and entirely different caches;
something which was not possible before this change.

gerhard-boden:
added minor styling improvements that jumped my eye, but want to take a closer look at this one when it's not that late😄 (especially that data loader part interests me)  

Found my way around in the code pretty fast, which is a good sign I guess 😉 Code looks very lean and accessible. Great work @skurfuerst!

packages/neos-ui/src/manifest.dataloaders.js
```diff
@@ -0,0 +1,160 @@
...
+
+            Takes the current context (workspace, dimensions) into account when doing the search, as this is passed in
+
+            OPTIONS:
+                - contextForNodeLinking: the current value of "selectors.UI.NodeLinking.contextForNodeLinking", required.
+                - nodeTypes: an array of node type names; if set, the search is restricted to these node types.
+        `,
+
+        _lru() {
+            if (!this._lruCache) {
```
nlx-sascha: Hey Sebastian, could you please explain me why you use external lib insteand of es6 Map?

PR #746, opened 2017-05-31
TASK: remove deprecation warnings

mstruebing:
- [x] Use prop-types package instead of React.PropTypes
- [x] Update react-codemirror
- [x] switch react-addons-test-utils to react-test-renderer see: https://github.com/airbnb/enzyme#installation
- [x] check if other react-addons are used
- [x] update react to 15.5.4
- [x] stylelint: 'block-no-single-line' has been deprecated
- [x] update a bunch of dependencies with react in its name
- [x] change all occurences of npm to yarn

-- `react-broadcast` is using React.PropTypes, there is already a commit that fixes that, but I don't want to use dev-master so we have to wait for a new release: https://github.com/mjackson/react-broadcast/commit/f376d87e72d3e9e83de48c247d493fe9213c3785
 ---> removed thx to @grebaldi 

Now there are only two deps left who uses the old PropTypes from react:
`$ grep -Hnir "^import .*PropTypes.* from 'react';"`

```
packages/react-ui-components/node_modules/babel-plugin-react-docgen/test/fixtures/case1/actual.js:1:import React, { PropTypes } from 'react';
packages/react-ui-components/node_modules/@kadira/react-split-pane/.storybook/stories/_utils.js:1:import React, { PropTypes } from 'react';
```

closes also #741 #724 #705 #704 #695


grebaldi:
We should probably get rid of `react-broadcast` since it doesn't seem to move. And it's experimental anyway...

Do you know if it's a third-party or are we depending on it directly? @mstruebing

grebaldi:
The remaining dependencies seem to be third-party dev-dependencies - so imho no show-stopper. I'd be fine with merging this, once it's been rebased. :+1:

Issue #688, opened 2017-04-13
"Edit / Preview" buttons slides as buttons are clicked


sorenmalling:
**Reproduce**
1. Click the "Edit / Preview" and display the **Editing mode**  and **Preview Central**
2. Click one of the Editig mode or previews and see the buttons slide from side to side, when click

**Other information**
It seems related to the css `transform: translate3d(..)` of the div inside `scroll-horizontal`

**Browser**
Google Chrome 
Version 57.0.2987.133 (64-bit)

skurfuerst:
Hey @sorenmalling,

thanks for your valuable feedback!

Could you add a short animated gif explaining the issue, f.e. by using [KAP](https://getkap.co/)?

Thanks, and all the best,
Sebastian

sorenmalling:
![neos-ui-sliding-buttons](https://cloud.githubusercontent.com/assets/531918/25001276/0073ebe8-2045-11e7-81ce-7fbe44b61459.gif)

skurfuerst:
Fancy 💃 interestingly I didn't see this yet

gerhard-boden:
So I did some research and it turns out that the error comes from a bug within external component for horizontal sliding:
https://github.com/hew/react-scroll-horizontal/issues/10

The proposed bugfix within the issue has already been merged, but does not fix our problem. Plus the component does not seem to be "heavily maintained" in general and needs a rewrite according to the creator.

I already tried replacing the component, but there aren't that many good lightweight horizontal scroll components out there as one would suppose. Already tried to replace it with a similar one, but didn't work right away so I wanted to gather some feedback first how we should proceed with this on.

As I see it our options are:

1. Help the creator of the used component to fix the issue
2. Find a different component and implement it
3. Write our own horizontal scroll component

dfeyer:
Hi, I think we can just abandon, the horizontal slider, it's a pure non sense in term of UX, a two select component, will work better

hew:
@gerhard-boden What's the issue (you are having)? I can't tell from the gif what exactly is going on. Regardless, I'll try to push a patch to the lib in the near future that addresses hew/react-scroll-horizontal#10. This lib definitely hasn't been maintained heavily up until this point, but as I see that no other lib has popped up (I honestly thought one would) but people are actually using this thing, I I'll try and be a bit more responsive.

gerhard-boden:
@hew Thanks a lot for reacting to this issue! 
[...]

PR #747, opened 2017-06-09
BUGFIX: Replace broken horizontal scroller for mode selection

gerhard-boden:
#fixes https://github.com/neos/neos-ui/issues/688


gerhard-boden:
So I went ahead and replaced the faulty component, just to find out I'm now using a faulty component as well. Very frustrating, because it in the end it's just a simple carousel. Might have to replace it once again. 

Any recommendations for carousel components in React?

PR #751, opened 2017-06-26
TASK: Adapt aspect to changed argument name in Neos

kitsunet:


robertlemke:
The symptom of this bug was:

image

PR #756, opened 2017-07-06
TASK: scroll content node into view when selecting a node from content tree

dimaip:
At first I wanted to use a polyfill for `scrollIntoView` with animation support, but all solutions I found didn't work with iframe scrolling (duh!), so I ended up animating the `scrollTop` property of the scrolling element.

**How to test:**

Select a node from content tree and make sure the page scrolls to reveal the active content node.


skurfuerst:
verrrry nice <3

PR #775, opened 2017-07-31
TASK: uncollapse content tree rootline when a node is focused

dimaip:
When selecting a content node within a page that node should be auto expanded and revealed in the content tree.

### DIFF
--- a/packages/neos-ui/src/Sagas/UI/ContentTree/index.js
+++ b/packages/neos-ui/src/Sagas/UI/ContentTree/index.js
@@ -1,8 +1,9 @@
 import {takeLatest} from 'redux-saga';
 import {put, select} from 'redux-saga/effects';
-import {$get} from 'plow-js';
+import {$get, $contains} from 'plow-js';
 
 import {actionTypes, actions} from '@neos-project/neos-ui-redux-store';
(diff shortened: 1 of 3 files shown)


packages/neos-ui/src/Sagas/UI/ContentTree/index.js
```diff
@@ -31,6 +32,28 @@ function * watchReloadTree({globalRegistry}) {
     });
 }
 
+function * watchNodeFocus() {
+    yield * takeLatest(actionTypes.CR.Nodes.FOCUS, function * loadContentNodeRootLine(action) {
+        const {contextPath} = action.payload;
+        const documentNodeContextPath = yield select($get('ui.contentCanvas.contextPath'));
+
+        if (contextPath !== documentNodeContextPath) {
```
grebaldi: This condition is implicit to the while loop below.

packages/neos-ui/src/Sagas/UI/PageTree/index.js
```diff
@@ -132,7 +124,7 @@ function * watchCurrentDocument() {
         if (contextPath !== siteNodeContextPath) {
```
grebaldi: same here

dimaip:
Thanks a lot for the review!

PR #782, opened 2017-08-01
BUGFIX: Let the userdropdown be over the codemirror editor

mstruebing:
screenshot_2017-08-01_10-48-15

Because Codemirror does something with it's z-index to do some stuff this is necessary

### DIFF
--- a/packages/neos-ui/src/Containers/PrimaryToolbar/style.css
+++ b/packages/neos-ui/src/Containers/PrimaryToolbar/style.css
@@ -7,6 +7,8 @@
     &:after {
         clear: both;
     }
+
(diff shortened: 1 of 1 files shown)


packages/neos-ui/src/Containers/PrimaryToolbar/style.css
```diff
@@ -7,6 +7,8 @@
     &:after {
         clear: both;
     }
+
+    z-index: 9999;
```
dimaip: Would be nice to extract all z-index constants into variables, to make the usage of them more predictable, like we did in the old UI.

PR #783, opened 2017-08-01
TASK: Add options to text input and area

samuelhauser:


samuelhauser:
checks failed because of #784

packages/neos-ui-editors/src/Boolean/index.js
```diff
@@ -47,7 +46,8 @@ BooleanEditor.propTypes = {
     label: PropTypes.string.isRequired,
     value: PropTypes.oneOfType([PropTypes.bool, PropTypes.string]),
     commit: PropTypes.func.isRequired,
-    isDisabled: PropTypes.bool
+    isDisabled: PropTypes.bool,
```
mstruebing: isn't this already in the options?

mstruebing: This line can be deleted I think.

mstruebing: This line can be removed as the disabled state comes from the options and isDisabled is also never used.

PR #789, opened 2017-08-01
TASK: respect loadingDepth in the trees

dimaip:
This PR switches from keeping track of uncollapsed nodes to tracking toggled nodes and XOR'ing them with the default collapsed state based on the loadingDepth configuration from the settings.


dimaip:
Reviewed by @Torsten85

PR #790, opened 2017-08-02
TASK: Hover state for publish button

samuelhauser:


packages/neos-ui/src/Containers/RightSideBar/Inspector/style.css
```diff
@@ -9,7 +9,8 @@
     text-align: center;
 }
 .publishBtn,
-.publishBtn:hover {
+.publishBtn:hover,
+.publishBtn:hover * {
```
dimaip: Are you sure this issue can not be fixed in the react-ui-components package on the original Button component?

PR #792, opened 2017-08-02
TASK: use new fontawesome declaration for contentcollection-icon

janinadanger:
to prevent error message for unknown icon

Related: #


janinadanger:
Hi dmitri,
I made this change in the main package as well. My idea was to fix this "bug" temporarily in the neos u in order to disable the error in the console. When the stable release is merged we should remove this issue

see: https://github.com/neos/neos-development-collection/pull/1655

mstruebing:
Hi Janina thx for your PR but we fix the issue like this: https://github.com/neos/neos-ui/pull/800
Also `folder-open` is a black icon, we would use `folder-open-o` which is the equivalent of `folder-open-alt`

PR #795, opened 2017-08-02
TASK: Set loading flag when the document creation is processing

### LINKED ISSUE #727: Loading indicator should be shown after adding a new page
Currently, if I add a new page it takes a couple of seconds until the page is shown in the left sidebar/page-tree. There's no indication that the page is currently being created.
After a few seconds the page is shown but until then it looks like something went wrong.
IMHO the page should be displayed right away with a loading indicator like it is shown when expanding pages. The indicator should disappear after the page has actually been persisted.
Not sure if that's a known issue but I think it could improve the UX.
Here's a gif that shows how it looks on my Neos setup:
https://cloud.passcreator.com/index.php/s/S5VacktLtji6rgu

akoenig:
This PR includes a mechanism for displaying the orange progress bar when the user creates a new document.

Closes #727

### DIFF
--- a/packages/neos-ui-redux-store/src/UI/ContentCanvas/index.js
+++ b/packages/neos-ui-redux-store/src/UI/ContentCanvas/index.js
@@ -13,6 +13,7 @@ const SET_PREVIEW_URL = '@neos/neos-ui/UI/ContentCanvas/SET_PREVIEW_URL';
 const SET_SRC = '@neos/neos-ui/UI/ContentCanvas/SET_SRC';
 const FORMATTING_UNDER_CURSOR = '@neos/neos-ui/UI/ContentCanvas/FORMATTING_UNDER_CURSOR';
 const SET_CURRENTLY_EDITED_PROPERTY_NAME = '@neos/neos-ui/UI/ContentCanvas/SET_CURRENTLY_EDITED_PROPERTY_NAME';
+const DOCUMENT_INITIALIZING = '@neos/neos-ui/UI/ContentCanvas/DOCUMENT_INITIALIZING';
 const DOCUMENT_INITIALIZED = '@neos/neos-ui/UI/ContentCanvas/DOCUMENT_INITIALIZED';
 const FOCUS_PROPERTY = '@neos/neos-ui/UI/ContentCanvas/FOCUS_PROPERTY';
 
@@ -25,6 +26,7 @@ export const actionTypes = {
     SET_SRC,
     FORMATTING_UNDER_CURSOR,
     SET_CURRENTLY_EDITED_PROPERTY_NAME,
+    DOCUMENT_INITIALIZING,
     DOCUMENT_INITIALIZED,
     FOCUS_PROPERTY
 };
@@ -34,6 +36,7 @@ const setPreviewUrl = createAction(SET_PREVIEW_URL, previewUrl => ({previewUrl})
 const setSrc = createAction(SET_SRC, src => ({src}));
 const setFormattingUnderCursor = createAction(FORMATTING_UNDER_CURSOR, formatting => ({formatting}));
 const setCurrentlyEditedPropertyName = createAction(SET_CURRENTLY_EDITED_PROPERTY_NAME, propertyName => ({propertyName}));
+const documentInitializing = createAction(DOCUMENT_INITIALIZING);
 const documentInitialized = createAction(DOCUMENT_INITIALIZED);
 
 //
@@ -45,6 +48,7 @@ export const actions = {
     setSrc,
     setFormattingUnderCursor,
     setCurrentlyEditedPropertyName,
+    documentInitializing,
     documentInitialized
 };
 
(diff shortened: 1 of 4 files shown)


packages/neos-ui-redux-store/src/UI/ContentCanvas/index.js
```diff
@@ -13,6 +13,7 @@ const SET_PREVIEW_URL = '@neos/neos-ui/UI/ContentCanvas/SET_PREVIEW_URL';
 const SET_SRC = '@neos/neos-ui/UI/ContentCanvas/SET_SRC';
 const FORMATTING_UNDER_CURSOR = '@neos/neos-ui/UI/ContentCanvas/FORMATTING_UNDER_CURSOR';
 const SET_CURRENTLY_EDITED_PROPERTY_NAME = '@neos/neos-ui/UI/ContentCanvas/SET_CURRENTLY_EDITED_PROPERTY_NAME';
+const DOCUMENT_INITIALIZING = '@neos/neos-ui/UI/ContentCanvas/DOCUMENT_INITIALIZING';
```
dimaip: Hey André!
Really nice to see you so active here!
Technically this PR works just as I would imagine it, great job!
One nitpick I have is with naming. Redux actions originally were supposed to represent something like commands, e.g. `FOCUS_PROPERTY` or `SET_SRC`. Then after we started using Redux Sagas we started to tolerate using Redux actions as a kind of events system, that represent that something has already happened, e.g. `DOCUMENT_INITIALIZED`. Those kinds of actions have nothing in their reducers, i.e. they do not chance the state usually, but are merely used to trigger some sage.
The `DOCUMENT_INITIALIZING` action that you added however doesn't feel right, it is actually a command to set document as loading, and I would rename it to something to reflect it better, e.g. `COMMENCE_DOCUMENT_INITIALIZATION` or `SET_DOCUMENT_AS_LOADING`, if you treat it merely as a setter.
Wdyt?

dimaip: And wait, there's one more thing.

akoenig: @dimaip I absolutely agree! To be honest, I was not sure how the team is handling the repective scenario. Personally, I also see actions as events (like in CQRS + EventSourcing) and name those in my projects in past tense as well.

So no worries regarding your "nitpick" :) It is highly appreciated and I will change it accordingly.

dimaip: I mean in redux originally actions are not events but rather commands in CQRS language, but we use it for both, regarding the context.

akoenig: @dimaip Regarding the `Saga` aspect: Yeah, that is absolutely correct and I totally agree on the fact that it would just be one line in the `ContentCanvas` reducer. On the other hand: This would lead to breaking the boundaries of the `ContentCanvas` reducer because it would react to an action type which shouldn't be the responsibility of that particular reducer. For me it looks more like a transaction (where the classical saga pattern fits nicely) for "communicating" or "mediate" between contexts.

What is your take on that? :)

grebaldi: Well, I think Events™ are the more suitable analogy for Action Types, while the identifiers for action creators should stay closer to Commands™.

`DOCUMENT_INITIALIZING` however is neither of those...

We are pretty inconsistent in this regard... But if we'd want to do things right from here on, I'd suggest to rename these action types to:

* `DOCUMENT_INITIALIZATION_STARTED`
* `DOCUMENT_INITIALIZATION_FINISHED`

And the corresponding action creators to:

* `startDocumentInitialization`
* `finishDocumentInitiazation`

... using the command form however, feels understandibly wrong at the place where the action is dispatched ...

dimaip: Ok, so you suggest to use Sagas to transcend reducer boundaries? Before we mainly tried to use Sagas for async actions.
So I see these options:
1) Ignore sub-reducer boundaries and allow to modify all state from any reducer. That would make a lot of logic very compact and explicit, also cleanup event log and give more semantical meaning to actions, that they are no longer a mere setters. On the other hand that would make the whole application a monolith.
2) Fire multiple actions straight from the create node dialog. That makes the logic scattered all over the place and action log noisy.
3) Use sagas to connect separate sub-reducer states by re-throwing actions based on other actions. The action log will still be noisy, but at least this logic is expressed in one place somewhere.

I myself would probably prefer 1), but it looks like we are drifting in the direction of 3) and use actions as mere setters, so maybe we should not resist the flow and go this way that feels natural to you.

Wdyt?

dimaip: @grebaldi regarding commands/actions, I don't really thing it makes sense to call action creators and actions differently, in our case they are basically exactly the same thing, they are hardly mapped/coupled.
It's clear that sometimes we use actions to tell the app to do something, and sometimes we just use them as signals to reflect that something has already happened and to listen to that signal via sagas.
E.g. `finishDocumentInitiazation` doesn't really work because document initialization has already been finished by this point, and we use an action merely to notify others of the already happened fact. If not for the sagas we would not even need that action...

One idea could be to use some different mechanism for those event-like actions, that would make them stand our from the stream of simple actions, or maybe bundle them all into a special sub-reducer e.g., e.g. maybe move them all into `System/`?

dimaip: @akoenig so we discussed it with Wilhelm and we decided to stay with 3) approach (as you have implemented it), and name actions to `START_LOADING` and rename `DOCUMENT_INITIALIZED` into `STOP_LOADING`. Wdyt?

akoenig: @dimaip Thanks for informing me about your conversation results. That is totally fine for me and I changed the PR respectively.
 
I also renamed `DOCUMENT_INITIALIZED` to `STOP_LOADING` which resulted in touching some other parts of the codebase as well. Therefore, reviewing those changes is highly appreciated :)

Thanks for all!

dimaip:
@akoenig good job! Thanks for the patience, a pleasure to code with you!

akoenig:
@dimaip Awesome, thanks for having me! Pleasure to code with you too! :)

Issue #794, opened 2017-08-02
Node still rendered in guest frame after deletion


akoenig:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

When deleting a node it is still rendered in the guest frame. What do you think of switching back to the root node afterwards?

![8d81328b-7078-47ec-ye34-5dce70bca3e7](https://user-images.githubusercontent.com/224910/28869076-dd7c902a-777b-11e7-802e-b826cf9bc32a.gif)

PR #798, opened 2017-08-03
WIP: Redirecting after node removal

akoenig:
Reference: https://github.com/neos/neos-ui/issues/794

### DIFF
--- a/packages/neos-ui/src/Sagas/UI/ContentCanvas/index.js
+++ b/packages/neos-ui/src/Sagas/UI/ContentCanvas/index.js
@@ -6,6 +6,17 @@ import {getGuestFrameDocument} from '@neos-project/neos-ui-guest-frame/src/dom';
 import {actionTypes, actions} from '@neos-project/neos-ui-redux-store';
 
 /**
+ * Observe the removal of a node
+ *
+ */
+function * watchNodeRemove() {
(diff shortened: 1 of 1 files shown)


dimaip:
Hey @akoenig this will be a very nice UX improvement. Do you plan to finish it?

akoenig:
@dimaip Yep, will work on that in the upcoming week.

dimaip:
@akoenig hey, any progress on this? If you want, I can take over.

akoenig:
@dimaip I planned to work on that at the weekend, but I guess in this case you will come up with a solution much quicker due to the fact that you are much deeper in the codebase.

I can give you some remarks I figured before: This server [feedback handler](https://github.com/neos/neos-ui/blob/master/packages/neos-ui/src/manifest.js#L213) gets overwritten by this [one](https://github.com/neos/neos-ui/blob/master/packages/neos-ui-guest-frame/src/manifest.js#L129) and therefore the dispatch of the [focus](https://github.com/neos/neos-ui/blob/master/packages/neos-ui/src/manifest.js#L221) actions does not go through.

dimaip:
@akoenig actually you are absolutely right, it was as simple as you said!
Fixed it here: https://github.com/neos/neos-ui/pull/931

Kind of annoying that you can register only one handler per feedback... Need to do smth about it!

PR #799, opened 2017-08-03
TASK: scroll the tree to focused node

dimaip:
**How to test**: click on a node inside content canvas that is not yet visible inside the content tree. The tree should scroll to reveal it. If the tree node is already in view the tree shouldn't scroll.

### DIFF
--- a/packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js
+++ b/packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js
@@ -71,12 +71,12 @@ export default class Node extends PureComponent {
         if (this.isFocused()) {
             const scrollingElement = findScrollingParent(this.domNode);
             if (scrollingElement) {
-                const nodePosition = this.domNode.getBoundingClientRect().top;
+                const nodeTopPosition = this.domNode.getBoundingClientRect().top;
                 const offset = 50;
                 const scrollingElementPosition = scrollingElement.getBoundingClientRect();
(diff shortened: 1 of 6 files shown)


packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js
```diff
@@ -49,6 +63,26 @@ export default class Node extends PureComponent {
...
 
+    componentDidUpdate() {
+        this.scrollFocusedNodeIntoView();
+    }
+
+    scrollFocusedNodeIntoView() {
+        if (this.isFocused()) {
+            const scrollingElement = findScrollingParent(this.domNode);
+            if (scrollingElement) {
+                const nodePosition = this.domNode.getBoundingClientRect().top;
```
mstruebing: This should get a clearer name as `scrollingElementPosition` is the BoundingClientRect it should be make clear that nodePosition has something to do with top.

packages/neos-ui/src/Containers/LeftSideBar/NodeTree/style.css
```diff
@@ -1,4 +1,8 @@
+// TODO: get rid of magick numbers here, they are not safe!
 .pageTree {
     overflow-y: auto;
-    max-height: calc(50vh - 88px);
+    max-height: calc(50vh - 85px);
```
mstruebing: please add `transition: max-height 0.5s;`

PR #804, opened 2017-08-03
TASK: Remove vscode settings from project repository

kitsunet:
If you need this, you should .gitignore the file in your
personal gitignore and put it there in your local installation.
This should never end up in the repository.

### DIFF
--- a/packages/.vscode/settings.json
+++ b/packages/.vscode/settings.json
@@ -1,4 +1,0 @@
-// Platzieren Sie Ihre Einstellungen in dieser Datei, um Standard- und Benutzereinstellungen zu überschreiben.
-{
-    "typescript.disableAutomaticTypeAcquisition": true
-}


mstruebing:
@skurfuerst :)

skurfuerst:
I disagree - why not make life easier for people using vs code?

kitsunet:
I mean, I don't feel SUPER strong about it, but if we do it with vscode, I could put my atom settings, someone else vim settings or whatever and it quickly becomes messy?

dimaip:
I don't mind it either, if that's something globally useful and not opinionated... But German comments? Well, I'm already learning German ;)

skurfuerst:
Hey,

If we remove this file, we'd also need to remove the file https://github.com/neos/neos-ui/blob/master/packages/jsconfig.json - as they belong together.

I'd personally would like to *re-add* it, and add documentation on how VSCode can be used in the README. Then it is transparent for others to use as well. I'd personally be fine if others do the same for Atom, IntelliJ, ...; so people have an as-friction-less-as-possible development experience.

Fine for you?

All the best,
Sebastian

kitsunet:
Sure :) With translated comment as @dimaip mentioned ;)

PR #812, opened 2017-08-09
BUGFIX: respect baseNodeType when rendering child nodes

dimaip:
No need to load filtered out document nodes into the state.

This change drastically improves performance if there are a lot of filtered nodes.

### DIFF
--- a/Classes/Neos/Neos/Ui/Fusion/Helper/NodeInfoHelper.php
+++ b/Classes/Neos/Neos/Ui/Fusion/Helper/NodeInfoHelper.php
@@ -74,7 +74,11 @@ class NodeInfoHelper implements ProtectedContextAwareInterface
             $nodeInfo['uri'] = $this->uri($node, $controllerContext);
         }
 
-        foreach ($node->getChildNodes() as $childNode) {
+        $documentChildNodes = $node->getChildNodes($this->baseNodeType);
+        $contentChildNodes = $node->getChildNodes('!Neos.Neos:Document');
+        $childNodes = array_merge($documentChildNodes, $contentChildNodes);
+
+        foreach ($childNodes as $childNode) {
             /* @var NodeInterface $childNode */
             $nodeInfo['children'][] = [
                 'contextPath' => $childNode->getContextPath(),
--- a/Classes/Neos/Neos/Ui/Fusion/Helper/NodeInfoHelper.php
+++ b/Classes/Neos/Neos/Ui/Fusion/Helper/NodeInfoHelper.php
@@ -74,7 +74,9 @@ class NodeInfoHelper implements ProtectedContextAwareInterface
             $nodeInfo['uri'] = $this->uri($node, $controllerContext);
         }
 
+        // child nodes for document tree, respecting the `baseNodeType` filter
         $documentChildNodes = $node->getChildNodes($this->baseNodeType);
+        // child nodes for content tree, must not include those nodes filtered out by `baseNodeType`
         $contentChildNodes = $node->getChildNodes('!Neos.Neos:Document');
         $childNodes = array_merge($documentChildNodes, $contentChildNodes);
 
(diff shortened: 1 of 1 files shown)


Classes/Neos/Neos/Ui/Fusion/Helper/NodeInfoHelper.php
```diff
@@ -74,7 +74,11 @@ public function renderNode(NodeInterface $node, ControllerContext $controllerCon
             $nodeInfo['uri'] = $this->uri($node, $controllerContext);
         }
 
-        foreach ($node->getChildNodes() as $childNode) {
+        $documentChildNodes = $node->getChildNodes($this->baseNodeType);
+        $contentChildNodes = $node->getChildNodes('!Neos.Neos:Document');
```
bwaidelich: Why not `$node->getChildNodes('Neos.Neos:Content');` or (if this needs to fetch all other nodes) `$node->getChildNodes('!' . $this->baseNodeType);`?

dimaip: @bwaidelich 
`Neos.Neos:Content` <-- doesn't includ ContentCollection nodes
`Neos.Neos:Content` <-- includes Document nodes that were filtered out by custom `baseNodeType`.

bwaidelich: Right, so this *should* include any other child nodes even if they are not rendering anything (like for example the FormElement definition NodeTypes I'm working on)? If so it might be clearer to rename `$contentChildNodes` to `$remainingChildNodes`!?

dimaip: @bwaidelich it should be all nodes needed for the content tree, no matter if rendered on the page or not. IMO `remainingChildNodes` doesn't better reflect the intent here, as these nodes are meant for the content tree, in contrast to the document tree.

dimaip: Added source code comments to explain the logic better

Classes/Neos/Neos/Ui/Fusion/Helper/NodeInfoHelper.php
```diff
@@ -74,7 +74,13 @@ public function renderNode(NodeInterface $node, ControllerContext $controllerCon
             $nodeInfo['uri'] = $this->uri($node, $controllerContext);
         }
 
-        foreach ($node->getChildNodes() as $childNode) {
+        // child nodes for document tree, respecting the `baseNodeType` filter
+        $documentChildNodes = $node->getChildNodes($this->baseNodeType);
+        // child nodes for content tree, must not include those nodes filtered out by `baseNodeType`
+        $contentChildNodes = $node->getChildNodes('!Neos.Neos:Document');
```
grebaldi: This breaks encapsulation, since we introduced the concept of nodeTypeRoles to address NodeTypes like this.

Have a look at https://github.com/neos/neos-ui/blob/ae609f49f4d8c9ae64db3e5b68446aed32878879/Configuration/Settings.yaml#L42

It could work to just replace the explicit `Neos.Neos:Document` with the value in `nodeTypeRole.document`, if we agree that the definition of content in any case is _not being a document_.

Otherwise the filter would have to include both, `nodeTypeRoles.content` and `nodeTypeRoles.contentCollection`

dimaip: Sure, I had my mind on cleaning this up, since we have this kind of mess in other lines of this file already (and all over the system in general). I think I'll go with the second option you suggested since this logic is pretty specific to the tree.

dimaip:
@grebaldi could you please take a look again, I tried to address your suggestion.

dimaip:
image

PR #818, opened 2017-08-12
WIP: Add missing tooltip, for elements which consist only of an icon

janinadanger:
needed to wrap each icon with an extra `<div>`, to align tooltip with parent element. Each tooltip should be centered on the icon. 

Therefore some icons are moved in their position (for example: **right sidebar toggle icon**) 

this PR needs rework


dimaip:
@janinadanger thanks a lot for getting this started! Closing in favor of #997

PR #819, opened 2017-08-13
BUGFIX: Remove button icon background

### LINKED ISSUE #814: Icons within buttons have background
image

mstruebing:
closes #814

### DIFF
--- a/packages/react-ui-components/src/Button/style.css
+++ b/packages/react-ui-components/src/Button/style.css
@@ -36,9 +36,11 @@
     padding: 0 8px;
 }
 
-.btn--clean,
-.btn--clean i {
+.btn--clean {
     background-color: var(--brandColorsContrastDarker);
+}
+.btn--clean i {
+    background-color: 0;
 }
 .btn--lighter,
 .btn--lighter i {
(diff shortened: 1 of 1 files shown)


packages/react-ui-components/src/Button/style.css
```diff
@@ -36,10 +36,12 @@
     padding: 0 8px;
 }
 
-.btn--clean,
-.btn--clean i {
+.btn--clean {
     background-color: var(--brandColorsContrastDarker);
 }
+.btn--clean i {
+    background-color: 0;
```
dimaip: Hey, what does `background-color: 0;` mean? Why is it necessary?

mstruebing: You are right :+1: 
`background-color: 0` is invalid
transparent or something like rgba(0, 0, 0, 0) is more what I was looking for. I will add this later

dimaip: sorry for knit-picking, but why is it even needed to set a transparent background? It should be transparent by default, no?!

mstruebing: Yep, you are right, sorry. PR updated :)

dimaip:
Thanks!

Issue #690, opened 2017-04-13
Embed webfont directly instead of importing from external source


sorenmalling:
The importing of the Work Sans webfont via Google Fonts can be avoided by shipping it along with the ui

A webfont kit can be downloaded from here

https://www.fontsquirrel.com/fonts/work-sans

skurfuerst:
@sorenmalling if you like, I guess that would be an easy one to get started contributing :)

PR #821, opened 2017-08-13
TASK: Extract fonts to load seperatelty

### LINKED ISSUE #816: Revalidate fonthandling
As discussed in Slack the delivered css and js are quite big files because of the direct include of fonts (font-awesome/work-sans) instead of loading them separately.

mstruebing:
closes #816 

@dimaip


dimaip:
Hey! What about that remark that Jonas made? As said, I'm not up to date on the topic. E.g. maybe our browser support would allow us to only ship woff2? That would reduce the download size.

mstruebing:
I've decided to include woff because of: http://caniuse.com/#search=woff vs http://caniuse.com/#search=woff2

IE11 isn't supporting woff2, but I think many of NeosCMS users has to be IE11 compatible.

I'm not up to date which would be better either. But I can make a research on that topic.

Maybe @jonas-siegl can say something more detailed about that?

mstruebing:
This is the only `useful` resource I could find right now: https://www.bramstein.com/writing/web-font-anti-patterns-inlining.html

But maybe Bram isn't telling us everything.

dimaip:
Yeah, I had suspected something like this, just wasn't sure enough to insist. Also we can cache fonts very aggressively, but that's not something we want to do with CSS.
I would just merge your PR, but wonder what @jonas-siegl and other more experienced frontend devs think. Maybe @aertmann has an opinion?

packages/build-essentials/src/webpack.config.js
```diff
@@ -33,8 +33,8 @@ const webpackConfig = {
                 loader: 'json-loader'
             },
             {
-                test: /\.(woff|woff2|eot|svg)$/,
-                loader: 'url-loader?limit=100000'
+                test: /\.(woff|woff2)$/,
```
grebaldi: I think this loader can be dropped, since it is removed from the package.json.

Any reasons, why it's still here?

jonas-siegl:
Hey,
I'm sorry for my late reply and the trouble. For now I would also recommend to remove the inlining completely, because it's not useful in combination with the current webfont integration approach. 
There are more advanced loading techniques to fight FOIT and FOUT where it make sense to use it. 
Here is a great article where you can read more about webfont loading strategies: https://www.zachleat.com/web/comprehensive-webfonts/

packages/build-essentials/src/webpack.config.js
mstruebing: Current implementation of font-awesome.

https://github.com/neos/neos-ui/tree/57b0655ef67c52168c3298869b1a2906a711eefe/packages/react-ui-components/src/Icon/fontAwesome

grebaldi: In that case you shouldn't remove the loader from the dependencies, right? https://github.com/neos/neos-ui/pull/821/files/b0242e333c0d12bf7fbeb031e3c6fe10a2831143#diff-6e183e897d30da3e86a99a9c16857780L52

mstruebing: It was working like this before I implemented the typeface-work-sans npm module and it is working again, the loader was added because of errors during the build with this module, why should I add a dependency I don't need?

Configuration/Settings.yaml
```diff
@@ -34,6 +34,15 @@ Neos:
           'Neos.Neos.UI:HostOnlyStyles':
             resource: '${"resource://" + Neos.Ui.StaticResources.compiledResourcePackage() + "/Public/Styles/HostOnlyStyles.css"}'
             position: 'start 900'
+          'WorkSans:200':
```
dimaip: Ehh you can't load the font like this, it ain't no webpack... ;)
You need to do smth like this CSS:
```
@font-face {
  font-family: "WorkSans";
  src: url("/...pathetofonts/work-sans-latin-200.woff") format("woff");
  font-weight: normal;
  font-style: normal;
}
```

dimaip: Also, if we switched from NotoSans to WorkSans (why?) you need to also change the family here: https://github.com/neos/neos-ui/blob/master/packages/neos-ui/src/styleHostOnly.css#L8

dimaip: Also, do we really need all three weights? As far as I can see in the UI we only use normal and bold weights.

mstruebing: I don't know why we switched. As I first picked up font stuff it were already work-sans.
These is were I picked up:
https://github.com/neos/neos-ui/pull/761
https://github.com/neos/neos-ui/issues/690

grebaldi: @dimaip We switched because of this, I think: https://github.com/neos/brand/blob/2f2fe306f0ce1c5ee2ca7345c51464134cfad247/index.js#L23

Is it possible, that the brand assets are used for both, the website and the UI?

dimaip: @grebaldi what do you mean by brand assets?

I don't mind WorkSans, we just have to use it here: https://github.com/neos/neos-ui/blob/master/packages/neos-ui/src/styleHostOnly.css#L8
Preferably by using brand constants of course.

Inkdpixels: @grebaldi @dimaip the referenced "brand-assets" is a package where the colors and fonts of Neos are defined, and yes they where used in the website back when I started it, this difference is ofc now kind of problematic. 

Was the change in font families meant to be just for the website or the whole brand of Neos? In the latter case ofc it makes sense to also change it in the UI doesnt it?

dimaip: @Inkdpixels hey nice to see you here again!
I'm sure it was just a mistake, we should use the definition from the brands package ofc.

packages/build-essentials/src/webpack.config.js
jonas-siegl: Because you are still using it. The loader "url?limit=100000" is a shortcut for "url-loader?limit=100000". Webpack automatically appends "-loader" to it when the first module lookup failed. Maybe it's still working because another package has the module "url-loader" as a dependency.
Anyway I would recommend to not use the short syntax because it can lead to strange issues if there is another module called "url". Therefore this logic has been removed in Webpack 2 by default. (see https://github.com/webpack/webpack/issues/2986)

dimaip:
Can we then remove the worksans npm package?

mstruebing:
I did: https://github.com/neos/neos-ui/pull/821/files#diff-2fce5f75e3e31d94e254e4466f727a84L92

dimaip:
Oh, sorry, missed it!

mstruebing:
So, everything alright then? :)

dimaip:
As you see, I approved your PR, but didn't push the green button myself, as I'm scared to take the final responsibility :laughing: 
Go ahead and push the button!

mstruebing:
boom-2028563_960_720

PR #787, opened 2017-08-01
TASK: Add options to text input and area

samuelhauser:

PR #822, opened 2017-08-14
BUGFIX: Use empty object to assign options to

mstruebing:
to eliminate sideeffects

like mentioned in #787


dimaip:
Good catch!

PR #839, opened 2017-08-28
TASK: add end-to-end tests with TestCafe

dimaip:
[TestCafe](https://devexpress.github.io/testcafe/) is really cool. It works without selenium in multiple browsers and even on mobile devices, it has a built-in set of [React selectors](https://github.com/DevExpress/testcafe-react-selectors) etc.

This PR lays a foundation for a TestCafe integration.

- [x] cleanup output of ./flow server:run
- [x] write more tests
- [x] testing withing portals: https://github.com/DevExpress/testcafe-react-selectors/issues/33 For now we can use id's.
- [x] speed up tests bypassing authentication?
- [x] fail tests on PropTypes validation fails https://testcafe-discuss.devexpress.com/t/catch-js-errors-and-console-output/421


.travis.yml
```diff
@@ -5,24 +5,88 @@ notifications:
...
+
+  # Execute the optional install shell script for the environment.
+  - if [ -f "Packages/Application/Neos.Neos.Ui/Build/TravisCi/InstallScripts/$TEST_SUITE.sh" ]; then bash Packages/Application/Neos.Neos.Ui/Build/TravisCi/InstallScripts/$TEST_SUITE.sh; fi
+
+  # Change into the repository directory where the environment based shell script will be executed.
+  - cd Packages/Application/Neos.Neos.Ui
+
+  # Since all environments depend on the node dependencies, install and
+  # afterwards prune them to remove extranous packages from previous/cached runs.
+  - yarn
```
mstruebing: You should install yarn at first

dimaip: Yarn is preinstalled on Travis CI when yarn.lock is present: https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn

mstruebing: Then forgot it, I thought it is only present in node_js

skurfuerst:
looks really really promising!

dimaip:
Guys, I've spent a ton of time debugging selectors within portals, and now they seem to work, thanks to TestCafe devs, they are super nice!

I have put all test code into one test for now, to avoid wasting time on re-authentication.
It could be possible to bypass authentication on our side, I tried to create a dummy authentication provider, but couldn't get it to work quickly. If some of you who know Flow better want to give it a try, go ahead.
But on the other hand, even with authentication bypassed, every test would still need to wait for the UI to load, which could accumulate to a lot of time.
The way I did it know is that it clicks through everything in one test. It has a downside that one part of test may influence the other, by accumulating state. But on the other hand it's even better, as it may catch stateful bugs as well.
So I propose to merge it this way for now, unless you want to do it better.

dimaip:
Actually the use of roles does skip authentication the second time now, don't know why it didn't work before.
But the question still remains, do we want to wait extra ~3s per each test? Also if you click-through all tests at once you got more chances of catching stateful bugs. On the other hand there's bigger chance of one part of test breaking another...
Thoughts? Maybe split tests into big parts, but not into too many parts?

dimaip:
Ok, merging as is. Got ideas how to make it better? Ping me or just submit a PR!
And for all good in the world, let's write more E2E tests, I can't do it all myself, and it's kind of fun task :)

PR #840, opened 2017-08-29
BUGFIX: Perform html decoding on node labels

### LINKED ISSUE #828: Umlauts are not displayed correctly in the inspector
If you use an umlaut in an title of a node, this title is not displayed correctly.

neos/neos-ui dev-master 9df0ee7

image

akoenig:
This PR fixes the problem with HTML encoded parts of a node label.

## Solution

The previous solution stripped the tags out of a node label, but didn't decode the HTML in the inner parts of the HTML string. This PR introduces [he](https://github.com/mathiasbynens/he) and shields the API by defining an own function in `utils-helpers`

Closes #828


packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js
```diff
@@ -1,10 +1,11 @@
 import React, {PureComponent} from 'react';
 import PropTypes from 'prop-types';
 import {$get} from 'plow-js';
+import {compose} from 'redux';
```
dimaip: I like your functional style! But just wondering, do you think it's a good idea to hijack Redux's helper function for this usecase? Maybe we should better use the same method say from lodash (`flowRight`)?
Feels weird to couple html decoding to redux...

akoenig: @dimaip Good catch, yes. Thought so too, but hesitated to bring in another dependency. (`lodash.flowright`) :)

dimaip: I think we can afford it, shouldn't be too fat :)

akoenig: @dimaip I figured that we have the "old" `lodash.compose` in there. Used that instead now.

packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js
```diff
@@ -1,9 +1,9 @@
 import React, {PureComponent} from 'react';
 import PropTypes from 'prop-types';
 import {$get} from 'plow-js';
-import {compose} from 'redux';
 import {connect} from 'react-redux';
 
+import compose from 'lodash.compose';
```
dimaip: In V4, `compose` is removed in favor of `flowRight`. Maybe use that right away, to save us from a breaking change in the future? https://github.com/lodash/lodash/wiki/Changelog
They are the same thing, just aliases.

dimaip: Or do we already use `lodash.compose` somewhere else? In this case it's ok.

packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js
dimaip: ah perfect

packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js
```diff
@@ -160,6 +161,11 @@ export default class Node extends PureComponent {
         };
     }
 
+    decodeLabel = compose(
```
dimaip: I think we can merge it as is, just a stylistic comment, since this `decodeLabel` is just an utility method and doesn't need access to `this`, it could be just written above the `class`, as a simple variable.

akoenig: @dimaip Okay, fair point. Will do that :)

akoenig:
You are very welcome. Thanks for merging!

PR #848, opened 2017-09-01
BUGFIX: Add missing url-loader dependency in neos-ui package

mstruebing:


skurfuerst:
Just out of curiosity: How come this has been removed? :)

PR #859, opened 2017-09-04
TASK: refactor dependencies to the top level

dimaip:
PROPOSAL:

Separate packages that are meant to be published to npm, and those that don't.
For now these are: `react-ui-components`, `neos-ui-extensibility`.

Other packages we use merely for internal structuring so it would be easier to import files from them.
For these internal packages I propose to merge all of their dependencies and put to the top level package.json.
For public packages we can still share all dev deps, and only duplicate production deps.

Pros:
- A lot faster, no need to duplicate deps in multiple places
- Easier to understand and maintain
- Would be possible to use Greenkeeper again (if we really want it, I don't)

Cons:
- It will be a bit harder to extract deps for a package if we'd like to publish it separately. But in fact that's not hard at all, just search for all "import" and you'll quickly see it... Besides I don't think we'd want to publish any new public packages any time soon.


dimaip:
Eh, there are issues with react-ui-componets, yarn start doesn't work. Will take care later, do not merge.

skurfuerst:
1) Yarn works, 2) Webpack build works 3) Storybook runs 🙂  - awesome!

PR #869, opened 2017-09-04
TASK: add support for selecting time to DateInput

dimaip:
The styles of the component are pretty messy, but since they are isolated in scope I didn't spend too much time cleaning it up.
I think UX-wise solution is OK, I somehow even like it better than what we had in the old UI.
Thoughts?

peek 2017-09-04 21-59


skurfuerst:
Looks really really good :)

dimaip:
Good to merge from my side.

dimaip:
@mstruebing ha, actually you found an unrelated (and very funny) bug. I fixed it.

PR #871, opened 2017-09-05
BUGFIX: scroll inline node into view only when requested from the tree

### LINKED ISSUE #835: Scrolling contentCanvas is a bit weird

See:
peek 2017-08-24 16-15

dimaip:
Fixes #835.

I couldn't think of a nicer way to do it. We imperatively request to scroll the node into view when selecting a node from the tree.


skurfuerst:
AH, Btw, can we add a comment on the action and the NodeToolbar explaining this imperative behavior and linking them together?

dimaip:
Good point, added a few comments.

PR #873, opened 2017-09-05
BUGFIX: fallback to the first tab when tab index is out of bounds

### LINKED ISSUE #826: When switching focused node, inspector becomes blank if the focused nodetype doesn't have that group

Steps to reproduce:
1. open in inspector some group that only this nodetype has, e.g. SEO group
2. focus a node that doesn't have such group, e.g. Text
3. contemplate inspector staying blank, instead of focusing to first available group in the new nodetype

image

dimaip:
Fixes #826

### DIFF
--- a/packages/react-ui-components/src/Tabs/tabs.js
+++ b/packages/react-ui-components/src/Tabs/tabs.js
@@ -73,13 +73,20 @@ export default class Tabs extends PureComponent {
         );
     }
 
+    getActiveTab() {
+        // If activeTab is out of bounds, choose the first tab
+        const {activeTab} = this.state;
+        const childrenCount = React.Children.count(this.props.children);
+        return childrenCount < activeTab ? 0 : activeTab;
+    }
+
     renderMenuItems() {
         const {
             IconComponent,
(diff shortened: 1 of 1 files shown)


packages/react-ui-components/src/Tabs/tabs.js
```diff
@@ -73,13 +73,20 @@ export default class Tabs extends PureComponent {
         );
     }
 
+    getActiveTab() {
```
skurfuerst: I'd call this "activeTabIndex"; and also change the state property to "activeTabIndex" - is more clear, wdyt?

dimaip: The problem with renaming state `activeTab` is that it has related prop `activeTab`, which is a part of component's API, and since we publish this package to npm and already version it, that would require a version bump... Frankly I'd just ignore it, but if you insist I can take care.

skurfuerst: OK then we keep it as is. Fine ;)

PR #881, opened 2017-09-05
BUGFIX: fix performance of trees by fixing makeChildrenOfSelector

dimaip:
Also fixes the `debug-reason-for-rendering` helper to support pure components by refactoring to HOC. 
It drastically improves performance! Before all nodes were rerendered on each update.

Before it used inheritance, which didn't work and in general is not a good idea with React components.

### DIFF
--- a/packages/neos-ui-redux-store/src/CR/Nodes/selectors.js
+++ b/packages/neos-ui-redux-store/src/CR/Nodes/selectors.js
@@ -73,6 +73,7 @@ export const makeChildrenOfSelector = allowedNodeTypes => createSelector(
     .map(
         contextPath => $get(contextPath, nodesByContextPath)
     )
+    .filter(node => node)
 );
 
 export const siteNodeSelector = createSelector(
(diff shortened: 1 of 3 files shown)


packages/neos-ui-redux-store/src/CR/Nodes/selectors.js
```diff
@@ -73,6 +73,7 @@ export const makeChildrenOfSelector = allowedNodeTypes => createSelector(
     .map(
         contextPath => $get(contextPath, nodesByContextPath)
     )
+    .filter(node => node)
```
skurfuerst: I do not understand yet how this change fixes the behavior. Do you have a hint / explanation on why this helps? :-)

dimaip: Sure. Without this line we get an immutablejs array full of `undefined`s, and (for some reason) when it compares two such arrays, with identical number of underineds it tells that they are difference, so forces a re-render.

skurfuerst: ah; I see! THX!

PR #883, opened 2017-09-05
TASK: Use Open Sans as the default UI font

dfeyer:


dfeyer:
Require https://github.com/neos/brand/pull/8 to be merged and a new version of the package released

packages/build-essentials/src/webpack.config.js
```diff
@@ -33,8 +33,8 @@ const webpackConfig = {
                 loader: 'json-loader'
             },
             {
-                test: /\.(woff|woff2)$/,
-                loader: 'url-loader?limit=100000'
+                test: /\.(woff|woff2|eot|ttf|svg)$/,
+                loader: 'url-loader?limit=10000&name=./Styles/Font-[hash].[ext]&publicPath=./../'
```
dfeyer: A bit ugly, but I'm not a big friend of Webpack, so maybe someone found a better solution

dfeyer: I love the value of limit, it does not make sens to embed the font in the CSS; HTTP2 make a better job with concurent request, and we don't update the font a lots ;)

packages/react-ui-components/.storybook/webpack.config.js
```diff
@@ -13,8 +13,8 @@ module.exports = {
                 loader: 'json-loader'
             },
             {
-                test: /\.(woff|woff2)$/,
-                loader: 'url?limit=100000'
+                test: /\.(png|woff|woff2|eot|ttf|svg)$/,
+                loader: 'url?limit=10000'
```
dfeyer: Not sure if this one need to be adapted, not tested

packages/neos-ui/src/styleHostOnly.css
```diff
@@ -1,25 +1,81 @@
...
+    font-weight: 300;
+    font-style: normal;
 }
 
 @font-face {
-  font-family: 'Work Sans';
-  src: url('../../../Resources/Public/Fonts/WorkSans-Bold.woff') format('woff');
-  font-weight: bold;
-  font-style: normal;
+    font-family: 'Open Sans';
```
dimaip: Do we need all of those font weights? That would be like tons of data...

dfeyer: @dimaip Based on the discussion on the UX Guild, I think we need more weights in the future. Currently we use only 2 weight (regular and bold), so only used weights are loaded.

Before this change the font is base64 encoded in the CSS, so the CSS was around 500Ko, now we have a really slim CSS, and even if all weights are loaded we save bandwidth. Fonts can be cached (http, or localstorage in the future, if we need to improve this)

packages/build-essentials/src/webpack.config.js
skurfuerst: ack :)

packages/neos-ui/src/styleHostOnly.css
dimaip: @dfeyer hey, but let's add font weights only when we need them, when the design changes? We ship something when it's not used?
Currently we don't base64 fonts afaik, @mstruebing fixed it a while ago.

dimaip: @dfeyer do you agree?

dimaip:
@dfeyer hey, do you still plan to work on this one?

skurfuerst:
We should implement this really soon :) @dimaip would you have time to do this? :)

dimaip:
Yes!

dimaip:
Don't seem to have push access here, closing in favor of #1089

PR #884, opened 2017-09-05
TASK: Remove top padding in Tabs component

dfeyer:


dfeyer:
Before


After

dimaip:
Thanks! Merged as is. Can you double check there are no other places that got affected by it in the bad way?

PR #886, opened 2017-09-05
BUGFIX: Fix right padding when SecondaryToolbar is fringe right

dfeyer:


dimaip:
What problem is it trying to solve? Don't see any visual bugs there.

dfeyer:
@dimaip The button to toogle the inspector is 46px wide, so the full screen button is partly hidden without this change

Issue #806, opened 2017-08-04
Mark a new change in the page/content tree


mstruebing:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

If I made some changes to some nodes I expect that these nodes are marked with an orange indicator like in the old ui, which is currently not the case.

![screenshot_2017-08-04_10-16-54](https://user-images.githubusercontent.com/12071529/28960229-0e14407e-78fe-11e7-87e6-b2fabc043487.png)

skurfuerst:
As that's an important UI concept from the old UI, we should replicate it!

PR #890, opened 2017-09-06
TASK: mark "dirty" nodes in the trees

dimaip:
Resolves: #806
image

### DIFF
--- a/packages/neos-ui-redux-store/src/CR/Workspaces/selectors.js
+++ b/packages/neos-ui-redux-store/src/CR/Workspaces/selectors.js
@@ -15,11 +15,12 @@ export const publishableNodesInDocumentSelector = createSelector(
     (publishableNodes, activeDocumentContextPath) => publishableNodes.filter(i => $get('documentContextPath', i) === activeDocumentContextPath)
 );
 
-export const makeIsDocumentNodeDirtySelector = documentContextPath => createSelector(
+export const makeIsDocumentNodeDirtySelector = () => createSelector(
     [
-        publishableNodesSelector
+        publishableNodesSelector,
+        (state, documentContextPath) => documentContextPath
     ],
-    publishableNodes => publishableNodes.filter(i => (
+    (publishableNodes, documentContextPath) => publishableNodes.filter(i => (
         $get('documentContextPath', i) === documentContextPath ||
         $get('contextPath', i) === documentContextPath
     )).count() > 0
--- a/packages/neos-ui-redux-store/src/CR/Workspaces/selectors.js
+++ b/packages/neos-ui-redux-store/src/CR/Workspaces/selectors.js
@@ -25,9 +25,10 @@ export const makeIsDocumentNodeDirtySelector = documentContextPath => createSele
     )).count() > 0
 );
 
-export const makeIsContentNodeDirtySelector = contextPath => createSelector(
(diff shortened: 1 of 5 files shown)


packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js
```diff
@@ -303,6 +307,7 @@ export const ContentTreeNode = withNodeTypeRegistry(connect(
             currentDocumentNode: selectors.UI.ContentCanvas.documentNodeSelector(state),
             focusedNodeContextPath: $get('cr.nodes.focused.contextPath', state),
             toggledNodeContextPaths: selectors.UI.ContentTree.getToggled(state),
+            isNodeDirty: selectors.CR.Workspaces.makeIsContentNodeDirtySelector($get('contextPath', node))(state),
```
skurfuerst: same here

packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js
```diff
@@ -273,6 +275,7 @@ export const PageTreeNode = withNodeTypeRegistry(connect(
             intermediateContextPaths: selectors.UI.PageTree.getIntermediate(state),
             loadingNodeContextPaths: selectors.UI.PageTree.getLoading(state),
             errorNodeContextPaths: selectors.UI.PageTree.getErrors(state),
+            isNodeDirty: selectors.CR.Workspaces.makeIsDocumentNodeDirtySelector($get('contextPath', node))(state),
```
skurfuerst: oups, that was too much ;-) the selector *factory* must be in the outer function (at the old location); but it is *not allowed to take any arguments*.

The argument "contextPath" must be passed into the selector as 2nd parameter after "state" :-)

This way you did it here means we do not memoize at all

dimaip:
@skurfuerst now I see what you mean by "second parameter after state". I never knew that there was such thing as a second parameter to a selector. Actually I think the convention is to pass props there, so it would work with default mapStateToProps. Maybe it would be better to do it this way everywhere -- pass the props and let the selector get what it needs?
If you don't think it's a good idea just merge it as is now.

dfeyer:
For me it work nicely for the structure tree, but show nothing in the document tree (if I edit a document title, or for document that content unpublished edited content

packages/neos-ui-redux-store/src/CR/Workspaces/selectors.js
```diff
@@ -14,3 +14,21 @@ export const publishableNodesInDocumentSelector = createSelector(
     ],
     (publishableNodes, activeDocumentContextPath) => publishableNodes.filter(i => $get('documentContextPath', i) === activeDocumentContextPath)
 );
+
+export const makeIsDocumentNodeDirtySelector = documentContextPath => createSelector(
```
skurfuerst: this selector still needs to be rewritten (to same way as the selector below  regarding arguments :) )

skurfuerst:
@dfeyer absolutely, can be explained by the code position I highlighted :)

dimaip:
@skurfuerst argh sorry, I got so caught up in the conceptual reselect topics that I forgot about the second selector! Fixed.

PR #902, opened 2017-09-07
BUGFIX: force dev-master version of Neos.Demo

dimaip:
For some reason currently it's installing the wrong branch.


skurfuerst:
looks good!

Issue #766, opened 2017-07-11
Pages can be added after main content collection


daniellienert:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

When selecting the main content collection in the content tree, press plus and select "insert after" the list of document nodes is shown.

This is technically correct but a bit confusing for editors. I would propose to disable the "insert after" mode there.

skurfuerst:
Hey,

I think forbidding this completely might not be the right approach; as there might be cases where Document Nodes will be directly editable on the page -> then it would feel natural to be able to insert them directly in-place in the content area.

I could imagine the following: We make it configurable on a per-node-type basis whether it is shown in the "Add Content" dialog or not. For Content, we enable it by default; for Document, we disable it by default.

What do you think?

All the best, Sebastian

daniellienert:
Hey Sebastian, 
that would be a good default behavior!

Cheers, 
Daniel

skurfuerst:
Hey,

marking this as "Must have for 1.0" as it is a major usability issue for editors.

All the best, Sebastian

dimaip:
Hey guys!
I don't think I'm in favor of adding another configuration option before 1.0.
I propose this behavior: when a document node is selectable inline, when you click an inline button, you'd get a dialog with only document nodes there, same as you would get from the document tree toolbar.
When you click add node button from content tree toolbar, you'd get just content nodes, consistent with the old UI.
Wdyt?

skurfuerst:
Good idea <3 Let's do it your way

PR #906, opened 2017-09-07
BUGFIX: only allow to create nodes of the same role as the reference node

dimaip:
Fixes: #766


skurfuerst:
Hm, need to investigate wyh this is the case; actually this has worked before I am quite sure

dimaip:
Hmm again I can not reproduce, sorry! Can you post an example NodeTypes.yaml snippet and exact description how to reproduce it.
I checked it with Sfi.Grid extension: https://github.com/psmb/Sfi.Grid/blob/master/Configuration/NodeTypes.yaml

image

dfeyer:
I can do more test later today

dimaip:
@dfeyer Can you also test if it was my changed that broke it for you or not. Because if not we can just merge this one.

PR #910, opened 2017-09-07
BUGFIX: update previewUrl and uris in state after editing uriPathSegment

### LINKED ISSUE #645: Editing URL path segment is not supported

If we rename the URL path segment, move to an other document, do back to the renamed ones, throw an exception (routing issue), if reload the UI everything fine.

if the renamed node has children all children are not accessible from the tree

dimaip:
Fixes #645

### DIFF
--- a/packages/neos-ui-redux-store/src/CR/Nodes/index.js
+++ b/packages/neos-ui-redux-store/src/CR/Nodes/index.js
@@ -159,10 +159,10 @@ const show = createAction(SHOW, contextPath => contextPath);
  * Update uris of all affected nodes after uriPathSegment of a node has changed
  * Must update the node itself and all of its descendants
  *
- * @param {String} oldUri
- * @param {String} newUri
+ * @param {String} oldUriFragment
+ * @param {String} newUriFragment
  */
-const updateUri = createAction(UPDATE_URI, (oldUri, newUri) => ({oldUri, newUri}));
+const updateUri = createAction(UPDATE_URI, (oldUriFragment, newUriFragment) => ({oldUriFragment, newUriFragment}));
 
 //
 // Export the actions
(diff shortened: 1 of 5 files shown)


dimaip:
I'm not super happy with how I implemented it, but that's the best I could think of now. Improvement suggestions are welcomed :)

packages/neos-ui-redux-store/src/CR/Nodes/index.js
```diff
@@ -230,7 +242,30 @@ export const reducer = handleActions({
     [CUT]: contextPath => $set('cr.nodes.clipboard', contextPath),
     [PASTE]: () => $set('cr.nodes.clipboard', ''),
     [HIDE]: contextPath => $set(['cr', 'nodes', 'byContextPath', contextPath, 'properties', '_hidden'], true),
-    [SHOW]: contextPath => $set(['cr', 'nodes', 'byContextPath', contextPath, 'properties', '_hidden'], false)
+    [SHOW]: contextPath => $set(['cr', 'nodes', 'byContextPath', contextPath, 'properties', '_hidden'], false),
+    [UPDATE_URI]: ({oldUri, newUri}) => state => {
+        const allNodes = $get('cr.nodes.byContextPath', state);
+        // Make sure to not include false positives by checking that the given segment ends either with "/" or "@"
+        const containsOldUriSegmentRegex = new RegExp(oldUri + '(/|@)');
```
skurfuerst: here, we should escape "oldUri" e.g. using https://lodash.com/docs#escapeRegExp, no?

skurfuerst: actually I would not use regexes here, as we only check for substring matching of two strings, right?

packages/neos-ui-redux-store/src/CR/Nodes/index.js
```diff
@@ -153,6 +155,15 @@ const hide = createAction(HIDE, contextPath => contextPath);
...
 const show = createAction(SHOW, contextPath => contextPath);
 
+/**
+ * Update uris of all affected nodes after uriPathSegment of a node has changed
+ * Must update the node itself and all of its descendants
+ *
+ * @param {String} oldUri
+ * @param {String} newUri
+ */
+const updateUri = createAction(UPDATE_URI, (oldUri, newUri) => ({oldUri, newUri}));
```
skurfuerst: can we rename this to "oldUriFragment" and "newUriFragment" for consistency? :)

packages/neos-ui-redux-store/src/CR/Nodes/index.js
dimaip: yep, will do so

dimaip:
@skurfuerst thanks for the review, fixed!

skurfuerst:
looks great :)

PR #915, opened 2017-09-08
FEATURE: Relogin on authentication error

### LINKED ISSUE #901: Show Re-Login Dialog if Session Timed Out

## Concept idea

wrap all endpoints internally in neos-ui-backend-connector such that once a request fails due to authnetication problems, we re-queue this request and all following requests to a queue.

Furthermore, we dispatch an action which in turn triggers the re-login dialog.

After re-login is successful, we will execute all requests serially.

## Related

I'd also add a callback if an error occured (in the same way); to display a message then (see #841)

skurfuerst:
Resolves #901 

To test, run `rm -Rf Data/Temporary/Development/Cache/Data/Flow_Session_*` while having the UI open

### DIFF
--- a/packages/neos-ui-backend-connector/src/Endpoints/index.js
+++ b/packages/neos-ui-backend-connector/src/Endpoints/index.js
@@ -1,8 +1,10 @@
 import {urlWithParams, searchParams} from './Helpers';
 
-const fetchJson = (endpoint, options) => fetch(endpoint, options).then(res => res.json());
+import fetchWithErrorHandling from '../FetchWithErrorHandling/index';
 
-const change = csrfToken => changes => fetchJson('/neos!/service/change', {
+const change = changes => fetchWithErrorHandling.withCsrfToken(csrfToken => ({
+    url: '/neos!/service/change',
+
     method: 'POST',
     credentials: 'include',
     headers: {
@@ -12,9 +14,10 @@ const change = csrfToken => changes => fetchJson('/neos!/service/change', {
     body: JSON.stringify({
         changes
     })
-});
+})).then(response => response.json());
 
-const publish = csrfToken => (nodeContextPaths, targetWorkspaceName) => fetchJson('/neos!/service/publish', {
+const publish = (nodeContextPaths, targetWorkspaceName) => fetchWithErrorHandling.withCsrfToken(csrfToken => ({
+    url: '/neos!/service/publish',
     method: 'POST',
     credentials: 'include',
     headers: {
@@ -25,9 +28,11 @@ const publish = csrfToken => (nodeContextPaths, targetWorkspaceName) => fetchJso
         nodeContextPaths,
         targetWorkspaceName
     })
-});
+})).then(response => response.json());
 
-const discard = csrfToken => nodeContextPaths => fetchJson('/neos!/service/discard', {
+const discard = nodeContextPaths => fetchWithErrorHandling.withCsrfToken(csrfToken => ({
+    url: '/neos!/service/discard',
+
     method: 'POST',
     credentials: 'include',
     headers: {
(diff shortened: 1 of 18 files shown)


packages/neos-ui/src/Containers/Modals/ReloginDialog/index.js
```diff
@@ -0,0 +1,110 @@
...
+        return (
+            <Dialog
+                title={<I18n id="Neos.Neos:Main:login.expired" fallback="Your login has expired. Please log in again."/>}
+                onRequestClose={emptyFn}
+                contentsClassName={style.dialog}
+                isOpen
+                >
+                <div className={style.modalContents}>
+                    <TextInput className={style.inputField} value={this.state.username} placeholder={i18nRegistry.translate('Neos.Neos:Main:username', 'Username')} onChange={this.handleUsernameChange}/>
+                    <TextInput className={style.inputField} value={this.state.password} placeholder={i18nRegistry.translate('Neos.Neos:Main:password', 'Password')} onChange={this.handlePasswordChange}/>
```
dimaip: uhm type=password is missing, so the password would be visible to anyone around...

packages/neos-ui/src/Containers/Modals/ReloginDialog/style.css
```diff
@@ -0,0 +1,12 @@
+.dialog {
+    max-width: 400px;
```
dimaip: I think it would be better to add smaller size of modal to Dialog component as a prop. We might need those smaller components in the future, and it's better to be consistent.

packages/neos-ui-backend-connector/src/Endpoints/index.js
```diff
@@ -159,57 +174,80 @@ const parseGetSingleNodeResult = requestPromise => {
...
+    data.set('__authentication[Neos][Flow][Security][Authentication][Token][UsernamePassword][username]', username);
+    data.set('__authentication[Neos][Flow][Security][Authentication][Token][UsernamePassword][password]', password);
+    // here, we
+    return fetch('/neos/login.json', {
+        method: 'POST',
+        body: data,
+        credentials: 'same-origin'
+    })
+    // parse the JSON if possible ...
+    .then(response => response.json())
```
dimaip: but won't it explode if json can't be parsed? Are you sure we don't need try/catch here?

packages/neos-ui-backend-connector/src/FetchWithErrorHandling/index.js
```diff
@@ -0,0 +1,133 @@
...
+     * FETCH example:
+     *
+     *     fn = csrfToken => (changes) => fetch('/your/url/here', {... fetch options here, use csrfToken ...})
+     *
+     * EXAMPLE of fetchWithCsrfTokenAndErrorHandling (rewrite your code like this):
+     *
+     *     fn = fetchWithCsrfTokenAndErrorHandling((changes) => csrfToken => ({
+     *       url: '/your/url/here',
+     *       ... fetch options here ...
+     *       ... use csrfToken properly...
```
dimaip: this comment is not clear

dimaip:
And yes, I actually tested the change, it works flawlessly!

skurfuerst:
Thanks for your comments, haha good catch with the clear text password :) :) will push follow up at latest on Monday morning!

packages/neos-ui-backend-connector/src/Endpoints/index.js
skurfuerst: no, if json can't be parsed an exception will be raised, which will lead to the promise failure -- which is correct in this case :)

dimaip:
I'm merging this since I need it for my work. If some problems arise just send a follow-up.

Issue #914, opened 2017-09-08
The title edit field on document creation writes right-to-left


daniellienert:
Tested on Mac in Chrome and Firefox.
Other input fields are working correct.

![strange title edit behavior](https://user-images.githubusercontent.com/642226/30211507-909a6014-94a1-11e7-9f52-60c0a194f82b.gif)

skurfuerst:
can reproduce it, will check

dimaip:
I fixed it, please check.

PR #924, opened 2017-09-09
BUGFIX: trash DisplayName component as it caused creationDialog to re-render and loose focus

dimaip:
Fixes: #914


skurfuerst:
@dimaip did you try to make the DisplayName component a PureComponent? That should also help I think. I kinda dislike the ID selectors all over the place (as IDs are only supposed to be on the page once etc...)

What do you think? :)

skurfuerst:
ah nevermind my comment!

PR #930, opened 2017-09-09
BUGFIX: don't scroll the tree unless focused node has changed

dimaip:
Without this fix, if you clicked on expanding a tree node, the tree would rewind to the focused node.


packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js
```diff
@@ -66,17 +66,34 @@ export default class Node extends PureComponent {
     constructor(props) {
         super(props);
 
+        this.state = {
+            shoulScrollIntoView: false
```
daniellienert: there is a 'd' missing in the variable

dimaip:
@daniellienert thanks, good catch!

Issue #791, opened 2017-08-02
Discard a newly created document node leads to errors


samuelhauser:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

Discarding a newly created document node, which never has been published, leads to errors. This is nothing new (comes from neos package), but we should handel this and redirect for example to the parent node.

dimaip:
Can confirm. Related to #794

PR #931, opened 2017-09-09
BUGFIX: misc bugs related to discarding and removing nodes

dimaip:
Fixes: #794
Fixes: #791 

- Move intialization of all serverFeedbackHandlers into one place, so they
don't override each other by accident, fixes redirect on node removal
- Allow discarding of removed nodes
- Correctly remove or re-create discarded nodes from UI

TODO:
- [x] write tests


dimaip:
Fixed one more bug along the way: not being able to discard deleted nodes.

PR #941, opened 2017-09-11
BUGFIX: escape space in package identifier

klfman:
Shells will otherwise treat the arguments separately


klfman:
Hey, you're welcome!
This is my first open source project, feels good to be around and to lend a hand!

Issue #939, opened 2017-09-11
When creating a new content node, it should get automatically focused


dimaip:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

PR #942, opened 2017-09-11
BUGFIX: focus content node on creation

dimaip:
Fixes: #939

### DIFF
--- a/packages/neos-ui/src/manifest.js
+++ b/packages/neos-ui/src/manifest.js
@@ -270,6 +270,7 @@ manifest('main', {}, globalRegistry => {
         const contentElement = (new DOMParser())
             .parseFromString(renderedContent, 'text/html')
             .querySelector(`[data-__neos-node-contextpath="${contextPath}"]`);
+        const fusionPath = contentElement.dataset.__neosFusionPath;
 
         if (!contentElement) {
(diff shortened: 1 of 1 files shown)


dimaip:
@mstruebing that's a good point, I've thought about it too. In the old UI it does do that. It's kind of hard to implement. If a node has multiple editables, which one should it focus?
In any case I propose to do that in an additional PR, maybe after 1.0. I'm trying hard not to go over-budget with all of these tiny bugfixes, so if it works the way as it did in the old UI IMO it's already good enough for now.

packages/neos-ui/src/manifest.js
```diff
@@ -270,6 +270,7 @@ manifest('main', {}, globalRegistry => {
         const contentElement = (new DOMParser())
             .parseFromString(renderedContent, 'text/html')
             .querySelector(`[data-__neos-node-contextpath="${contextPath}"]`);
+        const fusionPath = contentElement.dataset.__neosFusionPath;
```
skurfuerst: contentElement might be NULL, see check 3 lines below. So move this after this ;)

dimaip:
@skurfuerst Oh gosh, good catch! Fixed.

dimaip:
Merging, since the only review comment is now fixed.

Issue #877, opened 2017-09-05
Cleanup dependencies


dimaip:
If somebody has some spare time, would be nice to resolve these warnings at some point:

![image](https://user-images.githubusercontent.com/837032/30053155-7c88377c-9230-11e7-87a5-a2b8bea6fc7a.png)

markusguenther:
We have one warning actually regarding the react-codemirror and this could not be resolved. There is actually no new version. Otherwise I need to adjust this package and trigger Jet Wadson.

markusguenther:
Created a PR for react-css-themr to be react 16 compatible. Then this warning will be also gone :)

https://github.com/javivelasco/react-css-themr/pull/88

dimaip:
This issues is outdated

markusguenther:
@Dmitri i guess the react-collapse warning is still there

PR #946, opened 2017-09-12
TASK: Cleanup dependencies

mstruebing:
Remove codeclimate-test-reporter and nyc as it is not used until now
Remove Codecoverage and Codeclimate report from Readme
Remove lodash.isArray for Array.isArray
Remove lodash.compose for lodash.flowright

related #877 

It doesn't include all things mentioned from yarn but therefore it is not to big :)


dimaip:
Thanks a lot for taking care!

PR #951, opened 2017-09-12
TASK: Improve NodeInfoView

dfeyer:
This change add a line break between the title and the content and
display the current Node Type of the selected node.

### DIFF
--- a/packages/neos-ui-views/src/NodeInfoView/index.js
+++ b/packages/neos-ui-views/src/NodeInfoView/index.js
@@ -35,13 +35,34 @@ export default class NodeInfoView extends PureComponent {
             name: $get('properties._name', node) ? $get('properties._name', node) : '/'
         };
 
+        const nodeType = $get('nodeType', node);
+
         return (
             <ul className={style.nodeInfoView}>
-                <li><span>{i18nRegistry.translate('created', 'Created', {}, 'Neos.Neos')}</span> {new Date(properties.created).toLocaleString()}</li>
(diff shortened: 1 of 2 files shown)


dfeyer:
The rendering can be extract to a simple component, but that can be done in a separate PR

packages/neos-ui-views/src/NodeInfoView/style.css
```diff
@@ -4,21 +4,22 @@
...
-            float: left;
-            padding-right: 8px;
-            font-weight: bold;
-        }
+    &__title {
+        font-weight: bold;
+    }
+
+    &__content {
+        font-size: 13px;
```
dfeyer: A bit smaller font size, not used currently in the UI, but I think it make sense here to avoid line break in UUID.

dfeyer:
In this case, maybe event the "Metadata" headline does not make sense, and can be skipped, but not sure if we can do that currently

packages/neos-ui-views/src/NodeInfoView/style.css
dimaip: Hmm I guess we can merge it as an exception, but in general I'd be cautious about ad-hoc adding of new font-sizes like this, without touching the stylguide and react-ui-components.

packages/neos-ui-views/src/NodeInfoView/index.js
```diff
@@ -35,13 +35,34 @@ export default class NodeInfoView extends PureComponent {
             name: $get('properties._name', node) ? $get('properties._name', node) : '/'
         };
 
+        const nodeType = $get('nodeType', node);
```
dimaip: Better add this variable into properties object above, to make it more consistent

dimaip:
Nice, thanks! `nodeType` is really helpful actually

Issue #834, opened 2017-08-24
Editables with span tagName behaves strangely when pasting content


dimaip:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

When pasting content inside span editable it doesn't get saved:

![peek 2017-08-24 15-42](https://user-images.githubusercontent.com/837032/29666897-fbc2ea98-88e2-11e7-971f-5758c613bde5.gif)

dimaip:
@grebaldi this would be the area probably that you know best, right?

dimaip:
Actually such editables are completely unusable. Remove all content from it, and then the editable is gone. Unfocus it and you'd see the placeholder. Focus it again and it's gone again.

![peek 2017-09-06 17-02](https://user-images.githubusercontent.com/837032/30116056-28867ee0-9325-11e7-87d8-a288b09e3866.gif)

skurfuerst:
OK, have found a way to fix the placeholder handling. Copy/Paste coming :)

will all push tomorrow

dimaip:
You are a CKE magician! ;)
Also I noticed that CKE converts `«french quotes»` and other Unicode characters into html entities, which causes them to render as `&laquo;french quotes&raquo;`, if you don't use f:format.raw() in Fluid.

PR #959, opened 2017-09-13
FEATURE: Ckeditor inline element support

skurfuerst:
Fixes: #834


dimaip:
Just a thought: maybe at some point we should host our own version of CKE on npm, since they are too lazy to do that. Putting its source into our repo feels nasty.

Will review the rest now.

Issue #866, opened 2017-09-04
Content and Document tree should scroll when moving nodes through drag/drop


skurfuerst:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

When moving nodes in the document and content trees, you cannot move more than what you see as it is not possible to scroll the tree. We should implement an active area at the top/bottom which triggers scrolling appropriately.

This is needed by one of the core feature sponsors.

Issue #918, opened 2017-09-08
Moving a content node is not reflected in the structure tree


daniellienert:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

When the node is moved the change is not shown.
Saving indicator right top corner is showing "saving" - when the page is reloaded it shows the moved state.

Tested with #afc631c on Chrome / Mac

![moving node is not shown in strcuture tree](https://user-images.githubusercontent.com/642226/30216151-5428bdae-94b2-11e7-9c57-2932b105d188.gif)

dimaip:
Can not reproduce :(

dimaip:
Actually I tried in Firefox, and there drag-n-drop doesn't work altogether... I plan to re-implement dnd without useing html5 dnd api next week.

skurfuerst:
Might be related to / fixed by #969

dimaip:
@daniellienert can you check if you still have this problem on master, with the new tree? I will close this for now, feel free to re-open if you still have this issue.

Issue #830, opened 2017-08-21
Move nodes without prompting for position


dimaip:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)** 

When moving nodes now you are presented with a position selection dialog. I prefer how the move used to work in the old ui. Now every move requires a bunch of extra clicks and it's not easy to grasp the concept at first. Wdyt?

dimaip:
Feedback from Daniel: "I noticed the same, and share your opinion. It’s more clear what is happening im the one hand but a lot more clicks on the other. Makes a quick resorting of the tree impossible."

dimaip:
I would suggest to target this for 1.0, but only if we have time left. Still this is a big UX issue IMO.
@skurfuerst your thoughts?

daniellienert:
@dimaip / @skurfuerst - after working with the UI as an author for some days now, I would like to push this issue again. Reordering nodes or moving them into container nodes is really painful now.

dimaip:
Yep, we agreed to re-implement tree DND properly. Will take care.

dimaip:
@daniellienert reviews are welcome

PR #961, opened 2017-09-14
TASK: re-implement the tree DND

dimaip:
Fixes: #866 #830 #918

Re-implement the tree dnd with the help of react-dnd.
Allow moving nodes without prompting for position.

TODO:

- [ ] Optimistic updates or at least feedback on node move (the same is needed for delete and all other operations actually, so can be done as a separate PR)


packages/neos-ui/src/Sagas/CR/NodeOperations/moveDroppedNode.js
```diff
@@ -1,39 +1,21 @@
...
-
+export default function * moveDroppedNode() {
     yield * takeEvery(actionTypes.CR.Nodes.MOVE, function * handleNodeMove({payload}) {
-        const {nodeToBeMoved: subject, targetNode: reference} = payload;
-        const canBeInsertedAlongside = yield select(canBeInsertedAlongsideSelector, {subject, reference});
-        const canBeInsertedInto = yield select(canBeInsertedIntoSelector, {subject, reference});
+        const {nodeToBeMoved: subject, targetNode: reference, position} = payload;
 
-        const mode = yield call(
-            determineInsertMode,
```
skurfuerst: did we remove everything correctly related to the old behavior? IMHO "determineInsertMode" is not used anymore, is it? :) And the popup displaying the position selector can be removed as well, right?

dimaip: @skurfuerst it is used when (copy|cut)/pasting via the tree toolbar buttons

dimaip: https://github.com/neos/neos-ui/blob/32f82d6a9bf580d52a234cbc84f283b9207b745b/packages/neos-ui/src/Sagas/CR/NodeOperations/cutAndPasteNode.js#L39

skurfuerst: makes sense, thx <3

Issue #966, opened 2017-09-15
Tracking ticket for issues from punkt.de project (Core Feature Sponsor)


skurfuerst:

skurfuerst:
Bugfix: https://github.com/neos/neos-development-collection/pull/1721

skurfuerst:
Bugfix: https://github.com/neos/neos-ui/pull/1142

skurfuerst:
I'm closing this ticket now, as we have concluded the biggest part of the funded React rewrite. Thanks everybody :)

PR #969, opened 2017-09-15
BUGFIX: make decodeLabel() in node tree more robust

skurfuerst:
Previously, if the label was not given, the decode function
crashed with a NULL value.


skurfuerst:
tracked inside #966

PR #970, opened 2017-09-15
TASK: Adjust proptype validation for ReferenceEditor

johannessteu:
The nodetpye option in ReferenceEditor props can be a strig with a
single nodetype as well as an array


johannessteu:
Ah, should be fixed in ReferencesEditor as well, right?

PR #972, opened 2017-09-15
BUGFIX: Set correct icon in Reference(s)Editor

johannessteu:
Fixes: #973

### DIFF
--- a/packages/neos-ui-editors/src/Reference/index.js
+++ b/packages/neos-ui-editors/src/Reference/index.js
@@ -8,7 +8,8 @@ import {neos} from '@neos-project/neos-ui-decorators';
 
 @neos(globalRegistry => ({
     i18nRegistry: globalRegistry.get('i18n'),
-    nodeLookupDataLoader: globalRegistry.get('dataLoaders').get('NodeLookup')
+    nodeLookupDataLoader: globalRegistry.get('dataLoaders').get('NodeLookup'),
+    nodeTypeRegistry: globalRegistry.get('@neos-project/neos-ui-contentrepository')
 }))
 @connect($transform({
     contextForNodeLinking: selectors.UI.NodeLinking.contextForNodeLinking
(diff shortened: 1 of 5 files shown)


packages/neos-ui-editors/src/Reference/index.js
```diff
@@ -57,6 +61,11 @@ export default class ReferenceEditor extends PureComponent {
             this.setState({isLoading: true});
             this.props.nodeLookupDataLoader.resolveValue(this.getDataLoaderOptions(), this.props.value)
                 .then(options => {
+                    options.map(option => {
+                        option.icon = this.props.nodeTypeRegistry.getNodeType(option.nodeType).ui.icon;
```
dimaip: I'd first get a nodetype and then get it via `$get('ui.icon', nodeType)`. Otherwise you'd need to check if such properties exists.

packages/neos-ui-editors/src/Reference/index.js
```diff
@@ -81,6 +90,11 @@ export default class ReferenceEditor extends PureComponent {
             this.setState({isLoading: true, searchOptions: []});
             this.props.nodeLookupDataLoader.search(this.getDataLoaderOptions(), searchTerm)
                 .then(searchOptions => {
+                    searchOptions.map(option => {
+                        option.icon = this.props.nodeTypeRegistry.getNodeType(option.nodeType).ui.icon;
```
dimaip: same

packages/neos-ui-editors/src/Reference/index.js
dimaip: Wait, but how does it acutally work? You do `.map`, but don't assign it to anything.

johannessteu:
Looks like this now: 
bildschirmfoto 2017-09-15 um 17 11 27

bildschirmfoto 2017-09-15 um 17 11 35

PR #975, opened 2017-09-15
BUGFIX: replace history instead of pushing new entry

dimaip:
History is written on ifame url changes, so setting it again would cause
duplicate entries.

My previous attempt to fix it removed the url handling altogether...

### DIFF
--- a/packages/neos-ui/src/Sagas/Browser/index.js
+++ b/packages/neos-ui/src/Sagas/Browser/index.js
@@ -6,7 +6,7 @@ function * watchContentURIChange() {
     yield * takeEvery(actionTypes.UI.ContentCanvas.SET_CONTEXT_PATH, function * reflectChangeInAddressBar(action) {
         const {contextPath} = action.payload;
 
-        yield history.pushState({}, '', `?node=${contextPath}`);
+        yield history.replaceState({}, '', `?node=${contextPath}`);
     });
 }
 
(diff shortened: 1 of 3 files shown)


packages/neos-ui/src/Sagas/Browser/index.js
```diff
@@ -0,0 +1,15 @@
+import {takeEvery} from 'redux-saga';
+
+import {actionTypes} from '@neos-project/neos-ui-redux-store';
+
+function * watchContentURIChange() {
+    yield * takeEvery(actionTypes.UI.ContentCanvas.SET_CONTEXT_PATH, function * reflectChangeInAddressBar(action) {
+        const {contextPath} = action.payload;
+
+        yield history.replaceState({}, '', `?node=${contextPath}`);
```
Inkdpixels: We might want to use the [`call`](https://redux-saga.js.org/docs/api/#callfn-args) API from redux saga here to enhance the testability, wdyt @dimaip?

dimaip: Agree, if somebody was into writing tests for this he could change it. But since this saga is just 2 lines of code with one execution path I had no intentions to tests it, rather fix an obvious bug with the least amount of changes :)

dimaip: Btw, I get super excited when I see comments from you in this repo, good to see you back ;)

dimaip: Ah, and I didn't write that code, I just reverted the commit I deleted before with one small change.

PR #976, opened 2017-09-15
TASK: Change the aspect radio drop down to a more meaningful label

johannessteu:


packages/neos-ui-editors/src/Image/Components/Secondary/ImageCropper/index.js
```diff
@@ -178,7 +178,7 @@ export default class ImageCropper extends PureComponent {
                     </div>
 
                     <AspectRatioDropDown
-                        placeholder="Aspect Ratio"
+                        placeholder="Choose a Aspect Ratio"
```
dimaip: "an A"

dimaip: Wait, but actually why don't we have this stuff translated??

johannessteu: In the old ui it is not translated as well >.> I'll add a label in Neos.Neos.Ui

Issue #948, opened 2017-09-12
Image cropping frame hardly visible 


cwurditsch:
The cropping frame is hardly visible if the image has white background or very light
![bildschirmfoto 2017-09-12 um 11 55 27](https://user-images.githubusercontent.com/31848464/30319861-4f376d20-97b1-11e7-9f86-6c6d85a33af0.png)

PR #977, opened 2017-09-15
TASK: Make the crop border more visible

johannessteu:
#948


johannessteu:
Changed the border from 1px to 2px. This is how it looks with this change applied on a white and black background

bildschirmfoto 2017-09-15 um 20 54 59
bildschirmfoto 2017-09-15 um 20 54 20

PR #978, opened 2017-09-15
TASK: Update the image on upload as well in the CropEditor if it is open

johannessteu:


johannessteu:
That's how it behaves now:
http://recordit.co/1ES2DthbEH

packages/neos-ui-editors/src/Image/index.js
```diff
@@ -213,12 +214,18 @@ export default class ImageEditor extends Component {
     upload(files) {
         const {uploadAsset} = backend.get().endpoints;
         const {commit, siteNodePath} = this.props;
+        const {isImageCropperOpen} = this.state;
 
         const siteNodeName = siteNodePath.match(/\/sites\/([^/@]*)/)[1];
 
         return uploadAsset(files[0], siteNodeName).then(res => {
             this.setState({image: res});
```
Inkdpixels: Cool change @johannessteu! One little nitpick I've seen in this file which we could solve all together in this PR if you don't mind is that we should change the `setState` call here as well as on [L239](https://github.com/neos/neos-ui/pull/978/files#diff-47fb3536fb7a6b6d27f3cdc7976a4644R239) to reflect the asynchronous nature of it, e.g. move all of the following code into a callback as seen [here](https://github.com/johannessteu/neos-ui/blob/0564faa55b438fe2b837c43ecbea3adbe7110454/packages/neos-ui-editors/src/Image/index.js#L174).

Otherwise we might run into problems if the code gets extended/more complex :D :zap:

johannessteu: Thanks for your comment @Inkdpixels, i'm still getting into it :)
Should be fixed now!

Inkdpixels: @johannessteu No worries, I'm no better :)

PR #980, opened 2017-09-15
FEATURE: Set background color for the content canvas

### LINKED ISSUE #863: Make Background Color in ContentCanvas configurable

Currently, the Content Canvas flickers to "white" when a page load occurs. We should be able to add a background color in configuration to prevent this.

Although this is actually "minor", it is needed by one of the Core Feature Sponsors.

johannessteu:
This feature introduces a new Setting `Neos.Neos.Ui.configuration.contentCanvas.backgroundColor`. It can be set to any valid string for the css property `background` and defaults to a white color. With this setting set the background of the content canvas can be set to prevent it flickering on a page load.

Resolves #863

### DIFF
--- a/Configuration/Settings.yaml
+++ b/Configuration/Settings.yaml
@@ -35,6 +35,9 @@ Neos:
             resource: '${"resource://" + Neos.Ui.StaticResources.compiledResourcePackage() + "/Public/Styles/HostOnlyStyles.css"}'
             position: 'start 900'
 
+      contentCanvas:
+        backgroundColor: '#ffffff'
+
       #################################
       # INTERNAL CONFIG (no API)
       #################################
@@ -86,6 +89,7 @@ Neos:
             contentCanvas:
               src: '${Neos.Ui.NodeInfo.uri(documentNode, controllerContext)}'
               contextPath: '${q(documentNode).property(''_contextPath'')}'
+              backgroundColor: '${Configuration.setting(''Neos.Neos.Ui.contentCanvas.backgroundColor'')}'
             debugMode: false
             editPreviewMode: '${q(user).property("preferences.preferences")["contentEditing.editPreviewMode"] || Configuration.setting(''Neos.Neos.userInterface.defaultEditPreviewMode'')}'
             fullScreen:
--- a/Resources/Private/Fusion/Backend/Root.fusion
+++ b/Resources/Private/Fusion/Backend/Root.fusion
@@ -14,7 +14,6 @@ backend = Neos.Fusion:Template {
         nodeTree = ${Configuration.setting('Neos.Neos.userInterface.navigateComponent.nodeTree')}
         structureTree = ${Configuration.setting('Neos.Neos.userInterface.navigateComponent.structureTree')}
         allowedTargetWorkspaces = ${Neos.Ui.Workspace.getAllowedTargetWorkspaces()}
-        contentCanvas = ${Configuration.setting('Neos.Neos.Ui.contentCanvas')}
         endpoints = Neos.Fusion:RawArray {
             nodeTypeSchema = Neos.Fusion:UriBuilder {
                 package = 'Neos.Neos'
--- a/Resources/Private/Fusion/Backend/Root.fusion
+++ b/Resources/Private/Fusion/Backend/Root.fusion
@@ -14,6 +14,7 @@ backend = Neos.Fusion:Template {
         nodeTree = ${Configuration.setting('Neos.Neos.userInterface.navigateComponent.nodeTree')}
         structureTree = ${Configuration.setting('Neos.Neos.userInterface.navigateComponent.structureTree')}
         allowedTargetWorkspaces = ${Neos.Ui.Workspace.getAllowedTargetWorkspaces()}
+        contentCanvas = ${Configuration.setting('Neos.Neos.Ui.contentCanvas')}
         endpoints = Neos.Fusion:RawArray {
             nodeTypeSchema = Neos.Fusion:UriBuilder {
                 package = 'Neos.Neos'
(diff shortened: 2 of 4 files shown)


Configuration/Settings.yaml
```diff
@@ -86,6 +86,7 @@ Neos:
             contentCanvas:
               src: '${Neos.Ui.NodeInfo.uri(documentNode, controllerContext)}'
               contextPath: '${q(documentNode).property(''_contextPath'')}'
+              backgroundColor: '#ffffff'
```
dimaip: Hey, we keep only those things that will be changing on runtime inside of state. I think this should go to some other path in settings, that should be exposed here https://github.com/neos/neos-ui/blob/master/Resources/Private/Fusion/Backend/Root.fusion#L13

skurfuerst: ack, and furthermore the "*state" part of the settings is private (e.g. only for unplanned extensibility).

So really cool feature, just we need to move the settings key :)

johannessteu: working on it 👍

johannessteu:
Like this @dimaip  @skurfuerst ?

Resources/Private/Fusion/Backend/Root.fusion
```diff
@@ -14,6 +14,7 @@ backend = Neos.Fusion:Template {
         nodeTree = ${Configuration.setting('Neos.Neos.userInterface.navigateComponent.nodeTree')}
         structureTree = ${Configuration.setting('Neos.Neos.userInterface.navigateComponent.structureTree')}
         allowedTargetWorkspaces = ${Neos.Ui.Workspace.getAllowedTargetWorkspaces()}
+        contentCanvas = ${Configuration.setting('Neos.Neos.Ui.configuration.contentCanvas')}
```
dimaip: You could as well expose the whole configuration part, to not have to do it every time. Just need to think a good naming for it ;)

johannessteu: What about `uiConfiguration` ?

skurfuerst:
Hey everybody,

thanks for the change :-)

I actually thought though this some more; with the following results:

* I think the `configuration` key is debatable; maybe we should just remove this level altogether? I just checked the "old" UI - there, it would be `Neos.Neos.userInterface.contentCanvas` (if we were to place it) - and "userInterface" is already in the package name. So, my suggestion would be: `Neos.Neos.Ui.contentCanvas.backgroundColor`.

* I know @dimaip has argued about removing the config from the Redux state and put it into the `neos` part of the UI.... However, I for me this background color can *very well be* part of the Redux state; in the same manner as we put the current user infos in there (which does not change as of now as well). I think it has even a reduced complexity, as we do not need to introduce the concept of "immutable configuration"...

Thus, as a solution:

* I'd use the config key `Neos.Neos.Ui.contentCanvas.backgroundColor` as API to the user.
* I'd reference this config key in `Neos.Neos.Ui.state.backend.ui.contentCanvas.backgroundColor` using the `Configuration` Eel helper

This way, we have a public API and do not need to introduce a new concept for this config; and it would be even modifiable by extensions if people want to do that.

Fine for you? :-)

All the best,
Sebastian

johannessteu:
Sound's great! Will adjust later. Or, @dimaip ?

dimaip:
@skurfuerst 
- I'm 100% with you on the first point, thought about it myself, just didn't dare to suggest it as it would require bigger refactoring.
- Not convinced, really. I think it makes total sense to separate runtime changeable parts, from static configuration, we'll need that concept anyways for other things. Doing so reduces cognitive complexity when reasoning about the UI. Our state is huge enough as is, I would love to keep it as lean as possible to easy the debugging.

So my proposal:
1. Expose the whole configuration under `Neos.Neos.Ui` to the UI. Use it where needed.
2. Put this setting under `Neos.Neos.Ui.contentCanvas.backgroundColor`.

skurfuerst:
Hey @dimaip 

Interesting discussion :-)

1) Please, no.... I think this makes configuration more "immutable" than it needs to be. We currently do not expose any part of this configuration directly to the UI; and I think we live fine by it :-)

I am currently missing any "smoking gun" where we need this immutable configuration in the UI. Why should we introduce it now?

All the best,
Sebastian :-)

dimaip:
Hehe I'm thinking about it exactly the same way, but from the perspective that it should be proved that we need configuration to be mutable, rather than immutable.
My reasoning is simple: if you can segregate mutable stuff from immutable it lowers the complexity of the system, but separating the moving part, making it easier to reason about.
But OK, I'm not opinionated on it _too much_, so go ahead and handle it the way that makes most sense to you, I trust your intuition here, even though it differs from mine ;)

skurfuerst:
Cool; I can also fully understand your reasoning. My main motivation is keeping the number of concepts and building blocks one needs to understand low; so that's why I would handle it this way – and from a UI perspective I do not see a difference between e.g. the current user's name (which is stored in Redux) and the the canvas background color...

So, also from my side, no really hard feelings as well ;-)
All the best,
Sebastian

johannessteu:
Alright, will update it later 🎉

Resources/Private/Fusion/Backend/Root.fusion
```diff
@@ -14,6 +14,7 @@ backend = Neos.Fusion:Template {
         nodeTree = ${Configuration.setting('Neos.Neos.userInterface.navigateComponent.nodeTree')}
         structureTree = ${Configuration.setting('Neos.Neos.userInterface.navigateComponent.structureTree')}
         allowedTargetWorkspaces = ${Neos.Ui.Workspace.getAllowedTargetWorkspaces()}
+        contentCanvas = ${Configuration.setting('Neos.Neos.Ui.contentCanvas')}
```
skurfuerst: I think this line can be removed, right? :)

johannessteu:
@skurfuerst yeah, was a leftover after refactoring it back to state thingy

PR #985, opened 2017-09-16
BUGFIX: Allow the use of multiple classes in icons

### LINKED ISSUE #950: Multiple class for icons are not supported
This is allowed in the current UI, but the react UI return an error
```yaml
icon: 'icon-folder taxonomy-icon-blue'
```

This configuration create a valid icon with two classes.

In the react UI, this is the error message:
```
Failed prop type: Icon name "icon-folder taxonomy-icon-blue" was not a found in Font-Awesome 4.5.
Please use the icon names from the Font-Awesome website.
```

mstruebing:
fixes #950

@dfeyer

### DIFF
--- a/packages/react-ui-components/src/_lib/fontAwesome.js
+++ b/packages/react-ui-components/src/_lib/fontAwesome.js
@@ -2,6 +2,12 @@ export function makeValidateId(icons = {}) {
     const ICON_NAMES = Object.keys(icons);
 
     return function validateId(id = '') {
+        let iconClassesArray = [id];
+
+        if (id.includes(' ')) {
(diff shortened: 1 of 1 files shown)


dimaip:
@mstruebing could you please describe the logic of the new behavior?

mstruebing:
Hey, yes of course!
It will go through the whole list it gets passed from the NodeType configuration icon and looks if there is **only one** icon that matches to an existing font-awesome icon and put the other classes behind that and returns a valid icon. 
If there are multiple (or none) font-awesome icons it returns with invalid.

So it would be possible to add additional classes to a font-awesome icon  like @dfeyer wanted

PR #988, opened 2017-09-16
TASK: Show highlight status on Editors if a property is changed

### LINKED ISSUE #987: Check all Editors for a highlight flag

johannessteu:
Fixes #987 and #989


johannessteu:
Still need to check some more editors, so this is kind of WIP

PR #992, opened 2017-09-18
FEATURE: Filter nodetypes in SelectNodeTypeDialog

johannessteu:
If you have a lot of content nodetypes it can be complex to find the correct nodetype in our new node dialog. Therefore this change adds a possibility to filter nodetypes.

### DIFF
--- a/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
+++ b/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
@@ -62,9 +62,8 @@ export default class SelectNodeType extends PureComponent {
     };
 
     componentDidMount() {
-        const {filterSearchTerm} = this.state || {};
         this.setState({
-            filterSearchTerm: filterSearchTerm || '',
+            filterSearchTerm: '',
             insertMode: calculateInitialMode(
                 this.props.allowedSiblingNodeTypes,
                 this.props.allowedChildNodeTypes
@@ -72,10 +71,19 @@ export default class SelectNodeType extends PureComponent {
         });
     }
 
+    updateInsertMode() {
+        this.setState({
+            insertMode: calculateInitialMode(
+                this.props.allowedSiblingNodeTypes,
+              this.props.allowedChildNodeTypes
+            )
+        });
+    }
+
     componentDidUpdate(prevProps) {
         if (prevProps.allowedSiblingNodeTypes !== this.props.allowedSiblingNodeTypes ||
             prevProps.allowedChildNodeTypes !== this.props.allowedChildNodeTypes) {
-            this.componentDidMount();
+            this.updateInsertMode();
         }
     }
 
--- a/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
+++ b/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
@@ -12,6 +12,8 @@ import I18n from '@neos-project/neos-ui-i18n';
 
 import {InsertModeSelector} from '@neos-project/neos-ui-containers';
 import NodeTypeGroupPanel from './nodeTypeGroupPanel';
+import NodeTypeFilter from './nodeTypeFilter';
+import style from './style.css';
 
 const calculateInitialMode = (allowedSiblingNodeTypes, allowedChildNodeTypes) => {
     if (allowedSiblingNodeTypes.length) {
@@ -61,6 +63,7 @@ export default class SelectNodeType extends PureComponent {
 
     componentDidMount() {
         this.setState({
+            filterSearchTerm: '',
             insertMode: calculateInitialMode(
                 this.props.allowedSiblingNodeTypes,
                 this.props.allowedChildNodeTypes
@@ -79,7 +82,9 @@ export default class SelectNodeType extends PureComponent {
 
     handleCancel = () => {
         const {cancel} = this.props;
-
+        this.setState({
+            filterSearchTerm: ''
+        });
         cancel();
     }
 
@@ -107,20 +112,6 @@ export default class SelectNodeType extends PureComponent {
         }
     }
 
-    renderInsertModeSelector() {
-        const {insertMode} = this.state;
-        const {allowedSiblingNodeTypes, allowedChildNodeTypes} = this.props;
-
-        return (
-            <InsertModeSelector
-                mode={insertMode}
-                onSelect={this.handleModeChange}
-                enableAlongsideModes={Boolean(allowedSiblingNodeTypes.length)}
-                enableIntoMode={Boolean(allowedChildNodeTypes.length)}
-                />
-        );
-    }
-
     renderCancelAction() {
         return (
             <Button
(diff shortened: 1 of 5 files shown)


johannessteu:
This is how it looks/works right now:
http://recordit.co/rHdruLD21k

packages/neos-ui/src/Containers/Modals/SelectNodeType/nodeTypeFilter.js
```diff
@@ -0,0 +1,34 @@
...
+    constructor(props) {
+        super(props);
+
+        this.handleValueChange = this.handleValueChange.bind(this);
+    }
+
+    render() {
+        return (
+            <div className={style.nodeTypeDialogHeader__filter}>
+                <span className={style.nodeTypeFilter__label}>Filter:</span>
```
dimaip: Shall we add a filter icon here?

packages/neos-ui/src/Containers/Modals/SelectNodeType/nodeTypeGroupPanel.js
```diff
@@ -40,10 +47,21 @@ class NodeTypeGroupPanel extends PureComponent {
...
+            onSelect,
+            filterSearchTerm,
+            i18nRegistry
         } = this.props;
         const {name, label, nodeTypes} = group;
 
+        const filteredNodeTypes = (nodeTypes || [])
+            .filter(nodeType => {
+                const label = i18nRegistry.translate(nodeType.label, nodeType.label);
+                if (label.toLowerCase().search(filterSearchTerm.toLowerCase()) !== -1) {
```
dimaip: Don't we need to escapre regexp here? There's method for this in lodash i think.

packages/neos-ui/src/Containers/Modals/SelectNodeType/style.css
```diff
@@ -28,3 +28,18 @@
...
     padding: 0;
 }
+
+.nodeTypeDialogHeader {
+    display: flex;
+    justify-content: space-between;
+}
+
+.nodeTypeDialogHeader__filter {
+    margin-right: 40px;
```
dimaip: Don't we have spacing vars for this?

packages/neos-ui/src/Containers/Modals/SelectNodeType/style.css
```diff
@@ -28,3 +28,18 @@
...
+    display: flex;
+    align-items: center;
+}
+
+.nodeTypeFilter__label {
+    margin-right: 10px;
```
dimaip: And for this?

packages/neos-ui/src/Containers/Modals/SelectNodeType/nodeTypeFilter.js
johannessteu: Yeah! Good idea!

packages/neos-ui/src/Containers/Modals/SelectNodeType/nodeTypeGroupPanel.js
johannessteu: i'll check

packages/neos-ui/src/Containers/Modals/SelectNodeType/style.css
johannessteu: Ah, not sure. Could you point me to the place where all vars are set? Did'n found that earlier

johannessteu:
@dimaip Thanks for your comment! Regarding the height: i also was thinking about that. I'll try both variants this evening. First i tried to disable the nodes that does not match the filter. But then it would still be to complex to find the node. I'll play with the height!

dimaip:
Yeah I'm not sure what actually would be better: to fix the height or to animate it. Try it out and choose the option that would be the least irritating.

johannessteu:
Yeah. The screen cast is also a bit laggy and the animation looks faster that it actually is. But i'll play around with some variants

dimaip:
I suspected that. In this case use your best judgement here. If you'd choose to add animations, they need to be very quick and responsive (<<0.5s), hate slow animation effects :)

packages/neos-ui/src/Containers/Modals/SelectNodeType/nodeTypeGroupPanel.js
johannessteu: Should be fixed now!

packages/neos-ui/src/Containers/Modals/SelectNodeType/nodeTypeFilter.js
johannessteu: bildschirmfoto 2017-09-18 um 10 43 18

Way better!

johannessteu:
@dimaip Regarding the animation: I'd sugges to keep it the way it is right now. Feels smooth&fast.
Just discovered a little bug, if you got a filtered set of nodetypes and collapse a group you kind of reset the filter/state. Will look into that later

packages/react-ui-components/package.json
```diff
@@ -36,6 +36,7 @@
     "lodash.debounce": "^4.0.8",
     "lodash.isfunction": "^3.0.8",
     "lodash.omit": "^4.3.0",
+    "lodash.escaperegexp": "^4.1.0",
```
dimaip: Might make sense to use the same version as here: https://github.com/neos/neos-ui/blob/master/package.json#L87

johannessteu:
Ok, fixed it. This looks a bit dirty now since i had to handle that componentDidUpdate triggers compnentDidMount. Btw @dimaip shouldn't we avoid that a lifecycle hook triggers another livecyclehook?

dimaip:
Ah, wait, missed you last commit.

johannessteu:
It's your part to merge ;)

packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
```diff
@@ -60,7 +62,9 @@ export default class SelectNodeType extends PureComponent {
     };
 
     componentDidMount() {
+        const {filterSearchTerm} = this.state || {};
         this.setState({
+            filterSearchTerm: filterSearchTerm || '',
```
dimaip: Wait, but why is it even needed? Can you explain?

johannessteu: Because we use it later on. On initial load it might be null wich is a problem with the regex check etc wich needs a string. So we keep here the value in filterSearchTerm or set it to an empty string. The keep thing will kick in on every `componentDidUpdate` hook, the empty string when   `const {filterSearchTerm} = this.state || {};` is falsy

johannessteu: The main problem here is that `componentDidUpadte` triggers `componentDidMount` and we need to keep the state as well as initialize the state

packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
```diff
@@ -61,25 +63,37 @@ export default class SelectNodeType extends PureComponent {
...
         this.setState({
+            filterSearchTerm: '',
             insertMode: calculateInitialMode(
                 this.props.allowedSiblingNodeTypes,
                 this.props.allowedChildNodeTypes
             )
         });
     }
 
+    updateInsertMode() {
```
dimaip: But wait, this method is only called from one place, why bother to create the abstraction?

packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
```diff
@@ -61,25 +63,37 @@ export default class SelectNodeType extends PureComponent {
 
     componentDidMount() {
         this.setState({
+            filterSearchTerm: '',
```
dimaip: As said, the default value for `filterSearchTerm` should be set in the constructor

packages/neos-ui/src/Containers/Modals/SelectNodeType/style.css
dimaip: https://github.com/neos/neos-ui/blob/master/packages/build-essentials/src/webpack.config.js#L59

johannessteu:
This one is now good to go 🙃

packages/neos-ui/src/Containers/Modals/SelectNodeType/nodeTypeFilter.js
```diff
@@ -0,0 +1,63 @@
...
+class NodeTypeFilter extends PureComponent {
+    static propTypes = {
+        onChange: PropTypes.func.isRequired,
+        i18nRegistry: PropTypes.object.isRequired
+    };
+
+    constructor(props) {
+        super(props);
+
+        this.state = {filterSearchTerm: ''};
```
dimaip: Why do you duplicate the `filterSearchTerm` state here? Just pass it down via the props...

PR #997, opened 2017-09-21
FEATURE: Add new Tooltip Component

### LINKED ISSUE #802: Tooltips for buttons

The old UI have tooltips for buttons (i.e. add 'Create new' for the button to create new content elements) and we are currently having none.

johannessteu:
Fixes #802


johannessteu:
This is the Rework of form me and @janinadanger of her first draft for a tooltip component. We now use an external library. This is a first draft we'd like to know if the direction on this change is good or not. 

What we do have:
- Tooltip component wich can render a single label or another component and bigger html structure
- Integrated in or IconButton
- slim api to control the styling and functions that the tooltip library offers, we do not expose all optiosn, just some basic for now

What is missing right now:
-[] general styling
-[] some special styling for error boxes inside the inspector pane to make them sticky and not just on hover visible
-[] Integration other components

First preview:
http://recordit.co/bZHkrAaQHu

packages/neos-ui/src/Containers/RightSideBar/style.css
```diff
@@ -2,7 +2,7 @@
     padding-top: 41px;
     padding-bottom: 41px;
     transition: .2s ease transform;
-    will-change: transform;
+    /* will-change: transform; */
```
johannessteu: This introduces a bug in Chrome. With this property applied position of fixed elements is buggy. A tooltip with a `left: 500px` will not be offset 500px left from the viewport but from the sidebar.

grebaldi:
Hey @johannessteu,

this looks awesome! Especially the intelligent positioning :)

Generally, I like the integration, there's just a suggestion I'd like to make:

It would be great to provide a higher order component that injects tooltip functionality into other arbitrary components, instead of hard-wiring it like in `packages/react-ui-components/src/IconButton/iconButton.js`. This way, we could keep concerns separated and also provide a convenient way to add tooltips to further components.

I am however not too sure, if this is actually feasible in a general way. What do you think?

johannessteu:
I'll try to apply the hoc-pattern to the iconbutton push it and ping again 👍

johannessteu:
Just talked with @grebaldi about the HoC pattern. Won't be easy to do in this case since we can not just simply wrap the child component. The component must render some `data-*` attr. So we'd leave it as it is. I'll also add the implementation for the IconComonent as well. So IconButton and Icon can be rendered with a tooltip. Should we ship more Compnents with a tooltip integration? I'd say we should be good with just IconButton and Icon.

packages/neos-ui-editors/src/TextArea/index.js
```diff
@@ -9,14 +9,16 @@ const defaultOptions = {
...
+    const {value, commit, highlight, options, validationErrors} = props;
 
     const finalOptions = Object.assign({}, defaultOptions, options);
 
-    return (<TextArea
+    return (<TextAreaWithValidationResultRendering
         value={value}
         onChange={commit}
         highlight={highlight}
+        invalid={validationErrors && validationErrors.length > 0}
```
skurfuerst: I'd call this property "isValid"; thinking about double-negated properties gives me some headaches ;)

packages/neos-ui/src/Containers/LeftSideBar/NodeTreeToolBar/Buttons/CutSelectedNode/index.js
```diff
@@ -20,9 +22,11 @@ export default class CutSelectedNode extends PureComponent {
 
     render() {
         const {className, isDisabled} = this.props;
+        const tooltipLabel = <I18n id="Neos.Neos:Main:cut" fallback="Copy"/>;
```
skurfuerst: wrong fallback

packages/neos-ui/src/Containers/LeftSideBar/NodeTreeToolBar/Buttons/DeleteSelectedNode/index.js
```diff
@@ -21,9 +22,11 @@ export default class DeleteSelectedNode extends PureComponent {
 
     render() {
         const {className, isDisabled} = this.props;
+        const tooltipLabel = <I18n id="Neos.Neos:Main:delete" fallback="Copy"/>;
```
skurfuerst: wrong fallback

packages/neos-ui/src/Containers/LeftSideBar/NodeTreeToolBar/Buttons/HideSelectedNode/index.js
```diff
@@ -29,9 +30,11 @@ export default class HideSelectedNode extends PureComponent {
 
     render() {
         const {className, isDisabled, isHidden} = this.props;
+        const tooltipLabel = <I18n id="Neos.Neos:Main:hideUnhide" fallback="Copy"/>;
```
skurfuerst: wrong fallback

packages/neos-ui/src/Containers/LeftSideBar/NodeTreeToolBar/Buttons/PasteClipBoardNode/index.js
```diff
@@ -20,13 +21,12 @@ export default class PasteClipBoardNode extends PureComponent {
     }
 
     render() {
-        const {
-            className,
-            canBePasted
-        } = this.props;
+        const {className, canBePasted} = this.props;
+        const tooltipLabel = <I18n id="Neos.Neos:Main:paste" fallback="Copy"/>;
```
skurfuerst: wrong fallback

packages/neos-ui/src/Containers/LeftSideBar/NodeTreeToolBar/Buttons/RefreshPageTree/index.js
```diff
@@ -26,9 +27,11 @@ export default class RefreshPageTree extends PureComponent {
             [style.spinning]: isLoading,
             [className]: className && className.length
         });
+        const tooltipLabel = <I18n id="Neos.Neos:Main:refresh" fallback="Copy"/>;
```
skurfuerst: wrong fallback

packages/neos-ui/src/Containers/RightSideBar/index.js
```diff
@@ -44,6 +44,7 @@ export default class RightSideBar extends PureComponent {
         const toggleIcon = isHidden ? 'chevron-left' : 'chevron-right';
         const toggle = isFullScreen ? null : (
             <IconButton
+                tooltipLabel="Toogle Sidebar"
```
skurfuerst: use I18N helper here :)

packages/react-ui-components/src/ToggablePanel/toggablePanel.js
```diff
@@ -231,6 +231,7 @@ export class Header extends PureComponent {
                     {children}
                 </HeadlineComponent>
                 <IconButtonComponent
+                    tooltipLabel="Toogle the panel"
```
skurfuerst: I think this one is actually hard to localize; as we do not have a dependency from react-ui-components to the i18n package... So I'd leave it as is for now.

packages/react-ui-components/src/_lib/withValidationResultRendering.js
```diff
@@ -0,0 +1,33 @@
...
+             * An array of error messages
+             */
+            validationErrors: PropTypes.array
+        }
+
+        render() {
+            const {validationErrors, ...rest} = this.props;
+
+            const renderedErros = validationErrors && validationErrors.length > 0 && validationErrors.map((validationError, key) => {
+                return <div key={key}>{validationError}</div>;
```
skurfuerst: shouldn't the validation error be rendered as a tooltip; or did I misunderstand your change? :)

johannessteu: Yeah, we did plan that but doesn't work out really good so far. Hard to get it really sticky, overwriting styles is also not an easy task right now. Also i had some problem with a setState that is triggered inside the ReactTolltip Component and moun/unounting if they get rerendered. So just for now we thought might be better to get in v1 the tooltip as a component in and implement it in other parts of the ui in a follow up. This change also has not yet all tooltips set, just a few for a demo right now

packages/neos-ui-editors/src/TextArea/index.js
johannessteu: Hm, the textarea as well as the textfield component already got the invalid prop, should i change those also?

packages/react-ui-components/src/_lib/withValidationResultRendering.js
skurfuerst: OK :)

packages/neos-ui-editors/src/TextArea/index.js
skurfuerst: ah, I'd change them. but no blocker :)

johannessteu:
styling is still missing

skurfuerst:
@johannessteu any news? :) No pressure, just asking :)

johannessteu:
sry, much on the list, but will try to finish it this week!

janinadanger:
I tried to use the logic from the following url to render the tooltip.
https://github.com/mcumpl/react-lightweight-tooltip/blob/master/src/components/tooltip.js

The element which triggers the tooltip get's a wrapping`<div style="position:relative">`. Inside this `<div>` react renders the hovered element and the tooltip container. Here the problem begins, because some elements have an specific position style. For example: an absolute positioned button become a new position in the layout because the parent `<div>` with the relative position has changed.

Another problem is that some parent nodes cut the tooltip because 'overflow: hidden' is set. 

Do you have any idea to fix this problems? Should we really use an absolute positioned tooltip?

dimaip:
I don't have anything useful to say on the topic, I guess I would just mangle with CSS till it works :see_no_evil:

janinadanger:
The important tooltips are integrated. Some tooltips still missing. 
But all in all it works fine :)

Needed to restructure css in some places.
tooltip

dfeyer:
https://github.com/wwayne/react-tooltip is not maintained did we want to have this kind of dependencies ?

johannessteu:
@dfeyer we create our won tooltip component

johannessteu:
The dep is a leftover, need to be removed

dfeyer:
Need to test this one, but no time this week

What happens when the tooltip is open and the user scroll, it's a major issue in the current UI and don't want to see that on the rewrite ?

johannessteu:
btw, not fully finished right now. @janinadanger is working on the validation result rendering. The tooltip is only open on hover. So if you hover, the tooltip appears, if you scroll down it will disappear.

markusguenther:
ezgif com-video-to-gif

johannessteu:
I don't think that's that find of an issue for not merging it right now. This might be setup as well when all z-indexes are redone

kitsunet:
re-opening to hopefully trigger travis...

johannessteu:
I'll try to manage today to refactor the pr to the refactored tests

Inkdpixels:
@johannessteu Let me know if you need any help, will probably leave around 1PM for my flight though.

johannessteu:
@Inkdpixels if you want to take it would be great, guess i won't make it before this evening

Inkdpixels:
@johannessteu Will merge the current master into your PR and will try adjust the tests, thanks! :heart:

Inkdpixels:
FYI @johannessteu I changed a few bits in the component itself to make it easier to test and less complex, here is what I did.

* Renamed the props to have less repetition when using the component standalone, e.g. rename `tooltipLabel` into `label` in case you want to use `<Tooltip/>` itself
* Removed the `componentDidMount` method usage since it can be simplified during the state initialization, this is important since it was calling a `setState` effectively which caused an immediately re-render after the component got rendered the first time
* Removed the `assignOutsideTouchHandler` method since as far as I can tell the `react-click-outside` HOC can be used for this usecase

I hope you are okay with these changes :)

dimaip:
This looks fine to me in general. I think we should just merge it as it's big and gets conflicts quickly, and then fix any other bugs we find later with follow-ups.

dimaip:
@johannessteu I rebased the change.
But the tooltips look really weird:
image
On however they stick around.
I think we should leave the custom tooltips only where there are absolutely necessary (e.g. valtidation), and use native tooltips everywhere else.
Agreed? I can take care of it.

johannessteu:
Guess it has to do with some changes wich are made after the sprint. There were no sticky tooltips at that point at least. But it's not my decision tough. @janinadanger did the work on this one (she's on vacation right now and i'll be able to do some work earliest next week).

dimaip:
I had a quick discussion with @johannessteu and we came up with the following plan.

1. Implement simple tooltips for buttons as native tooltips. That would improve performance, make sure there are no glitches and in general is enough for the usecase. I'll take care of it a separate PR which hopefully will be quickly merged.
2. Use custom Tooltip component for things like validation messages and help messages. We should take the implementation from the current PR, fix it, and make sure it works well.

Any objections?

kitsunet:
Why did this never make it? Anything can be done to get this finally in? It's ages old by now and it's a bid sad we didn't mange to merge it.

johannessteu:
@kitsunet yeah unfortunately. For me it was several times good to go. But the decision was made at some point to "just user the browser tooltips"

dimaip:
@johannessteu we agreed to use this custom component for complex things like validation messages and help messages, leaving out simpler labels for the sake of performance and stability.
I'd just extract the component code from here and replace the existing component.

mstruebing:
ping, what to do about this?

Issue #833, opened 2017-08-23
After changing inspector property, you are not forced to apply changes


dimaip:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

i.e. it's possible to navigate away to another node, and there's no dialog prompting you to apply changes.

skurfuerst:
That's something I am quite unsure about -- the old behavior always felt kinda wrong (but ignoring it seems wrong as well); so maybe we have a clever idea how to change/fix this in a good way?

dimaip:
I think the way to improve it would be to have auto-saving inspector editors, but that's definitely not smth for 1.0 and would require a lot of UX considerations. For now I propose to implement it just the way it was, should be easy.

skurfuerst:
Totally agreed :)

PR #998, opened 2017-09-22
FEATURE: Prompt for decision if there are unapplied changes in the inspector

grebaldi:
Hey,

this is just a as-faithful-as-possible re-implementation of the famous "Throw away"-Dialog of the current UI, as decided in: #833

With one exception: This implementation is actually a bit more strict. It doesn't let you change the UI state at all (except for further editing in the inspector) until you either applied or discarded the changes.

I thought this might be more consequential, since the whole point of prompting in this case is, to remind the user of editing changes, that he/she might have forgotten, before the attempted UI state change destroys them irreversibly. This does not only apply to node selection, but also to any navigation that leads the user away from the main application.

solves: #833


packages/neos-ui/src/Containers/RightSideBar/style.css
```diff
@@ -2,7 +2,7 @@
     padding-top: 41px;
     padding-bottom: 41px;
     transition: .2s ease transform;
-    will-change: transform;
+    /*will-change: transform;*/
```
dimaip: Commented out code. Why?

packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
```diff
@@ -22,15 +22,23 @@ import style from './style.css';
...
-    return state => ({
-        focusedNode: selectors.CR.Nodes.focusedSelector(state),
-        node: selectors.CR.Nodes.focusedSelector(state),
-        isApplyDisabled: isApplyDisabledSelector(state),
-        isDiscardDisabled: selectors.UI.Inspector.isDiscardDisabledSelector(state)
-    });
+    return state => {
+        const isDirty = selectors.UI.Inspector.isDirty(state);
+        const shouldPromptToHandleUnappliedChanges = selectors.UI.Inspector.shouldPromptToHandleUnappliedChanges(state);
+        const shouldShowUnappliedChangesOverlay = isDirty && !shouldPromptToHandleUnappliedChanges;
```
dimaip: Just thinking, wouldn't it be too distracting if the editor is shown this dialog immediately once he starts typing? Maybe it should only appear on unfocus from any of the inspector editors? But yeah, tricky to implement I guess...

grebaldi: The dialog isn't shown immediately, but the transparent overlay that blocks everything but the inspector (and secondary inspector). Once the user clicks on that blocking overlay, the dialog pops up.

dimaip: Sorry, I didn't mean the dialog, I meant the overlay. I find it to be a bit distracting, and would prefer to see it only on unfocus, but I guess it's ok the way it is.

grebaldi: How is it distracting? It isn't visible :)

dimaip: Lol, by looking at the code I thought it is. So it's there only to protect from clicking outside    of inspector?

dimaip: Lol, by looking at the code I thought it is. So it's there only to protect from clicking outside    of inspector?

Issue #999, opened 2017-09-23
Tree state and content canvas do not get updated correctly if the currently focused document node is discarded


grebaldi:
**Steps to reproduce:**

On the demo site, do

1. Create a new document beneath the homepage
2. Switch to that document
3. Hit "Discard all"

After the Content Canvas has reloaded, you're left with a 404 in the iframe and the discarded document node is still present, selected and focused in the page tree.

**Expected behavior:**

If the currently focused document node happens to be newly created and unpublished and then gets discarded, the parent document should become automatically focused. The discarded node needs to be removed from the tree. I'd expect the exact same behavior as if the currently focused document node is removed (which behaves correctly in the new UI).

PR #1000, opened 2017-09-23
BUGFIX: Make e2e section helper function async

grebaldi:
I recognized that the e2e-Tests are currently false-positive. All of the test-scenarios run through immediately without executing any assertions, because the `section` function itself is not asynchronous.

I fixed that and immediately found this one thanks to the tests: #999

**Update**

This now includes a fix for #999, so that the e2e tests should be green now.

Fixes: #999

### DIFF
--- a/Tests/IntegrationTests/contentModule.js
+++ b/Tests/IntegrationTests/contentModule.js
@@ -3,9 +3,9 @@ import ReactSelector from 'testcafe-react-selectors';
 
 import Page from './pageModel';
 
-const section = (name, testFunc) => {
+const section = async (name, testFunc) => {
     console.log('\x1b[44m%s\x1b[0m', name);
-    testFunc();
+    await testFunc();
 };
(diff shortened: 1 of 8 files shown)


Tests/IntegrationTests/contentModule.js
```diff
@@ -3,9 +3,9 @@ import ReactSelector from 'testcafe-react-selectors';
 
 import Page from './pageModel';
 
-const section = (name, testFunc) => {
+const section = async (name, testFunc) => {
     console.log('\x1b[44m%s\x1b[0m', name);
-    testFunc();
+    await testFunc();
```
dimaip: How does running it async makes it any different in terms of false-positivity? If one of the async parts of code asserts an error, the whole test should die anyways, no?

grebaldi: The test runner calls all `section`s and then terminates without executing any assertions, because it doesn't wait for the promises to resolve. This results in a zero exit code.

The test is positive, because it has no chance to fail before it terminates :)

Issue #940, opened 2017-09-11
Clicking inside empty editable, doesn't focus that node


dimaip:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

You can edit that editable, but it wouldn't focus. It only focuses on second click, or when it already has some text.

dimaip:
![peek 2017-09-12 11-12](https://user-images.githubusercontent.com/837032/30315027-4b15c33c-97ab-11e7-85b9-52f2fb9721b5.gif)

PR #1001, opened 2017-09-23
BUGFIX: Simulate click after placeholder removal (CKEditor)

grebaldi:
This is a workaround for the focus issue that occurs, if you click on an empty inline editable.

Since it is unclear at which point the event propagation is stopped (looks like CK internals), I just simulate another click on the editable element after the neosPlaceholder is removed.

Fixes: #940


dimaip:
Is there maybe a way to use the capturing phase of the event?

skurfuerst:
@dimaip I am not sure this will work here, as we are inside a CKEditor event at this code point...

dimaip:
Yep you are right, I looked around a bit more :)

Issue #1003, opened 2017-09-24
Labels in structure tree in content language instead of UI language 


waelas:
**Actual**
The element labels in the structure tree are changing language dependent on the chosen content language (dimension).
For example it says "Content Collection" if content dimension is English and "Inhaltssammlung" if content language is German.
**Expected**
Labels respect the UI language, independent of content language.

PR #1004, opened 2017-09-24
BUGFIX: Use user-configured locale for node augmentation

grebaldi:
Fixes: #1003


skurfuerst:
I think this *may* have some side effects if people use a custom Node Implementation and accessing I18N there; but I think that's a very rare edge case (unless proven otherwise)

Issue #829, opened 2017-08-19
Text aligning options


daniellienert:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

In the old UI, aloha provided options to align text:
```
  properties:
    text:
      ui:
        aloha:
          alignment:
            'left': TRUE
            'center': TRUE
            'right': TRUE
            'justify': TRUE
```

These options seem to be missing for the new UI.

Also support "strikethrough" etc.

Needed by one of the Core Feature sponsors.

grebaldi:
Hi @daniellienert,

I addressed this in #1005, but the PR only handles the alignment options.

All other formatting options should be working imho. `strikethrough` should be available via:

```yaml
  properties:
    text:
      ui:
        aloha:
          'format':
            'del': TRUE
```

Can you confirm?

PR #1005, opened 2017-09-24
FEATURE: Enable alignment options for CKEditor

grebaldi:
This enables the `left`, `right`, `center` and `justify` alignment options for CKEditor.

**Attention when testing:** I recognized that by default, none of the alignment options is enabled for Text nodes on the demo page - although the old UI is showing them anyway. So, when you're testing this with the Demo page, you have to explicitly enable the options via:

```yaml
  properties:
    text:
      ui:
        aloha:
          alignment:
            'left': TRUE
            'center': TRUE
            'right': TRUE
            'justify': TRUE
```

Resolves: #829


dimaip:
Why does it happen?

skurfuerst:
great work :)

Issue #897, opened 2017-09-06
Unable to deselect the current content element, when body height is not 100%


dfeyer:
If the body does not have height set as 100%, clicking after the end the body element, does not deselect (or select the current document, like the current behaviour) the currently selected content element

PR #1010, opened 2017-09-24
TASK: Listen to click on guest frame document rather than body

grebaldi:
Fixes: #897


dimaip:
Nice! Please use github keywords in the description like "Fixes: ", so it would automatically close the related issue.

PR #1016, opened 2017-09-25
BUGFIX: Relogin Submit on Enter

### LINKED ISSUE #1011: Re-Login Dialog should submit on hitting enter
Currently it doesn't.

mstruebing:
This makes it possible to trigger submit via the enter key.
It also enables the textinput to receive 3 new props.
onEnterKey - which triggers a function when enter is pressed
type - like password
setFocus = to focus this input element when it is mounted.

I get a console warning: `Warning: Unknown prop `0` on <input> tag.` even I am omitting the `onEnterKey` prop. The warning is triggered from the searchinput. If anyone know why this happens please ping me.

fixes #1011

### DIFF
--- a/packages/react-ui-components/src/TextInput/textInput.js
+++ b/packages/react-ui-components/src/TextInput/textInput.js
@@ -1,6 +1,7 @@
 import React, {PureComponent} from 'react';
 import PropTypes from 'prop-types';
 import mergeClassNames from 'classnames';
+import omit from 'lodash.omit';
 
 class TextInput extends PureComponent {
     static propTypes = {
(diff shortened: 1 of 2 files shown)


packages/react-ui-components/src/TextInput/textInput.js
```diff
@@ -94,21 +120,37 @@ class TextInput extends PureComponent {
...
             </div>
         );
     }
 
+    handleKeyPress(e) {
+        const enterKeyCode = 13;
+        const keyCode = e.keyCode || e.which;
+        const {onEnterKey} = this.props;
+        if (keyCode === enterKeyCode) {
+            onEnterKey();
```
grebaldi: Imho, we don't know at this point, whether `onEnterKey` is set. Please add a check before executing it.

mstruebing:
Updated @grebaldi 
Regarding the console error: I omitted the props wronlgy, this is now fixed.

PR #1018, opened 2017-09-25
BUGFIX: fix constraints with auto-created child nodes

### LINKED ISSUE #932: Node Constraints are not applied in the node creation screen for a content collection defined in the document node
While the tested configuration works in the old UI and restrict the usable node types, they are not applied in the new UI.

Tested with: #931bb4f

skurfuerst:
This is broken since 9ca09aa6d80bfd2b45ca61684dc5f15bbe585a48 - but
fortunately easy to fix :)

Fixes #932


dimaip:
Good catch!

skurfuerst:
needed for #966

PR #1032, opened 2017-09-27
TASK: split the tests

dimaip:
I split the test cases via a native testcafe mechanism instead of our own custom `section` helper.

PROS:
- Less hackish, native way
- Tests are more isolated from each other thus more reliable

CONS:
- Slower, as for every test case the UI is reloaded (~5s extra time per each test)
- Will not catch stateful side-effects, when one test may cause other to fail.


dimaip:
Yep I think stability is more important in this case.

Issue #1040, opened 2017-09-27
Dimension switcher should close after selection


grebaldi:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

![screenshot_2017-09-27_09-54-35](https://user-images.githubusercontent.com/2522299/30903021-ac537090-a36d-11e7-8b6b-de2b215c7f92.png)

Currently, if you switch to another dimension, the dimension switcher stays open.

dimaip:
Can reproduce, will fix

PR #1046, opened 2017-09-27
BUGFIX: close DimensionSwitcher on select

dimaip:
Fixes: #1040

### DIFF
--- a/packages/neos-ui/src/Containers/SecondaryToolbar/DimensionSwitcher/index.js
+++ b/packages/neos-ui/src/Containers/SecondaryToolbar/DimensionSwitcher/index.js
@@ -79,13 +79,37 @@ export default class DimensionSwitcher extends PureComponent {
         activePresets: new Map()
     };
 
+    state = {
+        isOpen: false
+    };
+
+    handleSelectPreset = preset => {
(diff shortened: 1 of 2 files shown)


packages/neos-ui/src/Containers/SecondaryToolbar/DimensionSwitcher/index.js
```diff
@@ -79,13 +79,37 @@ export default class DimensionSwitcher extends PureComponent {
         activePresets: new Map()
     };
 
+    state = {
+        isOpen: false
+    };
+
+    handleSelectPreset = preset => {
+        this.setState({isOpen: false});
+        this.props.selectPreset(preset);
```
grebaldi: Looking at https://github.com/neos/neos-ui/blob/master/packages/neos-ui-redux-store/src/CR/ContentDimensions/index.js#L21, this needs to take a couple of more arguments :)

dimaip: Ouch! Thanks!

dimaip:
@grebaldi do you think it's fine to do it like this?

Issue #1042, opened 2017-09-27
PageTree is empty after switching to another dimension


grebaldi:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

![screenshot_2017-09-27_10-35-43](https://user-images.githubusercontent.com/2522299/30903677-d641ad70-a36f-11e7-989e-d69436909147.png)

PR #1051, opened 2017-09-27
BUGFIX: Replace node state on dimension switch

grebaldi:
So, this is a larger change.

Basically what it does is:

* Creating a new FlowQuery operation that allows to load the default nodes for the backend
* Exposing that FlowQuery operation to the UI
* <s>Replace the entire node state, whenever a dimension is switched, before the content canvas loads the new document state.</s>
* Load all default nodes of the target dimension into the store

<s>This is still crazy slow, but at least it works.</s>
I was able to parallelize loading the default nodes and the guest frame. Now it feels quite fast.

Fixes: #1042

### DIFF
--- a/packages/neos-ui-redux-store/src/CR/Nodes/index.js
+++ b/packages/neos-ui-redux-store/src/CR/Nodes/index.js
@@ -17,6 +17,7 @@ const COMMENCE_REMOVAL = '@neos/neos-ui/CR/Nodes/COMMENCE_REMOVAL';
 const REMOVAL_ABORTED = '@neos/neos-ui/CR/Nodes/REMOVAL_ABORTED';
 const REMOVAL_CONFIRMED = '@neos/neos-ui/CR/Nodes/REMOVAL_CONFIRMED';
 const REMOVE = '@neos/neos-ui/CR/Nodes/REMOVE';
+const REPLACE = '@neos/neos-ui/CR/Nodes/REPLACE';
 const COPY = '@neos/neos-ui/CR/Nodes/COPY';
 const CUT = '@neos/neos-ui/CR/Nodes/CUT';
 const MOVE = '@neos/neos-ui/CR/Nodes/MOVE';
@@ -38,6 +39,7 @@ export const actionTypes = {
     REMOVAL_ABORTED,
     REMOVAL_CONFIRMED,
     REMOVE,
+    REPLACE,
     COPY,
     CUT,
     MOVE,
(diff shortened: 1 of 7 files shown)


packages/neos-ui-redux-store/src/CR/Nodes/index.js
```diff
@@ -113,6 +115,11 @@ const confirmRemoval = createAction(REMOVAL_CONFIRMED);
 const remove = createAction(REMOVE, contextPath => contextPath);
 
 /**
+ * Remove all nodes from the store
+ */
+const replace = createAction(REPLACE, ({siteNode, documentNode, nodes}) => ({siteNode, documentNode, nodes}));
```
skurfuerst: I'd like to rename the variables a little, like: "siteNodeContextPath, documentNodeContextPath, nodes"; or at least document it here.

packages/neos-ui/src/Sagas/CR/ContentDimensions/index.js
```diff
@@ -1,66 +1,99 @@
...
-            yield put(actions.UI.ContentCanvas.setSrc(result.nodeFrontendUri));
-        } else if (nextAction.type === actionTypes.UI.NodeVariantCreationDialog.CANCEL) {
-            // Select old dimensions again if user cancelled
-            yield put(actions.CR.ContentDimensions.setActive(sourceDimensions));
-        }
+    if (nodeFound) {
+        return {nodeFrontendUri, nodeContextPath};
+    }
+
+    console.log('Open NodeVariantCreationDialog');
```
skurfuerst: remove console.log

packages/neos-ui/src/Sagas/CR/ContentDimensions/index.js
```diff
@@ -1,66 +1,99 @@
...
+import {put, select, race, take, call} from 'redux-saga/effects';
 import {$get} from 'plow-js';
 
 import {actions, actionTypes, selectors} from '@neos-project/neos-ui-redux-store';
 import backend from '@neos-project/neos-ui-backend-connector';
-function * watchSelectPreset() {
-    // Only process the latest selection and cancel previous ones
-    yield * takeLatest(actionTypes.CR.ContentDimensions.SELECT_PRESET, updateContentCanvasSrc);
+
+function * watchSelectPreset({configuration}) {
```
skurfuerst: add a comment here explaining in 1-2 sentences what the saga does, and especially the idea why we need to forget all nodes on dimension switch and reload the Node List.

dimaip:
Perhaps we could transfer initial state via an ajax request as well?
My ideal would be to have static lightweight html code to start loading all assets instantly, and move all changing parts out of it.

grebaldi:
I thought about that as well, but I'd like to open a separate PR for that.

packages/neos-ui/src/Sagas/CR/ContentDimensions/index.js
grebaldi: Actually we don't. I thought it would be a good opportunity to get rid of a few nodes to not let the node state grow too large. But after thinking about it, this might not be the correct place to handle this problem, so I'm going to remove it.

skurfuerst:
Hm, actually I like that we do not do so many ajax requests on loading... let's discuss this :)

Issue #1014, opened 2017-09-25
"Selected Element" is missing in Inspector


grebaldi:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

![screenshot_2017-09-25_11-44-30](https://user-images.githubusercontent.com/2522299/30802474-f625c322-a1e6-11e7-8900-3581b44fca93.png)

The current UI allows to select any element in the parent line of the currently selected element. This SelectBox shows up at the top of the inspector before any other view or editor, regardless of which tab is active.

PR #1055, opened 2017-09-28
FEATURE: Implement "Selected Element" for inspector

grebaldi:
Resolves: #1014

### DIFF
--- a/packages/neos-ui/src/Containers/RightSideBar/Inspector/SelectedElement/style.css
+++ b/packages/neos-ui/src/Containers/RightSideBar/Inspector/SelectedElement/style.css
@@ -1,0 +1,20 @@
+.label {
+    width: 100%;
+    overflow-x: hidden;
+    text-overflow: ellipsis;
+    padding: 0 40px 0 16px;
+    height: 40px;
+    margin: 0;
+    line-height: 40px;
+    font-weight: bold;
+    font-size: 14px;
+
+    span {
+        display: inline-block;
+        margin-left: .1em;
+    }
+}
+
+.content {
+    padding-bottom: 16px;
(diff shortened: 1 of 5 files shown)


packages/neos-ui/src/Containers/RightSideBar/Inspector/SelectedElement/style.css
```diff
@@ -0,0 +1,20 @@
+.label {
+    width: 100%;
+    overflow-x: hidden;
+    text-overflow: ellipsis;
+    padding: 0 40px 0 16px;
+    height: 40px;
+    margin: 0;
+    line-height: 40px;
```
dimaip: We have spacing constants for such stuff.

packages/neos-ui/src/Containers/RightSideBar/Inspector/SelectedElement/style.css
```diff
@@ -0,0 +1,20 @@
...
+    font-weight: bold;
+    font-size: 14px;
```
dimaip: Also would be great if we eventually gather all font-sizes as constants too... not necessarily now.
This would be especially important when we would start working on the design of the UI again.

packages/neos-ui/src/Containers/RightSideBar/Inspector/SelectedElement/style.css
```diff
@@ -0,0 +1,20 @@
...
+
+    span {
+        display: inline-block;
+        margin-left: .1em;
+    }
+}
+
+.content {
+    padding-bottom: 16px;
```
dimaip: Same

Issue #1059, opened 2017-09-28
Add-Button shouldn't be enabled when there are no node types to choose from


grebaldi:
Currently when due to constraints or other circumstances no node type would be allowed for insertion alongside or into the currently focused node, the "Add"-Button is still avaible, both in the content tree and in the inline node toolbar.

The button should be disabled to not confront the user with an empty node type dialog.

PR #1068, opened 2017-09-28
FEATURE: Disable "Add" button, if no node type options are available

grebaldi:
Resolves: #1059


packages/neos-ui-redux-store/src/CR/Nodes/selectors.js
```diff
@@ -198,6 +198,15 @@ export const makeGetAllowedChildNodeTypesSelector = (nodeTypesRegistry, elevator
 export const makeGetAllowedSiblingNodeTypesSelector = nodeTypesRegistry =>
     makeGetAllowedChildNodeTypesSelector(nodeTypesRegistry, parentNodeContextPath);
 
+export const makeHasAnyNodeTypeOptionsSelector = nodeTypesRegistry => createSelector(
```
skurfuerst: I think the name `HasAnyNodeTypeOptions` is somehow misleading. I'd suggest we use `isAllowedToAddChildOrSiblingNodes`

(also change at the other places)

dimaip:
Anyways, I merge it as is for now, may refactor when working on #1075

Issue #1060, opened 2017-09-28
"Open in target dimension" should be enabled after publishing to live


grebaldi:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

![screenshot_2017-09-28_14-20-45](https://user-images.githubusercontent.com/2522299/30966205-6f4d074c-a458-11e7-9a0b-97f0cb874b97.png)

After publishing a newly created document to live, the "Open in target dimension" button remains disabled, although all information should be present to render the correct URI.

PR #1071, opened 2017-09-29
BUGFIX: Reload document after publish

grebaldi:
Fixes: #1060


dimaip:
Hmm put that will... reload the document? Do we want that? Is there a way to prevent that? In the old UI we didn't do that.
Maybe we can send this info via a feedback?

grebaldi:
This will be very complicated to figure out on server side... Maybe we should add a condition, to only reload, if there's no `previewUrl` present.

dimaip:
That could be an option, though it's such a pity to reload the whole thing just for the sake of this preview url... Maybe we can get it from src by stripping context?

skurfuerst:
@dimaip @grebaldi what's the status here? :)

grebaldi:
@skurfuerst This feature is not mission critical, but I will look for a solution that doesn't require a full reload. 

I'll add a `WIP` prefix to the PR.

dimaip:
I'm quite surprised, but all of our editors seem to claim this bug *is* mission critical for them...
@grebaldi could you still find time to think it through and fix it in these two months? I think you understand the topic best.

dfeyer:
I think it's fine if we merge this, and create an issue to improve the reloading on a separate PR.

grebaldi:
I'm afraid, I have to disagree. The behavior introduced by this PR pretty much breaks the `auto-publish` feature. 

I'll try to put some thought on a better way to solve this.

dimaip:
Yep, no preview url is a lesser evil. But would be great to fix before the 1.0!

grebaldi:
I updated the PR, the page is no longer reloaded. The preview url gets updated independently, if the current content canvas document node is amongst the recently published nodes. 

I also disabled the preview button, if there's no preview url - before it has been disabled visually, but actually opened the UI in a new tab on click.

packages/neos-ui/src/Sagas/Publish/index.js
```diff
@@ -21,6 +22,19 @@ function * watchPublish() {
                 const feedback = yield call(publish, nodeContextPaths, targetWorkspaceName);
                 yield put(actions.UI.Remote.finishPublishing());
                 yield put(actions.ServerFeedback.handleServerFeedback(feedback));
+
+                //
+                // Reload the current content canvas document node, to retrieve a new preview uri,
```
dimaip: Update comment

PR #1072, opened 2017-09-29
BUGFIX: Handle secondary inspector state via redux

grebaldi:
Fixes: #1063


grebaldi:
Well, I think the original idea was to leave the concern of handling secondary inspectors to the respective inspector editors - which is a sort-of replication of what the old UI does. That's why we have this breach in the first place.

But thinking about it... What are our Secondary Inspectors currently? 

1. Choose image from media module
2. Display image information in media module
3. Image cropper
4. Code Editor

The first two will indeed be needed elsewhere (e.g. AssetEditor, AssetListEditor).

So, in hindsight - I agree, it makes sense to have those accessible via registry and get rid of the `renderSecondaryInspector` API.

Long story short: Will do :smiley:

skurfuerst:
awesome <3

grebaldi:
So, I've been on this for a while now and I came to the conclusion that the current logic is actually our best shot to provide a decent API to the outside.

If we would move the handling of secondary editor options into the store, we would have to figure out a way to handle callbacks like `onComplete` in the image cropper or `onClose` in the media details view.

This would move important parts of editor implementations to the store and would decrease flexibility of inspector editors overall, while increasing complexity for those, who want to implement their own editors. All this inmho for actually no significant advantage.

I think the way, we are handling this problem now is not that bad. This is clearly a sensible point to leave our redux context, since every secondary editor lives in a context of its own - that is not expected to have any effects in the rest of the application.

1. Secondary Inspector opens 
2. The user acts, as if this was a separate application embedded in the UI 
3. Secondary Inspector returns a result
4. Result is pushed to the store

I think this is perfectly fine.

However, what we have to think about is, how we can find an abstraction for the management of secondary editors (which is currently done solely in the Inspector component). In the long run, we will need other places to be capable of managing them as well (such as the NodeCreationDialog).

skurfuerst:
Thanks for your detailed explanation and thinking the issue through 👍 - so that's really good, then we merge this PR as soon as it builds :)

Issue #1044, opened 2017-09-27
New API for registries and more unplanned extensibillity


dimaip:
Currently the registries are flat key/value stores.
There are many cases where it makes sense to have them nested.

Take for example the topic of unplanned extensibility. Currently it's only possible to replace certain parts of the UI. But it would be cool to be able to insert new custom components anywhere between them.

We could take this syntax:
```js
containerRegistry.add('PrimaryToolbar/UserDropDown', UserDropDown);
containerRegistry.add('PrimaryToolbar/PublishDropDown', PublishDropDown);
...
```
But so that `containerRegistry.get('PrimaryToolbar')` would return a map of its inner keys, e.g.
```js
{
  UserDropDown: UserDropDown,
  PublishDropDown: PublishDropDown
}
```
Then we can just map over it and render all of the related components together.

But to support inserting more components later into the right position, we should support sorting.

It could work like this: 
```js
containerRegistry.add('PrimaryToolbar/MyCustomButton', MyCustomButton, 'after PublishDropDown');
```
We could use the positioning syntax from Fusion for example.

Should be relatively easy to implement, but would allow to do really crazy stuff extensibility-wise.

Would be also needed for things like registering multiple feedback handlers onto one feedback.

skurfuerst:
- Decision: we'll make it return an array here
- rename "add" to "set"
- getOrderedSubElements

PR #1084, opened 2017-09-30
FEATURE: implement positional array sorter and use it to sort the registries

dimaip:
Resolves: #1044 

Changes to the registry API:
- rename `add` to `set`.
- third parameter to `set` takes a position, implementing full positional array sorter spec
- `getChildren` and `getChildrenAsObject` return all keys starting with a certain string

Example:
```js
containerRegistry.set('LeftSideBar/Top/PageTreeToolbar', PageTreeToolbar);
containerRegistry.set('LeftSideBar/Top/PageTreeSearchbar', PageTreeSearchbar);
containerRegistry.set('LeftSideBar/Top/PageTree', PageTree);
```
Then somewhere from the user code:
```js
containerRegistry.set('LeftSideBar/Top/MyCustomBar', MyCustomBar, 'before LeftSideBar/Top/PageTree');
```
The you can render all containers, sorted, like this:
```js
containerRegistry.getChildren('LeftSideBar/Top').map((Item, key) => <Item key={key}/>);
```

### Differences so far from the orginal PositionalArraySorter:
1. Invalid positions are sorted as if they had no positions and only console warning is displayed. PHP sorter would throw. I think it's better to make frontend code more resilient to errors.
2. Chains of circular dependencies are detected and moved to the end. PHP sorter would hang.
3. `start 12 34` was considered an invalid key. Now it's interpreted as `start 12`
4. PHP sorter is recursive, this one is flat. One can always recursively apply if needed.
5. It was possible to add weights to `before` and `after`, now it's not.

TODO:

- [x] use `getChildren` in feedback handlers
- [x] write tests for the registry


packages/neos-ui-extensibility/src/registry/SynchronousRegistry.js
```diff
@@ -1,38 +1,74 @@
...
+            result[item.key] = item.value;
+        });
+        return result;
+    }
+
+    getChildren(searchKey) {
+        return this._getChildrenWrapped(searchKey).map(item => item.value);
     }
 
     has(key) {
```
skurfuerst: can you add docs to all the public methods here, how they shall be used?

packages/neos-ui-extensibility/src/registry/SynchronousRegistry.js
```diff
@@ -1,38 +1,74 @@
...
-        this._keys = [];
+        this._registry = [];
     }
 
-    add(key, value) {
-        // TODO: implement position in registry!
-        // TODO: "key" must be string!
-        this._registry[key] = value;
-        this._keys.push(key);
+    set(key, value, position = 0) {
```
skurfuerst: I am not sure we should allow having the same key twice. I think this should always be an error, as it makes the remaining API inconsistent / indeterminate

packages/utils-helpers/src/positionalArraySorter.spec.js
```diff
@@ -0,0 +1,179 @@
+import positionalArraySorter from './positionalArraySorter';
+
+test(`Middle`, () => {
```
skurfuerst: I suppose you ported the tests from the PHP side? (Just asking :-) )

packages/neos-ui-extensibility/src/registry/SynchronousRegistry.js
dimaip: Great point!

packages/neos-ui-extensibility/src/registry/SynchronousRegistry.js
dimaip: Will do.

packages/utils-helpers/src/positionalArraySorter.spec.js
dimaip: Erm... honestly... no. Will check how they compare.

packages/neos-ui-extensibility/src/registry/SynchronousRegistry.js
dimaip: On the other hand, what about if the user wants to override something in our UI? E.g. replace the tree with smth else?

skurfuerst:
perfect :) Awesome job!

Issue #874, opened 2017-09-05
File Upload Editor


kitsunet:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

PR #1086, opened 2017-10-02
FEATURE: Upload component

kitsunet:
Adds a component to facilitate asset uploads to Neos.
The component is used inside the Image editor but can be re-used
for other editors.

Fixes: #874


packages/neos-ui-editors/src/Library/index.js
```diff
@@ -0,0 +1,5 @@
+import AssetUpload from './AssetUpload';
+
+export default () => ({
+    AssetUpload
+});
```
skurfuerst: This is wrong; you need to do an "export AssetUpload" here :)

A named import in the form `import {....}` is *never* accessing the default import; they are two totally different things.

packages/neos-ui-editors/src/Editors/Image/index.js
```diff
@@ -220,21 +220,16 @@ export default class ImageEditor extends Component {
     }
 
     upload(files) {
```
dimaip: files defined but never used

packages/neos-ui-editors/src/Editors/Image/index.js
```diff
@@ -220,21 +220,16 @@ export default class ImageEditor extends Component {
...
-        const {commit, siteNodePath} = this.props;
+        const {commit} = this.props;
         const {isImageCropperOpen} = this.state;
 
-        const siteNodeName = siteNodePath.match(/\/sites\/([^/@]*)/)[1];
-
-        return uploadAsset(files[0], siteNodeName).then(res => {
-            this.setState({image: res}, () => {
-                commit(res.object);
+        this.setState({image: res}, () => {
```
dimaip: res is undefined

kitsunet:
@dimaip this has been updated, retested, linted and should just work now 😆  Please review

Issue #1075, opened 2017-09-29
constraints are not made visible in the tree when moving nodes


crazydrecksack:
![sep-29-2017 13-24-45](https://user-images.githubusercontent.com/683773/31014119-9f2d333a-a519-11e7-8ca9-7167b929ec4c.gif)

In the example I have constraints attached to "Werkbereiche" to only allow child nodes of a specific type. This is why I cannot move "Hauptseite" into "Werbereiche" however this is not indicated by the blue line (drop target) wich used to be red if the desired position of a node drag and drop is prohibited via constraints.

Dragging a node over another does not show any highlight to indicate if it is ok to drop the node inside or in case of a constraint violation that it is not allowed.

dimaip:
My bad, will fix.

crazydrecksack:
old UI
![sep-29-2017 13-50-01](https://user-images.githubusercontent.com/683773/31014807-32bbb452-a51d-11e7-9c18-8994e20c4457.gif)

dimaip:
Actually there are two problems at once here.
The first one is that constraints are not taken into account, but it may be addressed here: https://github.com/neos/neos-ui/pull/1068

And the second problem is how to style the disabled drop targets. When implementing the tree I chose to not display them at all, which makes for less visual clutter when dragging (really annoying to see red all the time), so the blue drop target is only displayed when you can drop something there.
If you think that we should still paint the denied drop targets red I can easily do that too.

dimaip:
See:
![peek 2017-09-29 15-56](https://user-images.githubusercontent.com/837032/31016800-d9b8b2d0-a52e-11e7-9640-745c7a70bafc.gif)

PR #1090, opened 2017-10-03
BUGFIX: only show sibling dropzones when drop is allowed

dimaip:
Fixes: #1075


dimaip:
Will open another PR with more stuff

Issue #817, opened 2017-08-10
Copy/Cut/Paste buttons should reflect the clipboard state


dimaip:
In the old ui, when you copy/cut a node, the respected button becomes highlighted, until you have pasted the node, while the paste button also becomes active. When you click on the copy/cut button again, it should clear the clipboard and unselect it.

That's not what currently happens in the new ui.

![image](https://user-images.githubusercontent.com/837032/29175081-c22bfbaa-7df0-11e7-95d2-dca56e6146db.png)

akoenig:
@dimaip Do you suggest that the insert mode should also be visible in the toolbar?

dimaip:
@akoenig certainly!

skurfuerst:
I think this needs a little discussion -- for the insert mode, we do not have these three-strip buttons anymore but we always solve it in a modal (for now).

Do you think that's a must-have for the new UI for now? For me that is rather a small UX change, with not a big impact. That's why I remove the must-have label for now... If you disagree, let's discuss :)

All the best,
Sebastian

dimaip:
IMO this makes the UI less visually responsive. I'd consider it a mid-minor UX issues and would try to fix it before 1.0 if possible, especially since it's pretty easy to fix.
I'd love if we could prioritize issues for 1.0, i.e. those which we tackle first and those which we can postpone but still good to make, e.g.: "high-pro", "mid-prio", "low-prio", "not-going-to-1.0".

PR #1091, opened 2017-10-03
BUGFIX: misc cut/paste logic fixes

dimaip:
Fixes: #1075
Fixes: #1092
Fixes: #817

Fix multiple bugs related to the logic of moving nodes:
- DragAndDrop of autoCreated nodes should be disallowed
- Moving into descendants of the node itself should be disallowed
- Before and after dropzones should not be highlighted when can't drop into them
- Cut/Copy buttons should reflect their state


dimaip:
Will open another PR with more stuff

Issue #755, opened 2017-07-06
Fatal Error in DimensionSwitcher on dimension with multiple values


frozenstupidity:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

The dimension switcher causes a white page with the following error when the content dimension in this case language contains more than one values.

> TypeError: Cannot read property ‘contains’ of undefined

This seems to be caused by the evaluated eel `${Neos.Ui.ContentDimensions.allowedPresetsByName(documentNode.context.dimensions)}` resulting in the array `['de']` being passed into the function when the values in the dimension configurations are `['de', 'de_DE', 'de_AT']` this ultimately causes the function not to match any of the dimensions and to return an empty array when initializing the allowedPresets prop.

frozenstupidity:
It looks like the following code:

```
$preset = $this->contentDimensionsPresetSource->findPresetByDimensionValues($dimensionName, 
$dimensionValues);
if ($preset !== null) {
    $preselectedDimensionPresets[$dimensionName] = array_shift($dimensionValues);
}
```

may be irrelevant as I think the `documentNode.context.dimensions` evaluates to an array of identifiers, however tests are showing this to cause some other errors, similar to those when switching from English (US) to Dutch in the demo installation so will require more debugging.

frozenstupidity:
It appears as though this issue is not so much an issue with the menu as much as it is the dimension being initialized with the value of Neos.ContentRepository.contentDimensions.language.default which does not match the values of the values of the preset.

frozenstupidity:
This issue appears to be related to issue neos/neos-development-collection#1204 which also identifies issues with the definition of default and defaultPreset in combination with presets.

skurfuerst:
Hey,

thanks for your comments and ideas ❤️ 

I just spent an hour figuring out what is going wrong; and I don't really understand the issue yet. I feel one pain point is that we do not all have the same configuration to debug from – Would you mind adjusting the Neos Demo Site in a way that it exposes the error, and post the Settings.yaml here – so we have the same state?

All the best, and thanks again,
Sebastian

frozenstupidity:
Hi Sebastian,

I have attached the whole demo settings which reproduces the problem., however the root cause of the problem can be narrowed down to 

```yaml

Neos:
  Demo:
    flickr:
      tagStreamUriPattern: 'http://api.flickr.com/services/feeds/photos_public.gne?format=json&nojsoncallback=1&tags=%s'
[...]

PR #1093, opened 2017-10-03
BUGFIX: fix loading of node with dimensions, use lastVisitedNode logic

skurfuerst:
The loading error occurred exactly if the *standard* dimension contained multiple
dimension values. That's what is fixed here.

Resolves: #755

## How to test

Important: After applying this patch, clear your caches and start by visiting the FRONTEND (such that the last-visited node can be properly memoized ,*including* dimension infos)


dimaip:
I have to agree with Dominique, it's kinda scary that we have to rely onto such implicit logic. If there is no way to make it clearer, I'd document it and create an issue with technical dept label.

skurfuerst:
Full ack :) I'll merge this now as it replicates the old behavior, but is something we need to clean up at some point.

dimaip:
Great! Please create a reminder issue for it.

PR #1099, opened 2017-10-04
FEATURE: Support keyboard navigation in SelectBox

### LINKED ISSUE #1036: Select Boxes should support keyboard navigation

mstruebing:
In this implementation it is possible to use the arrow keys to navigate
up and down and to chose one item via the enter key.
This replicates the behaviour of the old UI.

This doesn't work for SelectBoxes with a search box currently.

closes #1036

### DIFF
--- a/packages/react-ui-components/src/SelectBox/selectBox.js
+++ b/packages/react-ui-components/src/SelectBox/selectBox.js
@@ -2,8 +2,10 @@ import React, {PureComponent} from 'react';
 import PropTypes from 'prop-types';
 import DropDown from '../DropDown/index';
 import DefaultSelectBoxOption from './defaultSelectBoxOption';
+import keydown from 'react-keydown';
 import mergeClassNames from 'classnames';
 
+@keydown
 export default class SelectBox extends PureComponent {
 
     static defaultProps = {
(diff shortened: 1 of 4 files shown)


packages/react-ui-components/src/SelectBox/selectBox.js
```diff
@@ -2,8 +2,10 @@ import React, {PureComponent} from 'react';
 import PropTypes from 'prop-types';
 import DropDown from '../DropDown/index';
 import DefaultSelectBoxOption from './defaultSelectBoxOption';
+import keydown from 'react-keydown';
 import mergeClassNames from 'classnames';
 
+@keydown
```
grebaldi: Please have a look at https://github.com/glortho/react-keydown#monitor-only-key-codes-which-you-care-about

I think it would be better to restrict this. During inline editing we have tons of key events, that would trigger a redraw of every select box otherwise.

mstruebing: Hey @grebaldi I've updated my PR.
The events were only be captured when the selectbox is active but you are right, to restrict them would be better anyway.

PR #1103, opened 2017-10-04
BUGFIX: correctly redirect on removal of nested nodes

dimaip:
Steps to reproduce: Create a document node. Create another document node inside of it. Discard all and see all sorts of errors.

This change recursively looks for the parent node that still exists, and redirects to it.

### DIFF
--- a/packages/neos-ui/src/manifest.js
+++ b/packages/neos-ui/src/manifest.js
@@ -4,6 +4,8 @@ import {$get} from 'plow-js';
 
 import {actions, selectors} from '@neos-project/neos-ui-redux-store';
 
+import {parentNodeContextPath} from '@neos-project/neos-ui-redux-store/src/CR/Nodes/helpers';
+
 import manifest from '@neos-project/neos-ui-extensibility';
 import {SynchronousRegistry, SynchronousMetaRegistry} from '@neos-project/neos-ui-extensibility/src/registry';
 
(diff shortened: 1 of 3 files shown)


packages/neos-ui/src/manifest.js
```diff
@@ -247,11 +248,24 @@ manifest('main', {}, globalRegistry => {
...
+        // If we are removing current document node...
         if ($get('ui.contentCanvas.contextPath', state) === contextPath) {
-            const parentNodeUri = $get(['cr', 'nodes', 'byContextPath', parentContextPath, 'uri'], state);
+            let redirectContextPath = contextPath;
+            let redirectUri = null;
+            // Determine closest parent that is not being removed
+            while (!redirectUri) {
+                redirectContextPath = parentNodeContextPath(redirectContextPath);
+                // This is an extreme case when even the top node does not exist in the given dimension
+                // TODO: still find a nicer way to break out of this situation
```
grebaldi: We could fallback to the `siteNodeContextPath` for a start - that should be always present. But we can deal with that when we get rid of the hard-coded Uris alltogether.

dimaip: But it is not present in all dimensions... In any case the logic above would traverse up to the site node in given dimension, and if it isn't there it would rudely reload the whole thing (should be extremely rare). And yes, needs taking care when we want to get rid of hardcoded uris.

Issue #1073, opened 2017-09-29
Inline node toolbar needs repositioning when an image changes


grebaldi:
![screenshot_2017-09-29_11-47-18](https://user-images.githubusercontent.com/2522299/31010751-37677056-a50c-11e7-9bda-71cdadef0cc1.png)

Steps to reproduce (on Demo site):

1. Insert a new Image element beneath the Text "...future change we don’t know."
2. Select an image from the media module
3. Hit apply

After the reload is finished, the toolbar will have this weird position.

dimaip:
Can reproduce.

PR #1110, opened 2017-10-05
BUGFIX: Add 'load' eventListener to NodeToolbar

janinadanger:
to fix top position of toolbar after the page reloads

Fixes: #1073


skurfuerst:
sorry for the long delay for reviewing this one. Totally makes sense :)

Issue #832, opened 2017-08-23
HTML code editor doesn't expand to full height before starting to scroll


dimaip:
![peek 2017-08-23 11-20](https://user-images.githubusercontent.com/837032/29605927-4b44f4ba-87f5-11e7-8a7c-5fd2f4dd08f0.gif)

skurfuerst:
For me, that's more Medium-Prio than High-Prio (as the HTML editor is probably not used that often in Projects). I am not sure how easy to fix it is. If it's a 30 minute thing, then let's go for it; if it is 1 day then I'd skip it for now.

dimaip:
Agreed. It's super annoying actually, as you look at the editor and think that that's all code there is. Made me confused on real project and I was wondering for a while to what actually happened to my code...
Should be easy to fix.

Duske:
I guess the problem was the way the height of the codemirror-component was set. In the PR I used the `setSize()`-method of codemirror, which seems to solve the problem.

PR #1121, opened 2017-10-08
BUGFIX: set height of codemirror instance with ref callback

Duske:
Fixes: #832 
Before, sometimes not all lines were rendered in the editor because the setting the height manually did not trigger a rerender of the lines.
When setting the height with the [`setSize()`-method from codemirror](https://github.com/codemirror/CodeMirror/blob/0ea666fe3db9cd68dd951cc9178944cccdd9276d/src/edit/methods.js#L392), the number of displayed lines is newly calculated.

Setting the height this way seems to be advised with codemirror, as it mentioned here:
https://github.com/JedWatson/react-codemirror/issues/57#issuecomment-274236854


skurfuerst:
Thanks for your contribution ❤️

Issue #793, opened 2017-08-02
Extract z-index variables into constants


dimaip:
Currently, our usage of z-index is a jungle. Would be nice to extract those constants like we did in the old UI: https://github.com/neos/neos-development-collection/blob/30599d9a88aa537ab867b0a2b7c68690c6cfe00c/Neos.Neos/Resources/Private/Styles/Foundation/_variables.scss#L144

Duske:
I'd like to take this issue, but I guess I need some guidance here first:
1) Since we are not using SASS for the components, we could CSS variables to store these values, right?
2) Afaik, the CSS vars which are currently used are either in `@neos-project/brand` or directly set in the shared [`webpack.config.js`](https://github.com/neos/neos-ui/blob/master/packages/build-essentials/src/webpack.config.js#L55). Where should we put things like z-indices?

dimaip:
Thanks for the initiative to take care of it!
Needs some thinking through... My immediate thoughts are to extract css variables into a separate js file (for clarity), and put z-index constants there.
You may suggest better alternatives of course.

Duske:
So kinda like it's done for [`@neos-project/brand`](https://github.com/neos/brand/blob/master/index.js)? 
We could put the existing variables which are set in the `webpack.config.js` there, too.
The next question which comes to my mind is, where we put such a separate js file solely for styling.

dimaip:
Yes. But no, let's not put anything into brand, it's a separate package that is hard to manage, and these z-index constants are purely technical, they are not related to brand styles.
That's a good question where to put it. Probably most simple answer is just to put it alongside the webpack.config.js file in build-essentials, to avoid external dependencies.

Duske:
Totally agree, I did not mean to put it in a separate package since I've seen a git submodule abyss at work :D 
I guess I just create a PR and then we can look what needs to be changed/improved.

dimaip:
Yep, trust your intuition here :)

Duske:
I've created a PR but I am not really happy with the result. It reveals some aspects which should be discussed like naming of the variables and structure of the configuration object.

PR #1124, opened 2017-10-10
TASK: Extract z-index variables into CSS vars

Duske:
This PR is a first draft on how to move the z-indices to CSS vars as discussed in #793.
It reveals some issues which need to be solved:

- [x]  find a proper naming for the z-index-variables, e.g. scope the name to the corresponding component
- [x]  find a proper structure for the style config object
- [x]  refactor the existing [CSS vars](https://github.com/neos/neos-ui/blob/master/packages/build-essentials/src/webpack.config.js#L55) which are currently directly set in the webpack config. Maybe this out of scope of this PR which should deal with z-indices only.

### DIFF
--- a/packages/neos-ui-inspector/src/SecondaryInspector/style.css
+++ b/packages/neos-ui-inspector/src/SecondaryInspector/style.css
@@ -17,7 +17,7 @@
     right: 0;
 }
 .secondaryInspector--isElevated {
-    z-index: var(--zIndexSecondaryInspectorElevatedContext);
+    z-index: var(--zIndex-SecondaryInspectorElevated-Context);
 }
 .secondaryInspector .close {
     position: absolute;
--- a/packages/neos-ui-inspector/src/SecondaryInspector/style.css
+++ b/packages/neos-ui-inspector/src/SecondaryInspector/style.css
@@ -17,7 +17,7 @@
     right: 0;
 }
 .secondaryInspector--isElevated {
-    z-index: var(--zIndexSecondaryInspectorElevated);
+    z-index: var(--zIndexSecondaryInspectorElevatedContext);
 }
 .secondaryInspector .close {
     position: absolute;
--- a/packages/neos-ui-inspector/src/SecondaryInspector/style.css
+++ b/packages/neos-ui-inspector/src/SecondaryInspector/style.css
@@ -29,5 +29,5 @@
     font-size: 18px;
     border-right-width: 0;
     border-top-width: 0;
-    z-index: var(--zIndexSecondaryInspectorClose);
+    z-index: var(--zIndex-SecondaryInspector-Close);
 }
--- a/packages/neos-ui-inspector/src/SecondaryInspector/style.css
+++ b/packages/neos-ui-inspector/src/SecondaryInspector/style.css
@@ -17,7 +17,7 @@
     right: 0;
 }
 .secondaryInspector--isElevated {
-    z-index: 10001;
+    z-index: var(--zIndexSecondaryInspectorElevated);
 }
 .secondaryInspector .close {
     position: absolute;
@@ -29,5 +29,5 @@
     font-size: 18px;
     border-right-width: 0;
     border-top-width: 0;
-    z-index: 10;
+    z-index: var(--zIndexSecondaryInspectorClose);
 }
--- a/packages/neos-ui/src/Containers/RightSideBar/Inspector/style.css
+++ b/packages/neos-ui/src/Containers/RightSideBar/Inspector/style.css
@@ -19,7 +19,7 @@
     left: 0;
     width: calc(100% - var(--sidebarWidth));
     height: 100%;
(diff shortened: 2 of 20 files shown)


dimaip:
Great!

> e.g. scope the name to the corresponding component
What do you mean?

> Maybe this out of scope
As it is comfortable for you. I'd just go ahead and do it, it's just a copy paste and would allow to see the structure of the styles object clearer.

Duske:
> e.g. scope the name to the corresponding component

In the configuration object there are the properties `aspectRatioDropdownContents ` and `dropdownContents ` for example. They're basically dealing with the same stuff, but need different names (collision) as they are different components. So we could use some scoping like this:

```js
zIndex:{
  aspectRatio: {
    dropdownContents: '1'
  }
  someOtherComponent: {
    dropdownContents: '2'
  }
}

```

> Maybe this out of scope

Sounds good, just wanted to make sure that you're okay with it :)

Duske:
> e.g. scope the name to the corresponding component

The [configuration object](https://github.com/neos/neos-ui/pull/1124/files#diff-8d8c129d1fbe4d0e5d3cb6cfbcb0074aR4) contains the properties `aspectRatioDropdownContents ` and `dropdownContents `, which are dealing with the same stuff but are in located in different components. They needed a different name because of the flat hierarchy. So we could scope them to the belonging component:
```
zIndex:{
  aspectRatio: {
    dropdownContents: '1'
  }
  someOtherComponent: {
    dropdownContents: '2'
  }
}

```

> Maybe this out of scope

Sounds good, just wanted to make sure you're okay with it :)

dimaip:
oh you know what actually could make sense: group zIndexes around stacking contexts they divide into. Wdyt?

dimaip:
Smth like this could be helpful to understand it: https://chrome.google.com/webstore/detail/z-context/jigamimbjojkdgnlldajknogfgncplbh?hl=en

dimaip:
Also relevant: https://www.smashingmagazine.com/2014/06/sassy-z-index-management-for-complex-layouts/

Duske:
Just to verify that we're on the same page:
We group the various zIndex properties by their stacking contexts. So every element with a z-index which `parentStackingContext ` is `html` is set on the root level of zIndex. The property for these elements is called `context`. Child elements with z-index properties will be put inside.
```
zIndex:{
  dropdownContext: { // new stacking context
    context: '1',
    contents: '2',
    button: '3'
  }
}
```

Do you mean something like that? Should we receive the indices by using lists as described in the post?

dimaip:
Yep. I think it makes total sense. I hate those 99999 indexes really.

Duske:
I'm trying to group them by their stackingContext but its kinda hard. For example sometimes a a checkboxInput component has `html` as parentStackingContext, sometimes it is the surrounding dropdown component. Most of the components seem to have `html` as parentStackingContext, so the z-index is just 1. 
This results in a unnecessary complex structure for these simple cases. You can see the current work of the configuration object in [this gist](https://gist.github.com/Duske/45d4640bf957d1199f677f76b22206ba).

To sum it up: It reveals, just as you assumed, that most of these high indices are not needed. But structuring by their stackingContexts seems to be quite complicated and ambiguous.

Duske:
I've finished the grouping and used arrays to describe the z-index as you pointed out in Slack.
Since I'm not sure where `nodeToolBar`  is used, its z-index was not touched

Duske:
I think thats a good idea :) The names become quite long and a delimiter helps to read it more clearly

packages/neos-ui/src/Containers/RightSideBar/Inspector/style.css
```diff
@@ -19,7 +19,7 @@
     left: 0;
     width: calc(100% - var(--sidebarWidth));
     height: 100%;
-    z-index: 10000;
+    z-index: var(--zIndexUnappliedChangesOverlay);
```
dimaip: This one still doesn't have delimiting dashes

packages/neos-ui-inspector/src/SecondaryInspector/style.css
```diff
@@ -29,5 +29,5 @@
     font-size: 18px;
     border-right-width: 0;
     border-top-width: 0;
-    z-index: 10;
+    z-index: var(--zIndexSecondaryInspectorClose);
```
dimaip: same

dimaip:
@Duske so why is this still WIP?

Duske:
Oops you're right, I've updated the PR accordingly 😶

mstruebing:
@Duske Could you rebase your branch to the current master? I think there is nothing against merging it.

markusguenther:
Resolved the conflicts in a PR. Just because I am not able to modify the PR.

dimaip:
@Duske thanks a lot for your amazing effort! Will try to finally merge it with @markusguenther via his fork: #1361

PR #1131, opened 2017-10-11
FEATURE: Display Node Path in reference editor

skurfuerst:
betriebswirtschaftslehre und versicherung 2017-10-11 14-50-25

### DIFF
--- a/packages/neos-ui-editors/src/Editors/Reference/style.css
+++ b/packages/neos-ui-editors/src/Editors/Reference/style.css
@@ -1,0 +1,12 @@
+.referenceOption {
+    padding: var(--halfSpacing);
+    box-sizing: content-box;
+    height: var(--goldenUnit);
+    line-height: 20px;
+}
+
+.referenceOption__link {
+    color: var(--brandColorsContrastBright);
+    display: block;
+    font-size: 12px;
+}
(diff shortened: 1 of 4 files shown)


dimaip:
On the screenshot it seems right padding is missing?

packages/neos-ui-editors/src/Editors/Reference/style.css
```diff
@@ -0,0 +1,12 @@
...
+    padding: var(--halfSpacing);
+    box-sizing: content-box;
+    height: var(--goldenUnit);
+    line-height: 20px;
+}
+
+.referenceOption__link {
+    color: var(--brandColorsContrastBright);
+    display: block;
+    font-size: 12px;
```
dimaip: I'm a bit concerned how we use font-sizes around the codebase.
I will create an issue to gather all of these special font usages into a separate react-ui-components component, and use them to share styles. Otherwise our styles will quickly turn into a mess.

skurfuerst: :+1:

dimaip:
So what about missing right padding?

skurfuerst:
will fix the right padding

PR #1133, opened 2017-10-11
FEATURE: Frontend configuration API

dimaip:
This change introduces a place in settings which allows to transfer some configuration to UI components, including 3rd-party components.

E.g.:
```
Neos:
  Neos:
    Ui:
      frontendConfiguration:
        'Your.Own:Package':
          someKey: someValue
```

Then it may be accessed as:
```js
globalRegistry.get('frontendConfiguration').get('Your.Own:Package').someKey
```

I guess we should declare it as an experimental API that may change in the future in a breaking way.

TODO:

- [x] rename RenderState object to smth like RenderConfiguration


dimaip:
Usage example: https://github.com/psmb/Psmb.CreateNodeButton/commit/1abc4ab3b59b3a844c2f11e3b77a53e3da9056ba

PR #1142, opened 2017-10-13
BUGFIX: ensure JS backend information is not stale (fixes node tree after node creation)

skurfuerst:
## Problem Description

- create a document node A underneath root node R
- navigate back to root node R
- the node A disappears from the document node tree.

After a cache flush OR a modification on root node R, the node stays visible.

However, on the demo page, the cache is flushed nevertheless because of the tag
`NodeType_Neos.Neos:Document`, which the Menu Cache Segment listens to. If you
place a Menu on the page, then this menu has its own cache segment, which gets
flushed on ANY Document change -- and then we re-render the full page, because
ContentCache::replaceCachePlaceholders returns FALSE as soon as it cannot find
a single sub-entry.

In order to test the issue on the Demo page, you need to:
- remove all Menu and DimensionMenu elements from Fusion
- remove the Neos.Seo package (because that internally also uses a Menu)

## Bugfix

We need to ensure the JS backend information is always up to date, especially
when child nodes change.
Thus, we ensure that the JS backend information creates an own cache entry,
which is flushed whenever children are modified (using "DescendantOf_*")

### DIFF
--- a/Resources/Private/Fusion/Prototypes/Page.fusion
+++ b/Resources/Private/Fusion/Prototypes/Page.fusion
@@ -36,6 +36,28 @@ prototype(Neos.Neos:Page) {
                 inBackend = ${documentNode.context.inBackend}
                 newBackend = ${Neos.Ui.Activation.enableNewBackend()}
             }
+
+            // We need to ensure the JS backend information is always up to date, especially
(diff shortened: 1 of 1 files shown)


skurfuerst:
https://github.com/neos/neos-ui/issues/966

PR #1152, opened 2017-10-17
BUGFIX: Fix navigation in tree after an in-page navigation

skurfuerst:
## How to reproduce

- Go to page A in document tree
- navigate to page B in content area
- select page A in document tree again
- expected: selection to page A works
- actual: no navigation is triggered to PageA

## Bugfix

- ensure the redux state is always up to date
- do not trigger page refreshes if the URL is already correct in the iFrame

Additionally, the loading indicator does not need to be handled through SET_SRC
anymore, as this is now implemented differently (so that's a code cleanup).

### DIFF
--- a/packages/react-ui-components/src/Frame/frame.js
+++ b/packages/react-ui-components/src/Frame/frame.js
@@ -5,6 +5,7 @@ import ReactDOM from 'react-dom';
 
 export default class Frame extends PureComponent {
     static propTypes = {
+        src: PropTypes.string,
         mountTarget: PropTypes.string.isRequired,
         contentDidUpdate: PropTypes.func.isRequired,
         onLoad: PropTypes.func,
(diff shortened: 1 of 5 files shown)


packages/react-ui-components/src/Frame/frame.js
```diff
@@ -13,6 +14,18 @@ export default class Frame extends PureComponent {
...
         this.ref = ref;
+    };
+
+    componentDidMount() {
+        this.componentDidUpdate();
+    }
+
+    // we do not use react's magic to change to a different URL in the iFrame, but do it explicitely (in order to avoid reloads if we are already on the correct page)
+    componentDidUpdate() {
+        const win = ReactDOM.findDOMNode(this).contentWindow; // eslint-disable-line react/no-find-dom-node
```
dimaip: why not use a ref instead of findDOMNode?

skurfuerst: would be also fine. I can change it to ref.

packages/react-ui-components/src/Frame/frame.js
```diff
@@ -13,6 +14,22 @@ export default class Frame extends PureComponent {
 
     handleReference = ref => {
         this.ref = ref;
+    };
+
+    componentDidMount() {
+        this.componentDidUpdate();
```
dimaip: Please not like this! Extract the logic that should be executed both on mount and update into another method and call it from two places.

skurfuerst: ok if you like ^^

Issue #1150, opened 2017-10-17
ContentCollection empty handles cannot be removed


skurfuerst:
(needed for Punkt.de)

PR #1153, opened 2017-10-17
BUGFIX: If a content collection is rendered too big, do not display empty-collection-overlay

skurfuerst:
For custom rendering cases of Content Collections, is is weird
to have the "empty placeholder" rendered even though the content
collection is perfectly visible and selectable.

We now determine additionally based on the visual height of the collection 
whether the placeholder should be rendered or not.

Fixes: #1150

### DIFF
--- a/packages/neos-ui-guest-frame/src/dom.js
+++ b/packages/neos-ui-guest-frame/src/dom.js
@@ -135,18 +135,21 @@ export const markNodeAsVisible = contextPath => {
 
 //
 // Insert a placeholder element for content collections that don't have
-// any children yet
-//
+// any children yet and have a very small height (as they would not be clickable / selectable otherwise).
+// NOTE: If the element is "big enough" (i.e. more than 20 px), we do not render the placeholder either; as then
+// the user will very likely have created his own rendering.
 export const createEmptyContentCollectionPlaceholderIfMissing = collectionDomNode => {
     if (collectionDomNode) {
         const hasChildNodes = Boolean(
             collectionDomNode.querySelector('[data-__neos-node-contextpath]')
         );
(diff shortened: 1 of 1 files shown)


packages/neos-ui-guest-frame/src/dom.js
```diff
@@ -135,18 +135,21 @@ export const markNodeAsVisible = contextPath => {
 
 //
 // Insert a placeholder element for content collections that don't have
-// any children yet
-//
+// any children yet and have a very small height (as they would not be clickable / selectable otherwise).
+// NOTE: If the element is "big enough" (i.e. more than 20 px), we do not render the placeholder either; as then
+// the user will very likely have created his own rendering.
```
dimaip: I'm not so sure. ContentCollection overlay is a familiar affordance for the users. If we remove it, the integrators will need to re-implement it themselves.
It might make sense so there would be an option to not render the overlay, if the integrator wants it, but I'd rather not change the default behavior.
Anyways, it's just my opinion, might not see something right away.

skurfuerst: "It might make sense so there would be an option to not render the overlay, if the integrator wants it, but I'd rather not change the default behavior." - hm, I always fear that creating a new option is a bit of pain (creating maintenance burden etc etc); so if we can "intelligently" do it, that'd be my preference.

Could we merge this change and see if there are cases where the user misses the CC Overlays? I personally would not have noticed the change in our projects; but we probably need to see...

skurfuerst: (I hope it is OK that I merge this now, and we can discuss it afterwards // see if it breaks for people)

dimaip: Ok, let's see, easy to revert in any case. Will also test on my projects.

Issue #953, opened 2017-09-12
UI does not work when Neos is installed in a path (domain.com/path/neos!) 


Torsten85:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

It seems like all endpoint urls a hardcoded to `/neos!/...` like e.g. here: https://github.com/neos/neos-ui/blob/c7332968ccbb7366284b5417cb093abccd128fc9/packages/neos-ui-backend-connector/src/Endpoints/index.js#L6

This prevents the new UI to work properly in a path.

Also it is necessary to fix the LinkingService to create correct absolute node urls as described in my pull request here:
https://github.com/neos/neos-development-collection/pull/1695

PR #1163, opened 2017-10-19
BUGFIX: change all endpoint uris to start from root

dimaip:
Without this if you pick an image from media browser, the endpoint
request would be prefixed with the path of the media module, e.g.:
`/neos/media/browser/neos/content/image-with-metadata?image=***`

For some reason that happened only in Firefox for me.


johannessteu:
need to be adjusted again for #953

dimaip:
Sure, was just a quick fix

johannessteu:
jip, just wanted to mention the issue as a marker so we see this change in the issue

Issue #1166, opened 2017-10-20
Text overlaps chevron in SelectBox


dimaip:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

![image](https://user-images.githubusercontent.com/837032/31827899-795193a0-b5c1-11e7-96be-9f758c042014.png)

dimaip:
@crazydrecksack hey, do you still want to take care of things like these?

dimaip:
![image](https://user-images.githubusercontent.com/837032/31950920-66fef86a-b8e5-11e7-8dc9-6d3d027ebb5d.png)

mstruebing:
The issue of the second screenshot still exists

dimaip:
Yes, I just closed by accident and the re-opened.

PR #1169, opened 2017-10-23
BUGFIX: Various UX glitches

mstruebing:
adresses #1166 

Fixes the first and third mentioned issue there.
I can't reproduce the second one so it's hard to fix this.

### DIFF
--- a/packages/react-ui-components/src/CheckBox/style.css
+++ b/packages/react-ui-components/src/CheckBox/style.css
@@ -25,7 +25,6 @@
     border: 1px solid var(--brandColorsContrastBrighter);
     vertical-align: text-bottom;
     background: var(--brandColorsContrastDark);
-    color: var(--brandColorsContrastBrightest);
     font-size: 20px;
     font-weight: bold;
     color: #5b5b5b;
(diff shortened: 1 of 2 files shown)


dimaip:
Could you show how new checkboxes look please?

mstruebing:
checked
unchecked

packages/react-ui-components/src/CheckBox/style.css
```diff
@@ -25,7 +25,6 @@
     border: 1px solid var(--brandColorsContrastBrighter);
     vertical-align: text-bottom;
     background: var(--brandColorsContrastDark);
-    color: var(--brandColorsContrastBrightest);
     font-size: 20px;
     font-weight: bold;
     color: #5b5b5b;
```
dimaip: Btw, what's this color 5b5b5b? Haven't seen it anywhere before.
If it's really needed I guess we should add it to color constants, no?

mstruebing: It's color of the checkbox icon of the old ui. I'll add it to the color constants.

dimaip: Does it look really different from `brandColorsContrastBrightest` or any of other colors that we have? Maybe we can leave it normalize? Don't know, just asking.

mstruebing: bright

This would be bright. The other ones wouldn't fit. 
Do you think this is ok? Otherwise we need to merge this: https://github.com/neos/brand/pull/9

mstruebing: Anyway, @crazydrecksack merged my color PR, so I will use this.

PR #1175, opened 2017-10-24
BUGFIX: reset keys of publishableNodes after filtering

dimaip:
Otherwise if some entries were filtered out publishableNodes will get rendered as object to JSON, and not an array, which will break publishing.


dimaip:
Merged as a critical no-brainer.

Issue #1067, opened 2017-09-28
Table dialog is broken


grebaldi:
![screenshot_2017-09-28_17-49-54](https://user-images.githubusercontent.com/2522299/30976610-855a4c62-a475-11e7-8881-2a21aabb00a3.png)

skurfuerst:
I'll look into this

dimaip:
Also currently it's not possible to delete rows/columns, merge cells etc.
Probably we need smth like this? https://ckeditor.com/cke4/addon/tabletools

dimaip:
Sorry, I accidentally fixed it while working on smth else.

Issue #1009, opened 2017-09-24
CKEditor asks for non-existent CSS resource


grebaldi:
When loading the backend, there's a request to
```
/_Resources/Static/Packages/Neos.Neos.Ui/Plugins/ckeditor/skins/minimalist/editor.css
```
that originates from CKEditor. The Resource cannot be found.

We can fix this by just creating an empty CSS file at that location, but it might be sensible nonetheless to find out, why CKEditor is asking for the file.

dimaip:
I still wonder, why we don't use CKE from npm?
https://www.npmjs.com/package/ckeditor

PR #1176, opened 2017-10-24
BUGFIX: use correct skin in CKE and enable tabletools plugin

dimaip:
Fixes: #1067 #1009

Our skin was not used currently.
Besides fixing it I enabled `tabletools` plugin to allow editing tables:
image

It looks ugly as hell, but I don't think we should invest a lot into fixing it, since CKE5 will have much nicer implementation: https://github.com/ckeditor/ckeditor5/issues/610


dimaip:
I just merge this for now to have tables working in beta6. If you hate it and want to do smth about it -- go ahead ;)

PR #1195, opened 2017-10-26
TASK: Remove usage of `window.requestCallback()`

johannessteu:
This change removes the usage of `window.requestCallback()` as it is not supported in Safari right now.

### DIFF
--- a/packages/neos-ui-guest-frame/src/initializeGuestFrame.js
+++ b/packages/neos-ui-guest-frame/src/initializeGuestFrame.js
@@ -95,11 +95,20 @@ export default ({globalRegistry, store}) => function * initializeGuestFrame() {
             nodes
         });
 
-        // only of guest frame document did not change in the meantime, we continue initializing the node
-        if (getGuestFrameDocument() === node.ownerDocument) {
-            initializeCurrentNode(node);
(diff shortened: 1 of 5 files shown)


johannessteu:
Found a way to check if the function is available, can squash merge this one!

packages/neos-ui-guest-frame/src/initializeGuestFrame.js
```diff
@@ -95,13 +95,20 @@ export default ({globalRegistry, store}) => function * initializeGuestFrame() {
             nodes
         });
 
-        window.requestIdleCallback(() => {
-            // only of guest frame document did not change in the meantime, we continue initializing the node
+        if ('requestIdleCallback' in window) {
```
dimaip: Might make sense to add some kind of polyfil for requestIdleCallback, as it's a good idea to use it in other places.
In the simplest form you could just wrap this check into a function, and put it somewhere into utility or smth.

dimaip: See https://developer.mozilla.org/en-US/docs/Web/API/Background_Tasks_API#Falling_back_to_setTimeout

PR #1205, opened 2017-11-01
FEATURE: Improve ckeditor configurability

skurfuerst:
While building https://github.com/sandstorm/BlockStyles, I found the internal
API for configuring the RTE lacking some features. Specifically, the following
has changed with this PR:

## create more explicit inline editor configuration format

The old format looked like this:

```
ui.aloha:
  […]
    h2: true
    h1: false
```

This was then server-side transformed to:

```
aloha:
  […]: h2
```

-> i.e. false values were removed automatically.

Now, I wanted to make it possible to add specific options to the inline
editor instance (more than the "TRUE/FALSE" from above).

Thus, I created a new format:

```
ui.inline.editorOptions:
  formatting:
    h2: true
    h1: false
  placeholder: "placeholder string"
  autoparagraph: true
```

… which is used internally throughout the codebase; and the old "aloha"
format is converted to the new format.

This enables to pass in arbitrary objects for configuring a specific
editor instance.

## make FrontendConfiguration available in manifests

We now load the FrontendConfiguration early and pass them into the manifest

## !!! change way Store is passed to manifest

before, the store was passed to the manifest as 2nd parameter. To make this
more extensible, we now pass an object as 2nd parameter, currently containing
store and frontendConfiguration

### DIFF
--- a/packages/neos-ui/src/index.js
+++ b/packages/neos-ui/src/index.js
@@ -67,11 +67,16 @@ function * application() {
     yield system.getNeos;
 
     //
+    // Load frontend configuration very early, as we want to make it available in manifests
+    //
+    const frontendConfiguration = yield system.getFrontendConfiguration;
+
+    //
     // Initialize extensions
     //
     manifests
         .map(manifest => manifest[Object.keys(manifest)[0]])
-        .forEach(({bootstrap}) => bootstrap(globalRegistry, store));
+        .forEach(({bootstrap}) => bootstrap(globalRegistry, {store, frontendConfiguration}));
 
     const configuration = yield system.getConfiguration;
 
@@ -110,10 +115,6 @@ function * application() {
     const i18nRegistry = globalRegistry.get('i18n');
     i18nRegistry.setTranslations(translations);
 
-    //
-    // Load frontend configuration (edit/preview modes)
-    //
-    const frontendConfiguration = yield system.getFrontendConfiguration;
     const frontendConfigurationRegistry = globalRegistry.get('frontendConfiguration');
 
     Object.keys(frontendConfiguration).forEach(key => {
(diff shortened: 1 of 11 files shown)


dimaip:
Like what you are doing here, thought about the same things too.

skurfuerst:
@grebaldi please review; especially the 3rd part "change way Store is passed to manifest" is breaking for you I guess (as you introduced it).

packages/neos-ui/src/index.js
```diff
@@ -67,11 +67,16 @@ function * application() {
...
+    //
+    const frontendConfiguration = yield system.getFrontendConfiguration;
+
+    //
     // Initialize extensions
     //
     manifests
         .map(manifest => manifest[Object.keys(manifest)[0]])
-        .forEach(({bootstrap}) => bootstrap(globalRegistry, store));
+        .forEach(({bootstrap}) => bootstrap(globalRegistry, {store, frontendConfiguration}));
```
grebaldi: I'm generally okay with the second parameter being an object here. But I'd prefer to initialize the `frontendConfigurationRegistry` (see L118) before running the manifests, so the frontendConfiguration can be accessed that way - since this our mechanism of providing globals.

skurfuerst: @grebaldi I think that won't work, because the manifest.js is *creating* the frontendConfiguration registry; so that'd be somehow a "chicken-egg" problem. See? :-)

What we could do is the following:
- pass the frontendConfiguration to the registries as above in my change
- move the initialization logic (L118ff) into the manifest.js, right where the frontendConfiguration registry is created.

This way, other registries could rely on frontend configuration being available and initialized in their manifest.

skurfuerst:
See my last inline comment plz :)

packages/neos-ui/src/index.js
grebaldi: Agreed, this would be the easiest way for now. But I think this shows an issue with our initialization logic... We should address that after the release.

Issue #1137, opened 2017-10-12
Surround inspector editors and views with error boundaries, after upgrade to React 16


dimaip:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

The editors should fail without killing the whole UI and the error message should be shown.
See https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html
Related to #1083

Issue #979, opened 2017-09-15
Avoid overriding the constructor of React class extends by using class fields


Inkdpixels:
I just skimmed through the codebase a bit and found [a lot of usages of the typical `.bind(this)`](https://github.com/neos/neos-ui/search?utf8=%E2%9C%93&q=.bind%28this%29&type=) in the constructor's of React classes. To reduce the complexity and repetition we could just switch to [class fields](https://github.com/tc39/proposal-class-public-fields) in combination with arrow functions since they are automatically bound to the parent context :tada: 

In the same way I would also propose to assign the initial state to remove the usage custom constructors at all.

*before:*
```js
import React, {PureComponent} from 'react';

export default class Doge extends PureComponent {
    constructor(props) {
        super(props);

        this.handleWow = this.handleWow.bind(this);

        this.state = {isMuchCleaner: false};
    }

    render() {
        return (
            ...
        );
    }

    handleWow() {
        this.setState({isMuchCleaner: true});
    }
}
```

*after:*
```js
import React, {PureComponent} from 'react';

export default class Doge extends PureComponent {
    state = {isMuchCleaner: true};

    render() {
        return (
            ...
        );
    }

    handleWow = () => {
        this.setState({isMuchCleaner: true});
    }
}
```

dimaip:
Sure, why not.

dimaip:
I think we've resolved this issue by now. Thanks for the cool tip @Inkdpixels!

PR #1206, opened 2017-11-03
TASK: Inspector editors have error boundary

kitsunet:
With React 16 error boundaries were introduced to limit the effect
of errors on an react app. The EditorEnvelope is now a natural
error boundary and so effectively keeps errors in inspector editors
to the editor.

Fixes: #1137

### DIFF
--- a/packages/neos-ui-editors/src/EditorEnvelope/index.js
+++ b/packages/neos-ui-editors/src/EditorEnvelope/index.js
@@ -9,6 +9,8 @@ import style from './style.css';
     editorRegistry: globalRegistry.get('inspector').get('editors')
 }))
 export default class EditorEnvelope extends PureComponent {
+    state = {};
+
     static propTypes = {
         identifier: PropTypes.string.isRequired,
         label: PropTypes.string.isRequired,
@@ -22,11 +24,6 @@ export default class EditorEnvelope extends PureComponent {
 
         commit: PropTypes.func.isRequired
     };
-
-    constructor(props) {
-        super(props);
-        this.state = {};
-    }
 
     generateIdentifier() {
         return `#__neos__editor__property---${this.props.identifier}`;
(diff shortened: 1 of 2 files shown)


packages/neos-ui-editors/src/EditorEnvelope/index.js
```diff
@@ -62,6 +75,10 @@ export default class EditorEnvelope extends PureComponent {
     }
 
     render() {
+        if (this.state.error) {
+            return <div>{this.state.error && this.state.error.toString()}</div>;
```
dfeyer: Double test for this.state.error, no ?

packages/neos-ui-editors/src/EditorEnvelope/index.js
```diff
@@ -44,6 +49,14 @@ export default class EditorEnvelope extends PureComponent {
         return (<div>Missing Editor {editor}</div>);
     }
 
+    componentDidCatch(error, errorInfo) {
+        console.log(error);
```
dimaip: Should it be `console.error` then?

dimaip: Also could be `console.error(error, errorInfo)`

packages/neos-ui-editors/src/EditorEnvelope/index.js
```diff
@@ -22,6 +22,11 @@ export default class EditorEnvelope extends PureComponent {
         commit: PropTypes.func.isRequired
     };
 
+    constructor(props) {
+        super(props);
```
dimaip: see https://github.com/neos/neos-ui/issues/979

packages/neos-ui-editors/src/EditorEnvelope/index.js
dimaip: Should we give it some screaming red colors here? :scream_cat:

dfeyer: Any possibility to have an empty string here ? So maybe a default message can help ?

packages/neos-ui-editors/src/EditorEnvelope/index.js
kitsunet: 👍

PR #1218, opened 2017-11-05
TASK: Export data as JSON

kitsunet:
This exports the initial backend data as JSON directly instead
as HTML strings. This prevents double escaping (and potential issues
related to that) as well as parsing in our JS.

Values are stored in global JS variables until read and then deleted.


kitsunet:
These double escaped JSON strings were a thorn in my side since I saw them first. I like this way more (should probably also be faster), but obviously up for discussion.

PR #1219, opened 2017-11-05
TASK: Remove constructors from the neos-ui-editors package

kitsunet:
This removes constructors from all components in the `neos-ui-editors`
package and uses arrow functions as class properties instead
of binding. This makes the code a bit more concise.

Related: #979


kitsunet:
Will continue with other packages, but wanted to make smaller PRs instead of one huge thing.

mstruebing:
I would suggest one PR per package, like you currently did :+1:

dimaip:
Thanks for taking care!

PR #1220, opened 2017-11-06
TASK: Remove contructors from react-ui-components

kitsunet:
This removes constructors from all components in the
react-ui-components package and uses arrow functions
as class properties instead of binding. This makes
the code a bit more concise.

Related: #979


kitsunet:
Note: I was very liberal with removing. There were two places were props were put in state in the constructor. I removed that because I felt in both cases it was not really needed, but make sure you check.

kitsunet:
Apparently I broke some tests... Will look into it. Might need to bring one or two constructors back...

Inkdpixels:
@kitsunet Could I support you in this PR somehow?

kitsunet:
I just forgot 😆 Feel free to take over, I am right now entangled in something else... Can do later htough.

Inkdpixels:
@kitsunet alrighty, will take over, thanks a bunch :) :heart:

packages/neos-ui-guest-frame/src/InlineUI/NodeToolbar/index.js
```diff
@@ -37,10 +37,7 @@ export default class NodeToolbar extends PureComponent {
         isSticky: false
     };
 
-    constructor() {
-        super();
-        this.iframeWindow = document.getElementsByName('neos-content-main')[0].contentWindow;
-    }
+    iframeWindow = document.getElementsByName('neos-content-main')[0].contentWindow;
```
grebaldi: Maybe we can use this opportunity to replace this call with the helper method from `@neos-project/neos-ui-guest-frame/src/dom`

PR #1226, opened 2017-11-08
BUGFIX: correctly use default options in Select Boxes and Reference(s) editors

skurfuerst:
This has never worked before because the defaultOptions are static; so we
need to reference them statically as well.


dimaip:
Good catch!

Issue #867, opened 2017-09-04
Add extension point for overriding element Rendering in Node Trees


skurfuerst:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

One of our core feature sponsors needs a way to manipulate the rendering of the TreeNodes; i.e. he wants to display additional icons based on custom property values.

In order to support that, we at least need to put the tree node React elements in the registry, so they can be overridden from the outside.

dimaip:
Hmm I like the idea, but at the same time we should be careful about creating such ad-hoc APIs as we'd need to maintain them.
Can we maybe declare it as an unplanned extensibility?

skurfuerst:
yes, for sure, I'd definitely declare it as unplanned extensibility!

dimaip:
@skurfuerst can we close it now since you have this fancy extensibility thing in place?

mstruebing:
Not exactly, you'd have to override the whole tree, at the current state it's not possible to only override the tree elements.
I've experienced this while developing this: https://github.com/sitegeist/Sitegeist.TrafficLights

UI-specific code: https://github.com/sitegeist/Sitegeist.TrafficLights/tree/master/Resources/Private/TrafficLights/src

dimaip:
Ok, let's reopen then

PR #1230, opened 2017-11-08
TASK: Improve unplanned extensibility

skurfuerst:
Here is an example how the changes in here can be used:

https://github.com/sandstorm/NodeTreeStyling


skurfuerst:
I'd say this resolves https://github.com/neos/neos-ui/issues/867

Issue #1234, opened 2017-11-13
Move the babel and jest configurations into separate packages to ease the maintenance


Inkdpixels:
A smaller nitpick would be to move out the babel and jest configurations into preset packages within the mono-repo, that way we reduce the maintenance when adding / removing parameters.

PR #1240, opened 2017-11-13
TASK: Create preset packages and increase the test coverage

Inkdpixels:
Resolves: #1234 

- [ ] Use snapshot tests for all redux related code
- [ ] Use snapshot tests for all redux-sagas
- [ ] Use snapshot tests for all react components
- [ ] Use snapshot tests for the react-redux glue code(mapStateToProps)
- [x] Switch from sinon to jest's internal mock built-ins
- [ ] Use `describe` and `it` blocks to have more flexibility in regards to setting and cleaning up tests


Inkdpixels:
@mstruebing Thanks for the info, will take care of that! :heart:

PR #1255, opened 2017-11-14
TASK: Remove inline script tags after initialization

### LINKED ISSUE #1243: Throw away node metadata script tags after initialization
We're facing some interference with the script tags when :nth-child selectors or flexbox are used in the iframe.

Since we only need them during initialization it would be feasible to just remove the script tags afterwards. We'd still have a flash of wrongly styled content, but then at least the styles wouldn't be broken.

johannessteu:
Resolves #1243

### DIFF
--- a/Classes/Neos/Neos/Ui/Aspects/AugmentationAspect.php
+++ b/Classes/Neos/Neos/Ui/Aspects/AugmentationAspect.php
@@ -155,7 +155,7 @@ class AugmentationAspect
         $this->userLocaleService->switchToUILocale(true);
 
         $wrappedContent = $this->htmlAugmenter->addAttributes($content, $attributes, 'div');
-        $wrappedContent .= "<script>(function(){(this['@Neos.Neos.Ui:Nodes'] = this['@Neos.Neos.Ui:Nodes'] || {})['{$node->getContextPath()}'] = {$serializedNode}})()</script>";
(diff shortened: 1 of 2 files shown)


Classes/Neos/Neos/Ui/Aspects/AugmentationAspect.php
```diff
@@ -155,7 +155,7 @@ public function contentElementAugmentation(JoinPointInterface $joinPoint)
         $this->userLocaleService->switchToUILocale(true);
 
         $wrappedContent = $this->htmlAugmenter->addAttributes($content, $attributes, 'div');
-        $wrappedContent .= "<script>(function(){(this['@Neos.Neos.Ui:Nodes'] = this['@Neos.Neos.Ui:Nodes'] || {})['{$node->getContextPath()}'] = {$serializedNode}})()</script>";
+        $wrappedContent .= "<script data-neos-inlinescript='true'>(function(){(this['@Neos.Neos.Ui:Nodes'] = this['@Neos.Neos.Ui:Nodes'] || {})['{$node->getContextPath()}'] = {$serializedNode}})()</script>";
```
dimaip: we forgot to test, is `='true'` needed here. Probably not.

dimaip: And yes, naming. This is not just inline script. It's more like `data-neos-nodedata`.

johannessteu:
It's indeed not needed @dimaip. Adjusted. As well as the naming.

Issue #1257, opened 2017-11-15
After typing "<!--" in the editor, the backend breaks


skurfuerst:
If I type `<!--` in an inline editor and I reload the bckend, I cannot change the node anymore.

kitsunet:
Can confirm, this bug exists.

kitsunet:
we set config.basicEntities to `false` which is strongly discouraged by CK, don't know if that is on purpose, will try if something obviously breaks if I re-enable it.

skurfuerst:
thanks <3 maybe check via git blame when it was set to False :) @kitsunet

kitsunet:
Yep ofc, waiting for @dimaip because he set it but the reasons are unclear to me. Will check with him. Definitely fixes the issues though, so much I can say already.

skurfuerst:
Thanks <3

PR #1261, opened 2017-11-15
BUGFIX: Enable basic entities encoding

kitsunet:
This (re-) enables basic HTML entity encoding, to avoid
for example opening HTML comments breaking the site.

Fixes: #1257


kitsunet:
In person 👍 from @dimaip

PR #1268, opened 2017-11-15
BUGFIX: fix inclusion of CSS in extensibility modules

skurfuerst:
the webpack.config.js in @neos-project/neos-ui-extensibility makes use
of the ExtractTextPlugin.

however, if for some reasons, the instances used in "module.loaders" and
"plugins" are not the same ones, you get the folllowing error message 
when building plugins which need to load CSS:

	"Module build failed: Error: "extract-text-webpack-plugin" loader is
	 used without the corresponding plugin"

This happened for me in all cases where I want to load CSS in a custom extension.

### DIFF
--- a/packages/build-essentials/src/webpack.config.js
+++ b/packages/build-essentials/src/webpack.config.js
@@ -126,4 +126,15 @@ if (env.isProduction) {
     }));
 }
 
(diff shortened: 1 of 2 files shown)


skurfuerst:
Funny. not needed anymore when using "yarn" instead of npm to install the packages.... go figure. strange.

PR #1272, opened 2017-11-15
FEATURE: create new node from references editor

dimaip:
Create nodes straight from reference(s) editor in a predefined location of a predefined type.

Useful for quickly adding tags or authors to posts.

The "Create new" option is shown when `path` and `type` editor options are specified.

Example config:
```
  properties:
    references:
      type: 'references'
      ui:
        inspector:
          editorOptions:
            createNew:
              path: '/sites/neosdemo'
              type: 'Neos.Neos:Shortcut'
              titleProperty: 'name' # optional, defaults to title
```


packages/neos-ui-editors/src/Editors/Reference/referenceDataLoader.js
```diff
@@ -0,0 +1,134 @@
+import React, {PureComponent} from 'react';
+import PropTypes from 'prop-types';
+import {connect} from 'react-redux';
+import {$get, $transform} from 'plow-js';
+import {selectors} from '@neos-project/neos-ui-redux-store';
+import {neos} from '@neos-project/neos-ui-decorators';
+
+export default isMulti => WrappedComponent => {
```
Inkdpixels: Maybe wrap the `isMulti` bool into an object? Would be better readable when using the HOC and it's extendable in the future! :) :100:

Issue #1262, opened 2017-11-15
MasterPluginEditor and PluginViewsEditor is missing


samuelhauser:
The `MasterPluginEditor` and the `PluginViewsEditor` is missing

markusguenther:
I try that :)

markusguenther:
After PluginViewsEditor is merged, just the PluginViewEditor is missing.

So it turns out that three editors are missing and not just two.

PR #1276, opened 2017-11-16
Add the MasterPluginEditor

markusguenther:
The MasterPluginEditor was missing at the moment. So this PR will add the Editor.
The PR solves one part of the issue #1262 

Related: #1262


Resources/Private/Translations/en/Main.xlf
```diff
@@ -20,6 +20,12 @@
 			<trans-unit id="imageCropper__aspect-ratio-placeholder" xml:space="preserve">
 				<source>Choose an Aspect Ratio</source>
 			</trans-unit>
+			<trans-unit id="content.inspector.editors.masterPluginEditor.selectPlugin" xml:space="preserve">
+				<source>Select a Plugin</source>
```
dimaip: Just wondering, can't we use labels from the old UI here?

dimaip:
Looks good at first glance, left a comment.

Resources/Private/Translations/en/Main.xlf
markusguenther: Sure but yesterday someone mentioned that the dependencies should be low so I thought it is better to have the labels. But can change that.

markusguenther:
Yeah still struggling with the data fetching... the dimensions should be in a different format on PHP side.

PR #1287, opened 2017-11-16
BUGFIX: clear search value on select and createNew in SelectBox

dimaip:


packages/react-ui-components/src/SelectBox/selectBox.js
```diff
@@ -180,6 +180,22 @@ export default class SelectBox extends PureComponent {
...
+            this.props.onSearchTermChange('');
+        }
+    }
+
+    handleCreateNew = (...rest) => {
+        this.props.onCreateNew(...rest);
+        // Clear search box on creating new
+        if (this.props.onSearchTermChange) {
+            this.props.onSearchTermChange('');
+        }
```
Inkdpixels: Just a little nitpick, but most probably you could create another method that holds this optional reset of the `searchTerm` in one place instead of two :)

Issue #1277, opened 2017-11-16
Inspector can not handle type ``array``


gerhard-boden:
Currently the UI can not handle simple arrays within the inspector: 

Old UI:
![array_editor](https://user-images.githubusercontent.com/10533739/32888756-a7d5ed54-cac8-11e7-9eaf-76174e1b2c2c.PNG)

New UI:
![error_array_editor](https://user-images.githubusercontent.com/10533739/32888752-a33e7590-cac8-11e7-8037-06b4162971b0.PNG)

gerhard-boden:
Edit: @mficzel mentioned that this might be connected to a missing data source in the new UI. But we might need the old behaviour as well, because it was possible to change the order of values within arrays saved within the ``properties`` json

dimaip:
Blocked by #1362

skurfuerst:
@dimaip @gerhard-boden could you re-test plz? :)

dimaip:
Let's ask @gerhard-boden, since I don't have that kind of data in DB.

johannessteu:
@gerhard-boden is this still an issue?

gerhard-boden:
The error is the same I'm afraid `TypeError: Cannot convert undefined or null to object`. 

So the property is of type ``array`` and within the database the entry contains an JSON structured array (tried multiple variations like with and without index, but no luck)

markusguenther:
I tried to reproduce that and it works.

```
properties:
    styleOptions:
      type: array
      ui:
        label: 'Styling Options'
        inspector:
          group: 'document'
          editor: 'Neos.Neos/Inspector/Editors/SelectBoxEditor'
          editorOptions:
            allowEmpty: true
            placeholder: 'Select Styling Options'
            values:
              leftColumn:
                label: 'Show Left Column'
              rightColumn:
                label: 'Show Right Column'
```
                
               
               
<img width="320" alt="Screenshot 2021-10-12 at 22 21 28" src="https://user-images.githubusercontent.com/1014126/137023826-d05c459d-7d2d-4466-b756-d2cf4f29bc1f.png">

gerhard-boden:
I'm also pretty sure it's fixed by now, this stems was from the rather early days of the new UI :)

PR #1291, opened 2017-11-16
BUGFIX: Fix array handling for static data sources in inspector

gerhard-boden:
Adjust the ``SelectBoxEditor`` so it can handle simple ``array`` types as well

Fixes: https://github.com/neos/neos-ui/issues/1277


gerhard-boden:
Do not merge as of now we talked about utilizing the ``SelectBoxEditor`` in a different way (or not utilizing in that case)

PR #1292, opened 2017-11-16
BUGFIX: Fix selectbox storybook when no options exist

markusguenther:
The storybook for the selectbox component was broken.
When the selectbox get not options you get exceptions.

This PR fixes the selectbox component to fix the storybook

bildschirmfoto 2017-11-16 um 22 11 33


packages/react-ui-components/src/SelectBox/selectBox.js
```diff
@@ -171,6 +171,10 @@ export default class SelectBox extends PureComponent {
     createNewEnabled = () => this.props.onCreateNew && this.props.searchTerm;
 
     optionsCount = () => {
+        if (this.props.options === undefined) {
```
Inkdpixels: FYI I personally would not check against `undefined`, instead something like `Boolean(this.props.options) === false` is much safer

Inkdpixels:
FYI I added a `defaultProp` in this case since we check against the presence of the `options` prop in a bunch of cases, I would love if this could be double checked by @dimaip / @grebaldi or @skurfuerst.

PR #1295, opened 2017-11-16
BUGFIX: Make the reset button visible again

markusguenther:
The reset button was overplayed by the dropdown chevron. So this commit move the reset and loading button before the dropdown chevron.

Then it look like in the old UI.


markusguenther:
It turns out that the things behave differently in the storybook and the real neos UI.
So the distance of the button is way to far right in neos, but it fits in the storybook.

Issue #876, opened 2017-09-05
Asset Editor


kitsunet:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)**

PR #1296, opened 2017-11-16
FEATURE: AssetEditor

kitsunet:
Adds the AssetEditor for the inspector.

Resolves: #876

### DIFF
--- a/packages/neos-ui-editors/src/Editors/AssetEditor/index.js
+++ b/packages/neos-ui-editors/src/Editors/AssetEditor/index.js
@@ -83,9 +83,6 @@ export default class AssetEditor extends PureComponent {
         this.props.commit(value);
     }
 
-    handleCreateNew = () => {
-    }
-
     render() {
         return (<MultiSelectBox
             dndType={dndTypes.MULTISELECT}
@@ -93,7 +90,6 @@ export default class AssetEditor extends PureComponent {
             loadingLabel={this.props.i18nRegistry.translate('Neos.Neos:Main:loading')}
             displaySearchBox={true}
             optionComponent={AssetOption}
-            createNewLabel={this.props.i18nRegistry.translate('Neos.Neos.Ui:Main:createNew')}
             placeholder={this.props.i18nRegistry.translate(this.props.placeholder)}
             options={this.state.options || []}
             values={this.props.value}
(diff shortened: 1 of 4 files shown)


packages/neos-ui-editors/src/Editors/AssetEditor/index.js
```diff
@@ -0,0 +1,109 @@
...
+                });
+        }
+    }
+
+    handleValueChange = value => {
+        this.setState({searchOptions: [], searchTerm: ''});
+        this.props.commit(value);
+    }
+
+    handleCreateNew = () => {
```
gerhard-boden: do we need this empty method?

packages/neos-ui/src/manifest.dataloaders.js
```diff
@@ -192,8 +192,8 @@ manifest('main.dataloaders', {}, globalRegistry => {
             return Promise.all(
                 identifiers.map(identifier =>
                     this.resolveValue(options, identifier)
-                ).filter(promise => Boolean(promise)) // remove "null" values
-            );
+                )
+            ).then(results => [].concat(...results));
```
gerhard-boden: Is this 100% part of the feature PR?

packages/neos-ui-editors/src/Editors/AssetEditor/index.js
kitsunet: Currently yes, because it's a required prop, BUT I think that is actually a bug. I need to discuss it with @dimaip when he is in....

packages/neos-ui/src/manifest.dataloaders.js
kitsunet: Yes, otherwise this PR will not work and it was actually broken before.

gerhard-boden:
I've opened a couple of enhancement issues related to this PR:
#1304 
#1305 
#1306

packages/neos-ui-editors/src/Editors/AssetEditor/index.js
gerhard-boden: a right, because the ``MultiSelectBox`` needs the ``onCreateNew`` prop.. hmm... also think we should consider that within the ``MultiSelectBox``. But this is no blocker for now

Issue #1304, opened 2017-11-17
Open media model within asset list 


gerhard-boden:
In the old UI it was possible to pick assets from the media modal. Would be good to have that in the new UI as well 

relates: #1296

gerhard-boden:
Already worked on this one today, this is working, but not yet pushed

Issue #1306, opened 2017-11-17
Enable file upload for asset list


gerhard-boden:
In the old UI it was possible to directly upload images into the asset list. Would be nice to have that in the new UI as well.

relates: #1296

gerhard-boden:
Already worked on this one today, this is almost working, but not yet pushed

dimaip:
Seems like done, kind of.

PR #1316, opened 2017-11-18
FEATURE: Add file upload and asset selection in asset list editor

gerhard-boden:
This adds asset selection and file upload to the asset list editor.

It's also possible to just drag files onto the asset list.

#resolves: #1304 #1306

### DIFF
--- a/packages/neos-ui-editors/src/Editors/AssetEditor/Components/Controls/index.js
+++ b/packages/neos-ui-editors/src/Editors/AssetEditor/Components/Controls/index.js
@@ -1,0 +1,40 @@
+import React, {PureComponent} from 'react';
+import PropTypes from 'prop-types';
+import style from './style.css';
+import IconButton from '@neos-project/react-ui-components/src/IconButton/';
+import {neos} from '@neos-project/neos-ui-decorators';
+
+@neos(globalRegistry => ({
+    i18nRegistry: globalRegistry.get('i18n')
+}))
+export default class Controls extends PureComponent {
+    static propTypes = {
+        onChooseFromMedia: PropTypes.func.isRequired,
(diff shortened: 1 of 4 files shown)


gerhard-boden:
had no idea we had that, how "newly" is "newly"? When creating this I did similar like the ``ImageEditor`` handles uploads

kitsunet:
Which uses that upload component :)

https://github.com/neos/neos-ui/blob/master/packages/neos-ui-editors/src/Editors/Image/Components/PreviewScreen/index.js#L41

gerhard-boden:
Ah, haha 😄 Now that I see it, I remember it... I think I could also wrap the whole Asset List with that ``AssetUpload`` (like I currently do with the ``Dropzone``). Not sure why I did not do that in the first place, no hard reasons I hope.

kitsunet:
I think this can be change afterwards too.

dimaip:
Awesome! But would be nice to finish it the nice way before the release. @gerhard-boden Will you take care of it or should we just create a reminder ticket?

gerhard-boden:
@dimaip I'll take care of it, I just created a "reminder ticket" for myself
https://github.com/neos/neos-ui/issues/1329
Hopefully will do it tomorrow or even today. SHOULD not make a lot of trouble

dimaip:
@gerhard-boden no hurry now that we have it in, just wanted to make sure it doesn't get left behind. Thanks for taking care!

gerhard-boden:
The reason why I did not use ``AssetUpload`` in the first place was that it had no support for multiple files. This is also changed in https://github.com/neos/neos-ui/pull/1378 alongside with the proposed change for ``AssetList``

PR #1315, opened 2017-11-17
Feature: Plugin views editor

markusguenther:
Adding missing editor PluginViewsEditor.

bildschirmfoto 2017-11-17 um 17 04 39

Related: #1262

PR #1318, opened 2017-11-18
FEATURE: Plugin view editor

markusguenther:
This adds the last missing editor. The PR is a follow up of the PR and therefore does not based on the latest master.

_Related PR_: https://github.com/neos/neos-ui/pull/1315

Fixes: #1262


dimaip:
Hey, why is it a follow up, since the previous PR is not yet merged? Couldn't you just push the commits there or close the old PR?

kitsunet:
Yeah, both work. If we merge the other first it should be picked up by github and this PR should suddenly show less change.

Issue #1323, opened 2017-11-21
SelectBoxEditor is collapsed at the beginning


markusguenther:
Since the last codesprint the SelectBoxEditor changed the default behavior and is collapsed at the beginning. To make it easier for editors the default should be unfold again.

![ezgif com-video-to-gif-2](https://user-images.githubusercontent.com/1014126/33061423-5c2382a4-ce9c-11e7-834f-c088662bf2aa.gif)

PR #1326, opened 2017-11-21
TASK: Set isOpen default state to true

markusguenther:
The most ToggablePanel components does not provide the `onPanelToggle` prop and if the prop is provided, the component will not use the internal state, instead it will be controlled by the props.

The default state is false and so the most ToggablePanel does not open on default.
Therefore the default is changed to true.

Fixes: #1323


markusguenther:
Need to update snapshots, but don't know how to do that with the current setup.
The general jest needs just ```jest --updateSnapshot```

kitsunet:
Apparently you found out 😆

PR #1350, opened 2017-11-27
TASK: set button default type to button

kabarakh:
default type is submit if not given. Thus browsers are supposed to submit the form with the button which breaks react onClick handlers. By giving a default type of type="button" browsers won't submit automagically while we wait for react to do something (testable in Firefox)

### DIFF
--- a/packages/react-ui-components/src/Button/button.js
+++ b/packages/react-ui-components/src/Button/button.js
@@ -18,6 +18,7 @@ const Button = props => {
         hoverStyle,
         size,
         theme,
+        type,
         _refHandler,
         ...rest
     } = props;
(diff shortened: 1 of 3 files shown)


packages/react-ui-components/src/Button/button.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import mergeClassNames from 'classnames';
-import {makeFocusNode} from './../_lib/focusNode';
+import { makeFocusNode } from './../_lib/focusNode';
```
dimaip: No need for extra spaces here, breaks our code style

kabarakh: removed the spaces (was something phpStorm did automatically and I didn't even notice it...)
also removed the changes to yarn.lock ;)

dimaip:
Great! Also I think you need to update the snapshot test of the button component (`src/Button/button.spec.js`)

kabarakh:
Tests are green now, and I added one to see if the type can be overwritten

dimaip:
Yay!

Issue #1351, opened 2017-11-27
Root page type can't be changed


soee:
Hi,

i am trying the new UI and wanted to change page type of my root site to custom type `Vendor.Site:Home` but inside properties inspector the `Type` field that should be a `select` with available options is rendered as a single `div` with class `style__noOptionsAvailable___xxx`. 

After switching to old UI the same inspector field was rendered as a select element as it should.

It seems like it can be some bug in new UI.

PR #1352, opened 2017-11-27
BUGFIX: fix changing nodetype of a site node

dimaip:
Not the cleanest fix, but I think it's good enough workaround for this edge case.

Fixes: #1351

### DIFF
--- a/packages/neos-ui-editors/src/Editors/NodeType/index.js
+++ b/packages/neos-ui-editors/src/Editors/NodeType/index.js
@@ -22,8 +22,9 @@ import style from './style.css';
             subject: focusedNodeContextPath,
             reference: focusedNodeContextPath
         });
+        const isSiteNode = $get('cr.nodes.siteNode', state) === focusedNodeContextPath;
 
-        return {focusedNode, allowedSiblingNodeTypesForFocusedNode};
+        return {focusedNode, allowedSiblingNodeTypesForFocusedNode, isSiteNode};
     };
 })
 export default class NodeType extends PureComponent {
@@ -32,6 +33,7 @@ export default class NodeType extends PureComponent {
         commit: PropTypes.func.isRequired,
         highlight: PropTypes.bool,
 
+        isSiteNode: PropTypes.bool,
         focusedNode: PropTypes.object,
         allowedSiblingNodeTypesForFocusedNode: PropTypes.array,
         nodeTypesRegistry: PropTypes.object.isRequired,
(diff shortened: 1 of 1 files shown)


dimaip:
@grebaldi hey, thanks for the feedback.
The reason why I did not bother to check the policy here is that I think we have no way to actually define that policy currently. When we are at the site node, we have no parent node in our store that could possibly hold the policy. And I'm not even sure if you could define such policy on Flow side.

Anyways, I can hardly imagine anyone trying to restrict the number of nodetypes suitable for the site node, IMO we shouldn't bother about it.

grebaldi:
> The reason why I did not bother to check the policy here is that I think we have no way to actually define that policy currently. When we are at the site node, we have no parent node in our store that could possibly hold the policy. And I'm not even sure if you could define such policy on Flow side.

The relevant policy information is attached directly to the node, not to any of its ancestors. I don't see, why it wouldn't be possible to define that policy. Actually all it takes is a single `createdNodeIsOfType` matcher, I'd imagine.

> Anyways, I can hardly imagine anyone trying to restrict the number of nodetypes suitable for the site node, IMO we shouldn't bother about it.

No, but it would be possible to restrict the number of nodetypes in general. I know it's pretty edgy regarding the node type of the site node, but I'd find it wrong to add an exception to the rule at this place.

> And no, I don't think we should touch the logic of getAllowedNodeTypesTakingAutoCreatedIntoAccount, since that would actually allow creating sibling nodes next to the site node... Nodetype editor is kind of a special case, that's why I coded it this way.

I don't think that the node type editor is the special case here. CR-wise nothing speaks against siblings on site node level. It is Neos itself that adds the idea of sites on top of the CR, so actually the `add` and `paste` features introduce the speciality to not being able to create siblings to the site node.

I don't want to block this fix for an edge case though, so I would be fine with just adding a TODO comment to L61, that hints the issue. Then we can address it later on.

dimaip:
> The relevant policy information is attached directly to the node, not to any of its ancestors.
Yes, but that policy is related to creating descendant nodes, not siblings, if I understand correctly, no?
If not then I could indeed fix it.

> CR-wise nothing speaks against siblings on site node level.
Heh, that's an interesting perspective. Though we are now in the Neos land, and here the ability to create/paste nodes is much more common and happens from more places, then switching the nodetype, so from mere simplicity's POV it's easier to just handle this case as an exception.

grebaldi:
> Yes, but that policy is related to creating descendant nodes, not siblings, if I understand correctly, no?

You're right of course. I got the scenario wrong while thinking about it. So, we have indeed no possibility to consider the correct policy information here.

Thanks for taking the time to clear that up :)

Issue #801, opened 2017-08-03
Extract transition times into constants


mstruebing:
Like #793 the different times should be extracted into constants to make sure to use the same time if they are related transitions.

markusguenther:
Check this tomorrow ;)

PR #1354, opened 2017-11-28
FEATURE: Extract transition timings to constants

markusguenther:
Extracting the transition timings to constants.

Resolves: #801


markusguenther:
Trigger styleci again

dimaip:
Could we maybe put `transitionVars` into the `styleVars`?

markusguenther:
Yes we can ;)

packages/build-essentials/src/config/transitionConfig.js
```diff
@@ -0,0 +1,35 @@
+const upperFirst = require('lodash.upperfirst');
+
+const config = {
+    transition: {
+        fast: '.1s',
+        default: '.2s',
+        medium: '.3s',
```
dimaip: Btw, these kind of refactoring really raise the UX-related questions, e.g. is `.2s` really distinguishable from `.3s`?
But this should obviously be taken care of by the UX guild and not here.

markusguenther:
@dimaip You was to fast... the first was just resolving conflicts with the master via GitHub. But that did not affect the transition constants ;)

dimaip:
Ouch sorry, overlooked it :blush:

packages/build-essentials/src/config/transitionConfig.js
markusguenther: To be honest I see no difference ... in my point of view we could maybe drop 200ms and use 300ms as default... then we have just slow, fast and default.

markusguenther: Or 250 as default and 100 as fast and 500 as slow.

dimaip: Agreed.

markusguenther: Adjusted that now.

PR #1357, opened 2017-11-28
BUGFIX: Fix iframe control mechanism

### LINKED ISSUE #1324: white "Loading-Overlay-Screen" appears on all navigation actions in Node Tree
# How to reproduce

- navigate using the Document Tree
- you'll see the website changes directly to white; and only if the full website is loaded, it changes back.
- for big websites, this leads to a bad perceived performance; as the user might already e.g. scroll the website and orient himself on it while it is still loading.
- also it means the website only is visible after all images, banners, ... etc have appeared and finished loading

**to me this is critical, as it leads to bad perceived performance**

## Problem in Code Path

The critical path is https://github.com/neos/neos-ui/blob/master/packages/neos-ui/src/Containers/ContentCanvas/index.js#L137-L144

this condition is TRUE even if we navigate through the page tree.

## Does the feature currently work at all? :-)

I just tried to add a `<h1 onClick="window.location.href = 'http://spiegel.de';">HALLO</h1>` to the content output; and when I clicked it:

1) the overlay did appeared to white
2) a cross origin error occured `Host.js?27a5001b:7307 Uncaught DOMException: Failed to read the 'contentDocument' property from 'HTMLIFrameElement': Blocked a frame with origin "http://127.0.0.1:8081" from accessing a cross-origin frame.    at getGuestFrameDocument`

3) the old page did not re-appear after a while.

@grebaldi as you wrote the feature, could you shed some light on what/how it is supposed to work so we can fix it? :)

All the best,
Sebastian

grebaldi:
This PR fixes a couple of issues with the frame origin control mechanism for the Content Canvas frame:

* Primarily, the `beforeunload` event handler in the `<ContentCanvas>` container is removed (This is what caused #1324). An unfortunate side effect of that is, that if a site like "spiegel.de", which has no proper X-Frame headers (like e.g. "google.com"), is loaded in the frame, the site will be visible as long as it takes for the `load` event to fire. Afterwards, the UI will recognize that it has lost control and will navigate the user back to the last page, just as intended by the original feature.
* The `<Frame>` component did not handle cross-origin situations correctly. This PR adds error handling to the `updateIframeUrlIfNecessary` method to set the Iframe source instead of directly manipulating the iframe window's location object, in case a security error occurs.
* This PR also removes the `sandbox` attribute from the ContentCanvas frame, since it was likely to cause a lot of problems for currently existing projects.

Fixes #1324

### DIFF
--- a/packages/react-ui-components/src/Frame/frame.js
+++ b/packages/react-ui-components/src/Frame/frame.js
@@ -24,15 +24,21 @@ export default class Frame extends PureComponent {
         this.updateIframeUrlIfNecessary();
     }
 
-    // we do not use react's magic to change to a different URL in the iFrame, but do it explicitely (in order to avoid reloads if we are already on the correct page)
+    // we do not use react's magic to change to a different URL in the iFrame, but do it
+    // explicitely (in order to avoid reloads if we are already on the correct page)
     updateIframeUrlIfNecessary() {
         if (!this.ref) {
             return;
         }
 
-        const win = this.ref.contentWindow; // eslint-disable-line react/no-find-dom-node
-        if (win.location.href !== this.props.src) {
-            win.location = this.props.src;
+        try {
+            const win = this.ref.contentWindow; // eslint-disable-line react/no-find-dom-node
+            if (win.location.href !== this.props.src) {
+                win.location = this.props.src;
+            }
+        } catch (err) {
+            console.error(`Could not update iFrame Url from within. Trying to set src attribute manually...`);
+            this.ref.setAttribute('src', this.props.src);
         }
     }
 
--- a/packages/neos-ui/src/Containers/ContentCanvas/index.js
+++ b/packages/neos-ui/src/Containers/ContentCanvas/index.js
@@ -132,7 +132,7 @@ export default class ContentCanvas extends PureComponent {
     }
 
     handleFrameAccess = iframe => {
-        const {startLoading, requestRegainControl, requestLogin} = this.props;
+        const {requestRegainControl, requestLogin} = this.props;
 
         try {
             if (iframe) {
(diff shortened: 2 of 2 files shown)


packages/react-ui-components/src/Frame/frame.js
```diff
@@ -24,15 +24,21 @@ export default class Frame extends PureComponent {
...
-        const win = this.ref.contentWindow; // eslint-disable-line react/no-find-dom-node
-        if (win.location.href !== this.props.src) {
-            win.location = this.props.src;
+        try {
+            const win = this.ref.contentWindow; // eslint-disable-line react/no-find-dom-node
+            if (win.location.href !== this.props.src) {
+                win.location = this.props.src;
+            }
+        } catch (err) {
+            console.error(`Could not update iFrame Url from within. Trying to set src attribute manually...`);
```
dimaip: Should this really be console.error? Since we are recovering from it, maybe use console.warn at least?

packages/react-ui-components/src/Frame/frame.js
```diff
@@ -24,15 +24,21 @@ export default class Frame extends PureComponent {
...
     updateIframeUrlIfNecessary() {
         if (!this.ref) {
             return;
         }
 
-        const win = this.ref.contentWindow; // eslint-disable-line react/no-find-dom-node
-        if (win.location.href !== this.props.src) {
-            win.location = this.props.src;
+        try {
+            const win = this.ref.contentWindow; // eslint-disable-line react/no-find-dom-node
```
dimaip: Why ` react/no-find-dom-node`? I dont' see findDomNode here...

dimaip:
To be honest I'm not deep into this topic. We could merge it as is unless @skurfuerst wants to take a look.

Issue #1359, opened 2017-11-28
Upgrade moment.js to fix insecure dependency


markusguenther:
The dependency listing clarify the used moment.js version as insecure.

<img width="998" alt="bildschirmfoto 2017-11-28 um 20 14 07" src="https://user-images.githubusercontent.com/1014126/33339388-3da60f1c-d479-11e7-848d-dfc824fb4e89.png">
<img width="991" alt="bildschirmfoto 2017-11-28 um 20 14 24" src="https://user-images.githubusercontent.com/1014126/33339389-3dcd4d5c-d479-11e7-93f6-5fe96b54d494.png">

PR #1360, opened 2017-11-28
TASK: Upgrade moment.js to version 2.19.2

markusguenther:
The current used version 2.14.1 is marked as insecure.
Therefore it would be good to upgrade the package.

Fixes: #1359

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,8 +67,8 @@
     "url-loader": "^0.6.2",
     "watch": "^1.0.2",
     "webpack": "^3.8.1",
-    "webpack-merge": "^4.1.1",
-    "webpack-livereload-plugin": "^1.0.0"
+    "webpack-livereload-plugin": "^1.0.0",
+    "webpack-merge": "^4.1.1"
   },
   "dependencies": {
     "@neos-project/brand": "^1.1.0",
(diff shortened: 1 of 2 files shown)


markusguenther:
Restart TravisCI

markusguenther:
We don`t use the call moment.duration in combination with the format function. So first need to investigate were it really occurs and come from. :/

markusguenther:
Need to setup my local neos again ... because ee tests are actually falling for me in master. So something is wrong... but will update this soon.

mstruebing:
I think I've found something regarding this issue:

```
$ grep -Hnir 'this.moment.duration'
node_modules/testcafe-reporter-list/lib/index.js:94:            var durationStr = this.moment.duration(durationMs).format('h[h] mm[m] ss[s]');
node_modules/testcafe-reporter-spec/lib/index.js:95:            var durationStr = this.moment.duration(durationMs).format('h[h] mm[m] ss[s]');
```

So, testcafe needs an update that we can update `moment.js`

markusguenther:
Just updating test-cafe ist not enough ... but I am on track ;)

markusguenther:
Ok since testcafe had the problems I submitted PRs to the testcafe repositories.
Hope that they get merged soonish and find the way into a new release.
Otherwise this is blocked for a while ;)

https://github.com/DevExpress/testcafe-reporter-spec/pull/4
https://github.com/DevExpress/testcafe-reporter-list/pull/4

Issue #1264, opened 2017-11-15
ImageEditor does not support resize Feature for now


johannessteu:
Right now the resize feature is not supported as it is not integrated yet

PR #1369, opened 2017-11-30
FEATURE: implement resize in the Image editor

dimaip:
Resolves: #1264 

image


dimaip:
hmm, doesn't it happen? was supposed to work this way

PR #1371, opened 2017-12-01
BUGFIX: Filter all in asset editor

### LINKED ISSUE #1353: Filter "All" in asset editor doesn't work
If i have an asset editor in the inspector, when i select the "All" filter, document doesn't appear.
If i select "Document" filter, document appears. 

image

image

mstruebing:
It's still the hard coded URL but it fixes this bug for now.

closes #1353


dimaip:
Could you create an issue for that hardcoded url?

mstruebing:
done/ #1372

PR #1375, opened 2017-12-01
BUGFIX: Unable to create a Personal Workspace

dfeyer:


dfeyer:
See my message on Slack https://neos-project.slack.com/archives/C0U0KEGDQ/p1512164595000274

dimaip:
Could somebody with Flow knowledge comment on this?
/cc @kitsunet @grebaldi

dfeyer:
@kitsunet OK, but we need to solve this one before 1.0, right ?

dimaip:
@kitsunet so what about this one?

kitsunet:
Why should the user workspace not be there. It MUST be created while creating a user, if that is not happening that is the bug.

kitsunet:
Do we have a way to reproduce this?

dfeyer:
On my side: 

    flow site:prune && flow site:import Core.Website

And boom ...

dfeyer:
Oh not, I have the same bug, but it's not caused by site:prune / import ... maybe because the setup crash, caused by a broken Sites.xml ... so the CR was not in sane status. 

The problem currently, we don't have any way to get out of this without tweaking the DB.

dfeyer:
in \Neos\Neos\Service\BackendRedirectionService::createWorkspaceAndRootNodeIfNecessary we create the required workspace and node, the question is why this is not used in the current case ?

kitsunet:
Interesting. Maybe we need to have a look at site prune and import. Maybe it removed relevant unser workspace nodes or somthing :/

mstruebing:
Is this PR is still relevant? @dfeyer

dfeyer:
I will close it for now.

PR #1378, opened 2017-12-03
TASK: Utilize ``AssetUpload`` component

gerhard-boden:
The ``AssetList`` component now wrapped within the ``AssetUpload`` component which now supports drag & drop for multiple files.

This is a refactoring following https://github.com/neos/neos-ui/pull/1316 where drag & drop for the ``AssetList`` was introduced. Now the code is much more streamlined with other components that use drag & drop for assets like https://github.com/neos/neos-ui/blob/master/packages/neos-ui-editors/src/Editors/Image/Components/PreviewScreen/index.js#L41-L55

### DIFF
--- a/packages/neos-ui-editors/src/Library/AssetUpload.js
+++ b/packages/neos-ui-editors/src/Library/AssetUpload.js
@@ -25,7 +25,9 @@ export default class AssetUpload extends PureComponent {
         siteNodePath: PropTypes.string.isRequired,
         focusedNode: PropTypes.string.isRequired,
         highlight: PropTypes.bool,
-        children: PropTypes.any.isRequired
+        children: PropTypes.any.isRequired,
+        multiple: PropTypes.bool,
+        multipleData: PropTypes.oneOfType([PropTypes.string, PropTypes.object, PropTypes.array])
     };
 
     chooseFromLocalFileSystem = () => {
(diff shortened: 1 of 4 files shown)


dimaip:
@gerhard-boden hey! Is this still relevant after the SelectBox refactoring? Could you rebase it?

gerhard-boden:
@dimaip Yab, still considering code refactoring. Just rebased

gerhard-boden:
had to rebase this again today, but uploading files within the Asset list seems to broken on master (and therefore this rebased version). Maybe fix the issue after merging this one? 
Have to keep track of multiple files for each rebase, since I moved functions around within this change to reduce code duplication.

packages/neos-ui-editors/src/Library/AssetUpload.js
```diff
@@ -43,6 +45,36 @@ export default class AssetUpload extends PureComponent {
...
+        if (index < 0) {
+            if (onAfterUpload) {
+                onAfterUpload(values);
+            }
+            this.setState({
+                isLoading: false
+            });
+            return;
+        }
+        uploadAsset(files[index], this.props.propertyName, focusedNode, siteNodePath).then(res => {
```
dimaip: this will only work with image assets. For other assets it's missing a metadata parameter.
I'd make it a prop, so it could be used both from Image and Asset editors.

gerhard-boden:
looks like the third conflict fix is due 😄 
About 
>I'd make it a prop, so it could be used both from Image and Asset editors.

I don't fully understand where to create that prop or add that missing meta data. Inside the uploadAsset function? Do you have any pointer where general assets already work currently? I thought I had successfully tested it with a PDF after the first implementation, but I might be mistaking.

dimaip:
@gerhard-boden sure, take a look here: https://github.com/neos/neos-ui/blob/f97bba7a11c9d35b2b567ee1e7f96bd8e5350e48/packages/neos-ui-editors/src/Editors/AssetEditor/index.js#L163
The last argument to `uploadAsset` endpoint takes the name of metadata type, it should be `Asset` for file uploades, otherwise you'd get a backend exception.

dimaip:
An `AssetEditor` may have a prop that is undefined by default, that would be passed as that last argument to `uploadAsset`. Otherwise the `AssetEditor` can not be used for uploading regular non-image assets.

gerhard-boden:
Daily rebase done ;) 
@dimaip Thanks for explaining. It was clear to me what you mean as soon as I saw the sample you've provided.

I've also implemented the metadata prop and found two (unrelated) problems:
- There is no error message when the image is too big when uploading, but it will fail with an exception, that no asset was given (fooled me while testing)
- The buttons are currently super close together. Looks like a reset style is overruling the margin 
image

This works now. Would be great to merge this soonish, since a rebase is due almost every time a change occurs either in `AssetEditor` or `AssetUpload` since essential code has been move between those two components within this change.

dimaip:
Merging, so we would have more time to test.

Thanks!!!

gerhard-boden:
Thanks you @dimaip!!If something comes up related to the ``AssetList`` editor, don't hesitate to ping me, will gladly take care of it.

Issue #1379, opened 2017-12-03
Review changes option in the changes menu is missing


daniellienert:
Old UI:
![image](https://user-images.githubusercontent.com/642226/33528059-8e3fcd74-d85b-11e7-976b-7ad2503e9f93.png)

New UI:
![image](https://user-images.githubusercontent.com/642226/33528067-a16871f8-d85b-11e7-9cb6-38f05841711b.png)

Issue #1372, opened 2017-12-01
Hard coded URL for media browser


mstruebing:
Remove Hard coded URL 

https://github.com/neos/neos-ui/blob/master/packages/neos-ui-editors/src/SecondaryEditors/MediaSelectionScreen/index.js

PR #1381, opened 2017-12-03
TASK: replace hardcoded urls and add link to workspace review

dimaip:
Resolves: #1379 
Resolves: #1372

### DIFF
--- a/packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/index.js
+++ b/packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/index.js
@@ -15,7 +15,7 @@ import I18n from '@neos-project/neos-ui-i18n';
 import {actions, selectors} from '@neos-project/neos-ui-redux-store';
 import {neos} from '@neos-project/neos-ui-decorators';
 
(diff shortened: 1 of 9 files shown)


packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/index.js
```diff
@@ -91,6 +94,7 @@ export default class PublishDropDown extends PureComponent {
             neos
         } = this.props;
 
+        const workspaceModuleUri = $get('routes.core.modules.workspaces', neos);
```
mstruebing: This Uri lacks an `management` part, at least for me. I didn't know exactly where it comes from. 
But this results in a link to `neos.org/neos/workspaces/show?moduleArguments[workspace]=user-admin`
and this leads to a 404, whereas in the old ui it is: `neos.org/neos/management/workspaces/show?moduleArguments[workspace]=user-admin`

dimaip: duh, sorry about that, fixed

mstruebing: No problem, that what reviews are for. :)

PR #1392, opened 2017-12-07
TASK: Use neos serialization of properties

kitsunet:


kitsunet:
This is what I commented about. Hopefully the Image converter is the only one to be changed. This is a breaking change in a couple of ways though. First of all while the new UI is installed the old won't work with images anymore. Sadly the old couldn't handle JSON values for properties so it was delivered as string. Secondly any custom dataType converters that return JSON as string as in the image case will not work with the new UI and would have to be replaced with converters that return arrays of simple values. MEH, but either way the current situation is breaking and not acceptable.

The conversion from UI to node is also currently duplicating code that exists in Neos itself and we shouldn't release with this duplication and probably slightly different code. That's a separate change though...

skurfuerst:
Hey :)

@kitsunet "First of all while the new UI is installed the old won't work with images anymore." -- that's something I think we need to fix... 

Let's discuss the options together if you like :)

All the best,
Sebastian

kitsunet:
This should reduce the breakiness, still either way with or without change people using their own converters for anything will have a bad experience. But that will happen either way and with this change they at least have a chance to use a newly written mapper that works for the new backend.

dimaip:
So what do we do about this change?

kitsunet:
IMHO this is a blocker for 1.0. Many of the projects I work on won't work correctly without it.

dimaip:
What needs to be done until this could be merged? I don't really understand the problem.

kitsunet:
It's as ready as it gets. As said the new UI is incompatible on that specific level with the old UI anyway. That won't go away. This change just exposes it. It's not breaking anymore, you can use old and new UI side by side but for the special case of a custom configured type converter for a node property you would need to do something like we did and use different ones depending on the UI in use. Again that is just different behavior between the UIs, not related to this change. But at least you CAN configure a tpye converter which you cannot without this change.

dimaip:
So just merge it?

2case:
Please :) This should fix the problem i have for example with my custom DataSources. I use the entityConverter which produce a json string for my entities. The old ui simple use the string, the new one detect the json inside, so they break... That project was the base for this change :)

Issue #1382, opened 2017-12-03
Rethink z-indexes


dimaip:
We've merged a change to refactor usage of z-indexes: https://github.com/neos/neos-ui/pull/1361
However it caused a few regressions, e.g. see the screenshot.
The idea was to group z-indexes by their stacking contexts and stop making up their values arbitrarily.
However the problem is that currently we group z-indexes not always by the stacking contexts, which causes various issues.
So we need to fix those regressions in a clever way. In most cases it should be enough to force new stacking contexts where needed.

![image](https://user-images.githubusercontent.com/837032/33530482-240afdf8-d891-11e7-9f2a-e85206b90a77.png)

dimaip:
Maybe @Duske you have an opinion?

Duske:
I think introducing new stacking contexts would be most straightforward approach.
Currently I have no time to look into it in more detail 🙁

dimaip:
Yep, apparently there's even a dedicated property to do just that: https://developer.mozilla.org/en-US/docs/Web/CSS/isolation
We need to rethink a system that would handle nested z-indexes, essentially it should be a nested object.

E.g. this
```js
{
 primaryToolbar: {
   dropDown: true
 },
 contentCanvas: {
   some: {
     nesting: true,
     sibling: true
   }
 }
}
```
would transform into:

```js
--zIndex-primaryToolbar: 1
--zIndex-primaryToolbar-dropDown: 1
--zIndex-contentCanvas: 2
--zIndex-contentCanvas-some: 1
--zIndex-contentCanvas-some-nesting: 1
--zIndex-contentCanvas-some-sibling: 2
```
And of course this structure should resemble actual stacking contexts and don't be arbitrary.

jonnitto:
Also a topic if you have a tooltip with a following select box  
<img width="298" alt="screenshot 2017-12-22 16 11 40" src="https://user-images.githubusercontent.com/4510166/34302837-f9475fb6-e732-11e7-984d-c65b74a5f947.png">

johannessteu:
Those should be rendered inline/relative. At least i had it in my tooltip change. NOt sure what way @dimaip re-implemented those as the PR was closed. But in general those should fit in between and don't overlay the next property wich might has an validation error as well

dimaip:
See https://github.com/neos/neos-ui/issues/1426

mstruebing:
Should be fixed by #1937

dimaip:
[...]

PR #1397, opened 2017-12-07
TASK: Move the loading indicator on the top of the UI

dfeyer:
This change move the loading indicator from bellow the SecondaryToolbar to the top of the UI. 

This fix an issue when loading with the navigate component closed (wrong left position), but also make the loading indicator more global from the user perspective. We can use the loading indicator during publishing or any other action that require a bit of time. In the future, the loading a backend module can also use this indicator, or an action performed in a backend modules.

### DIFF
--- a/packages/neos-ui/src/Containers/SecondaryToolbar/LoadingIndicator/style.css
+++ b/packages/neos-ui/src/Containers/SecondaryToolbar/LoadingIndicator/style.css
@@ -1,10 +1,10 @@
 .loadingIndicator__container {
-    left: 50%;
-    top: 39px;
+    left: 0;
+    top: 0;
+    height: 2px;
     position: absolute;
-    transform: translateX(-50%);
-    padding: 0 320px;
-    width: 100%;
+    width: 100vw;
+    z-index: var(--zIndex-LoadingIndicatorContainer);
 }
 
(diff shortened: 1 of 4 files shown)


packages/neos-ui/src/Containers/SecondaryToolbar/LoadingIndicator/style.css
```diff
@@ -1,10 +1,9 @@
...
-    left: 50%;
-    top: 39px;
+    left: 0;
+    top: 0;
     position: absolute;
-    transform: translateX(-50%);
-    padding: 0 320px;
-    width: 100%;
+    width: 100vw;
+    z-index: 2;
```
dimaip: Hey, we have a clever system of managing z-index, put it there: https://github.com/neos/neos-ui/blob/master/packages/build-essentials/src/styles/styleConstants.js#L10

dfeyer:
@dimaip Now I use the clever things, in a non really cleverway ;)

dimaip:
Fine for now, should be tackled here: https://github.com/neos/neos-ui/issues/1382
This z-index thing wasn't really thought throw till the end, but we'll get there.

kitsunet:
Nice idea :)

Issue #1408, opened 2017-12-08
Various tiny visual regressions


dimaip:
- [ ] No border
![image](https://user-images.githubusercontent.com/837032/33755931-26740a64-dc04-11e7-9e30-e3521dfea3a0.png)

- [ ] Weird gap
![image](https://user-images.githubusercontent.com/837032/33755963-52888f9e-dc04-11e7-8a7e-5921d0aed66d.png)

- [ ] No separator "stick", looks boring
![image](https://user-images.githubusercontent.com/837032/33755989-67e37b24-dc04-11e7-9cba-f6a7374d6896.png)

mstruebing:
Regarind the first point(No border): When I zoom to 110% within chrome I can see the border. In Firefox I can see the border already at 100%.

PR #1412, opened 2017-12-09
TASK: Various tiny visual regressions

mstruebing:
references #1408 

Before:
before_select

After:
after_select

Before:
before_link

After:
after_link

Before:
selectbox_before

After:
ssss

### DIFF
--- a/packages/neos-ui-ckeditor-bindings/src/EditorToolbar/style.css
+++ b/packages/neos-ui-ckeditor-bindings/src/EditorToolbar/style.css
@@ -21,7 +21,7 @@
     width: 460px;
     border: var(--halfSpacing) solid var(--brandColorsContrastDarker);
     /* 8px border top and bottom */
-    height: calc(var(--goldenUnit) + 16px);
+    height: calc(var(--goldenUnit) + var(--spacing));
 }
 
 .linkIconButton__item {
--- a/packages/neos-ui-ckeditor-bindings/src/EditorToolbar/style.css
+++ b/packages/neos-ui-ckeditor-bindings/src/EditorToolbar/style.css
@@ -20,6 +20,8 @@
     z-index: var(--zIndex-SecondaryToolbar-LinkIconButtonFlyout);
     width: 460px;
     border: var(--halfSpacing) solid var(--brandColorsContrastDarker);
+    /* 8px border top and bottom */
+    height: calc(var(--goldenUnit) + 16px);
 }
 
 .linkIconButton__item {
--- a/packages/neos-ui-ckeditor-bindings/src/EditorToolbar/style.css
+++ b/packages/neos-ui-ckeditor-bindings/src/EditorToolbar/style.css
@@ -19,7 +19,6 @@
     position: fixed;
     z-index: var(--zIndex-SecondaryToolbar-LinkIconButtonFlyout);
     width: 460px;
-    height: 56px;
     border: var(--halfSpacing) solid var(--brandColorsContrastDarker);
 }
 
--- a/packages/neos-ui-ckeditor-bindings/src/EditorToolbar/style.css
+++ b/packages/neos-ui-ckeditor-bindings/src/EditorToolbar/style.css
@@ -19,6 +19,7 @@
     position: fixed;
     z-index: var(--zIndex-SecondaryToolbar-LinkIconButtonFlyout);
     width: 460px;
+    height: 56px;
     border: var(--halfSpacing) solid var(--brandColorsContrastDarker);
 }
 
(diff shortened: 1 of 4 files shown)


packages/react-ui-components/src/SelectBox/style.css
```diff
@@ -11,6 +11,7 @@
     composes: reset from './../reset.css', dropDown from './../DropDown/style.css';
     position: relative;
     display: inline-block;
+    height: 40px;
```
dimaip: there's constant for that

packages/neos-ui-ckeditor-bindings/src/EditorToolbar/style.css
```diff
@@ -19,6 +19,7 @@
     position: fixed;
     z-index: var(--zIndex-SecondaryToolbar-LinkIconButtonFlyout);
     width: 460px;
+    height: 56px;
```
dimaip: magic number. what is this number made of? could you `calc` it?

packages/react-ui-components/src/SelectBox_Header/style.css
```diff
@@ -3,11 +3,35 @@
     display: flex;
 }
 
+.selectBoxHeader:after {
+    display: block;
+    content: '';
+    position: absolute;
+    width: 1px;
+    height: 24px;
+    top: 8px;
```
dimaip: Could use constants here as well

packages/react-ui-components/src/SelectBox_Header/style.css
```diff
@@ -3,11 +3,35 @@
     display: flex;
 }
 
+.selectBoxHeader:after {
```
dimaip: I think this class would also affect other icons in different SelectBox configs, e.g. a search icon.
And also we don't need two lines there, only one. And it's better to render it as a separate html element, because it should be shown only when both the cross and the chevron are shown.

packages/neos-ui-ckeditor-bindings/src/EditorToolbar/style.css
mstruebing: Fixed, this is actually goldenUnit + 8px border-top + 8px border-bottom

packages/react-ui-components/src/SelectBox/style.css
mstruebing: Using goldenUnit now here.

mstruebing:
Fixed your comments. 
I've used seperators on both sides because this what how it looks like in the old UI
screenshot_2017-12-10_11-20-35

packages/react-ui-components/src/SelectBox_Header/style.css
mstruebing: I should introduce a constant for width/height/top even if they are not used anywhere else?

dimaip: No, definitely not, but a lot of numbers could be calc'ed from base constants.

dimaip:
Will take a look tonight, thanks!!!

packages/neos-ui-ckeditor-bindings/src/EditorToolbar/style.css
```diff
@@ -20,6 +20,8 @@
     z-index: var(--zIndex-SecondaryToolbar-LinkIconButtonFlyout);
     width: 460px;
     border: var(--halfSpacing) solid var(--brandColorsContrastDarker);
+    /* 8px border top and bottom */
+    height: calc(var(--goldenUnit) + 16px);
```
dimaip: 16px = var(--spacing)

packages/react-ui-components/src/SelectBox_Header/style.css
```diff
@@ -8,6 +8,18 @@
...
 }
 
+.selectBoxHeader__seperator {
+    display: block;
+    content: '';
+    position: absolute;
+    width: 1px;
+    height: 24px;
+    top: 8px;
+    right: 41px;
```
dimaip: calc(var(--goldenUnit) + 1px)

packages/react-ui-components/src/SelectBox_Header/style.css
```diff
@@ -8,6 +8,18 @@
     line-height: var(--goldenUnit);
 }
 
+.selectBoxHeader__seperator {
+    display: block;
+    content: '';
+    position: absolute;
+    width: 1px;
+    height: 24px;
+    top: 8px;
```
dimaip: 8px = --halfSpacing

dimaip:
As always, thanks for the patience!

Issue #1420, opened 2017-12-12
Extend documentation for end 2 end testing


markusguenther:
When you have no chrome you get the following error.
So we should add this requirement to the docs.

```
➜  Neos.Neos.Ui git:(master) ✗ yarn test:e2e
yarn run v1.3.2
$ yarn run testcafe -- chrome:headless Tests/IntegrationTests/* --selector-timeout=30000 --assertion-timeout=30000
warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.
$ /Users/markusgunther/Development/neos-development-distribution/Packages/Application/Neos.Neos.Ui/node_modules/.bin/testcafe chrome:headless Tests/IntegrationTests/contentModule.js Tests/IntegrationTests/pageModel.js --selector-timeout=30000 --assertion-timeout=30000
ERROR Was unable to open the browser "chrome:headless" due to error.

Error: Unable to run the browser. The browser path or command template is not specified.
    at checkBrowserPath$ (/Users/markusgunther/Development/neos-development-distribution/Packages/Application/Neos.Neos.Ui/node_modules/testcafe-browser-tools/lib/api/open.js:47:23)
    at tryCatch (/Users/markusgunther/Development/neos-development-distribution/Packages/Application/Neos.Neos.Ui/node_modules/testcafe-browser-tools/node_modules/babel-runtime/regenerator/runtime.js:72:40)
    at GeneratorFunctionPrototype.invoke [as _invoke] (/Users/markusgunther/Development/neos-development-distribution/Packages/Application/Neos.Neos.Ui/node_modules/testcafe-browser-tools/node_modules/babel-runtime/regenerator/runtime.js:334:22)
    at GeneratorFunctionPrototype.prototype.(anonymous function) [as next] (/Users/markusgunther/Development/neos-development-distribution/Packages/Application/Neos.Neos.Ui/node_modules/testcafe-browser-tools/node_modules/babel-runtime/regenerator/runtime.js:105:21)
    at tryCatch (/Users/markusgunther/Development/neos-development-distribution/Packages/Application/Neos.Neos.Ui/node_modules/testcafe-browser-tools/node_modules/babel-runtime/regenerator/runtime.js:72:40)
[...]

PR #1421, opened 2017-12-12
TASK: Extend README with integration tests section

markusguenther:
When you have a clean system without a chrome your end to end tests will fail. So we should add the information to the README to prevent frustration of the end users.

Resolves: #1420

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,12 @@ Adding unit tests is fairly simple, just create a file on the same tree level as
 
 Use `it.only(() => {})` and `describe.only(() => {})` if you want to run a specific test and not the whole test suite.
 
+#### Integration tests
+
+For end to end testing we use the headless chrome. So it is mandatory to install the chrome browser for integration tests.
+Since Chrome 59 the headless mode is integrated. So please install a Chome 59 or higher to execute the end to end tests.
+
+
 ## License
 THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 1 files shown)


markusguenther:
Yeah I think thats a point ... and had even the idea if we not can check if the server is running and if not just start the flow server.

dimaip:
Basically we can link people to the setup file: https://github.com/neos/neos-ui/blob/master/Build/TravisCi/InstallScripts/IntegrationTests.sh

We are really bad at keeping developer docs here up to date. Sometimes I think it might be better to just remove everything, rather than give people outdated information.

markusguenther:
Btw... it would be awesome if we can bootstrap such things with the user.
I know that from codeception. There I created a base class that bootstraps the TYPO3 with my Database and some fixture data and one part is obviously also the user stuff.

Then you have not the problem that the user need to care for that.
But don't know if we can setuo such things with testcafe ...

markusguenther:
But when you remove such information then it is hard to get people on board. The steps helped me at the beginning.

dimaip:
> Btw... it would be awesome if we can bootstrap such things with the user.

Didn't get the point. But go ahead and try.

mstruebing:
I didn't want to block this or make a big discussion out of that. I think it is enough at first to state the said points in the documentation(port, demo page, admin user).
And later on we can create a automated script or whatever.

markusguenther:
Was not ready yet but currently I stay in the base office and had no time yet :/

dimaip:
It's not critical, feel free to improve whenever you like

Issue #1290, opened 2017-11-16
Adjust routing to make react UI primary UI


kitsunet:
Decide on `legacy` backend route (`/neos-legacy`?)
Switch routes
Test
Adjust loading order

kitsunet:
From @skurfuerst in Slack:

FYI: over the weekend some ideas regarding how to "make the switch" from the old Ember UI to the new React UI. Before, it was still somewhat vague for me, but now that I had some time to think it through, I think I came up with a good and easy to   implement solution.

I propose the following:

1) The Neos React UI should add routes for /neos and /neos!. We furthermore configure Neos.Neos.Ui routes to be added *before* Neos.Neos Routes.
2) Furthermore, we add a transitional, unsupported setting "enableOldUi"; set it by default to "FALSE" -> if this setting is set to FALSE, the service which determines whether the old or new UI is enabled will always return the NEW UI.
3) this way, if people install the package, it will *replace* the old UI by the new one.
4) we're having some bigger corporate customers who need/want to extend the transition period some more and we'd help them a lot if we were able to keep the current behavior which we have now; but I would clearly state it is UNSUPPORTED. To get this behavior back, one would need to configure the Routes ordering differently, so that Neos.Neos is loaded before Neos.Neos.UI; and the "enableOldUi" switch needs to be set to TRUE.

kitsunet:
And forgot to tag the issue in hte PR: https://github.com/neos/neos-ui/pull/1423

PR #1423, opened 2017-12-12
TASK: Use new after login routing and set routes for new backend

kitsunet:
This re-configures the routes for use as default backend and overwrites
the default route after login to go to the new UI.

Fixes: #1290
Depends: https://github.com/neos/neos-development-collection/pull/1794


dimaip:
So how do you enable the old UI?

kitsunet:
@dimaip You overwrite the route settings back to what you see in the Neos change in your global settings for example.

kitsunet:
Sounds cumbersome, but then we all said this should be the exceptoin for special projects. And it opens up a lot of flexibility about this in the future.

dimaip:
@kitsunet but if you override the routes, will it be possible to use both UIs at the same time? How would the switching work then?

kitsunet:
I thought we don't support switching back and forth. I also don't think that's a good idea. But generally both seem to work if you call the respective URLs. The configuration "just" defines the entry point after login.

dimaip:
But that was the very point that @skurfuerst brought up, that we need to support switching back and forth for the transitional period.
Otherwise one could just remove the UI package, that would have the same effect...

So with the adjusted Routes, the behavior would be like this:
`/neos!` = new UI
`/neos` = old UI
Correct?
Will the login still work correctly?

kitsunet:
for the above I would have to adjust a few things again. I didn't understand that switching back and forth is a requirement and IMHO that is really a bad idea.

dimaip:
But that's clearly described as a requirement in @skurfuerst write up.
I personally am not opinionated on this, but I see why Sebastian asks for it.
As said, if you want the switch to be permanent then why bother, just remove the UI package and the old UI would start working, no?

kitsunet:
ok, 👎 from me for now. I need to figure something out that I missed. So we NEED the session value to determine which is the active UI. you cannot simply switch UIs and have the other UI. Just because the metadata rendering is not compatible, so the rendering must know which backend is the active one.

That means I need to provide an actual mechanism to switch around. AKA add a controller + route(s) that allows you to switch around which is the currently active UI and then redirect to that all the time. I have an idea but I couldn't wrap my head around implementing it tonight anymore.

dimaip:
Couldn't we just use the system that we already have in place (that redirect aspect)?

kitsunet:
Basically yes, but I don't see the need for an aspect. Will brew up something...

skurfuerst:
Hey everybody :)

@kitsunet I'd be up for discussing it tomorrow or so if you like :-)

All the best, Sebastian

kitsunet:
Ok, here we go. This should make every scenario possible in some way or the other.

a) new UI not installed, obviously works, nothing special here.
b) new UI installed with default settings -> All routes (eg /neos) will lead you to the new UI. You can switch to the old by using the switch in user menu. The "Content" entry in the module menu will always bring you back to the new UI though (which probably makes sense as it is default now)

c) new UI installed, overwrote the default route back to old UI like this:

```
Neos:
  Neos:
    userInterface:
      # The below configuration sets this to the "old" UI route.
      routeAfterLogin:
        values:
          '@package': 'Neos.Neos'
          '@controller': 'Frontend\Node'
          '@action': 'show'
          '@format': 'html'
```

Now you are in the trickiest state.  The route /neos and subsequent login will get you to the old UI. The "content" module entry will get you to the old UI. As we have no user interface element to switch now, you have two options `/neos!` will get you to the new UI as well as `neos/content` but you need to know that. We can obviously think about adding some UI element to the old UI but that is not part of this change then.

mficzel:
@kitsunet the solution seems fine to me. If i understand it correctly it should be possible to create a third-party-package that adds the route-settings plus the back and forth switch via unplanned extensibility. So imho no need to have that in the core.

kitsunet:
Only thing open I just noticed in discussion if we want to allow to remove the "go to old UI" button for cases if you fully switched to new UI.

dimaip:
Go to old UI should just be removed. This scenario of co-existence of UIs is still kind of marginal, so I don't think we should invest into it too much.

kitsunet:
Also fine. Maybe do that in a separate change then? Otherwise we can obviously do it here too if needed.

dimaip:
Separate change. @skurfuerst can we merge this?

skurfuerst:
Looks awesome <3

skurfuerst:
I'd keep the button to swithc to the old UI... for now... but that's something to be discussed

PR #1428, opened 2017-12-12
BUGFIX: remove margin from publish button

darmstaedter:
The publish button has a margin on the right hand side in safari. This bugfix sets the margin explicitly to 0.

Before: 
bildschirmfoto 2017-12-12 um 23 31 20

After:
bildschirmfoto 2017-12-12 um 23 31 37


dimaip:
Thanks!!!

PR #1433, opened 2017-12-13
BUGFIX: make table properties window usable

darmstaedter:
Before the table properties window wasn't really usable:
bildschirmfoto 2017-12-13 um 12 53 29

Now it's not perfect but at least you can enter some data:
bildschirmfoto 2017-12-13 um 12 54 56


dimaip:
Hmm, strange, I'm sure it did work at some point...
But thanks for the fix!

dimaip:
CKE5 would be a reworked table support btw, so we just need to wait for its final release, that's why we didn't really invest too much into current CKE4 integration.

PR #1436, opened 2017-12-13
BUGFIX: show padding in ListPreviewElement

darmstaedter:
Before:
bildschirmfoto 2017-12-13 um 14 37 52
bildschirmfoto 2017-12-13 um 15 02 33

After:
bildschirmfoto 2017-12-13 um 14 37 12
bildschirmfoto 2017-12-13 um 15 02 01


dimaip:
Hey, I'm not really sure about this one. E.g. in AssetEditor we intentionally removed the padding to make previews bigger.
I think those previews that have an image in them don't need the padding.

dimaip:
I think I found another way to fix it in #1446 
Thanks for your care!

PR #1450, opened 2017-12-15
BUGFIX: fix single AssetEditor

dimaip:
Also accidentally Fixes: #1435


gerhard-boden:
Tested again (post merge) and the AssetList components works as expected after this refactoring

PR #1451, opened 2017-12-15
TASK: Extract constants into one place

### LINKED ISSUE #1416: Move constants into one single place
Currently our constants are all over the place.

Some are in the `neos-ui-constants` package, some are in some post-css-config-file and others are coming from an external source (brandcolors), this is currently unmaintainable and unintuitive.

They should get merged into one single place to easier get found and used.

mstruebing:
This moves all constants from the brand-package inside the styleconstants for better maintance.
Also the usage of the constants is now more consistent

Resolves #1416


mstruebing:
Better now than never ;)

Issue #1430, opened 2017-12-13
Single Asset upload editor needs some love


dimaip:
E.g. it should use custom ListPreviewElement: https://github.com/neos/neos-ui/blob/665c37f050882a6ce92902c68aa7da831eba7ddf/packages/react-ui-components/src/SelectBox_Header/selectBox_Header.js#L61

Currently it looks awful:

![image](https://user-images.githubusercontent.com/837032/33927878-d9efb8d8-dff4-11e7-887f-3969943989f7.png)

PR #1452, opened 2017-12-15
TASK: use ListPreviewElement in SelectBox

dimaip:
Resolves: #1430 
Resolves: #1432 

image

image


dimaip:
Thanks for giving it a try!

PR #1453, opened 2017-12-15
BUGFIX: rework Navigate component CSS to use flexbox

skurfuerst:
This removes magic numbers from CSS; and fixes styling bugs
when new components are rendered in the Navigate Component using
unplanned extensibility (e.g. like https://github.com/sandstorm/FulltextSearchInNodeTree).

It should not have any UI/style implications; I checked the old
and new UI side-by-side and took screenshots of both of them, and
I think they match pixel-perfect.

By not relying on ToggablePanel and the inner logic thereof for
auto-sizing and animation; and instead using flexbox, the robustness
of the CSS in the navigate component could be improved and the logic
simplified.

### DIFF
--- a/packages/neos-ui/src/Containers/LeftSideBar/style.css
+++ b/packages/neos-ui/src/Containers/LeftSideBar/style.css
@@ -23,16 +23,18 @@
 
 .leftSideBar__top {
     flex-basis: 100%;
+    display: flex;
+    flex-direction: column;
 }
 
 .leftSideBar__bottom {
-    height: auto;
-    flex-basis: 100%;
+    flex: 0 0 50%;
     transition: flex-basis var(--transition-Slow), height var(--transition-Slow), overflow-y var(--transition-Slow);
+    display: flex;
+    flex-direction: column;
 }
 
 .leftSideBar__bottom--isCollapsed {
-    flex-basis: 0%;
-    height: auto;
-    overflow-y: visible;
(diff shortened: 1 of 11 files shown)


skurfuerst:
To see that this also fixes a bug, do the following:

- clone https://github.com/sandstorm/FulltextSearchInNodeTree to Sandstorm.FulltextSearchInNodeTree
- see that the Content Tree at the bottom cannot be reached anymore without this fix (as it is pushed off-screen).

with this modification, it all works nicely :)

skurfuerst:
!!!! I need to rebase based on https://github.com/neos/neos-ui/pull/1451/files -- there are some variable names which have changed names!

packages/neos-ui/src/Containers/LeftSideBar/style.css
```diff
@@ -23,16 +23,18 @@
...
     transition: flex-basis var(--transition-Slow), height var(--transition-Slow), overflow-y var(--transition-Slow);
+    display: flex;
+    flex-direction: column;
 }
 
 .leftSideBar__bottom--isCollapsed {
-    flex-basis: 0%;
-    height: auto;
-    overflow-y: visible;
+    flex: 0 0 40px;
```
dimaip: could use goldenUnit here

skurfuerst: yep, changed in followup :)

Issue #1439, opened 2017-12-13
Control Buttons are too close together because of reset styling


gerhard-boden:
![image](https://user-images.githubusercontent.com/10533739/33963877-e1f90f36-e056-11e7-9103-e38d9cd39d6f.png)

PR #1454, opened 2017-12-17
BUGFIX: Overwrite styling to have margin in control buttons

kitsunet:
Fixes: #1439


dimaip:
Uhh I'm so unhappy about !important there, but fine as a quickfix...

kitsunet:
Perfect. Neos fix let's the tests go through. merging...

PR #1467, opened 2017-12-20
BUGFIX: Do not focus link editor when navigating through text

skurfuerst:
To reproduce:
- create an external link in some text
- place the cursor after the text
- navigate using the arrow keys into the linked text

actual behavior (bug):
- focus jumps into the Link-Text-Box

expected behavior:
- focus stays in the main text

### DIFF
--- a/packages/neos-ui-ckeditor-bindings/src/EditorToolbar/LinkIconButton.js
+++ b/packages/neos-ui-ckeditor-bindings/src/EditorToolbar/LinkIconButton.js
@@ -191,7 +191,7 @@ class LinkTextField extends PureComponent {
                     placeholder="Paste a link, or search"
                     displayLoadingIndicator={this.state.isLoading}
                     displaySearchBox={true}
-                    setFocus={true}
+                    setFocus={!this.props.hrefValue}
                     showDropDownToggle={false}
                     allowEmpty={true}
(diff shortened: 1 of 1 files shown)


packages/neos-ui-ckeditor-bindings/src/EditorToolbar/LinkIconButton.js
```diff
@@ -191,7 +191,7 @@ class LinkTextField extends PureComponent {
                     placeholder="Paste a link, or search"
                     displayLoadingIndicator={this.state.isLoading}
                     displaySearchBox={true}
-                    setFocus={true}
+                    setFocus={!Boolean(this.props.hrefValue)}
```
mstruebing: `setFocus={!(this.props.hrefValue)}`

mstruebing: The linter complains :) and `!` is doing an implicit boolean casting IIRC

mstruebing: But it's working fine :)

skurfuerst: yep will fix

PR #1475, opened 2017-12-21
BUGFIX: Set an explicit width for the icons

### LINKED ISSUE #1470: Text and icons not aligned in main navigation
When content module is active, text and icons in the main navigation are not aligned correctly.

Screenshot unaligned text and icons

mstruebing:
so that the chars of the MenuItems are starting all
at the same place

closes #1470

### DIFF
--- a/packages/neos-ui/src/Containers/Drawer/style.css
+++ b/packages/neos-ui/src/Containers/Drawer/style.css
@@ -28,6 +28,12 @@
     text-align: left;
     font-size: 15px;
 }
+/* Set an explicit width for the icons so that the chars are starting */
+/* on the same place for every MenuItem */
+.drawer__menuItemGroupBtn i,
+.drawer__menuItemBtn i {
+    width: 2em;
+}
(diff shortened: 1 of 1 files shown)


packages/neos-ui/src/Containers/Drawer/style.css
```diff
@@ -28,6 +28,12 @@
     text-align: left;
     font-size: 15px;
 }
+/* Set an explicit width for the icons so that the chars are starting */
+/* on the same place for every MenuItem */
+.drawer__menuItemGroupBtn i,
+.drawer__menuItemBtn i {
+    width: 2em;
```
dimaip: @mstruebing just wondering, why do we sometimes set the widths in pixels, and sometimes in em?

mstruebing: I have set them to em because I thought it was wiser if we change the font-size somewhere

dimaip: I see, thanks for explaining

PR #1482, opened 2017-12-21
FEATURE: Version Information

### LINKED ISSUE #1478: Include Release-Version in built files
It would be cool if at a single place in the UI (e.g in the Main menu at the very bottom in grey), we would list which version of the neos-ui package people are using; to get better bug reports in the upcoming weeks.

mstruebing:
I'm not entirly sure if this is the correct place to call this script.
Otherwise I think it's working fine.

curr

closes #1478

### DIFF
--- a/packages/utils-helpers/src/getVersion.js
+++ b/packages/utils-helpers/src/getVersion.js
@@ -1,5 +1,5 @@
 const getVersion = () => {
-    return 'v1.0.3 ea52b957e6ed807bb36845766054eccb8ced678d';
+    return 'v1.0.3';
 };
 
 export default getVersion;
--- a/packages/utils-helpers/src/getVersion.js
+++ b/packages/utils-helpers/src/getVersion.js
@@ -1,0 +1,5 @@
+const getVersion = () => {
+    return 'v1.0.3';
+};
+
+export default getVersion;
--- a/packages/utils-helpers/src/getVersion.js
+++ b/packages/utils-helpers/src/getVersion.js
@@ -1,5 +1,5 @@
 const getVersion = () => {
-    return 'v1.0.3';
+    return 'v 1.0.3';
 };
 
 export default getVersion;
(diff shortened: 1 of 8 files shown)


packages/utils-helpers/src/getVersion.js
```diff
@@ -0,0 +1,5 @@
+const getVersion = () => {
+    return 'v1.0.3';
```
dimaip: maybe also add commit hash, so if it's on master, we could identify the correct commit?

skurfuerst: Instead of parsing this file again, let's just re-create it completely in the bash script :) Makes it more robust I'd say

mstruebing:
Commit hash is included now and the script recreates the file (you would manually need to add it to the repository - travis test will fail elsewise with this failing test: `should return the correct version, if this test fails try running ./Build/createVersionFile.sh`) 
curr

mstruebing:
Argh, Ok, It's clear that with this we have to re-create this file on every commit to let travis pass. 
We could automate this with git-hooks, but I'm not sure if we should. I'm removing the commit hash for now.

dimaip:
What would it show if you are running master and didn't run any scripts?

mstruebing:
@dimaip This script removes the `getVersion.js` file in `utils-helpers` and recreates it with the current version parsed from the `package.json` of the `neos-ui` subpackage (btw, is a any reason why the root package.json doesn't have a version?). 

The test runs this created file and checks it with the parsed version of the `package.json` from `utils-helpers` subpackage.

So if you are on master and the script was not run it should show the latest released version (unless you bumped the version and ran the script)

dimaip:
The root package.json doesn't have a version because it's not meant to be required by any 3rd-parties.

Btw, just an idea, an alternative to bash script could be to do it in Flow. If it's cached, it's not going to be expensive. Also we'd be able to provide more detailed information, e.g. current commit hash and possibly some more debug information in the future.

skurfuerst:
I'd say we merge it now as it is a debug tool; and we might revisit it later if needed. Fine for you @dimaip ? :)

dimaip:
Done

PR #1499, opened 2017-12-27
TASK: Respect `startingPoint` in ReferenceDataLoader

### LINKED ISSUE #1496: ReferenceEditor does not respect a `startingPoint`
Given this configuration:
```
    news:
      type: reference
      ui:
        reloadIfChanged: true
        inspector:
          editorOptions:
            nodeTypes:
              - 'Yeebase.t3nDe:Article'
            startingPoint: /articles
```

The startingPoint is not respected. Therefore the ReferenceEditor won't find any results

johannessteu:
Fixes #1496

### DIFF
--- a/packages/neos-ui-editors/src/Editors/Reference/referenceDataLoader.js
+++ b/packages/neos-ui-editors/src/Editors/Reference/referenceDataLoader.js
@@ -19,7 +19,8 @@ export default ({isMulti}) => WrappedComponent => {
             options: PropTypes.shape({
                 nodeTypes: PropTypes.oneOfType([PropTypes.string, PropTypes.arrayOf(PropTypes.string)]),
                 placeholder: PropTypes.string,
-                threshold: PropTypes.number
+                threshold: PropTypes.number,
+                startingPoint: PropTypes.string
             }),
 
             nodeLookupDataLoader: PropTypes.shape({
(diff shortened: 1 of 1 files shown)


packages/neos-ui-editors/src/Editors/Reference/referenceDataLoader.js
```diff
@@ -103,9 +104,13 @@ export default ({isMulti}) => WrappedComponent => {
         }
 
         getDataLoaderOptions() {
+            const contextNode = this.props.options.startingPoint ?
+                this.props.contextForNodeLinking.set('contextNode', this.props.options.startingPoint) :
```
kitsunet: you are modifying props there. I don't think that should EVER happen. You might need to create a deep clone of contextForNodeLinking if starting point is set and then set it on that copy.

johannessteu: Does it? contextForNodeLinking is a Immutable map and i tought that the set() creates a copy (https://facebook.github.io/immutable-js/ second code example)

johannessteu: It's initalizied here: https://github.com/neos/neos-ui/blob/master/packages/neos-ui-redux-store/src/UI/NodeLinking/index.js#L24

kitsunet: Oh, is it? In that case YAY, but IMHO the wording is awkward, not your fault though.. `set` shouldn't do `createDeepCloneWithThisPropertyChanged`

PR #1506, opened 2017-12-29
Update to latest moment js version 2.20.1

markusguenther:
The current used version 2.14.1 is marked as insecure.
Therefore it would be good to upgrade the package.

It turns out that testcafe has issues with the current momentjs and therefore Is submitted some patches to them.

Hope that they get merged soonish and find the way into a new release.
Otherwise this is blocked for a while ;)

https://github.com/DevExpress/testcafe-reporter-spec/pull/4
https://github.com/DevExpress/testcafe-reporter-list/pull/4

Fixes: #1359

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "lodash.throttle": "^4.1.1",
     "lodash.unescape": "4.0.1",
     "mime-types": "^2.1.10",
-    "moment": "^2.14.1",
+    "moment": "^2.20.1",
     "monet": "^0.8.10",
     "normalize.css": "^5.0.0",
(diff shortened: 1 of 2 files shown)


AlexanderMoskovkin:
Hi @markusguenther,

You can try to fix this without waiting for the new TestCafe release. Mainly the issue is related to the way how yarn installs modules in some cases. 
To fix the issue in your PR you can perform three steps:
1) remove `yarn.lock` locally;
2) add the `resolution` field into package.json:
```json
"resolutions": {
    "moment": "^2.20.1"
  }
```
3) install dependencies with yarn.

PR #1515, opened 2018-01-05
BUGFIX: Render Preview/Editing mode panel with at least one mode

dfeyer:
Without the change the UI is broken if the Editing or Preview panel
is empty.


mstruebing:
@dfeyer Could you provide some information in how to reproduce this behaviour?

dfeyer:
@mstruebing You can try to remove all Preview mode by example

mstruebing:
The same with this one, I'm not able to push. Sorry, I want to get stuff done and merged at this point, I hope you don't be mad at me

PR #1532, opened 2018-01-16
BUGFIX: Add 10px in order to fit the SelectBoxHeader inside the box

### LINKED ISSUE #1528: Link selector overlaps Container
image

## Tested On: 
Chrome / Mac
UI Version 1.0.4

mstruebing:
I'm sure this isn't the best possible solution but I didn't came up with anything else working.
So if you have an idea which might work let me know :)

closes #1528

screenshot_2018-01-16_06-47-20


markusguenther:
Looked into this some days ago and wondered why their is the gap of 10 pixels. It is maybe a problem with the rendering of the path. Sure this is a way to fix it but I bet that this is a thing that is related to the beloved and magical selectbox component.

Anyway it works out for this case :)

mstruebing:
It's better than before and no one else seems to have an eye on this, so I'm merging this for now.

markusguenther:
Had started with the fix and did not finish it ;) ...
When I am done with the dependency stuff I will try to concentrate on bugs.

Issue #1534, opened 2018-01-16
Localize Date Picker


lorenzulrich:
The date picker has two labels that are not localized:

* Today
* Apply

These need to be localized

lorenzulrich:
This can be closed IMO?

PR #1535, opened 2018-01-16
TASK: Localize Date Picker (refs #1534)

lorenzulrich:


dfeyer:
@lorenzulrich Happy to see your first contribution here ;)

Did you check if those labels exists for the old UI, if yes, just use the existing IDs

lorenzulrich:
@dfeyer Thanks ;-).

The label "today" doesn't exist, the label "Apply" does, but in a different context (applying changes made in the inspector). Are the labels used generally or context-specific?

mstruebing:
I would suggest to introduce a new label for the Datepicker-Apply to not mix contexts with the apply button. Make changes in the future easier.

mstruebing:
@lorenzulrich Please rebase your branch against the latest master, the master is repaired so that travis will run through if everything is fine

lorenzulrich:
Comment from @mstruebing:

> It's something with the dependencies fucked up. `lerna ERR! execute     Cannot find module '@neos-project/neos-ui-decorators' from 'dateInput.js'`

> Maybe it's enoug to add this dependencie to this subpackage, but I need to try.

lorenzulrich:
@mstruebing I think I did everything I could but it still doesn't work out. To me, it seems that the problem is that `react-ui-components` seems to be a kind of "standalone" package with an own Git repository:

```
  "repository": {
    "type": "git",
    "url": "https://github.com/neos/react-ui-components.git"
  },
```

(Which throws a 404 - is it maybe a protected repo?)

So this package is not aware of the packages that are not on NPM, but just a part of `Neos.Neos.Ui`.

mstruebing:
@lorenzulrich: I looked at your PR and it seems lerna isn't doing what it should. The `@neos-project/neos-ui-i18n` dependency should be in the `node_modules` of the
`react-ui-components` package, but isn't. So when I symlinked the package manually inside the `node_modules` the tests are running through. However, I've asked @grebaldi how to
handle this with lerna and the first thing was that the `react-ui-components` should not know about this, they are only presentational and didn't know anything, they get
 everything they need from outside. So the correct way should be to add the needed labels to the Props and give them in from the outside, in this case:
`neos-ui-editors/src/editors/DateTime` (that is already connected to the store for example).

This is also the case for https://github.com/neos/neos-ui/pull/1537

packages/react-ui-components/src/DateInput/dateInput.js
```diff
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import enhanceWithClickOutside from 'react-click-outside';
 import moment from 'moment';
 import mergeClassNames from 'classnames';
+import I18n from '@neos-project/neos-ui-i18n';
```
mstruebing: Please remove this dependency and pass the needed labels from the outside (`neos-ui-editors`)

packages/react-ui-components/src/DateInput/dateInput.js
```diff
@@ -96,8 +95,6 @@ export class DateInput extends PureComponent {
             theme,
             value,
             id,
-            applyLabel,
-            todayLabel,
```
mstruebing: This should be readded, sorry!

lorenzulrich:
Yeah! Thanks to your explanations, I now understand that the `react-ui-components` are presentational only. It seems that this change is now looking good and I hope it will get merged.

kitsunet:
@lorenzulrich The respective labels used here do not exist yet, right? Is there a corresponding PR in the neos-dev-collection?

lorenzulrich:
@kitsunet see here: https://github.com/neos/neos-development-collection/pull/1851

kitsunet:
Now this conflicts after the other change is merged. But if you can fix it quickly @lorenzulrich I would merge right after :)

kitsunet:
Waiting for travis here.

Issue #1503, opened 2017-12-28
The time picker is awful: click multiple times to choose the time


dimaip:
I tried hard to implement it differently at the time, but what we have now requires tens of clicks to pick the correct time.
The problem is that we'd be forced to use a completely different datatime editor, and I didn't find any nice React datepickers so far. In the worst case scenario we could implement it from scratch...

drlimbo:
24 hour time format would be very nice..

lorenzulrich:
What about this one:
https://github.com/YouCanBookMe/react-datetime

It is a fork of this one http://dev.quri.com/react-bootstrap-datetimepicker/ which is based on `bootstrap-datetimepicker` (which has no connection to React). I integrated the non-React variant some time ago and found it quite nicely. It also relies on Moment.js which is a nice library.

lorenzulrich:
Fun fact: That's the one already used in Neos UI. Strange that there is no 24h support...

dimaip:
Well, so maybe it's possible to enable the 24h format then?
I'm more concerned with how ugly the time picker is overall, e.g. you need 30 clicks to set time to 12:30...

lorenzulrich:
I'm currently looking into setting up the UI for development, but I'm not really a good React dev. I let you know about the results. I agree the UI could be better, but 24h time format should have priority.

lorenzulrich:
I pushed a first iteration only respecting the locale of the user interface

I need more guidance for overriding this through settings, especially where to place and how to name such a setting. There is almost no example in the current Settings.yaml of neos-ui.

dimaip:
Hey but it still is awful: you need to click 30 times to set 12:30, no?

lorenzulrich:
Yes, I agree the UX has not been improved. To my knowledge, there are not many good date/time picker components out there.

The current one could be integrated differently, by displaying the actual input field and not a fake input field and enable the `input` option, see readme at https://github.com/YouCanBookMe/react-datetime. This would allow to set the date and time manually in the input field.

jonnitto:
In the [documentation](
https://neos.readthedocs.io/en/stable/References/PropertyEditorReference.html#property-type-datetime-datetimeeditor-date-time-selection-editor), it is something written about `minuteStep` (default to `5`). I've tested it with the newest version of the UI (1.3.3), but this option gets ignored. Besides that also the `placeholder` option gets ignored

jonnitto:
[...]

PR #1537, opened 2018-01-16
TASK: Respect locale in date & time picker #1503

lorenzulrich:


packages/react-ui-components/src/DateInput/dateInput.js
```diff
@@ -3,7 +3,12 @@ import PropTypes from 'prop-types';
 import enhanceWithClickOutside from 'react-click-outside';
 import moment from 'moment';
 import mergeClassNames from 'classnames';
+import {connect} from 'react-redux';
+import {$transform, $get} from 'plow-js';
 
+@connect($transform({
+    interfaceLanguage: $get('user.preferences.interfaceLanguage')
```
dimaip: can't it have a value like `default` or smth?

lorenzulrich: There is always a default value, because it falls back to the Neos setting `Neos.Neos.userInterface.defaultLanguage` if it's not set in user preferences.

lorenzulrich: (This is set to `en` by default)

mstruebing:
@lorenzulrich Please rebase your branch against the latest master, the master is repaired so that travis will run through if everything is fine

lorenzulrich:
Comment my @mstruebing in Slack:

```lerna ERR! execute /home/travis/build/neos/neos-ui/packages/react-ui-components/src/DateInput/dateInput.js
lerna ERR! execute   110:13  error  'interfaceLanguage' is missing in props validation  react/prop-types
```
> That explains why code style fails (sorry that the output it this hard to read)
>
> And the unit tests are failing because of something related to the redux store I guess. `E.g: lerna ERR! execute     Invariant Violation: Could not find "store" in either the context or props of "Connect(DateInput)". Either wrap the root component in a <Provider>, or explicitly pass "store" as a prop to "Connect(DateInput)".`
>
> But I need to take a deeper look how to fix this.

packages/react-ui-components/src/DateInput/dateInput.js
mstruebing: Like in https://github.com/neos/neos-ui/pull/1535, please move this to the wrapping component (`neos-ui-editors`)

Issue #1552, opened 2018-01-20
Update dev dependency cross-env


markusguenther:
## Description

The react ui has an outdated dev dependency with cross-env. We use it for the storybook e.g.

PR #1553, opened 2018-01-20
TASK: Update cross-env to version 5.1.3

markusguenther:
Resolves: #1552

### DIFF
--- a/packages/react-ui-components/package.json
+++ b/packages/react-ui-components/package.json
@@ -63,7 +63,5 @@
   "jest": {
     "preset": "@neos-project/jest-preset-neos-ui"
   },
-  "dependencies": {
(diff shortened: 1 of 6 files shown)


packages/react-ui-components/package.json
```diff
@@ -62,5 +62,8 @@
   },
   "jest": {
     "preset": "@neos-project/jest-preset-neos-ui"
+  },
+  "dependencies": {
+    "cross-env": "^5.1.3"
```
mstruebing: Could you explain why you are adding this dependency here?

markusguenther: Yes I can ... it was to late and I found cross-env in the package.json of react-ui-components... but missed that it was not in dependencies :(

PR #1570, opened 2018-01-23
FEATURE: Add a signal after node creation handles are applied

daniellienert:
Node creation handlers are a new way to react on node creation triggered by a front-end action. One implementation is the node template package, which makes it possible to add arbitrary child nodes to the created node. We only have signals for nodeCreation of a single node, but none to react on the completion of the complete node creation process.

Usage Example: 
1. A node is created, a news node for example.
2. [Node templates](https://github.com/Flowpack/Flowpack.NodeTemplates) create a structure of nested nodes beneath the news node, including one that contains the date of the news.
3. After all nodes are created, the news node now needs to be moved into a date hierarchy using [archivist](https://github.com/punktDe/archivist) using this date.

### DIFF
--- a/Classes/Neos/Neos/Ui/Domain/Model/Changes/AbstractCreate.php
+++ b/Classes/Neos/Neos/Ui/Domain/Model/Changes/AbstractCreate.php
@@ -157,5 +157,19 @@ abstract class AbstractCreate extends AbstractStructuralChange
                 }
             }
         }
+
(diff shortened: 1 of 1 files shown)


dimaip:
Yep, I don't understand why you might need this?

daniellienert:
@dimaip, @kdambekalns thanks for the review! I thought the use case is obvious 😬. Added a detailed description. I hope that describes the use case.

PR #1580, opened 2018-01-27
TASK: Allow custom icon classnames

gerdemann:
Removes the FontAwesome warning and allows custom icon classnames.

Fixes 1572


gerdemann:
@mstruebing I updated the tests. I don't have much experience with the unit tests, is that so correct? :see_no_evil:

mstruebing:
Yep that's it. :)
Thank you @gerdemann

Issue #1508, opened 2017-12-29
Tree search is dead slow on large projects => rewrite the tree strucutre


dimaip:
I have a project with 50k nodes. In the old UI it took 3s to search the tree for a particular node. In the new UI it takes 30s.

The problem is not with searching itself, but in `renderNodesWithParents`.
We have 6000 sibling nodes, that hidden from the tree by default. When one of those nodes is found, its parent node is rendered too, and for its parent node, all children are rendered too, which results in all 6k nodes to be rendered as children.

I tried to quickly optimize it, but couldn't find an easy way to do it.

I think the way to do it, is to stop transferring (and storing in state) all children of a node, and rather just keeping a pointer to the node's parent, and then traversing all the nodes. In memory that is super fast (millions ops), so I guess that shouldn't cost us too much, especially compared to traversing all those nodes on PHP side + transferring a lot of redundant data via network.
Maybe we could use even more advanced algorithm for reconstructing the tree from child->parent pointers on the client side.

dimaip:
Actually currently much bigger issue is the performance of AOP, so I don't think this issues is relevant at the moment, before this is tackled: https://github.com/neos/neos-ui/issues/1646

bwaidelich:
Quick update:
`NodeInfoHelper::renderNodesWithParents()` really seems to be the major bottleneck because it loads all the nodes on the same level (and all parent levels) including their URLs etc..

As a quick shot I added a new parameter `$skipChildNodes` to `renderNode()` and set that to `true` in the `while ($parentNode->getNodeType()->isOfType($baseNodeTypeOverride) && $parentNodeIsUnderneathSiteNode) {` loop. That reduced the loading time dramatically but didn't result in a proper tree.

I'm still looking for a quick work around for this. In the long term that whole backend part needs a major rework IMO. It's doing crazy things currently.

Not exactly related (but I want to dump it from my brain): Why do we need so much info to display the tree at all? IMO a simple tree structure with id, label and dirty flag should suffice.. I probably miss some information here. But I'm quite sure we don't need to render the Frontend & Backend URIs only to display the tree.
In general the preview link could point to `/neos/preview?node=...` which would redirect to the final URI

dimaip:
> In general the preview link could point to /neos/preview?node=... which would redirect to the final URI

Cool idea, I like it.

hlubek:
[...]

Issue #1579, opened 2018-01-26
Initial load much slower than with legacy UI


kdambekalns:
### Description

On a site (I am working on) with a lot of nodes, the initial load of the React UI delivers a document of more than 1.75 MB - which takes 80 seconds to render (don't mind the absolute value, but keep it in mind for comparison.)

#### Expected behavior

The React UI should not load slower than before, especially because everyone claims it is so mch faster.

#### Actual behavior

<img width="594" alt="neos_content_node____" src="https://user-images.githubusercontent.com/95873/35434737-74e8950c-0288-11e8-82e5-19fd37572e74.png">

The old UI (with a `loadingDepth` of 4!) loads in 430kB and 2.5 seconds and then fetchs the tree in 320kB taking 4 seconds (but since the UI is alerady visible it feels much faster, perceived performance and all.)

So even when combined (750kB and 6.5 seconds) the old UI loads 12 times faster and produces only 42% the amount of data (give and take a little.)

### Affected Versions

Neos: 3.3.3
UI: 1.0.4

### Workaround

When lowering the `loadingDepth` of the node tree to 2 (from the default value of 4), the site goes down to 230 kB (and takes 8 seconds to produce.)

But that cannot be the solution we want our customers to be using, eh?

dimaip:
The issue here is not JS size. The PHP code to render nodes for the UI is super slow.
This is related: https://github.com/neos/neos-ui/issues/1508

kdambekalns:
Ok, no wonder… this is coming from
```      initialState:
        changes:
          pending: {  }
          processing: {  }
          failed: {  }
        cr:
          nodes:
            byContextPath: '${Neos.Ui.NodeInfo.defaultNodesForBackend(site, documentNode, controllerContext)}'
            siteNode: '${q(site).property(''_contextPath'')}'
```
calling

```    public function defaultNodesForBackend(NodeInterface $site, NodeInterface $documentNode, ControllerContext $controllerContext)
    {
        $flowQuery = new FlowQuery([$site, $documentNode]);
        $nodes = $flowQuery->neosUiDefaultNodes($this->baseNodeType, $this->loadingDepth)->get();

        $result = [];
        foreach ($nodes as $node) {
            $this->renderNodeToList($result, $node, $controllerContext);
        }

        return $result;
    }
```

which in turn… 🙈

dimaip:
See my analysis in the issue linked

dimaip:
[...]

PR #1581, opened 2018-01-27
WIP TASK: improve performance of initial state generation

dimaip:
Related: #1579 
Resolves: #1508

This PR does the following:
- reconstructs parent->child relations on the client for each change, for all nodes. But in-memory it's basically free, 0.3ms for 500 nodes, so no need to bother about that.
- loads default nodes for tree asynchronously. This is especially important, since now it's rather slow (due to policy enforcement)
- loads nodes for tree with `$omitMostPropertiesForTreeState`, which drastically reduces the payload size


dimaip:
So here comes actual performance analyses.
I've been testing on a website with **540 document nodes**, expanding the tree to show all of them.

1. The `neosUiDefaultNodes` FQ operation uses recursive `getChildNodes` calls and takes `2.7s`. While that is not super fast, I don't know how to improve that.

2. Now for the `renderNode` rendering itself.

The biggest cost are our own policy aspects: https://github.com/neos/neos-ui/blob/master/Classes/Neos/Neos/Ui/Aspects/PolicyAspect.php

The next thing is children rendering, which I removed in the PR above.

Then generating `preivewUris` takes a significan't amount of time.

If you remove all of that you end up with `0.58s` to render 540 nodes, which is OK.

Obviously we need to start with doing something about the policy aspect...

Here are the measurements themselves:

|Measurement|time, s|
|---|---|
|Original|24|
|No AOP|7.3|
|No AOP, No children|3.6|
|No AOP, No children, omitMostPropertiesForTreeState|2.6|
|No AOP, No children, omitMostPropertiesForTreeState, noPreviewUri|0.58|

Classes/Neos/Neos/Ui/Fusion/Helper/NodeInfoHelper.php
```diff
@@ -103,20 +105,6 @@ public function renderNode(NodeInterface $node, ControllerContext $controllerCon
                 $nodeInfo['previewUri'] = $this->uri($nodeInLiveWorkspace, $controllerContext);
             }
         }
-
-        // child nodes for document tree, respecting the `baseNodeType` filter
-        $documentChildNodes = $node->getChildNodes($baseNodeType);
```
skurfuerst: @dimaip does the BaseNodeType filter in the tree still work? :-) Just asking

dimaip: @skurfuerst yes, absolutely, just tested it with `baseNodeType: 'Neos.Neos:Document,!Neos.Neos:Shortcut'`:

image
`

dimaip:
Found a few edge cases, will fix tonight.

dimaip:
Duh, unfortunately I have to give up on this one. My PR was great, but it didn't consider one usecase: checking if nodes have children... And if I do extra request to check it, it would be not much different from how it works currently, performance-wise.
So I'm giving up in favor of #1640 

Anyways, the performance killer is checking for policies currently, before we don't fix that there's no sense in optimizing anything else...

PR #1585, opened 2018-01-28
TASK: Clear search term of SelectEditor when selecting an option #1486

lorenzulrich:


packages/react-ui-components/src/SelectBox/selectBox.js
```diff
@@ -274,6 +274,7 @@ export default class SelectBox extends PureComponent {
     handleChange = option => {
         const optionValueAccessor = this.getOptionValueAccessor();
         this.props.onValueChange(optionValueAccessor(option));
+        this.state.searchTerm = '';
```
kitsunet: need to use `this.setState`

lorenzulrich: Done.

Issue #1539, opened 2018-01-17
Node creation dialog element labels translation not working


DrillSergeant:
Seems labels with value "i18n" are not translated. Instead "i18n" is used as a fixed value.

**Steps to reproduce:**
* Add configuration to a node that adds an element to the node creation dialog:
```

    creationDialog:
      elements:
        date:
          type: DateTime
          ui:
            label: i18n
            editor: 'Neos.Neos/Inspector/Editors/DateTimeEditor'
            editorOptions:
              format: 'd.m.Y'
```
* Create a node (of the type you just added the creation dialog element for)

**Expected result:**
I see the element in the dialog with a translated label (or at least the i18n-key as fallback if no translation  exists yet.

**Actual result:**
The string "i18n" is used as label:
![image](https://user-images.githubusercontent.com/883552/35051207-fb827d98-fba4-11e7-975d-79249c36cddd.png)

lorenzulrich:
Just to let anyone know: I'm currently working on this. Most of it needs to be adapted inside neos/neos.

lorenzulrich:
I added this functionality for the label of the creationDialog elements. I'm not sure if the syntax inside the XLF file is OK, e.g. `creationDialog.title`. I think it's not too bad because it follows the known pattern `properties.foo`, `groups.bar` etc.. In contrast to `creationDialog`, they are in plural form, so an alternative - which is a little clumsy IMO - would be `creationDialogElements`.

lorenzulrich:
At the same time and as an example I made the `title` property of the default `Document` node type make use of the i18n functionality. I intentionally used `Title` instead of `Node title` because in the inspector of a page, the property is also named `Title`.

PR #1586, opened 2018-01-29
TASK: Change node title label in creation dialog to i18n

lorenzulrich:
Depends on this change: neos/neos-development-collection#1871

Fixes: #1539


mstruebing:
The behaviour is exactly the same with or without this PR currently.
No label in the Configuration given no label; any label given the label is used, but as the exact string.

So we don't need to wait for the depending change to get merged.

PR #1587, opened 2018-01-29
BUGFIX: Property group configured as collapsed must be collapsed

lorenzulrich:


lorenzulrich:
I don't understand why the frontend test fails. The test failing only has to do with collapsing the right side as a whole.

mstruebing:
Locally the tests are failing too. When I inspect the RightSideBar I see that the Property isHidden is behaving correctly.
When I remove the last asserting and click it is running through.
As soon as I add it is failing.

mstruebing:
maybe @dimaip can help here? :)

dimaip:
Will try to dig into it over the weekend.

mstruebing:
I rebased on latest master, see if this helps.

mstruebing:
Seems an old testcafe version was the problem or something.

Issue #1459, opened 2017-12-18
Make the initial screen Loading label translatable


soee:
It would be great if the initial screen/overlay layer with `Loading` text could be tranlsatable.

lorenzulrich:
As far as I can see from reading the code, "Loading..." is displayed at a very early stage of initializing the React application. Therefore the i18n labels are not fetched yet.

I suggest (and will push a PR) to use a spinner of the already existing `<Icon>` component.

![image](https://user-images.githubusercontent.com/1816023/35508350-595f82fe-04f0-11e8-9283-d3d82900a385.png)

PR #1588, opened 2018-01-29
TASK: Use loading spinner instead of (non-localized) text

lorenzulrich:
When loading the Neos UI, a text "Loading..." appears that is not
localized. It is displayed at a very early stage of UI initialization
when the language labels have not been fetched yet.

This change makes use of a spinning icon instead of text.

Fixes: #1459

### DIFF
--- a/packages/neos-ui/src/index.js
+++ b/packages/neos-ui/src/index.js
@@ -22,6 +22,8 @@ import Root from './Containers/Root';
 import apiExposureMap from './apiExposureMap';
 import DelegatingReducer from './DelegatingReducer';
 
+import Icon from '@neos-project/react-ui-components/src/Icon/';
+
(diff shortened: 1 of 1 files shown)


mstruebing:
ezgif com-video-to-gif

dfeyer:
Let's make it bigger ;) really. It's a bit small here

lorenzulrich:
Which one of these is it? http://fontawesome.io/examples/#animated

mstruebing:
Upps, sorry, I played around and forgot to undo my changes before recording. I've used just the first one. But I would be good with your one.

lorenzulrich:
OK, here you go. 30px it is, now!

mstruebing:
Awesome, thx!

Issue #1589, opened 2018-01-31
Preview Central does not immediately show all preview modes


gerdemann:
### Description

Preview Central does not immediately show all modes.
![Preview Central](https://cdn-pro.dprcdn.net/files/acc_496443/9lrZvb)

### Steps to Reproduce

1. Configure preview modes
2. Toggle Preview Central
3. Click an other preview mode

#### Expected behavior

Preview Central shows immediately all modes.

### Affected Versions

Neos: 3.3.3

UI: 1.0.4

PR #1591, opened 2018-01-31
BUGFIX: Show all modes in Preview Central

gerdemann:
The width of the slick slider was not set, so not all modes were displayed.

Fixes: #1589

### DIFF
--- a/packages/neos-ui/src/Containers/EditModePanel/Panel/slick-styles.vanilla-css
+++ b/packages/neos-ui/src/Containers/EditModePanel/Panel/slick-styles.vanilla-css
@@ -38,6 +38,7 @@
     top: 0;
     left: 0;
     display: block;
+    width: 5000px; /* Set initial width because Slick does not set it correct. (See https://github.com/neos/neos-ui/pull/1591) */
 }
 .slick-track:before,
 .slick-track:after


packages/neos-ui/src/Containers/EditModePanel/Panel/slick-styles.vanilla-css
```diff
@@ -38,6 +38,7 @@
     top: 0;
     left: 0;
     display: block;
+    width: 5000px;
```
dimaip: Hey, thanks for the fix! Magic number doesn't look good. Maybe something like `right: 0` could help here? Can't test right now.

gerdemann: Unfortunately, that doesn't help. Slick calculates a width via Javascript and writes it there inline.
I think because the slider is not visible at the beginning, Slick can't calculate a width.

mstruebing: Do the container has any width? if so could we use the width of the container?

gerdemann: No, slick-track must be as big as all slides in the slider. Slick actually calculates this himself, which unfortunately doesn't seem to work in this case.

mstruebing: Could you please at least write a comment into the code why we need this? 
This would be helpful in the future. Maybe also link to the PR in the comment.

gerdemann: I added a comment.

mstruebing: Thanks!

Issue #1593, opened 2018-01-31
No focus indicator


beardcoder:
### Description

No highlighting of focus elements.

### Steps to Reproduce

1. Open the neos-ui "mypage.xy/news"
2. Press the tab key

#### Expected behavior

After press the tab key you see the default browser visual focus indicator.

![neos ui](https://cdn.pbrd.co/images/H5uwXa0.png)

#### Actual behavior

No visual focus indiucator.

### Affected Versions

UI: dev-master 1.0.4

PR #1597, opened 2018-01-31
TASK: Enable focus indicator

beardcoder:
Resolves: #1593


beardcoder:
Hey, no problem.
The problem in Firefox is the new engine.

This engine use the :-moz-focusring selector. The default setting is a dark "ants line". This behavior is problematic on dark ui's. So wie must be define a special styling.

I think the best way is. At first we enable the default behavior and in the second step we need a global styling for all focus indicators.

mstruebing:
I agree your argumentation. Thanks!

PR #1602, opened 2018-02-01
TASK: Resolve warnings of composer validate

markusguenther:
Composer validate throws some warnings. This should resolve the
issues.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,11 +2,11 @@
     "name": "neos/neos-ui",
     "type": "neos-package",
     "description": "NEOS CMS UI rewritten in React",
-    "license": "GPL-3.0+",
+    "license": "GPL-3.0-or-later",
     "require": {
-        "neos/flow": "*",
-        "neos/neos": "*",
-        "neos/neos-ui-compiled": "dev-master"
+        "neos/flow": "~4.0",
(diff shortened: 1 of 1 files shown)


markusguenther:
Oversee the PR 8 days ago "Update SPDX license identifier in composer.json by kdambekalns · Pull Request #1575 · neos/neos-ui"

composer.json
```diff
@@ -2,11 +2,11 @@
...
     "type": "neos-package",
     "description": "NEOS CMS UI rewritten in React",
-    "license": "GPL-3.0+",
+    "license": "GPL-3.0-or-later",
     "require": {
-        "neos/flow": "*",
-        "neos/neos": "*",
-        "neos/neos-ui-compiled": "dev-master"
+        "neos/flow": "~4.0",
+        "neos/neos": "~3.3",
```
lorenzulrich: Not sure if these constraints are wise. AFAIK there is no 3.x branch of the new UI yet (which IMO is good because bugfix releases should be in focus) so this will cause trouble for devs using master.

markusguenther: Missed to update my local instance to the latest version and there @kdambekalns and @dfeyer already fixed the issues ;)

Therefore I closed the PR

Issue #1611, opened 2018-02-05
Improve handling of dimension constraints in multi-dimension setups


dimaip:
Currently constraints are only correctly applied on first load. Once you change one dimension, the second one should be re-filtered according to selected dimension's constraints.

Could be tricky to implement, as possibly it would require a server request to find out allowed dimensions.
In the old UI this logic feels really off too.

## Steps to reproduce

Take a similar dimension config:

```
    contentDimensions:
      country:
        label: 'Country'
        icon: icon-language
        default: eng
        defaultPreset: eng
        presets:
          eng:
            label: 'England'
            values:
              - eng
            uriSegment: eng
          rus:
            label: 'Russia'
            values:
              - rus
            uriSegment: rus
      language:
        label: 'Neos.Demo:Main:contentDimensions.language'
        icon: icon-language
        default: en_US
        defaultPreset: en_US
        presets:
          all: null
          en_US:
            label: 'English (US)'
            values:
              - en_US
            uriSegment: en
          en_UK:
            label: 'English (UK)'
            values:
              - en_UK
              - en_US
            uriSegment: uk
          de:
            label: German
            values:
              - de
            uriSegment: de
            constraints:
              country:
                'rus': true
                '*': false
          fr:
            label: French
            values:
              - fr
            uriSegment: fr
          nl:
            label: Dutch
            values:
              - nl
              - de
            uriSegment: nl
          da:
            label: Danish
            values:
              - da
            uriSegment: da
          lv:
            label: Latvian
            values:
              - lv
            uriSegment: lv
```

2. Select England as country. "German" language will be correctly hidden, due to constraints.
3. Choose Russia as country. "German" language would still be hidden, unless you apply selection of Russia and reload the page.

dimaip:
So yes, I've checked how it was implemented in the old UI: it was making an extra ajax request to check constraints. Will do the same.

PR #1615, opened 2018-02-06
FEATURE: dynamic updating of dimension constraints

dimaip:
Fixes: #1611

See issue for ways to test.

peek 2018-02-06 15-13

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "build": "./Build/createVersionFile.sh && cross-env NODE_ENV=production yarn build:ui",
     "build:ui": "NEOS_BUILD_ROOT=$(pwd) webpack --progress --colors",
     "build:ui:watch": "yarn build:ui -- --watch",
-    "build:ui:watch-poll": "yarn run build:ui --watch-poll --watch",
+    "build:ui:watch-poll": "yarn run build:ui -- --watch-poll --watch",
     "start-storybook": "lerna run --scope @neos-project/react-ui-components start",
     "test": "lerna run test --concurrency 1",
     "test:e2e": "yarn run testcafe chrome:headless Tests/IntegrationTests/* --selector-timeout=30000 --assertion-timeout=30000",
--- a/packages/neos-ui-backend-connector/src/Endpoints/index.js
+++ b/packages/neos-ui-backend-connector/src/Endpoints/index.js
@@ -121,6 +121,16 @@ export default routes => {
     })).then(response => response.json())
     .catch(reason => fetchWithErrorHandling.generalErrorHandler(reason));
 
+    const contentDimensions = (dimensionName, chosenDimensionPresets) => fetchWithErrorHandling.withCsrfToken(() => ({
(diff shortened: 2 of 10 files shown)


mstruebing:
Looks good at the first glance, will have a closer look tomorrow morning.

package.json
```diff
@@ -7,7 +7,7 @@
     "build": "./Build/createVersionFile.sh && cross-env NODE_ENV=production yarn build:ui",
     "build:ui": "NEOS_BUILD_ROOT=$(pwd) webpack --progress --colors",
     "build:ui:watch": "yarn build:ui -- --watch",
-    "build:ui:watch-poll": "yarn run build:ui --watch-poll --watch",
+    "build:ui:watch-poll": "yarn run build:ui -- --watch-poll --watch",
```
mstruebing: What about this?
`warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.`
Also for me, the resulting command is exactly the same, do you have an old yarn version?

packages/neos-ui-backend-connector/src/Endpoints/index.js
```diff
@@ -121,6 +121,16 @@ export default routes => {
...
     .catch(reason => fetchWithErrorHandling.generalErrorHandler(reason));
 
+    const contentDimensions = (dimensionName, chosenDimensionPresets) => fetchWithErrorHandling.withCsrfToken(() => ({
+        url: urlWithParams(`${routes.core.service.contentDimensions}/${dimensionName}.json`, {chosenDimensionPresets}),
+        method: 'GET',
+        credentials: 'include',
+        headers: {
+            'Content-Type': 'application/json'
+        }
+    })).then(response => response.json())
```
mstruebing: I'm just asking, but as we are on node 8.9.4 now, shouldn't we try to use async, await?

package.json
dimaip: The script just exits for me after the first run without it...

packages/neos-ui-backend-connector/src/Endpoints/index.js
dimaip: Wait, how does it relate to node? This is Babel code executed in the browser...      
But yes, could use async here, but I'd rather be consistent and replace it everywhere, so the code would be easier to understand

mstruebing: You are right, my fault.
As it runs through babel anywhere I would say let's keep it this way.

package.json
dimaip: Oh wait, I do! Will fix.

dimaip: `0.27.5` seems to be the latest version available for Ubuntu... Could we maybe leave this like that for now, and in some time remove `--` from both commands? Otherwise somebody on lower versions of yarn could also waste some time on this...

dimaip:
Thanks for the review!!!

Issue #1568, opened 2018-01-23
Missing support for Depending Properties


dfeyer:
The React UI does not support Depending Properties:

http://neos.readthedocs.io/en/stable/CreatingASite/NodeTypes/DependingProperties.html

We should implement at least the support without Backward compatibility, the current solution is lower level Ember based, so i think no way to have keep the compatibility here. In the documentation it's clearly noticed that the current API is experimental, so I think it's fine to break the compatibility. 

Reported on slack by @kdambekalns see https://neos-project.slack.com/archives/C0U0KEGDQ/p1516656865000167

dimaip:
With React, it's very simple to implement a declarative API, instead of a low-level imperative one like we did before.

I propose to pre-process nodetype definition in the following way: check if the value of a config begins with `UI_eval:` (or any other keyword, can't think of better one right now), if yes then `eval()` it, e.g.

```js
(nodeProperty, value, node) => value.indexOf('UI_eval') === 0 ? eval(value) : value
```

where `node` would be set to the current focused node, with transient values from inspector overlayed on top.

This would allow the following usage:

```
properties:
  serviceType:
    type: string
    ui:
      label: 'Service Type'
      inspector:
        group: product
        editor: 'Content/Inspector/Editors/SelectBoxEditor'
        editorOptions:
          allowEmpty: true
          placeholder: 'Service Type'
          dataSourceIdentifier: 'acme-servicetypes'
  contractType:
    type: string
    ui:
      label: 'Contract Type'
      # only render when serviceType property is set, also cast it to bool to prevent proptype errors
      disabled: 'UI_eval:node.properties.serviceType && true'
      inspector:
        group: product
        editor: 'Content/Inspector/Editors/SelectBoxEditor'
        editorOptions:
          allowEmpty: true
          placeholder: 'Contract Type'
          dataSourceIdentifier: 'acme-contracttypes'
          dataSourceAdditionalData:
            # use serviceType property as the search term
            searchTerm: 'UI_eval:node.properties.serviceType'
```

Should be trivial to implement.

Thoughts? Ideas? Suggestions?

mficzel:
How about using an eel subset with only limited helpers that can be evaluated in js? I thought i have seen experiments with js-versions of flowQuery an eel somewhere. 

```
      disabled: '${q(node).property("serviceType") ? true : false}'
```
The main advantage would be that we can evaluate that on the server aswell and prevent setting disabled properties.

[...]

PR #1618, opened 2018-02-07
FEATURE: dynamic client-side configuration processing

dimaip:
Resolves: #1568

I propose to introduce this API on the same terms as the old API, declaring that it's experimental and may be replaced with a more high-level something. Or if we like it this way, we can just leave it as is and declare support for it later.

---------------------

With React, it's very simple to implement a declarative API, instead of a low-level imperative one like we did before.

I propose to pre-process nodetype definition in the following way: check if the value of a config begins with `ClientEval:` (or any other keyword, can't think of better one right now), if yes then `eval()` it where `node` would be set to the current focused node, with transient values from inspector overlayed on top.

This would allow the following usage:

```
properties:
  serviceType:
    type: string
    ui:
      label: 'Service Type'
      inspector:
        group: product
        editor: 'Content/Inspector/Editors/SelectBoxEditor'
        editorOptions:
          allowEmpty: true
          placeholder: 'Service Type'
          dataSourceIdentifier: 'acme-servicetypes'
  contractType:
    type: string
    ui:
      label: 'Contract Type'
      inspector:
        # only render when serviceType property is set, otherwise move it to an invalid group to hide it
        group: 'ClientEval:node.properties.serviceType ? "product" : "not-rendered"'
        editor: 'Content/Inspector/Editors/SelectBoxEditor'
        editorOptions:
          allowEmpty: true
          placeholder: 'Contract Type'
          dataSourceIdentifier: 'acme-contracttypes'
          dataSourceAdditionalData:
            # use serviceType property as the search term
            searchTerm: 'ClientEval:node.properties.serviceType'
```

A quick demo demonstrating throttling/performance:
peek 2018-02-08 12-04

This config is used in the recording:

```'Neos.NodeTypes.BaseMixins:ImageMixin':
  properties:
    title:
      ui:
        label: 'ClientEval:node.properties.alternativeText'
```


lorenzulrich:
I must admit I never used this feature in Neos before, but in TYPO3 I used the similar `displayCond` a lot.

I think your proposal will work fine. I think it would be good to inject something more than the `node`, so I could e.g. also check if someone is administrator or to check a property of the parent node - but that might be out of scope.

dimaip:
@lorenzulrich yeah I think such features could potentially be implemented server-side.

dimaip:
I did a lot of experiments with performance. I tried to use immutable data, but it only made things more slow.
The best I got so far is to throttle the TextInput editor. Now it really feels snappy (even better than before), and seems to work fine.

dimaip:
To improve performance I've throttled every editor commit in the EditorEnvelope. Now editing feels a lot snappier than it used to, even without the change.
Processing itself only add 0.3ms to rendering, so it's negligible.

kitsunet:
I like it. Simple but effective! <3 But I feel the prefix is really odd. How about `UiCondition:` or `EvalUi` I feel that's more in line with what we do usually.

dimaip:
@kitsunet thanks, just was too focused on the feature to think about the naming. I like `EvalUi`, as it's really all about evaling any config, and not only conditions.

dimaip:
@kitsunet maybe `ClientEval:`?

dimaip:
@mficzel I've updated an example here, to hide the property by moving it to an invalid group

mficzel:
@dimaip interesting approach for hiding. I like that i will have the option to move a property field away aswell as setting it to readonly. Both makes sense depending on the specific context. 

I like that.

dimaip:
@kdambekalns yes, I just gave a promise to myself that I will not merge it without documentation (and also fix other missing documentation pieces)

dimaip:
Switched to immutableJS. Now full inspector re-eval takes 30ms (instead of 200ms), and as it's throttled, it's not even noticeable.
Will add the docs, polish, fix tests and so on.

dimaip:
OK, I was rather naive.
I throttled the Editors, but forgot to flush the throttle function before things like clicking apply. So it would fire the commit callback already after the component has been unmounted. It's easy to cancel throttled function on unmount, but hard to wait for all children to have fired it before clicking things like apply.
So I gave up on the idea, as it introduces a lot of potential for bugs.
What I did instead is debounce the processing of configuration itself. It's not as fast as fully debounced editor, but at least it doesn't slow things down compared to how they are now.

Will write the docs.

Please review!!!

dimaip:
Docs are ready! /cc @kdambekalns

lorenzulrich:
I went through the documentation which is pretty clear and left some comments regarding spelling.

dimaip:
Anyways, merging, more likely to spot some bugs in master as nobody wants to review.

Benjamin-K:
Have you thought about scenarios where one would like to start a request to an external api or sth? Like you would like to get LatLng of a location you enter into a field for example. Is this possible with the new configuration? 
Anyways, nice work. If the above will also work, even better.

mficzel:
> Have you thought about scenarios where one would like to start a request to an external api or sth? Like you would like to get LatLng of a location you enter into a field for example. Is this possible with the new configuration?

@Benjamin-K i think that would be a use-case for a custom editor implementation.

dimaip:
Yes, exactly, was about to write that. In a custom inspector you may reuse any Editors or UI components, so shouldn't be a lot of extra effort.

Benjamin-K:
Ok, sounds good to me. But custom editors aren't that well documented yet (see http://neos.readthedocs.io/en/stable/ExtendingNeos/CustomEditors.html).

dimaip:
Good point! Could you create an issue? We'll take care.

Benjamin-K:
Issue is here: https://github.com/neos/neos-development-collection/issues/1900

skurfuerst:
Really fancy feature!

PR #1622, opened 2018-02-09
TASK: Make NodeTreeFilter more robust

### LINKED ISSUE #1505: NodeTypeFilter glitches
image

- No close button in nodetypes filter
- Select nodetype filter, then click on the header, then try to select again: doesn't work releably

johannessteu:
Fixes #1505

### DIFF
--- a/packages/build-essentials/src/styles/styleConstants.js
+++ b/packages/build-essentials/src/styles/styleConstants.js
@@ -29,7 +29,7 @@ const config = {
         primaryToolbar: ['context'],
         checkboxInput: ['context'],
         dropdownContents: ['context'],
-        selectBoxContents: ['context'],
+        selectBoxContents: '3',
         notInlineEditableOverlay: ['context'],
         calendarFakeInputMirror: ['context'],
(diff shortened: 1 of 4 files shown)


johannessteu:
The issue with the dropdown is that the drag&drop targetzones got the same or even a higher zinde-x compared to the selectbox dropdown. So sometimes you click that zone instead of the item in your list

http://g.recordit.co/xyiNFwTaFF.gif

packages/build-essentials/src/styles/styleConstants.js
```diff
@@ -29,7 +29,7 @@ const config = {
         primaryToolbar: ['context'],
         checkboxInput: ['context'],
         dropdownContents: ['context'],
-        selectBoxContents: ['context'],
+        selectBoxContents: '3',
```
dimaip: Yep, this is OK for now, the z-index concept is kind of broken now, needs re-thinking

mstruebing: we have an own issue for that afaik :P

Issue #1631, opened 2018-02-10
Update testcafe


markusguenther:
### Description

We are using testcafe for acceptance testing. Actually we have version 0.18.0 in use.
The worked hard and released already 0.18.6 with lots of fixes and features.

PR #1632, opened 2018-02-10
Task update testcafe

markusguenther:
This PR updates testcafe and testcafe-react-selector to the latest versions.
In the current version of the issue 1738 is fixed and so we don't need the prop-type check in the integrationTest.sh anymore.

Resolves: #1631


dimaip:
Thanks a lot for taking care @markusguenther, at least somebody shares my love for TestCafe now :heart:

markusguenther:
@dimaip great that you like it :)

PR #1634, opened 2018-02-12
TASK: Update eslint config

mstruebing:
This introduces many new warnings like 'no-unused-vars' for React (see
Travis output). I'm not sure if we should go this way because that
would mean all no-unused-vars errors will be overseen more likely in the
future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "devDependencies": {
     "@neos-project/brand": "^1.1.0",
     "@neos-project/build-essentials": "*",
-    "@neos-project/eslint-config-neos": "^2.0.0",
+    "@neos-project/eslint-config-neos": "^2.1.1",
     "babel-cli": "^6.8.0",
     "babel-core": "^6.13.2",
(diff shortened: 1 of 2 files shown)


mstruebing:
Found a solution, marking as WIP until finished.

mstruebing:
Now the `no-unused-vars` warnings are gone and they will throw an error when you have a `real` unused var.

However there are still many new warnings introduced, I would propose to fix them and then change the setting from warning to error.

markusguenther:
Regarding the unused variables. I think from time to time we should fix this warnings.
I can take care of it ... it is a nice job for getting an overview over all the stuff.

And the goal should be not to warn users, they should see an error.
But like I mentioned already on slack. In the past try to update this someone mentioned that I should not refactor the whole codebase just for the linter. So the first iteration is to update to the current versions and  use the new features. And then fix the warnings or in general that are errors. in the packages.

Thanks Max for the additional react rules. Did not know that and so I disabled it via warning ;)
Think if I fix them package by package it will work out or not?

PR #1636, opened 2018-02-13
TASK: Add node as dependency and remove `.nvmrc`

mstruebing:
You can test that it works with simply add a script like this:
`"which": "node -v"` and run `yarn run which`
Please remember to do a `yarn install` before.
See here for source:
https://twitter.com/housecor/status/962347301456015360


dimaip:
Nice!

Issue #1464, opened 2017-12-20
When switching workspaces page tree is not refreshed


dimaip:
Also if the given page doesn't exist in the other workspace, you'll end up on 404 page.

Issue #887, opened 2017-09-06
Performance optimisation of inline editing


dimaip:
**[<- must-have for 1.0.0](https://github.com/neos/neos-ui/issues/687)** 

Demo on the current UI, site with a lot of nodes in the tree, production context (notice little hiccups when typing):
![peek 2017-09-06 12-12](https://user-images.githubusercontent.com/837032/30104009-0701f9de-92fd-11e7-8697-1f113c334509.gif)

**Goal**: when editing text inline, there should be 0 hiccups and fps drops. E.g. even when just pressing and holding a button it should just flawlessly type in more characters.

One obvious way to solve that is to just debounce ckeditor input with a big timeout. This is dangerous, since if the editor quickly edits something and then closes the tab, the data would be lost. And even this way, when the user starts typing during saving, he would feel that little hick up.

Proper way to solve it would be to make those updates to feel unnoticeable, that is to fit them into frame interval or close to it (16ms).

When we edit properties of some node, these things are re-rendered:
- Publish button indicator and other small things. That's cheap
- Given content node and all of its parents both in content and document trees. This happens because Node render also depends on chilNodes. Before all nodes used to re-render, but I fixed that bug (it was causing up to 1.5s hickups on sites with many nodes!)
- Basically all of inspector is re-rendered, because we pass a node everywhere, to every single inspector editor. If we only passed a property, and not the whole node, that would make inspector editor to re-render only when that particular property changes. But that would limit our API and just wouldn't work in some cases.

Ways to optimize:
- Somehow be clever and reduce the number of components that need to be re-rendered. E.g. try to make a node in the tree not dependent on its children.
With inspector editors it's a lot harder to do the same thing.
- Optimize the rendering of components. If components need to be re-rendered, maybe it's possible to make that happen quicker. E.g. extract all expensive calculations (like calculation of inspector tabs/groups) into external memoized functions.
- Do something very clever, which I'm not aware of ;-)

Anyways, I picked a low-hanging fruit by fixing the tree re-render, waiting to hear your thoughts about what else we should do next.

dimaip:
Looks like we'll have to postpone it, removing from 1.0.

Issue #1020, opened 2017-09-25
Improve performance of initial load by parallelizing fetch calls and iframe load


grebaldi:
![screenshot_2017-09-25_17-01-21](https://user-images.githubusercontent.com/2522299/30815473-9f4ead2a-a213-11e7-95d7-192740173643.png)

As you can see, this is currently pretty bad. We could improve the initial load performance by parallelizing those first fetch calls. We should also find a way to render the iframe earlier while everything else is still loading.

dimaip:
Should have been fixed.

PR #1640, opened 2018-02-21
TASK: refactor nodes loading

dimaip:
- Eases the pain of slow nodes state rendering by fetching default nodes asynchronously (#1579)
- Unify logic for re-fetching nodes state
- Reload nodes when switching workspaces, redirect to correct node if it doesn't exist in target (Fixes: #1464)
- Reloads the tree in one gulp, without extra requests
- Fixes loading indicator
- Loads default nodes + toggled nodes + nodes up to the current document node in one request
- Parallelizes JSON fetch requests for translations and other stuff (Improves #1020)
- Don't render hidden elements (improves render time from 1100ms down to 120-360ms!) (Fixes: #887 )

### DIFF
--- a/Classes/Neos/Neos/Ui/Controller/BackendServiceController.php
+++ b/Classes/Neos/Neos/Ui/Controller/BackendServiceController.php
@@ -102,7 +102,7 @@ class BackendServiceController extends ActionController
      * @param ResponseInterface $response
      * @return void
      */
-    public function initializeController(RequestInterface $request, ResponseInterface $response)
+    public function initializeController(RequestInterface $request, ResponseInterface $response): void
     {
         parent::initializeController($request, $response);
         $this->feedbackCollection->setControllerContext($this->getControllerContext());
@@ -114,7 +114,7 @@ class BackendServiceController extends ActionController
      * @param string $documentNodeContextPath
      * @return void
      */
-    protected function updateWorkspaceInfo($documentNodeContextPath)
+    protected function updateWorkspaceInfo(string $documentNodeContextPath): void
     {
         $updateWorkspaceInfo = new UpdateWorkspaceInfo();
         $documentNode = $this->nodeService->getNodeFromContextPath($documentNodeContextPath, null, null, true);
@@ -131,7 +131,7 @@ class BackendServiceController extends ActionController
      * @param ChangeCollection $changes
      * @return void
      */
-    public function changeAction(ChangeCollection $changes)
+    public function changeAction(ChangeCollection $changes): void
     {
         try {
             $count = $changes->count();
@@ -159,7 +159,7 @@ class BackendServiceController extends ActionController
      * @param string $targetWorkspaceName
      * @return void
      */
-    public function publishAction(array $nodeContextPaths, $targetWorkspaceName)
+    public function publishAction(array $nodeContextPaths, string $targetWorkspaceName): void
     {
         try {
             $targetWorkspace = $this->workspaceRepository->findOneByName($targetWorkspaceName);
@@ -192,7 +192,7 @@ class BackendServiceController extends ActionController
      * @param array $nodeContextPaths
      * @return void
      */
-    public function discardAction(array $nodeContextPaths)
+    public function discardAction(array $nodeContextPaths): void
     {
         try {
             foreach ($nodeContextPaths as $contextPath) {
@@ -248,8 +248,9 @@ class BackendServiceController extends ActionController
      * @param string $targetWorkspaceName,
      * @param NodeInterface $documentNode
      * @return void
+     * @throws \Exception
      */
-    public function changeBaseWorkspaceAction($targetWorkspaceName, $documentNode)
+    public function changeBaseWorkspaceAction(string $targetWorkspaceName, NodeInterface $documentNode): void
     {
         try {
             $targetWorkspace = $this->workspaceRepository->findOneByName($targetWorkspaceName);
@@ -286,7 +287,7 @@ class BackendServiceController extends ActionController
                 } else {
                     $redirectNode = $redirectNode->getParent();
                     if (!$redirectNode) {
-                        throw new \Exception('Wasn\'t able to locate any valid node in rootline in the target workspace.');
+                        throw new \Exception(sprintf('Wasn\'t able to locate any valid node in rootline of node %s in the workspace %s.', $documentNode->getContextPath(), $targetWorkspaceName), 1458814469);
                     }
                 }
             }
@@ -331,7 +332,7 @@ class BackendServiceController extends ActionController
      * @param boolean $includeRoot
      * @return void
      */
-    public function loadTreeAction(NodeTreeBuilder $nodeTreeArguments, $includeRoot = false)
+    public function loadTreeAction(NodeTreeBuilder $nodeTreeArguments, $includeRoot = false): void
     {
         $nodeTreeArguments->setControllerContext($this->controllerContext);
         $this->view->assign('value', $nodeTreeArguments->build($includeRoot));
@@ -343,7 +344,7 @@ class BackendServiceController extends ActionController
      * @param array $chain
      * @return string
      */
-    public function flowQueryAction(array $chain)
+    public function flowQueryAction(array $chain): string
     {
         $createContext = array_shift($chain);
         $finisher = array_pop($chain);
--- a/Classes/Neos/Neos/Ui/Controller/BackendServiceController.php
+++ b/Classes/Neos/Neos/Ui/Controller/BackendServiceController.php
@@ -14,14 +14,17 @@ use Neos\Flow\Mvc\Controller\ActionController;
 use Neos\Flow\Mvc\RequestInterface;
 use Neos\Flow\Mvc\ResponseInterface;
 use Neos\Flow\Persistence\PersistenceManagerInterface;
+use Neos\Neos\Domain\Service\ContentContextFactory;
 use Neos\Neos\Service\PublishingService;
 use Neos\Neos\Service\UserService;
+use Neos\ContentRepository\Domain\Model\NodeInterface;
 use Neos\ContentRepository\Domain\Repository\WorkspaceRepository;
 use Neos\Neos\Ui\Domain\Model\ChangeCollection;
 use Neos\Neos\Ui\Domain\Model\FeedbackCollection;
 use Neos\Neos\Ui\Domain\Model\Feedback\Messages\Error;
 use Neos\Neos\Ui\Domain\Model\Feedback\Messages\Info;
 use Neos\Neos\Ui\Domain\Model\Feedback\Messages\Success;
+use Neos\Neos\Ui\Domain\Model\Feedback\Operations\Redirect;
 use Neos\Neos\Ui\Domain\Model\Feedback\Operations\ReloadDocument;
 use Neos\Neos\Ui\Domain\Model\Feedback\Operations\RemoveNode;
 use Neos\Neos\Ui\Domain\Model\Feedback\Operations\UpdateWorkspaceInfo;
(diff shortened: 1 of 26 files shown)


Classes/Neos/Neos/Ui/Controller/BackendServiceController.php
```diff
@@ -237,10 +245,11 @@ public function discardAction(array $nodeContextPaths)
     /**
      * Change base workspace of current user workspace
      *
-     * @param string $targetWorkspaceName
+     * @param string $targetWorkspaceName,
+     * @param NodeInterface $documentNode
      * @return void
      */
-    public function changeBaseWorkspaceAction($targetWorkspaceName)
+    public function changeBaseWorkspaceAction($targetWorkspaceName, $documentNode)
```
daniellienert: Parameters should be annotated `string $targetWorkspaceName, NodeInterface $documentNode`

Classes/Neos/Neos/Ui/Controller/BackendServiceController.php
```diff
@@ -262,8 +271,35 @@ public function changeBaseWorkspaceAction($targetWorkspaceName)
...
+            // If current document node doesn't exist in the base workspace, traverse its parents to find the one that exists
+            $redirectNode = $documentNode;
+            while (true) {
+                $redirectNodeInBaseWorkspace = $contentContext->getNodeByIdentifier($redirectNode->getIdentifier());
+                if ($redirectNodeInBaseWorkspace) {
+                    break;
+                } else {
+                    $redirectNode = $redirectNode->getParent();
+                    if (!$redirectNode) {
+                        throw new \Exception('Wasn\'t able to locate any valid node in rootline in the target workspace.');
```
daniellienert: Exception should have an exception number

daniellienert: Add the documentnodes identifier to the exception to support debugging

Classes/Neos/Neos/Ui/Domain/Model/Feedback/Operations/Redirect.php
```diff
@@ -0,0 +1,101 @@
+<?php
+namespace Neos\Neos\Ui\Domain\Model\Feedback\Operations;
+
```
daniellienert: Licence header missing

Classes/Neos/Neos/Ui/Domain/Model/Feedback/Operations/Redirect.php
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * Get the description
+     *
+     * @return string
+     */
+    public function getDescription()
+    {
+        return sprintf('Redirect to node "%s".', $this->getNode()->getContextPath());
```
daniellienert: input type and return type declaration are missing at many placed, would be nice to have them

Classes/Neos/Neos/Ui/Domain/Model/Feedback/Operations/Redirect.php
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        return $this->getNode()->getContextPath() === $feedback->getNode()->getContextPath();
+    }
+
+    /**
+     * Serialize the payload for this feedback
+     *
+     * @param ControllerContext $controllerContext
+     * @return mixed
```
daniellienert: Array is returned here

Classes/Neos/Neos/Ui/FlowQueryOperations/NeosUiDefaultNodesOperation.php
```diff
@@ -55,28 +55,44 @@ public function canEvaluate($context)
      */
     public function evaluate(FlowQuery $flowQuery, array $arguments)
     {
-        list($site, $documentNode) = $flowQuery->getContext();
-        list($baseNodeType, $loadingDepth) = array_replace([
+        list($siteNode, $documentNode) = $flowQuery->getContext();
+        // set default values for arguments
+        list($baseNodeType, $loadingDepth, $toggledNodes) = array_replace([
```
daniellienert: That looks risky (or I don't get it). If $arguments contains 3 values the array replace could be skipped. If it contains less, How do you know here which arguments are set?

Classes/Neos/Neos/Ui/Domain/Model/Feedback/Operations/Redirect.php
dimaip: It's missing for all files in the package...

Classes/Neos/Neos/Ui/FlowQueryOperations/NeosUiDefaultNodesOperation.php
dimaip: Agreed, it was not my code, took me some time to figure out what it's doing actually. I think defaults are not needed here, will remove it.

dimaip:
@daniellienert thanks, fixed your comments.
Did you actually test the change itself? Would be great to get some extended testing on real projects, as I changed a lot of stuff.

Classes/Neos/Neos/Ui/Controller/BackendServiceController.php
```diff
@@ -262,8 +272,35 @@ public function changeBaseWorkspaceAction($targetWorkspaceName)
...
-            $this->feedbackCollection->add($reloadDocument);
+            // Construct base workspace context
+            $originalContext = $documentNode->getContext();
+            $contextProperties = $documentNode->getContext()->getProperties();
+            $contextProperties['workspaceName'] = $targetWorkspaceName;
+            $contentContext = $this->contextFactory->create($contextProperties);
+
+            // If current document node doesn't exist in the base workspace, traverse its parents to find the one that exists
+            $redirectNode = $documentNode;
+            while (true) {
```
skurfuerst: uhm, I have to check this while(true) logic in detail; it's always hard to see that it does not end up with an endless loop ;)

Classes/Neos/Neos/Ui/FlowQueryOperations/NeosUiDefaultNodesOperation.php
```diff
@@ -17,7 +17,7 @@
 use Neos\ContentRepository\Domain\Model\NodeInterface;
 
 /**
- * Renders nodes that are initially required by the user interface
+ * Fetches all nodes needed for the given state of the UI
```
skurfuerst: isn't this function actually *slower* than the one before? :)

packages/neos-ui/src/Containers/Drawer/index.js
```diff
@@ -113,7 +113,7 @@ export default class Drawer extends PureComponent {
                 onMouseLeave={this.handleMouseLeave}
                 aria-hidden={isHidden ? 'true' : 'false'}
                 >
-                {Object.values(menuData).map((item, index) => (
+                {!isHidden && Object.values(menuData).map((item, index) => (
```
skurfuerst: really nice :)

packages/neos-ui/src/Sagas/UI/PageTree/index.js
```diff
@@ -78,26 +78,6 @@ export function * watchNodeCreated() {
...
-        const nodesToReload = documentNodes.toArray().filter(node =>
-            uncollapsedContextPaths.includes(node.get('contextPath')) &&
-            node.get('children').filter(child => nodeTypesRegistry.hasRole(child.get('nodeType'), 'document')).size
-        );
-
-        for (let i = 0; i < nodesToReload.length; i++) {
-            const node = nodesToReload[i];
-            const contextPath = node.get('contextPath');
-
-            yield put(actions.UI.PageTree.requestChildren(contextPath, {unCollapse: false}));
```
skurfuerst: OK. in the old code, we had many AJAX requests, now we only have one.

packages/neos-ui/src/index.js
```diff
@@ -176,6 +182,14 @@ function * application() {
...
         appContainer
     );
+
+    const siteNodeContextPath = yield select($get('cr.nodes.siteNode'));
+    const documentNodeContextPath = yield select($get('ui.contentCanvas.contextPath'));
+    yield put(actions.CR.Nodes.reloadState({
+        siteNodeContextPath,
+        documentNodeContextPath,
+        merge: true
+    }));
```
skurfuerst: I am wondering if this change actually improves performance; or if it improves *perceived* performance. As long as the nodes are not loaded, can we even interact with the UI at all?

skurfuerst:
Following up on your measurements in https://github.com/neos/neos-ui/pull/1581#issuecomment-361018832:

- I think we should remove the PolicyAspect and *directly* inline the code in the NodeInfoHelper.
- is AOP slow or the policy check itself?

all the best,
Sebastian

skurfuerst:
btw - ❤️ for working on this - it is great to have that tackled!

Classes/Neos/Neos/Ui/Controller/BackendServiceController.php
dimaip: Sure, feel free to refactor if you know how to express it better. But I think it's safe.

Classes/Neos/Neos/Ui/FlowQueryOperations/NeosUiDefaultNodesOperation.php
dimaip: For default state, with no uncollapsed nodes, it's equivalent. It includes nodes on the rootline + all uncollapsed nodes so extra requests would not be needed when opening on a deep nested node, deeper then loadingDepth, or when reloading the tree.

packages/neos-ui/src/Sagas/UI/PageTree/index.js
dimaip: exactly, works a lot faster

packages/neos-ui/src/index.js
dimaip: It improves perceived performance. You can start interacting with everything, e.g. start inline editing. The nodes are seamlessly added when loaded. It takes 100ms to add 50 nodes to the tree, so it doesn't even lag.

dimaip:
@skurfuerst hey thanks a lot for the review!

> I think we should remove the PolicyAspect and directly inline the code in the NodeInfoHelper.

That will not help. AOP cost is negligible compared to how much it costs to evaluate EEL for each node. I did xdebug profiling.

I think to solve the problem we should work in multiple directions:

a) make tree loading async, as done in this PR: partially hides the problem, but it's not enough if the editors want to interact with the tree right away
b) separate loading node policies from loading node data, and make it also async. So in first pass the whole UI would be rendered, then with a tiny delay the tree nodes would appear and some time after the policies would kick in.
c) improve performance of actual policy checking and eel parsing. But it's a whole other topic.

dimaip:
@skurfuerst so here's what I suggest to do next after we merge this PR: https://github.com/neos/neos-ui/issues/1646

packages/neos-ui/src/index.js
kdambekalns: It's "actual perceived performance" ;) If you read about the timings in https://github.com/neos/neos-ui/issues/1579 you get an idea. So, the perceived performance increases (you don't just see the spinner for ages), and the actual performance as well, because you can actually start doing $something.

dimaip: @kdambekalns see my RFC how to increase real performance: https://github.com/neos/neos-ui/issues/1646

skurfuerst:
Awesome; I have a big instance now where I will test today :)

skurfuerst:
## Performance measurements (WIP)

I used [speedline](https://github.com/paulirish/speedline) to measure performance. The red line is when editing is possible.


## baseline - old UI - no optimizations

downloads zsh 2018-02-23 10-00-04

## new UI -- without this patch

downloads zsh 2018-02-23 10-01-28

## new UI - with patch (TODO)

skurfuerst:
@dimaip I found the following issue:

- as soon as the page finishes loading, the spinner in the node tree stops; although the node tree is not yet fully loaded. I'll push a followup for this.

kdambekalns:
> Let's merge as soon as Travis OKs it!

Well… 😏

dimaip:
@kdambekalns ?

PR #1643, opened 2018-02-21
TASK: Add consistent license header to all PHP files

daniellienert:
There have been some files with license headers. Unified them and
added headers to all files where they where missing


dimaip:
Thanks!

Issue #1523, opened 2018-01-09
Unpublished nodes can't be selected when moved


bwaidelich:
### Description

When a new (unpublished) document node is moved within the new UI it sometimes can't be selected/edited any longer

### Steps to Reproduce

1. Create a new Document node (e.g. in a fresh Demo Site)
![image](https://user-images.githubusercontent.com/307571/34719452-31bf46a4-f53b-11e7-9c56-d507afcd7b1d.png)
2. Drag the new (unpublished) document node into an existing (published) node
![image](https://user-images.githubusercontent.com/307571/34719467-41b32076-f53b-11e7-9cf4-276daca256f9.png)
3. Try to select the moved node in the tree

#### Expected behavior

The iFrame should load the representation of the moved node and it should be editable in the Backend

#### Actual behavior

(Sometimes) the new node can't be selected. Instead the homepage is selected (and loaded in the iFrame)

### Background

I pinned down the problem to be related to the `/neos/ui-services/change` AJAX request that returns an invalid `uri` for the moved nodes (sometimes):

It *should* be: `"uri": "http:\/\/domain.tld\/en\/download\/new@user-admin;language=en_US.html",`
but it *is*: `"uri": "http:\/\/domain.tld\/en@user-admin;language=en_US.html",`

I suspect some kind of race condition in the Routing because it is not always reproducible and never reproducible with disabled routing caches..

### Affected Versions

Neos: 

3.3+

bwaidelich:
Maybe this is also related to https://github.com/neos/neos-development-collection/pull/1721

bwaidelich:
After another debugging session with @kitsunet I'm still clueless about the source of this problem..
What I can say is that it is *not* related to the routing (cache) after all because it *is* reproducible even with disabled caches.

One part that made us very suspicious is the `clone` of the subject (= node to move) in `\Neos\Neos\Ui\Domain\Model\Changes\MoveInto`. Cloning nodes can have a lot of side effects due to runtime caches etc.. But even with removed `clone` the problem persists so this is at least not the (only) reason for the issue.

skurfuerst:
@bwaidelich maybe the change on https://github.com/neos/neos-ui/pull/1645 helps to fix the issue? Removing the clone is definitely wrong; we need information about the *old* location in the tree.

skurfuerst:
I might have been hit by a similarish issue; maybe it helps to track down the issue:

- create a new page A; do not publish it.
[...]

PR #1645, opened 2018-02-22
BUGFIX: make moving of nodes more robust

skurfuerst:
## Problem Description

when moving nodes, it sometimes happened that the node
on the original location was not removed.

## Root Cause Analysis

It turned out that the "Remove" Feedback (which is part
of the feedbacks returned on move) got transmitted
with the wrong contextNodePath (e.g. where the node was
*already moved*); thus it could not be removed in the UI.

For me in some cases it correctly worked; though I was
not able to track the issue down to specific cases. In
90% of the cases it did NOT work. I tested with a custom
4-column element; where moving in the columns 2-4 reliably
triggered the error. Interestingly, moving from Column1
mostly worked...

## Solution Fix

the "clone" only clones the node; without cloning the
corresponding NodeData object. The fix is to additionally
clone the NodeData object itself; thus we detach it
from persistence, thus it keeps the old NodePath (which is what we need).

Interesting this has not happened much more for people :-)

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/Classes/Neos/Neos/Ui/Domain/Model/Changes/AbstractMove.php
+++ b/Classes/Neos/Neos/Ui/Domain/Model/Changes/AbstractMove.php
@@ -11,6 +11,7 @@ namespace Neos\Neos\Ui\Domain\Model\Changes;
  * source code.
  */
 
+use Neos\ContentRepository\Domain\Model\Node;
 use Neos\ContentRepository\Domain\Model\NodeInterface;
 use Neos\Neos\Ui\Domain\Model\Feedback\Operations\RemoveNode;
 
@@ -42,5 +43,18 @@ abstract class AbstractMove extends AbstractStructuralChange
         $this->feedbackCollection->add($removeNode);
 
         parent::finish($this->getSubject());
+    }
+
+    protected static function cloneNodeWithNodeData(NodeInterface $node)
+    {
+        if ($node instanceof Node) {
+            $originalNode = $node;
+            $node = clone $originalNode;
+            $node->setNodeData(clone $originalNode->getNodeData());
+            return $node;
+        } else {
+            // do a best-effort clone
+            return clone $node;
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


skurfuerst:
might be related to #1523

Classes/Neos/Neos/Ui/Domain/Model/Changes/AbstractMove.php
```diff
@@ -43,4 +44,17 @@ protected function finish(NodeInterface $node)
 
         parent::finish($this->getSubject());
     }
+
+    protected static function cloneNodeWithNodeData(NodeInterface $node)
+    {
+        if ($node instanceof Node) {
+            $originalNode = $node;
+            $node = clone $originalNode;
```
dimaip: Just wondering, shouldn' Node be aware how to clone itself properly? Maybe we should add it to `__clone` method of the Node class?

skurfuerst: @dimaip I am not sure; we could do it, but I would like to keep the side effects to a minimum.

All the best, Sebastian

dimaip: I'm not sure I understand what you mean here by side effects. I mean to do it in the Neos.ContentRepository package itself. I think it's only natural that when you clone an object it should be aware how to clone itself, without having to manually do it from the outside.
Anyways, do as you see fit here, I'm not experienced with PHP so much.

skurfuerst:
Well this part of the code base (Node in CR) is extremely hard / fragile / ...
I fear that by introducing clone behavior, some other part of the code base breaks which might rely on the fact that it does not do a deep clone.

That is why I prefer to not touch the central Node object...

skurfuerst:
Thanks for your reviews!

Issue #1647, opened 2018-02-23
Re-add node type attribute to content element wrapping


bwaidelich:
Previously the `NodeType` was part of the `ContentElementWrapping`:
```html
<div data-node-_node-type="Some:Node.Type">...</div>
```
We use that for example in the Form Builder to improve Backend UX: https://github.com/neos/form-builder/#adjust-appearance-of-the-form-builder

With the new UI that attribute no longer exists so CEs can't be targetted via CSS/JS based on their type.

I suggest to re-add the attribute, but call it `data-__neos-node-type` to be in sync with the other attributes

lorenzulrich:
+1

PR #1648, opened 2018-02-23
BUGFIX: Adjust ContentElementWrapping aspect to recent changes

bwaidelich:
This adjusts the `AugmentationAspect` so that it respects the
`additionalAttributes` argument introduced with https://github.com/neos/neos-development-collection/pull/1919

Resolves: #1647


skurfuerst:
I have a slight preference currently to *not* making this a public API; in fact IMHO the CEWrappingService should *only* add things needed for technical reasons to the Neos UI.

However, I also do not have a clear idea how to realize what you want. For all Content Elements inheriting from Neos.Fusion:Tag, it is obviously easy; but for everything inheriting Template and Content, it is obviously more difficult.

What about making the Content Element Wrapping Service (and the fusion object) more configurable to make it possible to add additional attributes via Fusion? Then everybody relying on these kinds of styles can add them explicitely in the Fusion objects; but we do not make this a public "magic" API.

What do you think @bwaidelich and @lorenzulrich ? :-)

All the best,
Sebastian

bwaidelich:
I like the idea to make the ContentElementWrapping extensible! I'll try to implement this as a follow-up commit

bwaidelich:
This requires https://github.com/neos/neos-development-collection/pull/1919 to be merged!

bwaidelich:
Since neos/neos-development-collection#1919 is merged in the meantime, can we merge this one, @skurfuerst ?  (I merged current master and successfully tested it)

bwaidelich:
I consider this a no-brainer and the failing tests are most probably not related. So I decided to merge this myself for lack of reviews. I take full responsibility ;)

mstruebing:
It seems the tests are related to your PR - the e2e tests are currently broken in master

bwaidelich:
@mstruebing But how can that be related?

bwaidelich:
It was indeed related 🙈 #1773 fixes the CI

PR #1653, opened 2018-02-26
BUGFIX: Fix PHP 7.0 compatibility

hlubek:
Removes incompatible return type annotations (`void`) and adds PHP 7.0
to Travis CI to catch bugs for PHP 7.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ language: php
 
 php:
   - '7.0'
-  - '7.1'
 
 services:
   - mysql
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ notifications:
 language: php
 
 php:
+  - '7.0'
   - '7.1'
 
 services:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -15,6 +15,7 @@ notifications:
 language: php
 
 php:
+  - '7.0'
```
dimaip: Do we really need to run extra set of tests for PHP 7.0? I'm afraid it may slow things down... Since our PHP side is not so fat yet, I'd set php to the lowest supported version here, which is 7.0.

mstruebing: My intuition says we should check both. I've already received different kinds of errors only on specific PHP versions (7.0, 7.1, 7.2). But that will slow down Travis ofc as you already said.

There will always be a tradeoff.

dimaip: My dream is to have nightly build run that would check everything including performance metrics, and a lightweight build that we currently have for PRs...

PR #1681, opened 2018-03-02
BUGFIX: Add Lerna alias to Makefile

mstruebing:
to prevent MacOS errors while testing

@markusguenther Could you test this? :)


markusguenther:
Looks good to me @mstruebing ... thanks for fixing that

PR #1693, opened 2018-03-06
BUGFIX: make moving of nodes more robust

skurfuerst:
Moving of nodes did not work when the *rendered parent node* in DOM was not
the parent node in the node tree.

In this case, the following happened:

- the rendered parent node was transmitted as "parentDomAddress.contextPath"
  and made available in AbstractChange as getParentNode()
- in AbstractMove::canApply(), the "getParentNode()" method was called, and
  taken into account for constraint checks -- so if the rendered parent node
  was a different node than the actual parent (e.g. the grandparent), the
  constraints might not have matched and the move could not succeed.

To fix this, I changed the following:

- removed AbstractStructuralChange::getParentNode as this is not really the parent node, but the closest rendered node.
- pushed canApply() down; not relying on getParentNode() anymore
- for the Copy/Move INTO cases, add a specific property "parentContextPath" to the
  change; which contains the parent node without further processing.
- we ensure the renderedParentNode === parentNode (==is the ContentCollection),
  otherwise out of band rendering can not work (and we need to reload the full page
  instead).

### DIFF
--- a/Classes/Neos/Neos/Ui/Domain/Model/Changes/CopyInto.php
+++ b/Classes/Neos/Neos/Ui/Domain/Model/Changes/CopyInto.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace Neos\Neos\Ui\Domain\Model\Changes;
 
 /*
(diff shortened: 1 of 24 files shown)


Classes/Neos/Neos/Ui/Domain/Model/Changes/CopyInto.php
```diff
@@ -11,8 +12,57 @@
...
+use Neos\ContentRepository\Domain\Model\NodeInterface;
+
 class CopyInto extends AbstractCopy
 {
+
+
+    /**
+     * @var string
+     */
+    protected $parentContextPath;
```
dimaip: So this is in fact a target parent node, right? Maybe call it smh like `referenceNodeContextPath` or `targetNodeContextPath`?

Classes/Neos/Neos/Ui/Domain/Model/Changes/MoveAfter.php
```diff
@@ -15,6 +15,19 @@
...
+    /**
+     * "Subject" is the to-be-moved node; the "sibling" node is the node after which the "Subject" should be copied.
+     *
+     * @return boolean
+     */
+    public function canApply()
+    {
+        $nodeType = $this->getSubject()->getNodeType();
+
+        return $this->getSiblingNode()->getParent()->isNodeTypeAllowedAsChildNode($nodeType);
```
dimaip: Same here, we could just call `siblingNode` as a `referenceNode`, then it would have consistent naming between move before/after and into. I think that's the terminology we use in the UI itself.

packages/neos-ui/src/Sagas/CR/NodeOperations/helpers.js
```diff
@@ -49,6 +49,7 @@ export const calculateDomAddressesFromMode = (mode, contextPath, fusionPath) =>
             const element = findNodeInGuestFrame(contextPath, fusionPath);
 
             return {
+                parentContextPath: contextPath,
```
dimaip: Naming here is not cool. The method name says `calculateDomAddressesFromMode` and the actual returned keys mention "parent". Change either one or another.

dimaip:
Nicely done!
Left some naming nitpicks.

skurfuerst:
@dimaip good comments... currently I sadly do not have time to rework all this logic, and especially test it. (as it touches creation, moving, copy of nodes in all 3 cases each). Would you be OK with merging it nevertheless?

All the best, Sebastian

dimaip:
@skurfuerst sure. Just wanted to make the naming a bit clearer, but I understand it might not be the best time for you.

Issue #1479, opened 2017-12-21
Loading indicator must be visible when switch Preview/Editing mode


dfeyer:
Currently the loading indicator is not visible when you switch the mode

PR #1695, opened 2018-03-07
BUGFIX: fix loading indicator when navigating inline

dimaip:
In addition fixes: #1479


dimaip:
Feel free to post-review, I think there are more chances somebody spots any defects on master currently...

Issue #1028, opened 2017-09-26
Page Tree doesn't allow to select node, without loading it


grebaldi:
In the old UI, it is possible to select a node by clicking aside its label (a click directly on the label would trigger a page load). This behavior is important to perform operations like copy/cut and delete, without having to wait for a page load.

dimaip:
Wow, never knew of this behavior in the old ui :see_no_evil:

kitsunet:
IMHO not a planned feature but rather a "useful" bug?

I guess it would be cool to actually have a way to do that. But not something we urgently need to do?

PR #1696, opened 2018-03-07
TASK: instantly focus node when navigating from PageTree

dimaip:
We instantly switch context path to the new node and show the spinner in inspector, while full properties information is loaded.
It makes the UI feel more snappy and actually allows to create nodes or perform operations from node toolbar instantly, without waiting for content canvas to load.

Also I think it Fixes: #1028

peek 2018-03-07 10-04


dimaip:
Feel free to post-review, I think there are more chances somebody spots any defects on master currently...

Issue #1481, opened 2017-12-21
Inline NodeToolbar on inline Editable fields on document nodes


samuelhauser:
If you have an inline editable field on a document node, the NodeToolbar shows up (it shouldn't be)

skurfuerst:
Hey,

Thanks for your report!
@samuelhauser could you please provide more detail on how to reproduce the issue? maybe with screenshots etc?

Thanks ❤️ 
Sebastian

samuelhauser:
Hey @skurfuerst 

If you have a property on a documentNode as inline editable field, the toolbar appears, but it shouldn't

![documentnode](https://user-images.githubusercontent.com/1894064/34296243-fc4040fa-e711-11e7-8622-79dd5e81c4e7.png)

DrillSergeant:
I can confirm this behaviour. This blocks me from activating the new UI in production projects as we have a lot of inline editable properties.

dimaip:
Can confirm, will take care.

PR #1697, opened 2018-03-07
BUGFIX: don't show NodeToolbar for document nodes

dimaip:
Fixes: #1481


dimaip:
@skurfuerst do you think this is the best way to fix it?

dimaip:
Thanks!
@skurfuerst feel free to post-review then.

PR #1703, opened 2018-03-09
BUGFIX: set correct translation value to delete confirmation

suffle:
Fixes the translation in the Delete Confirmation Dialog by using the right identifier


mstruebing:
Thanks for this PR!

PR #1706, opened 2018-03-16
TASK: add ids

dimaip:
Add ids to most elements. Would help to simplify E2E tests. Also needed for some 3rd-party integrations.


dimaip:
Merging without reviews as no-brainer.

PR #1709, opened 2018-03-16
TASK: Refactor the PHP classes to psr-4 structure

daniellienert:


markusguenther:
Trigger travis.

mstruebing:
What to do about this?

dfeyer:
@mstruebing @daniellienert I think it's important to merge it at some point, no clue if it can break a lots of PR or not.

So this one need a rebase or a new `php-cs-fixer` run on the current master

dfeyer:
Did we need (i think we need) to retarget this to branch 1.0 ;)

And now clue about the failing test ... some package manager fight for more power

daniellienert:
Thanks for taking care guys!
I don't understand this error either - wanted to get the right guys together during the sprint - but forgot about it. 
I found nothing that relies on the missing Neos/Neos/Ui subfolders.

kitsunet:
The current test failure is more reasonable. I fear though it is due to how the tests are done. Flow and Neos by now use local path repositories to install the PR folder via composer. I assume autoloading goes horribly wrong in this travis build, as the master of the UI package is installed which has PSR-0 and then the new one is copied over it, so compser and neos at this point expect classes in PSR-0 folders but actually they are in PSR-4 folders. \o/
$Someone would have to adjust the travis scripts to do the same as is done in neos and flow collections, until then this will not be green. I would still like to merge it ASAP after a bit of local testing...

kitsunet:
Ah, found a stray class, probably from upmerging. Let's see if we can convince jenkins, but trying it locally seems fine now, so I might just merge it...

dfeyer:
Why did we have so much commit in this change ?

kitsunet:
Replaced by #1789

kitsunet:
No, can probably now go away!

Issue #1711, opened 2018-03-21
Missing CKEditor toolbar for inline editable `document node properties`


jobee:
### Description

Inline editable document properties don't show the CKEditor toolbar (upper left corner) when focused.

### Steps to Reproduce

1. Add an inline editable property to a document nodetype
2. Click on it / focus

#### Expected behavior

CKEditor toolbar should show up

#### Actual behavior

Content dimension navigation is shown, which means the document is still focused

### Affected Versions

Neos: 3.3

UI: > 1.0.7

dimaip:
This is probably the cause: https://github.com/neos/neos-ui/pull/1649/files

PR #1712, opened 2018-03-22
BUGFIX: fix CKE toolbar for properties of documentNode

dimaip:
Fixes: #1711


dimaip:
Merging without review as this one is critical and urgent. @skurfuerst please post-review when you got the time.

jobee:
@dimaip thank you for solving and merging that issue so fast!
If I click the inline property twice (second click after its already focused) the dimension menu still appears, I guess thats not an expected behaviour, isn't it?

dimaip:
@jobee thanks for testing, will fix!

PR #1714, opened 2018-03-23
BUGFIX: only re-focust when selecting a different node

dimaip:
Fixes https://github.com/neos/neos-ui/pull/1712#issuecomment-375562925


dimaip:
@jobee please test

PR #1722, opened 2018-03-26
TASK: Export more selectbox parts from ui-components package

kitsunet:
This allows using the respective components in extensibility
packages.

### DIFF
--- a/packages/react-ui-components/src/index.js
+++ b/packages/react-ui-components/src/index.js
@@ -20,31 +20,10 @@ import TextInput from './TextInput/index';
 import ToggablePanel from './ToggablePanel/index';
 import Tooltip from './Tooltip/index';
 import Tree from './Tree/index';
-
-export {
-    Badge,
-    Bar,
-    Button,
-    ButtonGroup,
-    CheckBox,
-    DateInput,
-    Dialog,
-    DropDown,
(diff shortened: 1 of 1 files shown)


packages/react-ui-components/src/index.js
```diff
@@ -68,5 +47,9 @@ export default {
     TextInput,
     ToggablePanel,
     Tooltip,
-    Tree
+    Tree,
+    MultiSelectBox,
+    MultiSelectBox_ListPreviewSortable,
```
dimaip: I understand about `MultiSelectBox`, but why would you want to expose smaller internals? These things are sure to be trashed during the next FrankenBox refactoring...

kitsunet:
Interestingly this not only fails to build but also breaks something in the UI itself, but I am not sure how or what. HALP

kitsunet:
Ok, apparently we need the export AND default export in there. Will push update to the PR, but WHY, cna someone explain that to me, because it seems strange to me.

mstruebing:
I cant explain this to you I can only say: export is named and export default is unnamed.

Anyway, should we merge it?

dimaip:
I don't really mind, albeit looks a bit fragile that we export so many implementation details.

Issue #1719, opened 2018-03-26
Inspector groups keep closing themselfes


MCStreetguy:
### Description

Whenever you edit some properties in the inspector, clicking on an editor instance (such as the LinkEditor where this is required) makes the corresponding group collapse.
In addition, clicking the "Apply" Button makes all groups collapse at once.

(The "document" group is somehow not affected by this behaviour)

### Steps to Reproduce

1. Edit some properties on an element
2. Click either within an editor or on the "Apply" button

#### Expected behavior

All groups should stay in the same state as before.

#### Actual behavior

As you can see in my gif below:

![neos-ui-inspector-groups-error](https://user-images.githubusercontent.com/8970747/37893618-e7b2b168-30db-11e8-88c4-87a5765798e5.gif)

### Affected Versions

Neos: 3.3.9

UI: 1.0.9

PR #1723, opened 2018-03-26
BUGFIX: keep toggled state of inspector panels in state

dimaip:
Fixes: #1719


dimaip:
@MCStreetguy could you please test if it fixes your problem.

MCStreetguy:
@dimaip I'll check it in a moment when I'm home and then leave feedback here.

MCStreetguy:
@dimaip Took me far longer than expected to test, but I could just confirm that this solves my problem. Thanks a lot for the quick fix.

PR #1727, opened 2018-03-30
BUGFIX: don't allow to publish/discard while publishing/discarding/saving

dimaip:
Currently it's possible to publish/discard twice, while another publish is in progress. That lead to corruption of CR sometimes (which should be a separate issue actually).


DrillSergeant:
Works for me.

PR #1730, opened 2018-04-03
BUGFIX: In- & outdent can be disabled outside of lists

ComiR:
**What I did**
Indent & outdent buttons of CKEditor show:
* only if they are possible (outdent only if indented)
* if cursor is in `ol` or `ul` OR
* if configured to `true`

**How I did it**
Adjusted `isVisibleWhen` condition.

**How to verify it**
indent


ComiR:
I couldn't figure out, what was meant to be implemented. So I chose this option, which made the most sense for me ;)

Before this, it only checked if `ol` or `ul` was allowed, which was kind of irrelevant.
Also there was a copy & paste bug only checking if `indent` is possible (even for outdent) 😉.

Issue #1733, opened 2018-04-03
Update end 2 end testing library testcafe


markusguenther:
### Description

The version 19.0 and 19.1 has been released since our last testcafe update. The new version is now
able to make screenshots. This issue take care of the dependency upgrade.

PR #1734, opened 2018-04-03
TASK: Update to testcafe 19.2.0

markusguenther:
Updates testcafe to the lastest version 19.1.0 with lots of bugfixes and some new features like screen shots. Also updates the testcafe-react-selectors.

Resolves: #1733


dimaip:
I suspect you also need to update their react selectors library...

dimaip:
Ah, sorry, you actually do update it...

markusguenther:
@dimaip Yeah first updated to the latest 1.1.0 but that did not work out and the 1.0.6 is already linked to testcafe 19.1.0. Need to investigate this. Because it turns out that the new versions should fix the testcafe issue #2015 but I now have the problem that this pops up in the  version. https://github.com/DevExpress/testcafe/issues/2015

Already comment the issue and hopefully they can help ... but will investigate this soonish ;)

markusguenther:
The testcafe fix is already approved but is waiting for merge and release. When the version gets released I will upgrade the package and hopefully then everything is working in the right way :)

PR #1735, opened 2018-04-04
BUGFIX: Improve performance of switching between nodes in NodeTree

skurfuerst:
without this fix, all nodes in the node tree get redrawn everytime
the user selects a different node; which leads to prohibitive performance
when switching nodes.


dimaip:
Post-review: looks good!

PR #1737, opened 2018-04-04
BUGFIX: improve performance for PolicyAspect

skurfuerst:
With this change, we are actually checking whether there are policies configured for the given PrivilegeTargets; and if not, we skip these code blocks.

For a node tree with 1000 nodes, this improves the node tree loading time from 15 to 5 seconds (AJAX request time)

### DIFF
--- a/Classes/Neos/Neos/Ui/Aspects/PolicyAspect.php
+++ b/Classes/Neos/Neos/Ui/Aspects/PolicyAspect.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace Neos\Neos\Ui\Aspects;
 
 /*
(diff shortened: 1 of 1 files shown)


Classes/Neos/Neos/Ui/Aspects/PolicyAspect.php
```diff
@@ -52,12 +56,44 @@ class PolicyAspect
...
+     * @var ObjectManagerInterface
+     */
+    protected $objectManager;
+
+    /**
+     * @param ObjectManagerInterface $objectManager
+     * @return array the key is a Privilege class name; the value is "true" if privileges are configured for this class name.
+     * @Flow\CompileStatic
+     */
+    public static function getUsedPrivilegeClassNames($objectManager)
```
dimaip: Just wondering, is this method fast? Perhaps its results could be cached? Did you try to profile it?

dfeyer: @dimaip Look like this method is statically cached, so should be fast ;)

skurfuerst: ack :) see the @Flow\CompileStatic annotation :)

dimaip: Guys, pardon my Flow ignorance, but doesn't this annotation just skips the step of Flow compilation, but would it cache its results? Or would it get re-executed for each and every node?

PR #1741, opened 2018-04-04
BUGFIX: Tab must respect position ordering in RightSideBar

dfeyer:


packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
```diff
@@ -250,6 +250,8 @@ export default class Inspector extends PureComponent {
                             $get('views', group).count() > 0
                         ), false))
 
+                        .sort((a, b) => $get('position', a) > $get('position', b))
```
dimaip: Maybe it makes sense to support positional array sorter here? I'm not sure if it was the case in the old UI though, but could be useful.

dfeyer: It was not the case, but yes can be a nice feature. 

If we can proceed with the bugfix first. Then the feature ;)

dimaip: Ok!

PR #1742, opened 2018-04-05
WIP BUGFIX: Improve tree performance for huge trees by using async rendering (hidden behind experimental feature flag)

skurfuerst:
NOTE: When enabling this, this currently leads to warnings in development mode in the console such as:

```
componentWillReceiveProps: Please update the following components to use static getDerivedStateFromProps instead: Connect(Node), DragSource(DropTarget(Header)), DropTarget(Header), DropTarget(NodeDropTarget), ThemedContents, ThemedDragDropContainer, ThemedIcon, ThemedNode, ThemedTree

componentWillUpdate: Please update the following components to use componentDidUpdate instead: Connect(Node)
```

We cannot fix this, ourselves,as this needs to be fixed in react-css-themr, react-dnd and react-redux. On the other hand, the deprecation warning is just a warning without negative side
effects for React 16.


dimaip:
Should we perhaps wrap inspector as well with `AsyncMode`?

skurfuerst:
I did some further tests; D&D kinda works but not always... not sure whether we should then merge this or not :-/

dimaip:
@skurfuerst uhm no, console warnings is one thing, but breaking DND is definitely a no-go.

dimaip:
Feature flags raise complexity so would be nice to have it fixed eventually, but let's go with it for now :+1:

skurfuerst:
We cannot merge this yet, as it depends on an unreleased version of react-redux, AND the tree does currently not render. However, the react-dnd improvements might be doone now :) For now, it's merely a WIP.

lorenzulrich:
@skurfuerst I assume that's still work in progress - are you using this in some patched instances? I'm asking because I experience quite some speed deterioration after upgrading an installation from 3.2 to 4.3. The tree is especially slow when there are many nodes (in my example ~700) under the same page.

dimaip:
Was it due to server request taking longer or due to some changes in neos-ui? If it's about neos-ui, it would be super helpful if you could git-bisect it.

skurfuerst:
@lorenzulrich hm, nope, I am not using this actively currently. IMHO this also fixes the "symptom", so trying to fix the original problem first would be better :)

skurfuerst:
@markusguenther let's close it for now

PR #1745, opened 2018-04-05
BUGFIX: add meta viewport tag to avoid resizing viewport

### LINKED ISSUE #1716: Inspector does not hide on mobile
#### Steps to Reproduce
1. Open UI in chrome or open on mobile device (then skip next step)
2. Go to "mobile view"
3. Hide Inspector

#### Expected behavior
Inspector is hidden fine
inspektor

#### Actual behavior
Site gets zoomed out due to inspector being right of the UI:
inspektor_fail

### Affected Versions
Neos: 3.3.9
UI: 1.0.9

maltemuth:
**What I did**

I avoided resizing the viewport on mobile devices when the right side bar slides out. This fixes #1716 

**How I did it**

I added a viewport meta tag setting the viewport width to device-width

**How to verify it**

Use any device / simulator of your choice, login to /neos, toggle right sidebar.

### DIFF
--- a/Resources/Private/Templates/Backend/Index.html
+++ b/Resources/Private/Templates/Backend/Index.html
@@ -3,7 +3,7 @@
 <html>
     <head>
         <title>Neos CMS</title>
-        <meta name="viewport" value="width=device-width">
+        <meta name="viewport" content="width=device-width,initial-scale=1.0">
         <script>_NEOS_UI_configuration = {configuration -> f:format.raw()};</script>
         <script>_NEOS_UI_nodeTypes = {nodeTypes -> f:format.raw()};</script>
         <script>_NEOS_UI_frontendConfiguration = {frontendConfiguration -> f:format.raw()};</script>
--- a/Resources/Private/Templates/Backend/Index.html
+++ b/Resources/Private/Templates/Backend/Index.html
@@ -3,6 +3,7 @@
 <html>
     <head>
         <title>Neos CMS</title>
+        <meta name="viewport" value="width=device-width">
         <script>_NEOS_UI_configuration = {configuration -> f:format.raw()};</script>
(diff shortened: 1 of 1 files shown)


Resources/Private/Templates/Backend/Index.html
```diff
@@ -3,6 +3,7 @@
 <html>
     <head>
         <title>Neos CMS</title>
+        <meta name="viewport" value="width=device-width">
```
mstruebing: I think this should be `<meta name="viewport" content="width=device-width, initial-scale=1.0">`

maltemuth:
Hey @mstruebing,

many thanks for your kind feedback. I'm very sorry that this PR is/was incomplete, I must've skipped some testing :|

I think there should be some strategic discussion about how the "mobile" backend should work, which will lead to additional changes in the meta-viewport-tag (like user-scalable=???). At the same time I think that having this meta-tag in with any value is actually better than having none.

Also, I think there should be some feedback from testing on actual mobile devices, since I don't trust e.g. Chrome devtools to get the situation exactly right. A sketch of test procedures might look like this:
- Login to neos backend on mobile device
- "toggle out" the right side bar
- confirm that no rescaling occurs
- reload the backend (with right side bar still "toggled out")
- confirm that no rescaling has happened
- toggle right side bar in
- confirm that no rescaling has happened

Finally, I'm terribly curious: Is there actual production use of the neos backend on tablets or even smartphones?

maltemuth:
@mstruebing: Unfortunately, I can't :( I might be able to in a few weeks, and the setup won't be professional; maybe someone else can jump in in the meantime?

maltemuth:
After some more Devtools-Testing, neither initial-scale nor device-width seem to be the solution; I don't know what I've seen when I opened this pull request, but this needs more investigation.

I'll update #1716 once I actually have some kind of solution.

Issue #1525, opened 2018-01-09
Crop is not enforced when uploading images


bwaidelich:
When configuring an image property with only one crop setting like:
```yaml
  properties:
    'teaserImage':
      type: 'Neos\Media\Domain\Model\ImageInterface'
      ui:
        inspector:
          editorOptions:
            features:
              crop: true
            crop:
              aspectRatio:
                locked:
                  width: 8
                  height: 1
```
The resulting image should be cropped automatically to the given ratio. Instead the crop settings are ignored, *unless the crop editor is actively used*.

johannessteu:
Should we force open the crop editor here so the editor notices that the uploaded image will be cropped? If not he might uploads an image and doesn't really recognize that it is force cropped

bwaidelich:
> Should we force open the crop editor here

yes, I think that would be a nice behavior

DrillSergeant:
Not sure if I shall put that into another issue, but also defaultOption does not work:

```
  properties:
    teaserImage:
      type: Neos\Media\Domain\Model\ImageInterface
      ui:
        label: i18n
        inspector:
          group: 'teaser'
          editorOptions:
            features:
              crop: TRUE
            crop:
              aspectRatio:
                options:
                  teaser:
                    width: 1000
                    height: 500
                    label: 'Teaser'
                enableOriginal: TRUE
                allowCustom: TRUE
                defaultOption: 'teaser'
```

I expect here that the image is cropped to 1000/500 when I select am image from the media library or upload an image. But in both cases no cropping at all happens.

I comment into this issue because it may be caused by the same thing. If not, I can open another ticket.

dimaip:
Oh sorry, GitHub closed it for me :)

DrillSergeant:
I just realized, it's not working as expected. The cropping area is selected when you open the cropping-window. But you need to click into it to apply cropping.
I would expect that the image is cropped already (just by uploading or selecting it) with any clicking on the cropping-tool neeed.

dimaip:
@DrillSergeant yes, that's right, as said the fix that I created only partly resolves the problem, thus the issue is still open. Will try to tackle it this week, but auto cropping is not that easy.

PR #1753, opened 2018-04-09
BUGFIX: respect default option in cropping

dimaip:
Partially fixes https://github.com/neos/neos-ui/issues/1525#issuecomment-376122477
It respects the default crop option, but does not enforce it by default yet.


drellstab:
Works in my project, default cropping option is selected as expected.

Issue #1748, opened 2018-04-06
CodeMirror CodeEditor can't be closed when containing 2 or more LoC


ComiR:
### Description
CodeMirror uses `z-index` 1 to 6. The close button of the `SecondaryInspector` gets assigned a `z-index` of 2. The lines in CodeMirror therefore overlap the close button, making it unusable.

### Steps to Reproduce
1. Open the CodeEditor for any property
2. write 2+ lines of code

#### Expected behavior
CodeEditor can be closed

#### Actual behavior
Close button cannot be clicked.

### Affected Versions
Neos: 3.3.9
UI: 1.0.9

markusguenther:
The index needs to be just 3, the others are not that important. The PR will raise the index and then the code close button is working again. The four other secondary inspectors are still working :)

PR #1755, opened 2018-04-09
BUGFIX: Increase close button z- index

markusguenther:
The secondary inspector has two constants for z index. The close
button needs to be incresed to value higher than 3. And the
possible iFrame has not z index yet. So we define a iframe zindex constant and so we also increase the close button constant.

Fixes: #1748


markusguenther:
The value needs to be increased to a value of 3 or higher. To not adjust the generation method for the constants or to not use a especial constant key with a decent value. I just added a constant for the possible iframe.

If it is not ok we should discuss which other version is preferred ;)

PR #1765, opened 2018-04-11
FEATURE: Add global keyboard shortcuts

theilm:
This PR adds a global hotkey registry. You can register hotkeys in the manifest.hotkeys.js that can trigger actions. Currently, only one hotkey is registered ("g i") that toggles the right sidebar (inspector).


mstruebing:
I like this.
I think this is the way to go, the only thing I would love to have would be to be able to set/add/manipulate shortcuts via `Settings.yaml` and it would be cool to not only be able to execute redux actions but any arbitrary js function - I didn't looked so much into sagas right now but I think it should be possible (like in my command palette example to open the preview in a new window isn't a redux action -- but depends on the redux store)

@theilm Do you plan to work further on this? Otherwise I would like to give it a try and take it over from you.

theilm:
Thanks for your comment @mstruebing. Using arbitrary js functions should be possible already (although redux actions would be the better choice in most cases probably).
Configuring hotkeys via Settings is a good idea. In order to provide a really accessible solution, we even have to make all hotkeys configurable for each user via the UI. Do you think we have to add configuration possibilities already to this PR or can we tackle these things in the next step?
I'll try to work further on this, but help is welcome, of course - so feel free to add commits!

dimaip:
@theilm I'd really try to avoid having long-lasting PRs, so please merge as soon as it's really tested in the current state and proceed from there...

packages/neos-ui/src/Sagas/UI/Hotkeys/index.js
```diff
@@ -0,0 +1,43 @@
...
+import {getGuestFrameDocument} from '@neos-project/neos-ui-guest-frame/src/dom';
+
+export function * handleHotkeys({globalRegistry, store}) {
+    yield take(actionTypes.System.READY);
+
+    const hotkeyRegistry = globalRegistry.get('hotkeys');
+    const items = hotkeyRegistry.getAllAsList();
+    let mousetrapPaused = false;
+
+    for (let i = 0; i < items.length; i++) {
```
dimaip: Just a style nitpick, but we strongly prefer `forEach` to loops... You may ignore this comment.

packages/neos-ui/src/Sagas/UI/Hotkeys/index.js
```diff
@@ -0,0 +1,43 @@
...
+        const waitForGuestFrameInteraction = yield race([
+            take(actionTypes.UI.ContentCanvas.STOP_LOADING),
+            take(actionTypes.UI.ContentCanvas.SET_CURRENTLY_EDITED_PROPERTY_NAME)
+        ]);
+        const action = Object.values(waitForGuestFrameInteraction)[0];
+
+        // Bind mousetrap to guest frame after content canvas stopped loading
+        if (action.type === actionTypes.UI.ContentCanvas.STOP_LOADING) {
+            const mousetrapGuest = new Mousetrap(getGuestFrameDocument());
+            for (let i = 0; i < items.length; i++) {
```
dimaip: Same

dimaip:
We could perhaps put defaults for shortcuts into Settings.yaml and then provide an option to override them from the user module later on...

theilm:
@dimaip Defaults via Settings.yaml and overrides via user module is the way to go. I'll create tickets for it after this PR has been merged. (And thanks for the other comments, I changed it to forEach.)

mstruebing:
I will have a closer look tomorrow morning, as always :)

mstruebing:
Merged!!!

PR #1767, opened 2018-04-12
TASK: Changes to support fontawesome5 in Neos.Neos.Ui

theDex:
This commit enables fontawesome 5 in the new UI. It depends on the according PR in the neos development collection. 

TODO:
extend react-fontawesome to support accessibility features like aria


packages/react-ui-components/src/Icon/index.js
```diff
@@ -1,16 +1,22 @@
...
 import style from './style.css';
 import Icon from './icon.js';
 
+fontawesome.library.add(brands, solid, regular);
+
 const ThemedIcon = themr(identifiers.icon, style)(Icon);
 
 export default injectProps({
-    iconMap: icons,
+    //TODO: What to do about the icon map?
```
johannessteu: Can't we just remove it?

packages/react-ui-components/src/Icon/style.css
```diff
@@ -20,10 +20,10 @@
     font-size: .5em;
 }
 .icon--paddedLeft {
```
johannessteu: Should we adjust the classname as well?

theDex: We talked about that with @dimaip. He said the word "padded" just means the icon has a space next to it and does not correspond with the css-property "padding". So according to him it does not matter if we use css-margin or css-padding to realise what this classname means.

packages/react-ui-components/src/Icon/index.js
theDex: I hope so. I'm not a 100% sure what it did with the old icons, and if we still need it somehow. So we kept it and wanted to leave the decision to remove it to somebody who knows more about the new Ui.

packages/neos-ui/src/Containers/EditModePanel/Panel/index.js
```diff
@@ -32,7 +33,9 @@ export default class Panel extends PureComponent {
             speed: 300,
             slidesToShow: Math.round(modes.length / 2), // https://github.com/kenwheeler/slick/issues/1207#issuecomment-105663300
             variableWidth: true,
-            adaptiveHeight: true
+            adaptiveHeight: true,
+            nextArrow: <div><IconComponent icon="angle-right" size="big" /></div>,
+            prevArrow: <div><IconComponent icon="angle-left" size="big" /></div>
```
theDex: This still produces an error output on the console even though it works. I could not find out how to solve that problem, because I'm not good enough with react yet. Maybe someone else can figure that out ... ```Warning: React does not recognize the `currentSlide` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `currentslide` instead. If you accidentally passed it from a parent component, remove it from the DOM element.```

kitsunet:
Would be awesome to get the errors fixed that travis complaints about.

markusguenther:
If it is ok I fork the PR and take over to fix the listing stuff. The commit from @theDex will be kept but as Contributor I am not able to adjust this here.

theDex:
@markusguenther You can do that, or you can now just push into my fork. I added you as a contributor.

theDex:
I fixed the linting in Icon/icon.js and Icon/index.js and am curious what Travis says about that now ... These are the files me and Maya changed, so I think I knew what to do there. All other linting errors the last build showed appered in files we never touched.

markusguenther:
@theDex You are able to check that locally as well ... ```make lint``` and everything is fine.
When you editor is using the editorconfig file you will also see the `errors` in your IDE. Sry did not make it yesterday. But if is something left I will take a look.

Issue #1779, opened 2018-04-18
Don`t disable createDialog button


markusguenther:
### Description

While fixing issue #1320 the create button of the create Dialog has been disabled.
It would be better for the user experience to not disable the button. Then it is possible to get early
feedback when you have a date time editor for instance with a not empty validator.

If the create button is disabled you will not understand why you can not continue. If the button is active you click it and will then get the validation error.

### Steps to Reproduce

1. Create create dialog like in issue #1320
2. Create new page
3. Try to create page and you will get no feedback what you have to do.

#### Expected behavior

Create button is active and triggers validators so that you get user feedback.

### Affected Versions

Neos: 3.3

UI: current master (current release is 1.0.11)

mstruebing:
My intention was to not let the editor click the button when there are existing errors and let him know that he/she need's to fix it - but if the majority is for not doing this I'm in!

dfeyer:
I think disabling a button is to communicate an unavailable action. If I can choose two editing mode A and B, it's OK to disable the current mode. Your need to switch to the other mode, to be able to select again the other.

From my POV, if there is some errors, and the user click the button you should remind the user that he needs to solve the errors. Disabling the button is bit like saying "This action is not possible go away".

mstruebing:
To be clear: The button is active first and only is disabled if an actual validation error is present. 

![peek 2018-04-20 16-08](https://user-images.githubusercontent.com/12071529/39055619-feec1e40-44b4-11e8-81df-878742954b21.gif)

wrybit:
Usually in our GUI project we **disable** a Button until all monetary Fields are filled in and all possible Client-Side validation was successful. On rollover we give the user some Feedback, what he needs to accomplish, to activate the Button. This works quite well in most cases.

In UX you wish to help users to avoid seening error messages - by education and support while he works with the GUI. An initaly active BTN seems to be wrong - in this case - to me. As it's fake.
This will make the user feel bad, if he just tries to push the Button. If possible, avoid that. :)

UX Law: "Feedback generates Trust" to fulfill that Law, you will have to add Feedback before the Error happens. Thats why we add a Hover (or Touch) Interaction to the disabled Button with additional Ix Informations for the user.

wrybit:
I also like Progressive Hovering (sadly there is no Mobile Fallback invented yet) and that could support the user, too. Example here: https://tympanus.net/Development/ProximityFeedback/ hover the Button.

skurfuerst:
[...]

PR #1781, opened 2018-04-18
TASK: Don`t disable createDialog save button

markusguenther:
Fixes: #1779


mstruebing:
Why is my review requested @jonnitto?

This discussion is related: https://github.com/neos/neos-ui/issues/1779

I don't know what we want to do about it now :)

markusguenther:
Talked with @jonnitto at the sprint ... would be nice if a UX expert decides what to do.

dimaip:
Based on feedback from #guild-ux let's move on in this direction: https://github.com/neos/neos-ui/issues/1779#issuecomment-387462960
Thanks for bringing it up!

Issue #1780, opened 2018-04-18
Add important to date time input field again.


markusguenther:
### Description

In the PR to issue #1320 we removed the important on the invalid css classes. This works out for
the create dialog but not for form elements in the right inspector bar.

### Steps to Reproduce

1. Create a inspector with a date time with an validator
2. Trigger the validator in edit the page or so

#### Expected behavior

The date time input will be rendered with a red border.

#### Actual behavior

Don't get the red border.

### Affected Versions

Neos: 3.3

UI: 1.0.11

PR #1782, opened 2018-04-18
BUGFIX: Fix red borders for invalid datetimes

markusguenther:
The datetime form field does not render the invalid border
color red when the input field is in the right inspector bar.

In the create dialog it works without important!

Fixes: #1780


mstruebing:
That's what the important was for ....

PR #1785, opened 2018-04-18
<!-- Thanks for your contribution, we appreciate it! -->

dfeyer:
**What I did**

**How I did it**

**How to verify it**


dfeyer:
This change also remove the second implementation of the FullScreenButton. The second one (for closing the full screen) is not needed now that we have the toolbar always on.

packages/neos-ui/src/Containers/ContentCanvas/style.css
```diff
@@ -23,7 +23,7 @@
     padding-top: 192px; /* height of EditModePanel added */
 }
 .contentCanvas--isFullScreen {
-    padding: 0;
+    padding: 40px 0 0 0;
```
dimaip: I think there's constant for 40px

mstruebing:
Personally, I like it. Should we decide this on our own or do we need to ask the whole team? :)

packages/neos-ui/src/Containers/ContentCanvas/style.css
mstruebing: `var(--spacing-GoldenUnit)`

dimaip:
Would be nice to at least ask #guild-ux if there are no objections to this change, but IMO it's really cool and goes well together with inline structured editors concept.

aertmann:
Rasmus did consider doing this at some point in relation to distraction free writing. Also this works better in the new UI now that the page is inside an iframe as there's no top margin altering the page to display the toolbar. So I'm def for it.

If no paragraph is being edited, is the content dimension selector shown?

dfeyer:
> If no paragraph is being edited, is the content dimension selector shown?

Yes

PR #1789, opened 2018-04-21
TASK: Refactor the PHP classes to PSR-4 structure

kitsunet:


kitsunet:
Cherry picked all relevant commits into 1.0, this should be clean to merge, although travis will probably still complain due to the autoloading problem in tests described in #1709

kitsunet:
This became a bit less urgent to merge now if it goes into 1.0 anyway... But if we can before the release even better.

kdambekalns:
The test failure looks genuine. Probably only some namespace import issue, but has to be fixed.

kitsunet:
@kdambekalns it is not... See my comments here https://github.com/neos/neos-ui/pull/1789#issuecomment-383272424 and here https://github.com/neos/neos-ui/pull/1709#issuecomment-383169639

kitsunet:
You can successfully test this locally by either installing the branch via composer or letting composer dump autoloaders after checking out the branch.

kitsunet:
ahhhhhh, and why do I have so many commits in here now

bwaidelich:
69 Commits, 171 Changes 😱  - a short description would make this a bit easier to review I guess. Seems to be more than just a refactoring

kitsunet:
Changing the branch messed it up....

PR #1790, opened 2018-04-22
TASK: Add mapper for Neos icon names to FA5

kitsunet:


kitsunet:
And yes, icon could get a little refactoring to make the code more compact and easier to read, but not in this PR ;)

packages/react-ui-components/src/Icon/mapper.js
```diff
@@ -0,0 +1,505 @@
...
+        'yc-square',
+        'yelp',
+        'yoast',
+        'youtube',
+        'youtube-play',
+        'youtube-square'
+    ];
+
+    if (!icon.startsWith('icon-')) {
+        return icon;
```
dimaip: Wait, but why do you return icons with `icon-` prefix? I thought you should have stripped it, no?

kitsunet: I return icons WITHOUT icon- prefix. Maybe those are third party icons or just something special. at least we cannot detect it here and need to take it as it is.

dimaip: Duh, sorry, missed the `!`

daniellienert:
This is not an replacement for #2001 but for my addition to #2001. Would you please remove the migration in this pr also? And shouldn’t this be targeted to #2001 as it is not functional by itself?

kitsunet:
@daniellienert this replaces the migration completely if I am not mistaken?

daniellienert:
@kitsunet: As already written: you only replaced my PHP Icon-name mapper from https://github.com/theDex/neos-development-collection/pull/1 with a JS Mapper - If your mapper also handles all icon names that could be defined in the node type yaml. Your mapper is missing the legacy name mapping from FA3. If that is covered we can remove the node migration from  https://github.com/neos/neos-development-collection/pull/2001 - but the other about 30K LOC changed in https://github.com/neos/neos-development-collection/pull/2001 are IMHO still needed.

Issue #1751, opened 2018-04-06
Reloading backend not working when documentNode is not available in default dimension


DrillSergeant:
### Steps to Reproduce

1. Have a setup with multiple dimensions (like country and language)
This is the one of my current project:
```
Neos:
  ContentRepository:
    contentDimensions:
      country:
        label: 'country'
        icon: icon-language
        default: 'global'
        defaultPreset: global
        presets:
          global:
            label: 'Global'
            values:
              - global
            uriSegment: 'global'
          ch:
            label: 'Schweiz'
            values:
              - ch
              - global
            uriSegment: 'ch'
            constraints:
              language:
                'de': true
                'fr': true
                '*': false
          de:
            label: 'Deutschland'
            values:
              - de
              - global
            uriSegment: 'de'
            constraints:
              language:
                'de': true
                '*': false
      language:
        label: 'language'
        icon: icon-language
        default: 'en'
        defaultPreset: en
        presets:
          en:
            label: 'English'
            values:
              - en
            uriSegment: 'en'
          de:
            label: 'Deutsch'
            values:
              - de
            uriSegment: 'de'
          fr:
            label: 'Français'
            values:
              - fr
            uriSegment: 'fr'
````
2. Go to dimension combination that is not the default one (in my case I go to global_de)
3. Create a node (so that the node does not exist in the default combination)
4. Reload the backend (F5 or Ctrl+R)

#### Expected behavior
The backend is loads again an I am at the same spot (node, dimension-combination)

#### Actual behavior
I get an error-message:
"Validation failed while trying to call Neos\Neos\Ui\Controller\BackendController->indexAction()."

### Workaround
As a workaround I can go to the browser-address-bar and delete everyting behind "https://www.domain.tld/neos" and hit enter. It brings me exactly on the spot I wanted to be.

### Affected Versions
Neos: 3.3.9
UI: 1.0.9

PR #1793, opened 2018-04-22
BUGFIX: fix urls with multiple dimensions

dimaip:
Fixes: #1751 

Fully encodes the node context path argument so `&` between different dimensions doesn't break it.
The url would look like `/neos/content?node=%2Fsites%2Fneosdemo%2Fnode-4nriy1uwtm2x0%40user-admin%3Blanguage%3Dde%26country%3Dde%2Cglobal`
Ugly but editors shouldn't care about contextPath anyways...


kitsunet:
Right, that definitely makes sense.

Issue #771, opened 2017-07-26
Remove quick fix for ContentElementWrapping


mstruebing:
When this is merged and in a stable release we could remove the quickfix from the neos-ui

see:
https://github.com/neos/neos-ui/commit/1fb4ac5f57a180f42bb06fed4391b430e0412e42
https://github.com/neos/neos-development-collection/pull/1635

dimaip:
This is merged. Anyone took care already?

skurfuerst:
I'd for now keep it, as it makes the React UI compatible with Neos 3.0 and up.

markusguenther:
Is this still relevant. Mean the new UI works now on 3.3 and 4.0 ... so backwards compatibility is not needed anymore. Or what is the goal? Should the UI also work on 3.2

dimaip:
I'd say let's go for it

PR #1796, opened 2018-04-22
TASK: Remove quickfix for ContentElementWrapping

markusguenther:
The quickfix for ContentElementWrapping for neos 3.2 could be
remove. The handling has been adjusted. So that the
ContentElementWrappingImplementation of fusion already take
care of it.


mstruebing:
closes #771 (if merged)

Issue #1804, opened 2018-04-23
JavaScript warnings for the EditPane Sliders


markusguenther:
### Description

In the react-slick package is an issue regarding the arrow buttons.

![bildschirmfoto 2018-04-23 um 10 20 00](https://user-images.githubusercontent.com/1014126/39114804-2ba7b96c-46e0-11e8-9882-4863747a811f.png)

This can be fixed by using react-components instead of plain HTML elements.

### Steps to Reproduce

1. Just load the backend

#### Expected behavior

Editpane slider buttons work without warnings :)

#### Actual behavior

Warnings in the console.

[What actually happened]

### Affected Versions

Neos: 3.3

UI: 1.0.11

dimaip:
Maybe even better to fix it upstream?

markusguenther:
They had lot of discussions there and it seems that the outcome is that theses properties don't belong to HTML and therefore they use React-Components now. They also do this now in there examples.

markusguenther:
PR has been merged :)

PR #1805, opened 2018-04-23
TASK: Resolve Javascript warning is in the console

markusguenther:
The react-slick component has an issue with plain HTML arrows. The solution is to use react components as wrappers.


markusguenther:
The warnings are gone. But I need to polish the PR a bit and maybe @dimaip or @mstruebing give a opinionated comment regarding creation a component in our package or just create a component in the EditPane folder because it is actually only used there?

Because when I make it as regular component I will also create storybook stuff and so on.

dimaip:
Yeah it doesn't make a lot of sense to put it as a reusable component IMO, I would just create it inline.

dimaip:
But up to you to judge

mstruebing:
If it is only used there and don't even considered to be reused it should be in EditPane in my opinion.

mstruebing:
What about this @markusguenther? :)

markusguenther:
Warnings has been resolved but the buttons does not work anymore. Need to adjust that.

markusguenther:
I finally made it to update this PR. Sorry that it took so long.

markusguenther:
Ah and think forgot to mention that this fixes #1804

PR #1808, opened 2018-04-23
Task/core migrations Neos.Neos.ui

mgoldbeck:
https://github.com/neos/neos-development-collection/issues/2014


markusguenther:
Ending in the composer file is crashing the style check

PR #1827, opened 2018-04-26
TASK: Remove useless return

mstruebing:
related https://github.com/neos/eslint-config-neos/pull/16

### DIFF
--- a/packages/neos-ui/src/Sagas/CR/NodeOperations/addNode.js
+++ b/packages/neos-ui/src/Sagas/CR/NodeOperations/addNode.js
@@ -108,7 +108,7 @@ export default function * addNode({globalRegistry}) {
                     }]));
                 }
 
-                default: return;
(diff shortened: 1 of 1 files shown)


mstruebing:
Basically: eslint was updated by @markusguenther which introduced and adjusted many rules, no one cares about our linter setup and I wanted to remove any linter warning which isn't a `TODO`-warning (reducing linter noise) but @dimaip proposed to adjust the rules to our needs :)

dimaip:
Yeah I would adjust the rules when I have the time and inspiration for such work :)
Or somebody else could take care as well.

dimaip:
I would be so much in favor of prettier, but also there were some people against it, but I could not possibly bash about code style and linting tools (getting old, eh?)

mstruebing:
@dimaip @markusguenther is this helpful for further adjustments or can I close this PR?

PR #1835, opened 2018-04-28
WIP FEATURE: Read keyboardShortcuts from Settings.yaml

mstruebing:


packages/neos-ui/src/Sagas/UI/Hotkeys/index.js
```diff
@@ -1,20 +1,36 @@
...
+    const getActualAction = actionAsString => actionAsString.split('.').reduce((action, actionStringPart) => {
+        if (action && actionStringPart in action) {
+            return action[actionStringPart];
+        }
+
+        return null;
+    }, actions);
+
+    for (const keyboardShortcut of Object.values(keyboardShortcuts)) {
+        const actualAction = getActualAction(keyboardShortcut.action.substr(keyboardShortcut.action.indexOf('.') + 1));
```
dimaip: I'm not quite sure about coupling keyboard actions to redux actions. Alternatively we could still keep the registry of all actions callable via keyboard. Would be more flexible, but also more boilerplate to write.
Anyways, not opinionated here. @theilm how do you feel about this?

mstruebing: I'm not sure whether it would be better to be able to call arbitrary js-functions or only redux-actions.
I mean everything that is meaningful should be represented in the store and so manipulatable via a redux action anyway, or am I wrong?

dimaip: Perhaps you are right.

mstruebing: I'm not sure which path we should follow.
Having a registry where we store all available keyboard shortcuts which the 'user' can use would make it a lot safer but will increase our work amount. At the same time, it is more limiting to the user who maybe comes up with ideas we didn't even have considered how to use something.
In a registry, we could define more complex functions which would be hard with only settings.yaml.

So, what are your thought dumps?

Issue #1729, opened 2018-04-03
"Backend in Backend" - UI loads in guest frame (again)


ComiR:
### Description
As already described in #1332 and #1225, the UI sometimes loads itself again in the guest frame. I could determine two scenarios when this happens.

#### Session timeout
1. Use backend
2. Let the session time out
3. Refresh login via timeout modal
#### Using legacy backend
1. Use backend
2. In a new tab, open `/neos/legacy`
3. Switch page in tab # 1

#### Expected behavior
"Normal" backend behaviour.

#### Actual behavior
![image](https://user-images.githubusercontent.com/11410095/38238849-754b7442-372b-11e8-8732-508414049246.png)

The whole backend gets loaded in the guest frame. So there now is a "backend in backend" situation.

### Affected Versions
Neos: 3.3.9
UI: 1.0.9

klfman:
can confirm this, seems to be a regression? I reported this a few months ago and I think it was fixed

PR #1836, opened 2018-04-28
BUGFIX: prevent "backend in backend" behavior after session expires

dimaip:
Fixes: #1729 

With this change we invert the behavior of the session parameter and make the new UI activated by default, and storing `__neosLegacyUiEnabled__` into the session with the user clicks "Switch to old UI".

Obviously it won't alter the behavior of old systems without the neos-ui package installed.


dimaip:
I'll just merge it. If somebody is not happy with the behavior, should be easy to revert.

Issue #1823, opened 2018-04-25
Allow configuration of keyboard shortcuts via Settings.yaml


theilm:
### Description

We have a hotkey registry for global keyboard shortcuts (see #1765). The shortcuts are defined in the manifest.hotkeys.js file at the moment; they should be configurable via Settings.yaml.

mstruebing:
This is already the case see https://github.com/neos/neos-ui/pull/1839

mstruebing:
This is already the case see https://github.com/neos/neos-ui/pull/1839

PR #1839, opened 2018-04-29
FEATURE: Make keyboard shortcuts configurable via Settings.yaml

theilm:
This was an alternative try to #1835 for solving #1823

### DIFF
--- a/Configuration/Settings.yaml
+++ b/Configuration/Settings.yaml
@@ -55,7 +55,7 @@ Neos:
 
         hotkeys:
           'UI.RightSideBar.toggle': 't i'
-          'UI.FullScreen.toggle': 'f'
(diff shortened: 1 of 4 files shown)


theilm:
@dimaip @mstruebing Could you have a look? I am fine with the other solution, too. Would just be curious to know if it's ok to use a registry like that (passing configuration). Thanks!

dimaip:
Well yes, that was exactly my gut feeling to do it this way: feels somewhat cleaner, albeit more boilerplate code to write (which is OK).
I think even if we use hotkeys purely for triggering actions, this is still the way to go.

dimaip:
Good to go from my side.
@mstruebing please review and merge if you like it.

PR #1854, opened 2018-05-04
TASK: safely parse json response and show meanigful error on failure

dimaip:
Currently if backend breaks really badly it shows a general exception, and not an json encoded error message.
This PR detects non-valid JSON response, strips tags from response body and displays an error message.


mstruebing:
@dimaip do you have a cool way to reproduce this?

dimaip:
Haha cool no. I just inserted some `var_dump('break this JSON')` inside the property operation.

mstruebing:
Do we need to wait for @grebaldi // @skurfuerst here?

dimaip:
I don't think so, it's easy to revert if I did something stupid :)

PR #1873, opened 2018-05-09
FEATURE: sort options in dimension menu

samuelhauser:
Sort the options in the dimension menu for better user experience with many dimension values (more then 50 languages for example).

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,7 @@
     "lodash.isstring": "^4.0.1",
     "lodash.memoize": "^4.1.2",
     "lodash.merge": "^4.6.0",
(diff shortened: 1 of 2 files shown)


samuelhauser:
I'm not sure. Should we use the loadsh _.orderBy function?

dimaip:
@samuelhauser yes, why not

mstruebing:
But better `_sortBy` as it is more lightweight? :)

samuelhauser:
Does someone know, why the tests are failing? With the local tests I don't have any errors and the output from travis does not say much.

PR #1881, opened 2018-05-18
BUGFIX: Open preview in same tab

### LINKED ISSUE #1749: "Open in target dimension" does always open a new browser tab
The button "Open in target dimension" behaves differently in the new UI.
image

### Steps to Reproduce

1. Hit the "open in target dimension" button while in backend
2. A new browser tab should have opened showing the frontend
3. Go back to the backend-tab (dont close the frontend-tab)
4. Hit the "open in target dimension" button again

#### Expected behavior
The frontend-tab should have been reloaded

#### Actual behavior
A new frontend-tab has opened

### Affected Versions

Neos: 2.2.9

UI: 1.0.9 (also master-branch), 

It is working as expected when I switch to old UI.

### LINKED ISSUE #1880: PreviewButton always opens a new browser tab
### Description

With https://github.com/neos/neos-development-collection/commit/cbb7f9dc619decd17f06304804157f9646c4314b we told Neos to always show the preview in the same browser tab.
This behavior seems to be lost with the React UI: Every click on the previewButton opens a new tab.

### Steps to Reproduce

1. Hit the preview link: 
image

#### Expected behavior

The first time a new browser tab should be opened to display the content.
Successive clicks should re-focus the previously opened preview tab and reload the content.

#### Actual behavior

Every click opens a new browser tab

### Affected Versions

UI: All up to `v1.1.0`

mstruebing:
I don't know about security here, if someone has an opinion about
that let me know :)

closes #1749

### DIFF
--- a/packages/neos-ui/src/Containers/SecondaryToolbar/PreviewButton/index.js
+++ b/packages/neos-ui/src/Containers/SecondaryToolbar/PreviewButton/index.js
@@ -34,7 +34,6 @@ export default class PreviewButton extends PureComponent {
                     id="neos-PreviewButton"
                     href={previewUrl ? previewUrl : ''}
                     target="neosPreview"
-                    rel="noopener noreferrer"
                     className={previewButtonClassNames}
                     aria-label={i18nRegistry.translate('Neos.Neos:Main:showPreview', 'Show Preview')}
                     >


dimaip:
See the discussion at #1749
Seems like a minor risk though, but not ready to decide here.

dfeyer:
I agree with what @bwaidelich said on Slack:

> wow, that was fast. But security-wise it might be an issue indeed (https://mathiasbynens.github.io/rel-noopener/) OTOH we only open internal pages anyways so that should not be a problem

We control what page we want to open, so I'm fine to merge this. It's predictable for the editor and a regression compared to the Ember UI

dimaip:
We open internal pages, but the user may navigate from that page to a 3rd-party page.

Anyways, seems like a low-risk thing, but a high usability improvement.

bwaidelich:
> We open internal pages, but the user may navigate from that page to a 3rd-party page.

One could argue: If the integrator doesn't protect their links it is not our problem (not meant in a polemic way)

PR #1883, opened 2018-05-24
BUGFIX: Missing property in Property class

dfeyer:


dimaip:
Wow, how did it even work?

Issue #1885, opened 2018-05-29
AssetEditor renders an error message when no options are given at config


beheist:
### Description

AssetEditor without configured editorOptions renders only an error message.
**!!! This issue may exist for other editors as well, as the problematic line is there too, e.g. in the ImageEditor. !!!**

### Steps to Reproduce

The following config:
```
'My.NodeType:Audio':
  superTypes:
    'Neos.Neos:Content': true
  properties:
    audioAsset:
      type: Neos\Media\Domain\Model\Audio
      ui:
        inspector:
          editor: 'Neos.Neos/Inspector/Editors/AssetEditor'
```
produces an error message in the Neos UI instead of a rendered editor.
The problem is likely this line in AssetEditor:102:
`const features = Object.assign({}, DEFAULT_FEATURES, this.props.options.features);`
this.props.options is null, so what we get in the console is:
```
TypeError: Cannot read property 'features' of undefined
    at AssetEditor.isFeatureEnabled (Host.js?316debf5:14)
    at AssetEditor.renderAssetUpload (Host.js?316debf5:14)
    at AssetEditor.render (Host.js?316debf5:14)
    at h (Host.js?316debf5:6)
    at beginWork (Host.js?316debf5:6)
    at d (Host.js?316debf5:6)
    at f (Host.js?316debf5:6)
    at g (Host.js?316debf5:6)
    at m (Host.js?316debf5:6)
    at w (Host.js?316debf5:6)
```

It can be fixed by adding:
```
          editorOptions:
            features: []
```

#### Expected behavior

A rendered AssetEditor

#### Actual behavior
![image](https://user-images.githubusercontent.com/10347669/40637741-398c2fe8-6307-11e8-88b8-8764df4c3860.png)

### Affected Versions

Neos:

UI:

You can get this information by running `composer show` or using the package management module within Neos.
The UI version can also be found inside the left sidebar after version 1.0.3.

PR #1886, opened 2018-05-29
BUGFIX: Fix object traversal in AssetEditor (#1885)

beheist:
See #1885


dimaip:
But `$get` would not return empty object in case of non-existence, no?

mstruebing:
I've thought it would, but in this case, it's fine like it is.

dimaip:
Looking at this code it should just return `false` https://github.com/grebaldi/plow-js/blob/master/src/projections/atoms/get/index.js#L23

beheist:
So should I change it or leave it? :)

PR #1888, opened 2018-05-31
TASK: Add inline.editorOptions to propertyDefinition

kdambekalns:


kdambekalns:
@dimaip This could very well be expanded to include the mergin of `inline.editorOptions` with `aloha` as discussed in https://neos-project.slack.com/archives/C0U0KEGDQ/p1527756453000030

packages/react-proptypes/src/editorOptions.spec.js
```diff
@@ -0,0 +1,9 @@
+import editorOptions from './editorOptions';
+
+test(`should export a function.`, () => {
+    expect(typeof (editorOptions)).toBe('function');
```
dimaip: no braces necessary, just `typeof editorOptions`

packages/react-proptypes/src/editorOptions.js
```diff
@@ -0,0 +1,36 @@
+import PropTypes from 'prop-types';
+
+export default PropTypes.shape({
+    autoparagraph: PropTypes.bool,
+    placeholder: PropTypes.string,
+    formatting: PropTypes.shape({
+        strong: PropTypes.bool,
+        b: PropTypes.bool,
```
dimaip: `b` and `i` don't work with CKE I think (`strong` and `em` instead). If so then should we provide a migration or just leave as is and remove from the schema?

packages/react-proptypes/src/editorOptions.js
```diff
@@ -0,0 +1,36 @@
...
+        h4: PropTypes.bool,
+        h5: PropTypes.bool,
+        h6: PropTypes.bool,
+        pre: PropTypes.bool,
+        removeFormat: PropTypes.bool,
+        table: PropTypes.bool,
+        a: PropTypes.bool,
+        ol: PropTypes.bool,
+        ul: PropTypes.bool,
+        right: PropTypes.bool,
```
dimaip: shouldn't we support the old schema where alignment and links and lists are placed outside of formatting? Or just break and ignore again?

kitsunet: Good question, this is just documenting the status quo that currently exists in code. I think we have to decide if we say a) yes we want to have this and document it now or b) we go back say only use aloha config for now because we are still looking at how the new config might look like, then discuss and finalize it so it is stable and then document it when it gets implemented that way. I am fine with either but we need to decide that.

kdambekalns:
The thing is, it *is* documented already, but not in a way that is complete. I‘d be for officially deprecating aloha, bit quickly clarify the wanted syntax for this.

kdambekalns:
Hi, maybe @dimaip or @kitsunet can help me solve this conflict…

kdambekalns:
Ok, I think the "solution" is right, but please check again to see if this is still needed and working as expected.

packages/react-proptypes/src/editorOptions.js
kdambekalns: Is this solved with the current code? Then we could just drop them here?

packages/react-proptypes/src/editorOptions.js
kdambekalns: Again, might be solved with the current code already!?

packages/react-proptypes/src/editorOptions.spec.js
kdambekalns: I'm pretty sure I just copied that from somewhere… ;)

dimaip:
This shall no longer be relevant in light of TypeScript migration.
Thanks for the care though!

Issue #1715, opened 2018-03-23
Add special class to content shining through


ComiR:
### Description
When a content element exists in a fallback dimension, but not in the current one, it shines through.
The UI of `neos/neos` has a special class `neos-contentelement-shine-through` for these elements but with `neos/neos-ui` you can't differentiate between normal content and content shining through.

### Steps to Reproduce
1. Create content element in one dimension
2. Switch dimension to one with dimension above as fallback

#### Expected behavior
You can differentiate elements "only" shining through from those actually existing in the selected dimension.

#### Actual behavior
You can't.

### Affected Versions
Neos: 3.3.*, tested with 3.3.9
UI: <=1.0.9, tested with 1.0.9

dimaip:
These kind of things were never an official API, but I still believe we should fix it, so PRs are welcome.

PR #1892, opened 2018-06-07
TASK: Legacy class for content shining through

Elliryanna:
**What I did**
Added reference to old Node.php, now the elements get a class called "neos-contentelement-shine-through" like in the old one if they are not matching with the current dimension.

**How I did it**
Added a reference in NodeInfoHelper.php -> renderNode() and created a const in initializeContentDomeNode.js -> called: matchesCurrentDimensions.

**How to verify it**
You need to take a look at the html object of your UI, if the dimension shines trough the css-class is added.

solves #1715

### DIFF
--- a/Classes/Fusion/Helper/NodeInfoHelper.php
+++ b/Classes/Fusion/Helper/NodeInfoHelper.php
@@ -11,6 +11,7 @@ namespace Neos\Neos\Ui\Fusion\Helper;
  * source code.
  */
 
+use Neos\ContentRepository\Domain\Model\Node;
 use Neos\ContentRepository\Domain\Model\NodeInterface;
 use Neos\Eel\FlowQuery\FlowQuery;
 use Neos\Eel\ProtectedContextAwareInterface;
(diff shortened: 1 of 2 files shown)


Classes/Fusion/Helper/NodeInfoHelper.php
```diff
@@ -102,7 +103,9 @@ public function renderNode(NodeInterface $node, ControllerContext $controllerCon
             'depth' => $node->getDepth(),
             // TODO: 'uri' =>@if.onyRenderWhenNodeIsADocument = ${q(node).is('[instanceof Neos.Neos:Document]')}
             'children' => [],
+            'matchesCurrentDimensions' => $node instanceof Node && $node->dimensionsAreMatchingTargetDimensionValues(),
```
dimaip: Checking for `$node instanceof Node` shouldn't be required here, right, since it's enforced by the argument type hint?
And even if it was, you check for `NodeInterface`, not `Node`.

ComiR: The argument type hints to `NodeInterface` and `dimensionsAreMatchingTargetDimensionValues()` is only defined in `Node`. See [`Neos\Neos\Service\ContentElementWrappingService::addCssClasses()`](https://github.com/neos/neos-development-collection/blob/master/Neos.Neos/Classes/Service/ContentElementWrappingService.php#L171).

dimaip: Ah, true, sorry.

kitsunet: Seems that method should find it's way into the interface?

kitsunet:
Thanks @Elliryanna :)

PR #1894, opened 2018-06-11
FEATURE: Expose neos-ui-views

nlx-lars:
**What I did**
I exposed neos-ui-views so the dataLoader can be used in custom DataSource Views

**How I did it**
I added a shim in neos-ui-extensibility, a webpack alias and added that to the apiExposureMap.js.

### DIFF
--- a/packages/neos-ui/src/index.js
+++ b/packages/neos-ui/src/index.js
@@ -43,7 +43,7 @@ createConsumerApi(manifests, apiExposureMap);
 require('./manifest');
 require('@neos-project/neos-ui-contentrepository');
 require('@neos-project/neos-ui-editors');
-require('@neos-project/neos-ui-views');
+require('@neos-project/neos-ui-views/src/manifest');
 require('@neos-project/neos-ui-guest-frame');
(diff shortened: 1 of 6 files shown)


packages/neos-ui/src/index.js
```diff
@@ -43,7 +43,7 @@ createConsumerApi(manifests, apiExposureMap);
 require('./manifest');
 require('@neos-project/neos-ui-contentrepository');
 require('@neos-project/neos-ui-editors');
-require('@neos-project/neos-ui-views');
+require('@neos-project/neos-ui-views/src/manifest');
```
dimaip: Why is this necessary? Couldn't it be just `require('@neos-project/neos-ui-views');`? Then you'd need to export needed views from `index.js`.

nlx-lars: That's how it's done for @neos-project/neos-ui-i18n a few lines below that as well. I took that as a reference.

dimaip: Ok, never mind then, could always be refactored.

PR #1896, opened 2018-06-13
TASK: Require node with "engines" in package.json instead of as a dep

beheist:
According to https://stackoverflow.com/questions/29349684/how-can-i-specify-the-required-node-js-version-in-packages-json, this is the "correct" way to make sure users are running the right node version.

!!! ATTN !!!
If you still have a "node" binary inside your `./node_modules/.bin`, you might have to remove it manually.
Recommendation: use "nvm" to manage your node versions: https://github.com/creationix/nvm

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ In order to start contributing, follow the following steps:
 
 1) Ensure you have the `dev-master` version installed (see above).
 
-2) We require [Chrome](https://www.google.com/chrome/browser/desktop/index.html) as well as the `yarn`(https://yarnpkg.com/en/) command and GNU Make(https://www.gnu.org/software/make/) to be installed on your system.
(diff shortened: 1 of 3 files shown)


mstruebing:
Summary how I thought the setup should be:

- yarn installed globally via your OS's package-manager (https://yarnpkg.com/en/docs/install)
- installing correct node version via yarn specified in the `package.json`
- everything works and you only need yarn installed as a prerequisite

grebaldi:
What's the bug exactly? I don't think switching back to .nvmrc is a good idea and I'd like to understand, why you think it's necessary.

mstruebing:
@grebaldi if you have an older node version it isn't possible to install even if you have yarn installed globally via your OS's package manager and not via npm.

Try it yourself: 

```
cd <NEOS_UI_ROOT>
make clean
nvm use 6
# verify that yarn is installed somewhere globally 
# and not in an npm/yarn directory
which yarn

# see it breaking
 make install
```

---
Currently I see it's because of `flow-mono-cli` 
I'll try to remove it as I think we are not using it anyway.

mstruebing:
I'll close this for now.

PR #1897, opened 2018-06-13
WIP: VERY primitive and simple implementation of error boundaries

beheist:
First draft (and a very basic one) of how error boundaries could be pulled in into several UI layers, so that we avoid white-screens of death.
Open questions, to discuss at the sprint:
* In which package should we put the ErrorBoundary component?
* How do we style the error message?
* In which levels do we want error boundaries at all?

### DIFF
--- a/packages/neos-ui/src/Containers/App.js
+++ b/packages/neos-ui/src/Containers/App.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import ErrorBoundary from './ErrorBoundary/index';
 import FlashMessages from './FlashMessages/index';
 
 const App = ({globalRegistry, menu}) => {
@@ -16,6 +17,8 @@ const App = ({globalRegistry, menu}) => {
     const RightSideBar = containerRegistry.get('RightSideBar');
     const LoadingIndicator = containerRegistry.get('SecondaryToolbar/LoadingIndicator');
 
+    console.log('foo');
+
     // HINT: the SecondaryToolbar must be *BELOW* the
     // ContentCanvas; to ensure the SecondaryToolbar is rendered
     // afterwards and can overlay the ContentCanvas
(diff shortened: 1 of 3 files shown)


dimaip:
Would make sense to also take this boundary into account: https://github.com/neos/neos-ui/blob/1a4346e2d1844706b6af6ff6075b2f58605eb3fe/packages/neos-ui-editors/src/EditorEnvelope/index.js#L56
It already provides some styling.

packages/neos-ui/src/Containers/ErrorBoundary/index.js
```diff
@@ -0,0 +1,28 @@
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+
+export default class ErrorBoundary extends Component {
+    static propTypes = {
+        errorMessage: PropTypes.string
```
dimaip: Why is the error message coming from outside, and not from `componentDidCatch`?

dimaip: Ah, I see. So you don't want to display the actual exception to the user, right?

packages/neos-ui/src/Containers/ErrorBoundary/index.js
```diff
@@ -0,0 +1,28 @@
...
+    };
+
+    constructor(props) {
```
dimaip: We don't use constructor for this, you can just:
```
state = {
  hasError: true
}
```

packages/neos-ui/src/Containers/ErrorBoundary/index.js
```diff
@@ -0,0 +1,28 @@
...
+        // Display fallback UI
+        this.setState({hasError: true});
+        // You can also log the error to an error reporting service
+        console.error(error, info);
+    }
+
+    render() {
+        if (this.state.hasError) {
+            // You can render any custom fallback UI
+            return <h2>{this.props.errorMessage || "This UI element could not be loaded. Check the console output for details."}</h2>;
```
dimaip: Maybe also promt to reload the UI and the reload button?

packages/neos-ui/src/Containers/App.js
```diff
@@ -16,6 +17,8 @@ const App = ({globalRegistry, menu}) => {
     const RightSideBar = containerRegistry.get('RightSideBar');
     const LoadingIndicator = containerRegistry.get('SecondaryToolbar/LoadingIndicator');
 
+    console.log('foo');
```
dimaip: console.log

dimaip:
Hi @beheist!
I think this change is super important. Do you need any help to get it finished?

beheist:
I agree - unfortunately it's a matter of time for me ;) Likely won't happen in the next 3 weeks, as I'll be very busy. If you have time currently, you can take this over if you want?

dimaip:
I'll focus on other things for now, so I guess this can wait. You've spent more time with this issue, so you know it better.

markusguenther:
Hi guys do you have any ideas how we could push this forward or is it not that important anymore?

markusguenther:
As we had no action for so long, I would close it.

PR #1899, opened 2018-06-14
TASK: Improve "Changes" saga

kitsunet:
In order to make behavior slightly more predictable and
avoid race conditions on the server the changes are now
sent in order and with only one request at a time.

We collect changes from change actions until one of the
following happens:

* we have taken 3 actions (to avoid too many unpersisted changes)
* 2 seconds have passed

If at that point another request is still running, the wait is
prolonged until that request has finished at which point the
new batch of changes will be persisted.

### DIFF
--- a/packages/neos-ui/src/Sagas/Changes/index.js
+++ b/packages/neos-ui/src/Sagas/Changes/index.js
@@ -44,7 +44,7 @@ export function * watchPersist() {
         if (action) {
             changes.push(...action.payload.changes);
         }
-        
+
         const state = yield select();
         if (changes.length > 0 && !$get('ui.remote.isSaving', state)) {
             yield call(persistChanges, changes);
--- a/packages/neos-ui/src/Sagas/Changes/index.js
+++ b/packages/neos-ui/src/Sagas/Changes/index.js
@@ -1,5 +1,5 @@
 import {take, race, put, call, select} from 'redux-saga/effects';
-import {delay} from 'redux-saga'
+import {delay} from 'redux-saga';
 import {$get} from 'plow-js';
 
 import {actionTypes, actions, selectors} from '@neos-project/neos-ui-redux-store';
--- a/packages/neos-ui/src/Sagas/Changes/index.js
+++ b/packages/neos-ui/src/Sagas/Changes/index.js
@@ -36,6 +36,15 @@ export function * watchPersist() {
     let changes = [];
 
     while (true) {
+        const {action} = yield race({
+            action: take(actionTypes.Changes.PERSIST),
+            time: call(delay, 1500)
+        });
+
+        if (action) {
+            changes.push(...action.payload.changes);
+        }
+
         const action = yield take(actionTypes.Changes.PERSIST);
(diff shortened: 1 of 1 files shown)


kitsunet:
@hlubek this is what we talked about in regards to change requests. Should help our case.

dimaip:
Wow Christian you are so cool! I'll be able to review tomorrow.

packages/neos-ui/src/Sagas/Changes/index.js
```diff
@@ -1,34 +1,72 @@
...
-                const publishableNodesInDocument = publishableNodesInDocumentSelector(state);
-                yield put(actions.CR.Workspaces.publish(publishableNodesInDocument.map($get('contextPath')), baseWorkspace));
-            }
-        } catch (error) {
-            console.error('Failed to persist changes', error);
+        /*
+         *  We actually persist IF:
+         *  - we have changes AND
+         *  - there is not another request running
+         *  - either the "saveDelay" time has passed OR we have taken 3 actions (to avoid huge change queues locally)
```
dimaip: Just wondering, how did you come up with this number of 3? Did you test having longer queue? Modern JS engines are really powerful and keeping even a few MBs in memory would be nothing.

dimaip: On the other hand I don't understand the need for `2s` delay here? What does it give us? The inputs that require delay should have already been debounced.

kitsunet: yes for inline that is true, for inspector not so much it seemed (although I didn't check the code yet). It just seemed sensible to "collect" changes happening within a certain timeframe.

kitsunet:
Ok, so I had a discussion with @dimaip about this here:
https://neos-project.slack.com/archives/C0U0KEGDQ/p1529309298000386

And now tried to remove the delay. But then UI inconsistencies can happen (which I guess are due to the change request return values).

Generally the following easily triggers problems for me:
Take inline editable element, click in, position mouse over "hide" button in the element toolbar, start typing, while typing press "hide". This is not 100% but pretty often I get the following behaviors:

* without this change -> 403 exception probably due to race conditions in session handling
* with this change as is -> all seems to work smoothly
* with change but removed delay -> no exceptions but you can bring the UI in an inconsistent state (eg. hidden button not depressed but element "greyed out" or other way around, sometimes neither happens while the elemetn is still hidden)

So tl;dr; I woudl prefer the change to go in as is.

kitsunet:
Ok, these UI inconsistencies can even happen with delay. It might be a bit harder to coerce them, but it happens. So we should look into that at some point but that means the delay doesn't really have a benefit here. I removed it in favor of just sending chnages when possible.

WARNING DO NOT MERGE, I just figured we still need to race with a fixed time to see if a chnage is still queued, but I need to run now, will fix later.

kitsunet:
Alright, this should do it. We frequently check every 1.5 secs for open changes, which is needed to actually persist a change that came in while a request was running if no other changes happen afterwards.

PR #1900, opened 2018-06-15
BUGFIX: Let ImageEditor only accept images

### LINKED ISSUE #1879: ImageEditor should only accept images
#### Expected behavior
For the upload, I'd expect to be shown a choose file dialog like the one of `<input type="file" accept="image/*"/>`.

For the media browser I at least would expect it to have the filter pre-set to "Images".

Since the possibility of an editor still choosing a file of the incorrect type remains, the handling of such cases should be more graceful.
In a particular case, an editor uploaded a PDF file and published it with that file somehow being used as an image. This broke the site completely and resulted in a `The Flow application could not be launched.`.

#### Actual behavior
Upload a file that is not an image:
upload
I can select any file type I want, but if it isn't an image, it shows no preview.

If you open the media browser (by clicking the camera icon), you can choose any file type. If you don't choose an image, let's say a PDF, you'll get an error message due to `Argument 1 passed to Neos\Neos\Controller\Backend\ContentController::imageWithMetadataAction() must implement interface Neos\Media\Domain\Model\ImageInterface, instance of Neos\Media\Domain\Model\Document given`.

### Affected Versions
Neos: 3.3.9
UI: 1.1.0

mstruebing:
closes #1879

### DIFF
--- a/packages/neos-ui-editors/src/Library/AssetUpload.js
+++ b/packages/neos-ui-editors/src/Library/AssetUpload.js
@@ -78,7 +78,7 @@ export default class AssetUpload extends PureComponent {
     }
 
     render() {
-        const {isLoading, highlight} = this.props;
+        const {isLoading, highlight, multiple, children} = this.props;
 
         const classNames = mergeClassNames({
             [style.thumbnail]: true,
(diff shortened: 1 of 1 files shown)


packages/neos-ui-editors/src/Library/AssetUpload.js
```diff
@@ -73,12 +73,22 @@ export default class AssetUpload extends PureComponent {
...
     }
 
-    getUploadMetaData() {
+    getUploadMetaData = () => {
         return this.props.imagesOnly ? 'Image' : 'Asset';
     }
 
+    acceptedFileTypes = () => {
+        if (this.props.imagesOnly) {
+            return ['.jpg', '.jpeg', '.png'].reduce((fileTypeString, fileType) => {
```
mstruebing: I'm not sure which image formats I should allow here. Do we have a definition somewhere?

We could/should make it configurable later on I think.

kdambekalns: Well, seeing SVG more and more. Absolutely needs to be configurable. What about jp2, webm, …?

mstruebing: Good question.
I work on a configurable solution soon.

packages/neos-ui-editors/src/Editors/Image/Components/PreviewScreen/index.js
```diff
@@ -64,6 +65,7 @@ export default class PreviewScreen extends PureComponent {
                     highlight={highlight}
                     ref={this.setAssetUploadReference}
                     imagesOnly={true}
+                    accept={accept === undefined ? 'image/*' : accept}
```
dimaip: Could as well be `accept || 'image/*'`?

PR #1902, opened 2018-06-15
TASK: Use Elm for Checkbox

mstruebing:
:trollface: 

Don't get it serious - I just wanted to try if and how that works :)

//cc @dfeyer


dimaip:
Oups, it came sooner then I hoped :rofl:

Build/Dependencies/react-elm-components/example/Chat.elm
```diff
@@ -0,0 +1,139 @@
+port module Chat exposing (main)
```
dimaip: Why commit these examples?

mstruebing: Like I said, it's more a joke than serious and I simply cloned the `react-elm-components`-package to change one line in their code.

kitsunet:
No!?!

PR #1903, opened 2018-06-18
BUGFIX: Determine closest documentNode for `reloadPageIfChanged`

mficzel:
Without this change the ui caused an "Could not resolve a route for building an URI for the given resolve context." error whenever `reloadPageIfChanged` was used on a non-document nodes because it tried to route the actual content node to the frontend controller.


mficzel:
This was fixed in master already but inside the reloadDocumentFeedback object

Issue #1890, opened 2018-06-07
Version number displayed in menu interferes with menu items if the list grows too long


grebaldi:
![neos-ui-version-number](https://user-images.githubusercontent.com/2522299/41079769-7473e092-6a23-11e8-9857-852649570d74.png)

This is likely to happen in a multi-site setup with lots of sites.

PR #1912, opened 2018-06-26
BUGFIX: Use flexbox to align the drawer menu

markusguenther:
The version number needs to be sticky at the bottom of the drawer.
So we use now flexbox to align the menu item groups and the version number.
If the List of menu items become to long we now get a scrollbar.

Fixes: #1890

### DIFF
--- a/packages/neos-ui/src/Containers/Drawer/style.css
+++ b/packages/neos-ui/src/Containers/Drawer/style.css
@@ -55,5 +55,5 @@
 .drawer__version {
     color: var(--colors-ContrastBright);
     padding: var(--spacing-Full);
-    margin-top:auto;
+    margin-top: auto;
 }
--- a/packages/neos-ui/src/Containers/Drawer/style.css
+++ b/packages/neos-ui/src/Containers/Drawer/style.css
@@ -1,19 +1,33 @@
 .drawer {
     position: fixed;
-    top: 41px;
+    display: flex;
+    flex-flow: column;
+    flex-shrink: 0;
+    justify-content: flex-start;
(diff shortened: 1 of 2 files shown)


packages/neos-ui/src/Containers/Drawer/style.css
```diff
@@ -1,19 +1,33 @@
...
     position: fixed;
-    top: 41px;
+    display: flex;
+    flex-flow: column;
+    flex-shrink: 0;
+    justify-content: flex-start;
+    top: calc(var(--spacing-GoldenUnit) + 1px);
     left: 0;
     z-index: var(--zIndex-Drawer-Context);
     height: 100%;
```
mstruebing: this needs to be lower because of the top bar, otherwise you could'nt scroll down enough to see the version number when there are many items. (I think something like `100% - goldenUnit` or smth)
screenshot-2018-6-26 neos demo site

markusguenther:
@mstruebing Good point ... also moved the spacing at the bottom.

packages/neos-ui/src/Containers/Drawer/style.css
```diff
@@ -40,6 +55,5 @@
 .drawer__version {
     color: var(--colors-ContrastBright);
     padding: var(--spacing-Full);
-    bottom: var(--spacing-GoldenUnit);
-    position: absolute;
+    margin-top:auto;
```
mstruebing:
```
src/Containers/Drawer/style.css
 58:16  ✖  Expected single space after ":"   declaration-colon-space-after
```

PR #1915, opened 2018-06-26
TASK: Render custom icon component to visualize node status

### LINKED ISSUE #1500: Disabled/Timed DocumentNodes are not marked correct in the nodetree
If a document node is disabled it is marked as hidden right now. In the old ui there was a `x`-icon.

Old:
bildschirmfoto 2017-12-27 um 15 25 22

New:
bildschirmfoto 2017-12-27 um 15 25 42

### LINKED ISSUE #1895: DocumentTree Icons does not differentiate between hidden and hiddenInIndex
image

Neos: 3.3.*
UI: 1.1.*

johannessteu:
This change will add some icons on top of the nodetype icon in the tree to visualize a `hidden` and `hiddenBefore/After` status.

Resolves #1500 
Resolves #1895

### DIFF
--- a/packages/react-ui-components/src/Tree/node.js
+++ b/packages/react-ui-components/src/Tree/node.js
@@ -97,6 +97,7 @@ export class Header extends PureComponent {
         hasError: PropTypes.bool.isRequired,
         label: PropTypes.string.isRequired,
         icon: PropTypes.string,
+        customIconComponent: PropTypes.node,
         iconLabel: PropTypes.string,
         level: PropTypes.number.isRequired,
         dragAndDropContext: PropTypes.shape({
@@ -152,6 +153,7 @@ export class Header extends PureComponent {
             isLoading,
             label,
             icon,
+            customIconComponent,
             iconLabel,
             level,
             onClick,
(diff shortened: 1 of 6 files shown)


johannessteu:
Icons can now be easily set/adjusted for different states as well

bildschirmfoto 2018-06-26 um 13 25 51

packages/react-ui-components/src/Tree/node.js
```diff
@@ -197,7 +199,10 @@ export class Header extends PureComponent {
                             style={{paddingLeft: (level * 18) + 'px'}}
                             >
                             <div className={theme.header__labelWrapper}>
-                                <IconComponent icon={icon || 'question'} label={iconLabel} className={theme.header__icon}/>
+                                {customIconComponent ?
+                                  customIconComponent :
+                                  <IconComponent icon={icon || 'question'} label={iconLabel} className={theme.header__icon} />
```
mstruebing: Not the right amount of left-padding spaces, you need 4, not 2

johannessteu: meh >.>

johannessteu: But why 4 instead of 2 🤷‍♀️ ... anyways pushing a fix

mstruebing: also: `make lint` executes the linter locally :)

mstruebing: I'm not very opinionated about this but it's because of this line: https://github.com/neos/neos-ui/blob/master/.editorconfig#L7

:)

johannessteu: Yeah, wondering why atom doesn't respect that setting for me

Issue #1906, opened 2018-06-25
Gaps between buttons in Edit/Preview Panel missing


maya-shines:
While updating a project, I noticed in direct comparison, that the buttons for the editing mode and the preview central looked off. The space between them got lost somewhere between 3.3 and 4.0 and the slider icons became very small (I may very well have contributed to this during the font awesome 5 update...)

old:
![image](https://user-images.githubusercontent.com/6552092/41853048-1befaa6e-788d-11e8-83a1-4fcd32cc153f.png)
new:
![image](https://user-images.githubusercontent.com/6552092/41853067-27efc6d2-788d-11e8-9722-e9b97fac8a7d.png)

PR #1917, opened 2018-06-26
BUGFIX: Take care that the button margin is visible

markusguenther:
The buttons in the edit pane always overrides the margin-right.
So we now define a lastItem class that only resets the margin-right
for the last button in the edit pane.

Fixes: #1906

### DIFF
--- a/packages/neos-ui/src/Containers/EditModePanel/style.css
+++ b/packages/neos-ui/src/Containers/EditModePanel/style.css
@@ -38,8 +38,8 @@
     margin-right: 1rem;
     white-space: nowrap;
 
-    &:last-child {
-        margin-right: 0;
+    &:.editModePanel__lastButtonWrapper {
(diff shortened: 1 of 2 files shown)


packages/neos-ui/src/Containers/EditModePanel/style.css
```diff
@@ -38,8 +38,8 @@
     margin-right: 1rem;
     white-space: nowrap;
 
-    &:last-child {
-        margin-right: 0;
```
mstruebing: Maybe we should just remove that property and live with the fact that the last item has also a 1rem margin-right - this would reduce complexity, any thoughts on this?

markusguenther: Would be also a valid solution. I don't care if the last item has the margin or not ;)

PR #1918, opened 2018-06-26
TASK: Let jenkins test before releasing

mstruebing:
I'm not sure if we should do that as this should already be done via travis and so would extend the actual release time with tasks that are already done.

Any thoughts on this?


johannessteu:
Doesn't hurt at all. I'd say go for it

PR #1919, opened 2018-06-26
TASK: Use CachingHelper instead of hardcoded tags

johannessteu:


johannessteu:
Related to https://github.com/neos/neos-development-collection/pull/2097

The CachingHelper will get some more "magic" to resolve that issue. Therefore we must use it in all our packages of course

Issue #1920, opened 2018-06-27
Update outdated react-collapse


markusguenther:
### Description

The library react-collapse is outdated and needs an update.
latest version is 4.0.3 and we are using 2.4.1.

#### Expected behavior

Use latest react-collapse version

#### Actual behavior

Some libraries are outdated like react-collapse

PR #1921, opened 2018-06-27
TASK: Update react-collapse to version 4.0.3

markusguenther:
Updating the libray to version 4.0.3.

Resolves: #1920

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "react-click-outside": "^2.3.1",
     "react-close-on-escape": "^2.0.0",
     "react-codemirror2": "^5.0.1",
-    "react-collapse": "^2.4.1",
+    "react-collapse": "^4.0.3",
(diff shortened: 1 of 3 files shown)


dimaip:
Uhm, guys, but did is this issue now resolved with 4.0.3? We didn't upgrade for a reason...
https://github.com/neos/neos-ui/pull/1874

markusguenther:
Hmm I was wondering because I had in mind that I already upgraded this library. Sorry did not see that it was downgraded again.

dimaip:
@markusguenther no need to be sorry, there was no way to know that. I'm just wondering how to prevent somebody else upgrading it in the future (till the problem is solved)... There's no way to leave comments in package.json....

JamesAlias:
Ooops, I did it again 😅

Issue #1646, opened 2018-02-22
RFC: lazy-load policies


dimaip:
Currently rendering node information is dead slow due to policy checking. We can't instantly make it fast, it's a hard task to speed it up.

So I propose to lazy-load it.
On initial page load we just render info about current documentNode.
Then we lazily request all nodes needed to render the tree. It happens quickly, as policies are not applied at this moment (with an exception of `NodeTreePrivilege`).
And then the policies are lazily loaded as another json request. It doesn't really matter how long it would take, as you only notice policies when creating new nodes and doing stuff like that.

skurfuerst:
Makes sense :) I also checked the numbers (and after a wrong benchmark on my side) came to the same conclusions :)

PR #1927, opened 2018-06-27
FEATURE: Asynchronously load policy information for nodes

kitsunet:
Avoid evaluating policies when fetching node information from
server in order to improve performance of loading/display of nodes.

Policies for Documents are only fetched when the respective node is
selected in the tree. Policies for content nodes are fetched after
the page is loaded in the iframe.

Related: #1579
Fixes: #1646

### DIFF
--- a/Classes/Service/NodePolicyService.php
+++ b/Classes/Service/NodePolicyService.php
@@ -1,7 +1,6 @@
 <?php
 namespace Neos\Neos\Ui\Service;
 
-use Neos\ContentRepository\Domain\Model\NodeType;
 use Neos\ContentRepository\Domain\Service\NodeTypeManager;
 use Neos\ContentRepository\Security\Authorization\Privilege\Node\CreateNodePrivilege;
 use Neos\ContentRepository\Security\Authorization\Privilege\Node\CreateNodePrivilegeSubject;
@@ -47,7 +46,7 @@ class NodePolicyService
      * @return array the key is a Privilege class name; the value is "true" if privileges are configured for this class name.
      * @Flow\CompileStatic
      */
-    public static function getUsedPrivilegeClassNames($objectManager)
+    public static function getUsedPrivilegeClassNames(ObjectManagerInterface $objectManager): array
     {
         $policyService = $objectManager->get(PolicyService::class);
         $usedPrivilegeClassNames = [];
@@ -86,12 +85,11 @@ class NodePolicyService
         if (!isset(self::getUsedPrivilegeClassNames($this->objectManager)[NodeTreePrivilege::class])) {
             return true;
         }
-        $hasNodeTreePrivilege = $this->privilegeManager->isGranted(
+
+        return $this->privilegeManager->isGranted(
             NodeTreePrivilege::class,
             new NodePrivilegeSubject($node)
         );
-
-        return $hasNodeTreePrivilege;
     }
 
     /**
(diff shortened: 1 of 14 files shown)


kitsunet:
This also changes usage of the Nodes.ADD action to Nodes.MERGE to retain loaded policy information that would be erased by ADD.

Classes/Service/NodePolicyService.php
```diff
@@ -0,0 +1,176 @@
...
+     * @var ObjectManagerInterface
+     */
+    protected $objectManager;
+
+    /**
+     * @param ObjectManagerInterface $objectManager
+     * @return array the key is a Privilege class name; the value is "true" if privileges are configured for this class name.
+     * @Flow\CompileStatic
+     */
+    public static function getUsedPrivilegeClassNames($objectManager)
```
mstruebing: typehint's would be cool I think

packages/neos-ui/src/Sagas/CR/Policies/index.spec.js
```diff
@@ -0,0 +1,3 @@
+test(`write tests`, () => {
+    expect(true).toBe(true);
```
mstruebing: :1st_place_medal:

dimaip: What's the point of these empty test files? We never write any tests anyways...

kitsunet: :D I just followed best practice 😆

Classes/Service/NodePolicyService.php
kitsunet: right, will do!

packages/neos-ui-redux-store/src/CR/Nodes/index.js
```diff
@@ -67,6 +69,12 @@ const add = createAction(ADD, nodeMap => ({nodeMap}));
  */
 const merge = createAction(MERGE, nodeMap => ({nodeMap}));
 
+/**
+ * Specific action for node information coming from the guest frame,
+ * to be able to trigger more specific sagas based on that information.
```
dimaip: > "to be able to trigger more specific sagas"
These ain't cool. I'd rather fire an additional action to inform about that, or listen to one of the existing actions, if there are any

kitsunet: Mmm, maybe my explanation is bad? So far there is only merge and add which is very generic. We don't know where it comes from. In theory we could also use those BUT as we trigger a merge after we fetched policies that might result in a possible recursion? That's why I added this more specific action.

packages/neos-ui/src/Sagas/CR/Policies/index.js
```diff
@@ -0,0 +1,53 @@
...
+    const {endpoints} = backend.get();
+    while (true) {
+        const action = yield take(actionTypes.CR.Nodes.MERGE_FROM_GUEST_FRAME);
+        const {nodeMap} = action.payload;
+
+        const nodesWithoutPolicies = Object.keys(nodeMap).filter(contextPath => {
+            const getNodeByContextPathSelector = selectors.CR.Nodes.makeGetNodeByContextPathSelector(contextPath);
+            const node = select(getNodeByContextPathSelector);
+            const policyInfo = $get('policy', node);
+            return (!policyInfo || policyInfo === false);
```
dimaip: I don't understand this condition. Isn't it equal to `!policyInfo`?

PR #1928, opened 2018-06-27
TASK: Add Hover to PreviewScreen

mstruebing:
screenshot_2018-06-27 ix the mock turtle s story - the book - neos demo site


johannessteu:
Why actually? And if it's to get a feedback to the user that this element is clickable what about a blueish layer? The dashed border looks like it's cropable within the previewscreen.

mstruebing:
screenshot_2018-06-27 ix the mock turtle s story - the book - neos demo site 1

PR #1936, opened 2018-06-28
TASK: Fix several prop validation errors

johannessteu:


johannessteu:
Related to https://github.com/neos/neos-development-collection/pull/2108

johannessteu:
`FontAwesomeIcon` only allows those values: 
```
Warning: Failed prop type: Invalid prop `size` of value `big` supplied to `FontAwesomeIcon$1`, expected one of ["lg","xs","sm","1x","2x","3x","4x","5x","6x","7x","8x","9x","10x"].
```

and 2x was to large, lg does fit well

Issue #1651, opened 2018-02-26
NodeTypes properties validation feedback kicks in only after properites are populated


DrNik:
Hi Guys,
as mentioned on this post: https://discuss.neos.io/t/neos-react-ui-nodetype-property-validation-issue/3097 there seems to be a bug in Noes React UI property validation mechanism.
Basically validation feedback kicks in only if the properties that needs to be validated are populated (fill in + apply) first.

Steps to reproduce:

1. Have a NodeType with a property that must be validated, like:
```yaml
properties:
  [PROPERTY_NAME]:
     [...]
     validation:
        'Neos.Neos/Validation/NotEmptyValidator': []
```

2. Apply values to properties leaving the _must-be-validated_ property blank -> modifications get applied without warnings or “failed validation” feedback

3. Fill in mandatory property and Apply

4. Empty out the _must-be-validated_ property -> the “validation failed” popup is shown correctly

Tested versions:
Neos: 3.3.4
Neos.UI: 1.0.7

Thanks,
Nicola

markusguenther:
Is this still a problem? We worked on validation stuff at the Hamburg Codesprint and with regular fields like the title or url segment it is working fine. Also the create dialog handles this validations.

Issue #1426, opened 2017-12-12
Tooltips for validation and help messages have messed-up z-index


dimaip:
This PR added a better tooltip component: https://github.com/neos/neos-ui/pull/997

While we don't want to use it for all the tooltips (for performance and other reasons), it makes sense to extract the updated tooltip component from there and use it for validation and help messages.

dimaip:
Minimal thing that should be done is to fix the z-index of tooltips:

![image](https://user-images.githubusercontent.com/837032/34303445-6d094d7a-e746-11e7-8941-e34f58db8771.png)

mstruebing:
Is this still the case?

mstruebing:
Closed by #1937

PR #1937, opened 2018-06-28
TASK: Adjust responsibilities for validation result rendering

### LINKED ISSUE #1769: Some Editors does not render validation errors
### Description

If you create a custom createDialog like in issue #1320 you will recognize that some editors will not render the validation errors. It seams that only the textInput and textArea actually respect the validationError property and render the tooltip.

The DateTimeInput has been adjusted already in issue #1320 ... so probably all others excerpt the
three will have this validation error rendering issue.

### Steps to Reproduce

1. Configure a createDialog and add some additional fields (see the 1320 issue)
2. Create a new page without entering stuff

#### Expected behavior

For all possible editors the validation errors should be rendered. The validationErrors prop will exist for all on a higher level. The EditorEnvelope pass thru the props for all.

#### Actual behavior

Validation will be done, but the user gets no visual feedback for some field types.

[What actually happened]

### Affected Versions

Neos: 3.3

UI: 1.0.11 and lower

johannessteu:
This change takes moves the validation result rendering to a central place into EditorEnvelope. Also core ui components like a select box or a textfield should not care about stuff like `isHighlighted` oder rendering validation results. Those options are removed and will be handled from `EditorEnvelope`

Resolves #1769  #1651 #1426 (as we render tooltip with validation inline)

### DIFF
--- a/packages/neos-ui-editors/src/EditorEnvelope/index.js
+++ b/packages/neos-ui-editors/src/EditorEnvelope/index.js
@@ -67,7 +67,7 @@ export default class EditorEnvelope extends PureComponent {
             // We pass down a classname to render a highlight status on the editor field
             const classNames = mergeClassNames({
                 [style['envelope--highlight']]: highlight && !isInvalid,
-                [style['envelope--invalid']]: validationErrors
+                [style['envelope--invalid']]: validationErrors && validationErrors.length > 0
             });
 
             return (
@@ -149,7 +149,7 @@ export default class EditorEnvelope extends PureComponent {
                 </span>
 
                 {this.renderEditorComponent()}
-                {this.props.validationErrors && <Tooltip renderInline asError>{validationErrors}</Tooltip>}
+                {this.props.validationErrors && this.props.validationErrors.length > 0 && <Tooltip renderInline asError>{validationErrors}</Tooltip>}
             </Fragment>
         );
     }
--- a/packages/neos-ui-editors/src/EditorEnvelope/index.js
+++ b/packages/neos-ui-editors/src/EditorEnvelope/index.js
@@ -1,9 +1,13 @@
-import React, {PureComponent} from 'react';
+import React, {PureComponent, Fragment} from 'react';
 import PropTypes from 'prop-types';
+import omit from 'lodash.omit';
+import mergeClassNames from 'classnames';
+
 import Label from '@neos-project/react-ui-components/src/Label/';
 import {Tooltip} from '@neos-project/react-ui-components';
 import I18n from '@neos-project/neos-ui-i18n';
 import {neos} from '@neos-project/neos-ui-decorators';
+
 import style from './style.css';
 
 import {Icon} from '@neos-project/react-ui-components';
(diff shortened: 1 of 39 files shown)


johannessteu:
Resolves #1769  #1651 #1426 (as we render tooltip with validation inline)

johannessteu:
Some examples with faked validation errors / highlighted styles

bildschirmfoto 2018-06-28 um 21 32 17
bildschirmfoto 2018-06-28 um 21 32 08
bildschirmfoto 2018-06-28 um 21 32 13
bildschirmfoto 2018-06-28 um 21 31 23
bildschirmfoto 2018-06-28 um 21 31 15

packages/neos-ui-editors/src/EditorEnvelope/index.js
```diff
@@ -54,15 +60,22 @@ export default class EditorEnvelope extends PureComponent {
...
             const EditorComponent = editorDefinition && editorDefinition.component;
 
+            const restProps = omit(this.props, ['validationErrors', 'highlight']);
+            const {highlight, validationErrors} = this.props;
+            const isInvalid = validationErrors && validationErrors.length > 1;
+
+            // We pass down a classname to render a highlight status on the editor field
+            const classNames = mergeClassNames({
+                [style['envelope--highlight']]: highlight && !isInvalid,
+                [style['envelope--invalid']]: validationErrors && validationErrors.length > 0
```
mstruebing: Shouldn't this be the the same as `isInvalid` ?

packages/neos-ui-editors/src/EditorEnvelope/index.js
```diff
@@ -54,15 +60,22 @@ export default class EditorEnvelope extends PureComponent {
         if (editorDefinition && editorDefinition.component) {
             const EditorComponent = editorDefinition && editorDefinition.component;
 
+            const restProps = omit(this.props, ['validationErrors', 'highlight']);
+            const {highlight, validationErrors} = this.props;
+            const isInvalid = validationErrors && validationErrors.length > 1;
```
mstruebing: Shouldn't this be `> 0`? - as soon as one error occurs it is invalid I think

packages/neos-ui-editors/src/EditorEnvelope/index.js
```diff
@@ -54,15 +60,22 @@ export default class EditorEnvelope extends PureComponent {
         if (editorDefinition && editorDefinition.component) {
             const EditorComponent = editorDefinition && editorDefinition.component;
 
+            const restProps = omit(this.props, ['validationErrors', 'highlight']);
+            const {highlight, validationErrors} = this.props;
```
mstruebing: `const {highlight, validationErrors, ...restProps} = this .props`
That way you wouldn't need the `omit` function call

packages/neos-ui-editors/src/EditorEnvelope/index.js
johannessteu: fixed

packages/neos-ui-editors/src/EditorEnvelope/index.js
johannessteu: fixed

packages/neos-ui-editors/src/EditorEnvelope/index.js
johannessteu: fixed

PR #1941, opened 2018-06-29
TASK: Convert stateless function root into a component

johannessteu:


packages/neos-ui/src/Containers/Root.js
```diff
@@ -1,30 +1,34 @@
...
 import PropTypes from 'prop-types';
 import {Provider} from 'react-redux';
 import {withDragDropContext} from '@neos-project/neos-ui-decorators';
 import style from './style.css';
 
 import Neos from './Neos/index';
 
-const Root = ({store, globalRegistry, configuration, menu, routes}) => {
-    const containerRegistry = globalRegistry.get('containers');
+class Root extends Component {
```
mstruebing: Can we use `PureComponent` here instead of `Component`?

johannessteu: Sure

Issue #1889, opened 2018-06-06
Update CKEditor to a recent version, fixes strange editing behavior in Chrome


hlubek:
### Description

See https://github.com/neos/neos-development-collection/issues/2076 for a description of the editing issue. It seems it is caused by a regression for (at least) Chrome in CKEditor when editing certain inline elements.

A quick test of replacing it with the current version 4.9.2 seems to fix the issue.

UI: 1.1.0

dimaip:
I don't think it makes sense to maintain CKE4, as CKE5 integration is quite mature already and we should make it the default soonish.
Re-open if you disagree.

PR #1942, opened 2018-06-29
FEATURE: CKE5 integration

dimaip:
The default text editor is not configurable in Settings.yaml via `Neos.Neos.Ui.frontendConfiguraion.defaultInlineEditor`.
To test this editor set it to `ckeditor5`.

TODO:
- [x] sub/sup
- [x] placeholder
- [ ] removeformat
- [ ] table inline balloon UI, table styles
- [x] cleanup and documentation
- [ ] remove the CKE UI for editing the link


kitsunet:
- Table styles so table is always visible
- Double click in link leads to fatal error
- New text element has no toolbar

dimaip:
@kitsunet works on my machine (tm)

dimaip:
@kitsunet I'll be away from internet for a week, but then I'll try to ping you to debug what happens on your machine, while it works perfectly fine on mine. Also please test the current state on master, maybe it has somehow got better?

bwaidelich:
Did this resolve #1889 ?

mstruebing:
I would say no as the CK5 integration is currently experimental and more or less hidden via a setting.

kitsunet:
How do we deal with this problem (see thread)? 
https://neos-project.slack.com/archives/C0U0KEGDQ/p1530608559000131

Can we make sure that people don't need the respective ckeditor5 theme to compile their plugins?

mstruebing:
I will look into it tomorrow!

KPeschke:
Have also tested:
- The new link tooltip has different positions, mostly outside the node. It is currently not positioned as in your example video
- Underline and strike through missing
- The error reported by Christian with the double clicked link I had several times too. Unfortunately, it is difficult to reproduce the steps to cause this errors.
- trouble with non-block elements like span. It generates <p>-tags within the span
- missing hover and focus classes on element (blue dashed and solid outlines)
- it seems a automated spell check is active and marks the entire text in chrome

dimaip:
Hi @KPeschke! Thanks a lot for thorough testing!

- Link tooltip will be gone, replaced by completely custom UI.
- Underline and strikethrough seem to work for me. Do you have the enabled in NodeTypes.yaml?
- Will try to debug non-block elements tonight
- Same with hover, forgot about it
- Not sure about spell check, do you know what could be triggering it? I don't think I have it personally.

KPeschke:
Hi @dimaip! Thanks for your reply :)

- forgot to mention that I tested on Neos v3.3.12 with UI v1.2.2
- Underline and strikethrough: had enabled all possible "aloha" options
- spell check: "old" ckeditor has an attribute `spellcheck="false"` set

dimaip:
@KPeschke 
This seems to work for me:
```
      ui:
        aloha:
          'format':
            'underline': true
            'strikethrough': true
```
Though maybe this change is still in master and not 1.2.2.

Regarding spellchecker, could be cool to be able to enable it, but we don't know the content's language...

dimaip:
Please leave your test feedback here from now on: https://github.com/neos/neos-ui/issues/2002

PR #1950, opened 2018-06-29
TASK: Use stateless compoents

mstruebing:
Convert a bunch of components to stateless components whenever possible.


kitsunet:
Oh nice, so we get that in the release as well :)

skurfuerst:
See the comment at https://github.com/neos/neos-ui/pull/1980#issuecomment-405165808 why we revert this change

PR #1966, opened 2018-07-01
BUGFIX: Don't try to get policies for removed nodes

kitsunet:
Due to the property mapping the request would fail for
removed nodes but we actually don't even need the policies
for those so I just filter them before requesting.


kitsunet:
I will merge this in latest 24h as it basically breaks your editing experience if you have removed nodes.

mstruebing:
Sure, sorry used the full day to just chill after the sprint :)

PR #1967, opened 2018-07-01
TASK: Remove old ckeditor samples

kitsunet:
Due to the php publishing resources can fail and it's not
nice to have these samples laying around publicly anyway.


kitsunet:
See discussion here: https://neos-project.slack.com/archives/C0U0KEGDQ/p1529567550000396

PR #1980, opened 2018-07-04
TASK: Convert to Stateless components

mstruebing:
Convert the remaining possible classes to stateless components

### DIFF
--- a/packages/react-ui-components/src/SelectBox_CreateNew/selectBox_CreateNew.js
+++ b/packages/react-ui-components/src/SelectBox_CreateNew/selectBox_CreateNew.js
@@ -1,5 +1,5 @@
 /* eslint-disable camelcase, react/jsx-pascal-case */
-import React, {PureComponent} from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import SelectBox_Option_SingleLine from '../SelectBox_Option_SingleLine/index';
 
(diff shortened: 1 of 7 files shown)


skurfuerst:
@kitsunet @mstruebing We need to discuss this change! IMHO this destroys many performance optimizations - as all callback handlers are created ON EVERY RENDER instead of only once.

Or did I miss anything?

All the best,
Sebastian

mstruebing:
I was not aware of this :/
sorry, so we should revert this change?

skurfuerst:
a summary of  our discussion on slack:

I (Sebastian) wrote:
> So for me the performance problem is two-fold:
> 1st As far as I know (and it is started in the docs) SFCs currently behave not like PureComponent, but like Component with shouldComponentUpdate always returning true — thus the component always rerenders!
> 2nd - imho the pattern of SFCs encourages to „just quickly“ create new callback function handlers inside of render(); which effectively also forced nested PureComponents to always rerender. And in my experience, while components often start with simple render functions, at some point you often need handler functions - which are then more work to properly integrate, as you always need to convert the component from an SFC to PureComponent.
> 
> To me, there are various further general problems of using just Component (and thus SFCs as they behave the same) compared to PureComponent: it masks illegal state mutation and „mostly“ behaves as one would like; but you sometimes have to deal with lost updates and/or excessive rerenders again...
> 
> That’s why I always recommend and teach that one should always use PureComponent; and essentially forget the other two ways of writing components.
> 
> AFAIK the react community is a little split on this; there are people advocating using Component as default; and PureComponent only when needed „for performance reasons“; but I strongly disagree with this argumentation - for the reasons explained above.
> 
> 
> Bottom line - sorry that I only saw this right now and not earlier; and I would propose to always use PureComponent in our code base for the reasons above :)
>
> All the best - Sebastian
> PS do not get me wrong; you are all doing an awesome job with the UI and you fix and improve way more than I do currently :+1::kissing_heart: hope you do not mind that I am voicing my opinion so loudly ;)
> PPS: Some addendum: It is also possible to write PureComponents which break shouldComponentUpdate, e.g. by creating inline callback functions in render() or by calling .bind(this) inside the Render function -- however in our code we do not do this and IMHO there are eslint rules to prevent this.

@dimaip wrote:

> there are pros and cons to every approach.
> The only Pro of SFCs that Christian brought up is that you can tell from the first glance that a component is stateless. That's cool.
> The main disadvantage for me personally is whenever you need to add state or event handlers to a component you need to re-write it from scratch. People argued that it rarely happens, but for me it happened quite a bit even in our codebase, and it's a bit annoying. Also the PropTypes are written after the component definition, and I always need to scroll down to see what component does.
> 
> PureComponent has downsides too. E.g. it may actually decrease performance. If the components that are often meant to re-render, it would first need to do the `shouldComponentUpdate` and only then do the rendering, so the whole operation would actually be more expensive.
> Basically I tend to share the position to use Component as a default and only use PureComponent when necessary, but I don't really care so much, as I'm sure it doesn't make sense to have performance bike-shedding without doing actual measurements.
>
> Bottom line: I'd follow whatever you guys decide, but personally I'd rather not use SFCs at all.
>

I (Sebastian) wrote:
>
> imho measuring this is quite hard, as e.g. depends on the number of nodes in the system. Some problems f.e. Only occur on huge trees etc - would be of course great to have some numbers, but I would not spend too much time on it if it turns out to be hard.
> Btw one thing I think to keep in mind is that PureComponents shouldComponentUpdate is incredibly cheap - it is just a bunch of reference comparisons... what react does during render is way more expensive (and still fast)

PR #1981, opened 2018-07-04
BUGFIX: Use submodule path to check for availability

kitsunet:
The helper needs the full module path to check availability,
interestingly it apparently returns true for wrong paths,
that should be fixed in the future.


kitsunet:
Sure, just change the Neos.Neos policy file and remove a grant for any SUB module (eg. history). Without the change it will still be visible in the menu and then fail hard when you click it. With the change it won't show up in the menu at all.

PR #1984, opened 2018-07-04
BUGFIX: Correctly check for allowed NodeTypes and properties

kitsunet:
The filter should make sure that the collection of NodeTypes or
Node properties only contains those that are not allowed in the
end, which means the condition must be reversed to get the right
result.


kitsunet:
Will merge as soon as travis is done as the error is pretty obvious when you have a project with policies.

PR #1989, opened 2018-07-05
TASK: Update to ckeditor 4.10.0

kitsunet:
As there are problems with the currently used ckeditor 4.7.x and
ckeditor 5 is not yet fully integrated we should upgrade to a
recent version fo the time being.

Fixes: #1889


bwaidelich:
Unfortunately the update does _not_ seem to fix #1889

kitsunet:
nope, sadly not. Might still be a nice idea to upgrade?

dimaip:
Makes total sense. I propose to merge it in master and just wait about a week before releasing, to get more testers, as RTE bugs can be quite critical.

skurfuerst:
@dimaip sounds great :)

skurfuerst:
@dimaip so let's merge it, right?

Issue #1990, opened 2018-07-09
Interaction with Neos Backend is not possible without filtering for the NodeType


mgoldbeck:
### Description

In the documentation https://neos.readthedocs.io/en/4.0/ExtendingNeos/InteractionWithTheNeosBackend.html?highlight=Neos.NodeCreated#javascript-events is the interaction with the Neos Backend described. 

Essentially for this is to know which kind of NodeType is handled in the event. In the old ui the whole node was added to the node wrap in the backend

#### Expected behavior

I know there was a decision which I think is absolut right to NOT render the whole node into the wrapping div. But for interacting with the backend and to handle javascript based contentelements like sliders or tabelements it is necessary to have the information about which NodeType is handled in the fired event. Otherwise it would be necessary to add the NodeType information to the class or something else in the NodeType definition itself. 

#### Actual behavior

![node_wrap](https://user-images.githubusercontent.com/1336044/42436344-ef585528-8359-11e8-82ce-a4a70eb1d906.jpg)

### Affected Versions

Neos: 4.0.5

UI: 1.2.2

mgoldbeck:
As the PR will not be merged the documentation should be updated

PR #1991, opened 2018-07-09
TASK: add node type to contentelement wrap

mgoldbeck:
This adds `data-__neos-node-nodetype="Acme.Site:NodeTypeName"` to the wrapping div so it is possible to filter for the NodeType in the here described https://neos.readthedocs.io/en/4.0/ExtendingNeos/InteractionWithTheNeosBackend.html?highlight=Neos.NodeCreated#javascript-events eventhandling without the integrator is forced to add NodeType information to the markup like in classnames.

So it it would be possible to get the NodeType name with something like 
`const nodeTypeName = event.detail.element.dataset.__neosNodeNodetype;`

I think it is not the most elegant way but as far I do remember there was a decision to not render the whole node so this solution would be a good comprimise.

Of course the documentation needs to be adjusted then as well.

https://github.com/neos/neos-ui/issues/1990

### DIFF
--- a/Classes/Aspects/AugmentationAspect.php
+++ b/Classes/Aspects/AugmentationAspect.php
@@ -146,6 +146,7 @@ class AugmentationAspect
         $attributes = $joinPoint->isMethodArgument('additionalAttributes') ? $joinPoint->getMethodArgument('additionalAttributes') : [];
         $attributes['data-__neos-node-contextpath'] = $node->getContextPath();
         $attributes['data-__neos-fusion-path'] = $fusionPath;
+        $attributes['data-__neos-node-nodetype'] = $node->getNodeType();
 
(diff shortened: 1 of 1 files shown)


skurfuerst:
I do not have a 100% pro or con opinion on this; but I tend to follow @kitsunet 's reasoning - I'd personally like to reduce the "dom clutter" to a minimum; and only add attributes we actually need to the DOM.

All the best,
Sebastian

mstruebing:
I would personally prefer to have as less in the dom as needed too

kitsunet:
Seems there is no other way to do this though? So as long as there isn't an alternative in place we might as well...

skurfuerst:
@kitsunet you can add this via Fusion I think; and I'd go that route instead if you need it.

mgoldbeck:
Ok so I will close this

Issue #1988, opened 2018-07-05
After creating documents no sibling-documents can be created on the same level until parent is reloaded


mficzel:
### Description

After a new document was created (insert-into, -after or -before) no more sibling-documents can be created on the same level until the parent-document was loaded once. 

- If the sibling-document does not allow child-documents the whole plus icon stays disabled. 
- If the sibling-doucment allwows children the plus is enabled but in the create node-dialog insert-before and insert-after are disabled.

That way editors cannot create multiple documents on the same level in a single flow without reloading the backend or selecting the parent in between.

### Steps to Reproduce

1. Select a Document that already has Child-Documents 
2. Add another child Document and wait until it is loaded
3. Try adding a new child-document (notice that insert-before, -after are not available or the whole insert button ins disabled for leaf-documents)
4. Select a Sibling Document and wait until it is loaded
5. Try adding another new child-document (notice that insert-before, -after are not available or the whole insert button ins disabled for leaf-documents)
6. Select the parent document and afterwards repeat the attempts from 3. and 5. and notice that now sibling-documents can be created before and after.

### Affected Versions

Neos: I reproduced this in 3.3 + 4.0 but i assume master is affected aswell.
UI: I reproduced this in 1.2.2 and master but this might be older.

kitsunet:
AS described in #1985

mficzel:
@kitsunet correct. I looked up totally different terms when i searched wether this already was reported. Maybe leave this open for others who run into this as a pointer.

PR #1992, opened 2018-07-09
BUGFIX: Should not overwrite policy information on node add

kitsunet:
We cannot use add as that would override policy information or
in general already loaded data. Merging the information results
in working inserts right after adding a node again.

Fixes: #1988


kitsunet:
Will merge soon, so finally get rid of these policy problems.

Issue #1985, opened 2018-07-04
Async Policy problems


kitsunet:
We need to load policies for other pages in the tree to allow moving and adding above/below, both are not possible due to the lack of respective policies for other nodes.

We need to avoid overwriting policies after adding content.

PR #1993, opened 2018-07-09
BUGFIX: Do no load policies blocking

kitsunet:
The policies saga needs to catch all relevant events and shouldn't
be blocking, therefore refactoring while making sure that no nodes
are requested twice.

Resolves: #1985


kitsunet:
Narf, I think we need to fix this test somehow. I am pretty sure it is a timing issue in the test. I have no problems switching dimensions and the like.

PR #2006, opened 2018-07-13
TASK: Don’t render script tags for unstructured nodes

johannessteu:


kitsunet:
Literally nodes without nodetype, but I am pretty sure this is not the only place we should take care.

johannessteu:
Yeah, just a node without a nodetype set. Just like `/site` . In our case it's a kind of root node to hold all meta nodes of a kind.

johannessteu:
Close in favor for #2009

PR #2007, opened 2018-07-14
FEATURE: new link editor

dimaip:
This is a complete rewrite of the link editor.
The advanced pallet would allow to set anchors, title, rel and maybe event custom attributes configurable via settings, not yet ready.

Please give feedback on the UX concept!

peek 2018-07-17 09-15

Configuration of available link options:
```
'Neos.NodeTypes.BaseMixins:TextMixin':
  properties:
    text:
      ui:
        inline:
          editorOptions:
            linking:
              anchor: true
              title: true
              relNofollow: false
              targetBlank: false
```

TODO:

- [x] advanced options pallete
- [x] enter should submit the link
- [x] make advanced pallete options configurable
- [x] i18n, placeholders etc
- [ ] polish styling and UX (please MORE feedback!!)
- [ ] integrate the new editor into inspector
- [x] unlinking should also unlink title and other attributes


dimaip:
Current looks of the advanced options pallet:
image

daniellienert:
Great!!!!

DrillSergeant:
I like it!

KPeschke:
Looks really great. Love it. 

Is it possible to integrate the [rel="noopener](https://developers.google.com/web/tools/lighthouse/audits/noopener)" attribute for external links or set that attribute automatically with the target="_blank"? Or is this a feature request for the [neos-development-collection](https://github.com/neos/neos-development-collection) ([ConvertUrisImplementation](https://github.com/neos/neos-development-collection/blob/9179b3f7abfb8fd26994f41193db0dc5acea60fa/Neos.Neos/Classes/Fusion/ConvertUrisImplementation.php))?

dimaip:
@KPeschke I think so yes, that these kind of transformations that don't require editor choice are better to be left for server side. And if it's really required to do it in RTE, it's possible with a custom plugin (as that would be a really exotic usecase).

rolandschuetz:
Looks great!

I would recommend three small changes:

- The right 3-points icon should be a settings icon
- Empty links should automatically be removed when clicking away
- If it looks like a URL without a schema "Should this be formatted as URL starting with `http://`?" could be asked, similar for email “Should this be formatted as an email?”

Can't wait to get this new feature :)

dimaip:
@rolandschuetz thanks a lot for the feedback!
- a setting icon? you mean like a gear? Could you explain why you prefer it to 3 dots? What everybody else is thinking, I don't really mind personally.
- agreed, will fix
- yeah that would be pretty neat. Do you know any apps that do it right where I could get some inspiration from?

rolandschuetz:
@dimaip 

- The `cog` icon is generally understood as settings. The three dots suggest a submenu and the icon is lesser known. Since you configure the link I think cog makes sense.
- Cool
- Squarespace and Wordpress right now don't convert anything, they assume `http://` and add it as a prefix. Google Docs do convert E-Mail links automatically adding `mailto:`

image showing Squarespace link editor

dimaip:
@rolandschuetz I'll switch to cog icon then, if there are no other opinions.
Regarding auto-suggest, I'm not sure how to do it not in annoying way. I'd wait for somebody to come up with a solid UX concept how it should work (with mockups etc).

rolandschuetz:
I think the easiest way to now be annoying is to provide onentry in the
autosuggest with two lines:

Is this an email?
mailto:example@example.de

So it would look like the auto-suggest of Neos pages. But I would still
have a prompt for users, a bit annoying but in 99% of cases this will be
what the editor wants...


Dmitri Pisarev <notifications@github.com> schrieb am Mo., 23. Juli 2018,
12:36:

dimaip:
@rolandschuetz smth like this?

image

dimaip:
see https://github.com/neos/neos-ui/pull/2026

PR #2023, opened 2018-07-20
FEATURE: CKE5 table initial implementation

dimaip:
peek 2018-07-23 17-16

image

TODO:
- [x] render contextual dropdowns only when table is under cursor (no clue how to do that yet)
- [x] find icons for contextual dropdowns
- [x] make sure it's easy to add text before and after table and also to remove the whole table
- [x] add visual feedback to header commands


beheist:
Loving that we get a great table UI with this, thank you! The only thing I find a bit unsuitable is that yellow/purplish border around the table - at least from the short GIF, I didn't really understand what it does, and I wouldn't immediately understand that I can click the little nook at the top. Do you think we could make the functionality of this more clear somehow?

kitsunet:
I really like the bold colours there to make clear you selected a cell or the table. It behaves nicely. I think people will want to select multiple cells at some point or full rows/cols, although you don't really need to. So I don't tihnk it's a problem.

kitsunet:
I would probably go and merge this soon-ish

PR #2026, opened 2018-07-23
TASK: improve link editor

dimaip:
peek 2018-07-23 20-31

What this PR does:
- cog icon instead of dots
- suggestions for email and http links
- don't create empty link on first link icon click


kitsunet:
I have several problems with this, first of all it hardcodes `node://` and `asset://`  pretty deeply. Additionally it just assumes `http://` for external links, although `https://` is becoming more and more of a standard. Finally the double multi-line ternary is just unreadable (it's already bad in the current state IMHO)

PR #2029, opened 2018-07-25
BUGFIX: Remove PHP 7.0 incompatible ``void`` return type

gerhard-boden:
Since Neos 3.3 still supports PHP 7.0, we can't make use of the ``void`` return type yet.


gerhard-boden:
wow, that was fast! Thanks guys 😄

PR #2033, opened 2018-07-25
TASK: Ensure that LinkPreview does not overlap the controls

### LINKED ISSUE #2032: CSS issues in LinkEditor
bildschirmfoto 2018-07-25 um 21 55 53

If you select a Reference wich is pretty "long" it will overlay the apply/options-panel button

johannessteu:
Fixes #2032 2032


dimaip:
Hey unfortunately this solution is not very reliable, e.g. think what happens when there's no cog icon. But `max-width:0` would solve it, will do it in a different PR.

Issue #2036, opened 2018-07-27
CKEditor5 inserts bogus "@" character for empty editables on Document nodes


lisakampert:
### Description

When using CKEditor in Document-NodeTypes, the inline-editing behaviour isn't correct: Instead of showing the Placeholder-Text, the value is set to `@`. When deleting it, the placeholder (acutally just `i18n`)  is shown.

### Steps to Reproduce

1. Define Document-NodeType with Inline-Editing-Property:
```
Foo.Bar:Document.Location:
  superTypes:
    'Neos.NodeTypes:Page': true
  properties:
    teaserText:
      type: string
      defaultValue: ''
      ui:
        inlineEditable: true
        inline:
          editorOptions:
            placeholder: i18n
...
```
2. Create new Document in Neos-Backend
3. View `teaserText`-Field in Editing-Area

#### Expected behavior

Show the (translated) placeholder

#### Actual behavior

The value is set to `@`, when deleting it, the placeholder is shown.
`@` as value:
<img width="85" alt="bildschirmfoto 2018-07-27 um 09 26 28" src="https://user-images.githubusercontent.com/31652404/43307672-30922342-917f-11e8-82b8-c3107b1952a9.png">
after removing the `@`: 
<img width="113" alt="bildschirmfoto 2018-07-27 um 09 26 24" src="https://user-images.githubusercontent.com/31652404/43307673-30ab4f52-917f-11e8-8312-583fa6c5acd3.png">

### Affected Versions

Neos: 4.0.6

UI: 1.3.0

PR #2041, opened 2018-07-27
BUGFIX: prevent double init of inline editables

dimaip:
Fixes: #2036


dimaip:
Merging without reviews as a) no-brainer b) quite critical bugfix for some projects

PR #2048, opened 2018-07-31
FEATURE: NodeTree Presets

bwaidelich:
Adds basic support for "NodeTree presets" (see https://github.com/neos/neos-development-collection/issues/1952):
If additional presets are configured such as:

    Neos:
      Neos:
        userInterface:
          navigateComponent:
            nodeTree:
              presets:
                'default':
                  baseNodeType: 'Neos.Neos:Document,!Some.NodeType:ToBeHiddenByDefault'
                 'preset1':
                   ui:
                     label: 'Some Preset'
                     icon: 'icon-some'
                   baseNodeType: 'Some.NodeType:Foo'
                 'preset2':
                   ui:
                     label: 'Some Other Preset'
                     icon: 'icon-some-other'
                   baseNodeType: 'Some.NodeType:Bar'

Instead of displaying all Document node types, the NodeType filter will display the presets.

### DIFF
--- a/packages/neos-ui/src/Containers/LeftSideBar/NodeTreeSearchBar/NodeTreeFilter/index.js
+++ b/packages/neos-ui/src/Containers/LeftSideBar/NodeTreeSearchBar/NodeTreeFilter/index.js
@@ -1,5 +1,6 @@
 import React, {PureComponent} from 'react';
 import PropTypes from 'prop-types';
+import {$get} from 'plow-js';
 
 import {neos} from '@neos-project/neos-ui-decorators';
 import {SelectBox} from '@neos-project/react-ui-components';
@@ -17,6 +18,7 @@ export default class NodeTreeFilter extends PureComponent {
     static propTypes = {
         i18nRegistry: PropTypes.object.isRequired,
         nodeTypesRegistry: PropTypes.object.isRequired,
+        neos: PropTypes.object.isRequired,
         onChange: PropTypes.func.isRequired,
         value: PropTypes.any
     }
@@ -30,36 +32,49 @@ export default class NodeTreeFilter extends PureComponent {
     }
 
     render() {
-        const {i18nRegistry, nodeTypesRegistry, onChange, value} = this.props;
+        const {i18nRegistry, nodeTypesRegistry, neos, onChange, value} = this.props;
         const label = i18nRegistry.translate('filter', 'Filter', {}, 'Neos.Neos', 'Main');
 
-        const documentNodeTypes = nodeTypesRegistry
-            .getSubTypesOf('Neos.Neos:Document')
-            .map(nodeTypeName => nodeTypesRegistry.getNodeType(nodeTypeName))
-            .filter(i => i);
-        const options = documentNodeTypes.map(nodeType => ({
-            value: nodeType.name,
-            label: i18nRegistry.translate(nodeType.label),
-            icon: nodeType.ui.icon
-        }));
+        const presets = $get('configuration.nodeTree.presets', neos);
+        let options = Object.keys(presets)
+            .filter(presetName => (presetName !== 'default'))
+            .map(presetName => ({
+                    value: presets[presetName].baseNodeType,
+                    label: presets[presetName].ui && presets[presetName].ui.label || '[' + presetName + ']',
+                    icon: presets[presetName].ui && presets[presetName].ui.icon || null,
+                })
+            );
+
+        if (options.length === 0) {
+            const documentNodeTypes = nodeTypesRegistry
+                .getSubTypesOf('Neos.Neos:Document')
+                .map(nodeTypeName => nodeTypesRegistry.getNodeType(nodeTypeName))
+                .filter(i => i);
+
+            options = documentNodeTypes.map(nodeType => ({
+                value: nodeType.name,
+                label: i18nRegistry.translate(nodeType.label),
+                icon: nodeType.ui.icon
+            }));
+        }
 
         return (
             <div id="neos-NodeTreeFilter" className={style.searchBar}>
-                <SelectBox
-                    placeholder={label}
-                    placeholderIcon={'filter'}
-                    onValueChange={onChange}
-                    allowEmpty={true}
-                    value={value}
-                    options={searchOptions(this.state.filterTerm, options)}
-                    displaySearchBox={true}
-                    searchTerm={this.state.filterTerm}
-                    onSearchTermChange={this.handleFilterTermChange}
-                    threshold={0}
-                    noMatchesFoundLabel={this.props.i18nRegistry.translate('Neos.Neos:Main:noMatchesFound')}
-                    searchBoxLeftToTypeLabel={this.props.i18nRegistry.translate('Neos.Neos:Main:searchBoxLeftToType')}
-                    />
-            </div>
-        );
+    <SelectBox
+        placeholder={label}
+        placeholderIcon={'filter'}
+        onValueChange={onChange}
+        allowEmpty={true}
+        value={value}
+        options={searchOptions(this.state.filterTerm, options)}
+        displaySearchBox={true}
+        searchTerm={this.state.filterTerm}
+        onSearchTermChange={this.handleFilterTermChange}
+        threshold={0}
+        noMatchesFoundLabel={this.props.i18nRegistry.translate('Neos.Neos:Main:noMatchesFound')}
+        searchBoxLeftToTypeLabel={this.props.i18nRegistry.translate('Neos.Neos:Main:searchBoxLeftToType')}
+        />
+        </div>
+    );
     }
 }


packages/neos-ui/src/Containers/LeftSideBar/NodeTreeSearchBar/NodeTreeFilter/index.js
```diff
@@ -30,36 +32,49 @@ export default class NodeTreeFilter extends PureComponent {
...
-                    displaySearchBox={true}
-                    searchTerm={this.state.filterTerm}
-                    onSearchTermChange={this.handleFilterTermChange}
-                    threshold={0}
-                    noMatchesFoundLabel={this.props.i18nRegistry.translate('Neos.Neos:Main:noMatchesFound')}
-                    searchBoxLeftToTypeLabel={this.props.i18nRegistry.translate('Neos.Neos:Main:searchBoxLeftToType')}
-                    />
-            </div>
-        );
+    <SelectBox
```
dimaip: What happened to indentation here?

packages/neos-ui/src/Containers/LeftSideBar/NodeTreeSearchBar/NodeTreeFilter/index.js
```diff
@@ -30,36 +32,49 @@ export default class NodeTreeFilter extends PureComponent {
...
-            value: nodeType.name,
-            label: i18nRegistry.translate(nodeType.label),
-            icon: nodeType.ui.icon
-        }));
+        const presets = $get('configuration.nodeTree.presets', neos);
+        let options = Object.keys(presets)
+            .filter(presetName => (presetName !== 'default'))
+            .map(presetName => ({
+                    value: presets[presetName].baseNodeType,
+                    label: presets[presetName].ui && presets[presetName].ui.label || '[' + presetName + ']',
```
dimaip: easier to use $get here to save checking if the path in object exists

packages/neos-ui/src/Containers/LeftSideBar/NodeTreeSearchBar/NodeTreeFilter/index.js
```diff
@@ -30,36 +32,49 @@ export default class NodeTreeFilter extends PureComponent {
...
+                    icon: presets[presetName].ui && presets[presetName].ui.icon || null,
+                })
+            );
+
+        if (options.length === 0) {
+            const documentNodeTypes = nodeTypesRegistry
+                .getSubTypesOf('Neos.Neos:Document')
```
dimaip: Is it necessary to hardcode `'Neos.Neos:Document'` here? Can't one of the roles be used instead?

packages/neos-ui/src/Containers/LeftSideBar/NodeTreeSearchBar/NodeTreeFilter/index.js
```diff
@@ -30,36 +32,49 @@ export default class NodeTreeFilter extends PureComponent {
...
+                .map(nodeTypeName => nodeTypesRegistry.getNodeType(nodeTypeName))
+                .filter(i => i);
+
+            options = documentNodeTypes.map(nodeType => ({
+                value: nodeType.name,
+                label: i18nRegistry.translate(nodeType.label),
+                icon: nodeType.ui.icon
```
dimaip: safer to use $get

packages/neos-ui/src/Containers/LeftSideBar/NodeTreeSearchBar/NodeTreeFilter/index.js
dimaip: I mean this: https://github.com/neos/neos-ui/blob/master/Configuration/Settings.yaml#L104

packages/neos-ui/src/Containers/LeftSideBar/NodeTreeSearchBar/NodeTreeFilter/index.js
bwaidelich: that was an accident, already reverted ;)

packages/neos-ui/src/Containers/LeftSideBar/NodeTreeSearchBar/NodeTreeFilter/index.js
bwaidelich: This is the old implementation

dimaip: Fair point. Would be a good opportunity to refactor though :)

packages/neos-ui/src/Containers/LeftSideBar/NodeTreeSearchBar/NodeTreeFilter/index.js
dimaip: I mean this: `$get([presetName, 'ui', 'label'], presets)`

bwaidelich:
@dimaip Thanks for the review. I tried to address your remarks.

As for an example: It's quite simple.
By default the filter looks like this (as before):
image

Now, given the following preset configuration:
```yaml
Neos:
  Neos:
    userInterface:
      navigateComponent:
        nodeTree:
          presets:
            'default':
              baseNodeType: 'Neos.Neos:Document,!My.Site:Mixin.HideInBackendByDefault'
            'legalPages':
              ui:
                label: 'Legal pages'
                icon: 'icon-gavel'
              baseNodeType: 'My.Site:Document.Imprint,My.Site:Document.Terms'
            'landingPages':
              ui:
                label: 'Landing pages'
                icon: 'icon-bullseye'
              baseNodeType: 'My.Site:Mixin.Legal'
```
it would look like this:
image

bwaidelich:
In https://github.com/neos/neos-development-collection/issues/1952 a lot more features were discussed but this could be a first basic improvement

packages/neos-ui/src/Containers/LeftSideBar/NodeTreeSearchBar/NodeTreeFilter/index.js
```diff
@@ -40,41 +40,41 @@ export default class NodeTreeFilter extends PureComponent {
             .filter(presetName => (presetName !== 'default'))
             .map(presetName => ({
                     value: presets[presetName].baseNodeType,
-                    label: presets[presetName].ui && presets[presetName].ui.label || '[' + presetName + ']',
-                    icon: presets[presetName].ui && presets[presetName].ui.icon || null,
+                    label: $get('ui.label', presets[presetName]) || '[' + presetName + ']',
```
dimaip: I tried to provide an example, I meant this: `$get([presetName, 'ui', 'label'], presets)`

dimaip:
@bwaidelich ah so it kind of hijacks the filter menu and replaces it with tree presets menu if they are defined? I personally don't mind this, but could it be the case that people would want to have both: filters and presets?

Also left one more comment.

packages/neos-ui/src/Containers/LeftSideBar/NodeTreeSearchBar/NodeTreeFilter/index.js
bwaidelich: Ah right. But is that any better than `$get('ui.label', presets[presetName])`? Anyways, I'll change it

bwaidelich:
@dimaip I tweaked the "plow js" syntax as suggested.
I'm not sure why the CI fails. It seems to complain about a lot of unrelated things?!
Regarding the general behavior: Good point, I meant to share my reasoning but forgot to (: 
In my first attempt I combined presets and node types (the `SelectBox` wouldn't allow for separators or different rendering of the two so I used groups instead):
image
In the end I decided against this approach though because of several reasons:
- The groups use more space and the rather low level titles might be confusing to editors
- When NodeType presets are required it's probably because there is a large list of Node Types in the first place. Adding all of these on top of the presets creates a huge list leading to visual clutter and a noticeable performance hit
- Lastly, the value of the entries is the `baseNodeType`. If they overlap, you'll get weird results:
image

In general this could use a lot more sophistication (see discussion on https://github.com/neos/neos-development-collection/issues/1952) but I hope that this approach is a simple first step that won't prevent further improvements

packages/neos-ui/src/Containers/LeftSideBar/NodeTreeSearchBar/NodeTreeFilter/index.js
```diff
@@ -30,18 +32,31 @@ export default class NodeTreeFilter extends PureComponent {
...
-            icon: nodeType.ui.icon
-        }));
+        const presets = $get('configuration.nodeTree.presets', neos);
+        let options = Object.keys(presets)
+            .filter(presetName => (presetName !== 'default'))
+            .map(presetName => ({
+                value: $get([presetName, 'baseNodeType'], presets),
+                label: $get([presetName, 'ui', 'label'], presets) || '[' + presetName + ']',
+                icon: $get([presetName, 'ui', 'icon'], presets)
+                })
```
dimaip: wrong indentation

dimaip:
Thanks for taking the time to explain. Intuitively I agree with the reasoning.
Could you also create a PR for the docs?
Also would be nice to have people who actually discussed this feature to approve the concept of this PR here /cc @aertmann @hlubek @radmiraal

radmiraal:
thanks for the work, and thanks for the highlighting :-)

I think I agree to @bwaidelich that if one starts to use the presets the filters on nodetypes do not really make sense anymore. I personally don't use the filter that much because (even though it has autocomplete) it shows far too much stuff...

To me the whole presets stuff was two-fold, one being the presets, the other having the possibility to influence the tree structure itself. It's great to see the first part is now landing in this change :-) So functionality wise: awesome!

bwaidelich:
Thanks!
I added some basic documentation with https://github.com/neos/neos-development-collection/pull/2128
Adjustments to the settings schema were not necessary apparently

PR #2049, opened 2018-08-01
BUGFIX: fix changes saga again

dimaip:
Way to test: type a letter, wait 350ms for the request to start saving, type another letter while it's saving. The second request would not get triggered without this change.


kitsunet:
By reading

PR #2062, opened 2018-08-09
BUGFIX: Correctly align hidden nodes in node tree

### LINKED ISSUE #2051: Setting a "hide before" datetime breaks node alignment in tree
See screenshot - icon padding and alignment break once you set a "hide before" date.
image

h1Modeling:
**What I did**
Added node__header__icon___1KHge class to parent span of icon

**How I did it**

**How to verify it**
Hide a node in the node tree. It should now be correctly aligned with the non hidden nodes.

This fixes #2051


h1Modeling:
There is so much to learn here. I'll try to understand how it works and will come back to you hopefully with a better fix / or questions. Thanks for your reply.

mstruebing:
There is a `styles.css` mostly which is imported into the `.js` file and the webpack build is transforming these classes used by `styles.className` into these unique classnames so that no overlapping classnames exists over the whole codebase and nothing can accidentally influence other styles.

mstruebing:
Almost, I would prefer to have a `.css` file in `packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node` called `styles.css` or similiar, this would be duplicated (css)code but on the other hand your approach would increase coupling between the `react-ui-components` package and the `neos-ui` package.

Isn't it possible to add these needed styles inside the `react-ui-components` package? That would be my favourite.

h1Modeling:
Currently the span which needs the additional class is defined in the neos package. 
By your last sentence @mstruebing do you mean creating a new Component (something like ```LayeredIcon``` in the ```react-ui-components``` package?
Or do I misunderstand you?

mstruebing:
No I meant into an existing component.

But if you say that the style needs to be set on this item I think it's okay to do so :)

mstruebing:
Hey @h1Modeling great you did it!

With your change applied the Icons are aligned more to the right in my testcase (demo page)

screenshot_2018-08-12_17-03-51

As you have more looked into this issue as I am, maybe you can tell if it is possible to have the icons all aligned correctly or if there are always some downsides regarding position?

mstruebing:
Sorry for all this back and forth by your first PR :)

My feeling says option one would be better but I'm curious what @dimaip has to say to this?

dimaip:
Ah sorry missed the questions in this thread.
@h1Modeling so I think it definitly should be 1): align both icons and labels and keep the margin between them variable.

mstruebing:
I like how it looks now.

I think there is no specific reason, feel free to change it but I think thous would be a separate PR

dimaip:
Looks good to me, merging!

Thanks a lot for taking care, a lot of relief from seeing more people joining as contributors!

Issue #2057, opened 2018-08-07
White page after Discard all


samuelhauser:
If you make changes in an inline editable field and discard them all and click again in this inline editor field, while the inspector is still loading, the ui crashes with a white page.

Console Error: `Error: Key must be a string react-dom.production.min.js:188`

Seems to be similar to https://github.com/neos/neos-ui/issues/1898

![neos_ui_bug](https://user-images.githubusercontent.com/1894064/43760006-2e064dba-9a21-11e8-84d7-4995e774899d.gif)

Issue #1898, opened 2018-06-13
White page after clicking on node types during dimension loading process


mirkokaufmann:
If you switch the dimension in the backend and click during the loading process of the target dimension you will receive a white page. Editors dont wait until the page is fully loaded. The loading bar often runs through several times, so the editor dont know when its finished. After a reload the site works as aspected. 

**Tested browsers:** Firefox (60.0.2), Chrome (67.0.3396.79)

**Logs:** No exception logged

**Steps to reproduce**

1. Login into backend
2. Go to a page which exists in 2 dimensions
3. Switch the dimension
4. During the loading process click on any node type

**Error Stack**
```javascript

columnNumber: 595576
fileName: "https://www.website.loc/_Resources/Static/Packages/Neos.Neos.Ui.Compiled/JavaScript/Host.js?bust=5954d41e?538ce429"
lineNumber: 6
message: "Key must be a string"
[...]

PR #2066, opened 2018-08-13
BUGFIX: replace some exceptions with console errors to make UI feel more robust

dimaip:
Fixes: #2057 #1898

We shouldn't be so strict about reading from registries (writing is a different topic), especially since we don't have a proper type system at our disposal.
That causes all sort of runtime exceptions that hit the editors on the head (surrounding them with error boundaries is another topic).

We, the developers would still see that something is wrong just by looking at the console output.


dimaip:
:smile:

Issue #2020, opened 2018-07-19
Collapsed NodeType group isn't collapsed


mstruebing:
```
Neos:
  Neos:
    nodeTypes:
      groups:
        plugins:
          position: 200
          label: 'Neos.Neos:Main:nodeTypes.groups.plugins'
          collapsed: true
```

This doesn't render the group as collapsed.

PR #2068, opened 2018-08-13
BUGFIX: respect collapsed setting of nodetype group

dimaip:
Fixes: #2020

### DIFF
--- a/packages/neos-ui/src/Containers/Modals/SelectNodeType/nodeTypeGroupPanel.js
+++ b/packages/neos-ui/src/Containers/Modals/SelectNodeType/nodeTypeGroupPanel.js
@@ -30,7 +30,8 @@ class NodeTypeGroupPanel extends PureComponent {
         group: PropTypes.shape({
             name: PropTypes.string.isRequired,
             label: PropTypes.string.isRequired,
-            nodeTypes: PropTypes.array.isRequired
+            nodeTypes: PropTypes.array.isRequired,
+            collapsed: PropTypes.bool
         }).isRequired,
         onSelect: PropTypes.func.isRequired,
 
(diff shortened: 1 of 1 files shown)


packages/neos-ui/src/Containers/Modals/SelectNodeType/nodeTypeGroupPanel.js
```diff
@@ -57,9 +58,12 @@ class NodeTypeGroupPanel extends PureComponent {
                 return false;
             });
 
+        // Take `collapsed: true` group seting into account
```
bwaidelich: typo => "setTing"

packages/neos-ui/src/Containers/Modals/SelectNodeType/nodeTypeGroupPanel.js
```diff
@@ -57,9 +58,12 @@ class NodeTypeGroupPanel extends PureComponent {
...
+        const isOpen = $get('collapsed', group) ? collapsedGroups.includes(name) : !collapsedGroups.includes(name);
```
bwaidelich: I'm not into this one, but this line looks a bit weird to me:
The panel is *not* collapsed if `group.collapsed` is `true` and it is included in `collapsedGroups` *or* if `group.collapsed` is _not_ `true` and it is _not_ included in `collapsedGroups`.
Is that correct and does it make sense?

dimaip: Hey @bwaidelich, sorry for writing unclear code, was just lazy to do the proper refactoring.
The collapsedGroups in this case become more like toggledGroups: it takes the default value specified by the setting, and flips it. So it's basically an XNOR:
`isOpen = $get('collapsed', group) XNOR toggledGroups.includes(name)`

I've created a PR to make the code a bit more readable: https://github.com/neos/neos-ui/pull/2070

PR #2074, opened 2018-08-14
BUGFIX: fix various edgecases regarding switching dimensions

dimaip:
If user switched dimensions through inline navigation (and not dimensions selector), then switching dimensions with creating new nodes would be broken.

**Steps to reproduce**

1. Go to front page of the demo site, switch to Latvian via dimensions selector
2. Switch back to English via the site's language menu
3. Go to some other page, e.g. Downloads and switch language to Latvian via dimensions selector, click "Create and copy" (or whatever it's called)

That would end up with a failing ajax request + console errors, and nothing happening.

Also this PR fixes another problem:
1. Navigate to a different dimension via the site's language menu
2. State is not reloaded, e.g. tree doesn't have all needed nodes


dimaip:
Please don't merge yet, I have an idea

dimaip:
Nevermind

Issue #1994, opened 2018-07-09
Custom icon classnames are gone again


jobee:
### Description

Missing icons in backend

### Steps to Reproduce

1. Change a document icon via NodeTypes.yaml e.g. `icon: 'fas fa-tag super-important'`
2. Refresh the nodetree

#### Expected behavior

Nodetree icon in front of the document label incl. the custom class

#### Actual behavior

No icon is shown for the solid iconset (fas).
However the brand iconset (fab) still works, eg. `icon: 'fab fa-product-hunt super-important'`

### Affected Versions

Neos: 3.3.12

UI: >=1.1.2

mstruebing:
Hey sorry to hear that, I'm sure someone will work on this soonish  to fix this.

h1Modeling:
Hi @mstruebing, I've looked into this and also found the reason why it doesn't work. 
I'm just wondering how the functionality is supposed to be:
1. Should icons in general be provided without the prefixes? Or should both be possible?
2. Should all FA5 icons be inside the mapping file or is there a reason why there is a reduced amount in that file (on the FA webpage it lists over 1000 icons and in the mapping file there are just around 500)

I will provide a PR after I got the information. Thanks in advance.

h1Modeling:
I'll provide a PR for this...

PR #2075, opened 2018-08-16
BUGFIX: Make custom icon classNames possible again

h1Modeling:
Fixes: #1994 

**What I did**
Removed the custom className from the icon in the mapper method.

**How I did it**
.split(' ')[0]

**How to verify it**
Create a node with the icon 'fas fa-tag super-important'

### DIFF
--- a/packages/react-ui-components/src/Icon/mapper.js
+++ b/packages/react-ui-components/src/Icon/mapper.js
@@ -489,11 +489,11 @@ const mapper = icon => {
      *  otherwise the mapping would'nt work for brand icons
      */
     if (icon.startsWith('fas fa-')) {
-        icon = icon.substr(7);
+        icon = icon.substr(7).split(' ')[0];
     }
 
     if (icon.startsWith('icon-')) {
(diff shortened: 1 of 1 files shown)


dimaip:
Hey, sorry, I don't understand a bit how it works. So with this change in place if I write `fas fa-tag super-important` it would show an extra class name on the icon? Or does it just strip it and does nothing?

h1Modeling:
I hope I can clarify with the following:

Current behavior (without PR)
- User is able to add a custom className to the Icon.
- Neos currently doesn't find the icon with the fas or icon prefix if it has an additional class set (happens in the mapper function)
- It works with fab icons because the current logic in the mapper function returns then just the full icon name

Goal of the PR
- Let user still add additional classNames
- Remove the additional className in the mapper function, that Neos finds the right icon

The additional class is set by Neos because it just appends the content of the original icon Name to the className.
So this PR adds no new functionality it just tries to make the same with fas icons possible what is already possible with fab icons.

dimaip:
Ok, let's give it a try then.

Issue #2069, opened 2018-08-14
Deactivating node tree preset leaves stale nodes in tree


kdambekalns:
### Description

With configured presets (as per https://github.com/neos/neos-ui/pull/2048) limiting the shown nodes to a preset works fine. But when resetting the preset, the nodes that were shown for that preset are simply kept at the end of the "new tree content" that is loaded with the reset.

Only when reloading the tree afterwards, those nodes disappear.

### Steps to Reproduce

Limit to a preset:

<img width="158" alt="limited-to-preset" src="https://user-images.githubusercontent.com/95873/44075187-0730fbce-9f9c-11e8-9211-66af7ac597f9.png">

Click the preset reset icon:

<img width="160" alt="preset-reset" src="https://user-images.githubusercontent.com/95873/44075243-53da876a-9f9c-11e8-9c48-a06cd306328a.png">

The nodes formerly shown are still there. Reloading the tree once makes the disappear.

### Affected Versions

Neos: 4.0.7

bwaidelich:
meh, too bad.. I can't look into this one right now, but I'll check asap if noone else does before!

bwaidelich:
It seems like this bug already existed before:
When hiding certain document types per default via
```
Neos:
  Neos:
    userInterface:
      navigateComponent:
        nodeTree:
          presets:
            default:
              baseNodeType: 'Neos.Neos:Document,!Neos.Demo:Chapter'
```
They still appear in the tree when changing the filter  forth and back:
![image](https://user-images.githubusercontent.com/307571/44258849-3cead700-a210-11e8-8a15-770ca59aafe1.png)

![image](https://user-images.githubusercontent.com/307571/44258862-45dba880-a210-11e8-9737-5e6807344969.png)

![image](https://user-images.githubusercontent.com/307571/44258887-51c76a80-a210-11e8-8abe-d90872bd25fe.png)

Probably a regression of the React UI!?

PR #2076, opened 2018-08-17
BUGFIX: Avoid "stale" nodes when resetting NodeType filter

bwaidelich:
Fixes: #2069


bwaidelich:
Unfortunately that doesn't do the trick yet..

Here's what I found out and tried before:
We have a mismatch between: "no nodeType filter is active" and "use the default nodeType filter"
That's why we have the `presets.default.baseNodeType` hard coded in several places:

https://github.com/neos/neos-ui/blob/master/packages/neos-ui/src/Sagas/CR/NodeOperations/reloadState.js#L17

https://github.com/neos/neos-ui/blob/master/packages/neos-ui/src/Sagas/UI/PageTree/index.js#L42

I first tried to extend the `SelectBox` component so that you can set a `defaultOption` that is selected when the "reset button" in clicked, but that introduced several issues (i.e. the reset button was not removed, ..).
In general I think it's a good idea to separate the *selected* filter from the *effective* filter but I'm not sure where the best place for that logic would be.

With the current change `matchingNodes` already contains the correct nodes, but it still won't hide the previously filtered ones for some reason

dimaip:
Maybe it happens because you are doing merge? `yield put(actions.CR.Nodes.merge(nodes));`

bwaidelich:
@dimaip Thanks for the input but where do I do a merge and where to change this? And what you wrote is exactly the content of line 151!?

dimaip:
@bwaidelich yes, you merge at line 151. Maybe that is the cause of your troubles, so if you would set the state instead of merging it would work better?
Also, do the nodes that are fetched when filtering have the correct policy information?
Sorry, I'm not deep into how this thing works, but if you need more direct help we could chat on slack or have a hangout some time.

bwaidelich:
> you merge at line 151. Maybe that is the cause of your troubles, so if you would set the state instead of merging it would work better?

Yes. If you mean "we" because I didn't change that part of the code :)
What puzzles me: That exact same code is executed if you select a filter initially and that seems to work (it hides the other nodes).

> Also, do the nodes that are fetched when filtering have the correct policy information?

The only thing I changed is not to break if the filter is reset but set it to the default preset.


Thanks for your input. A quick chat would be great. This should be fairly easy to solve if we look at it together

dimaip:
Sorry for wrong wording, I didn't separate "you" from "we" here, so whatever I refereed to wasn't necessarily caused by any of your changes. Ping me on slack whenever you'd need me.

bwaidelich:
I think that was it: 4e4f39dd179333c828b2bc15a0cfbe78f70f3004

bwaidelich:
We experienced some buggy behavior with the "Refresh" button, but this is unrelated (see #2093). Merging this one now

kdambekalns:
@bwaidelich This could be cherry-picked to 4.0, no?

dimaip:
@kdambekalns sorry, I didn't understand your comment. We only have master branch in this wild wild west corner of Neos development.

kdambekalns:
No worries, my fault. I didn't realise in which corner I was in. :D

kdambekalns:
Still… what about a new release?

dimaip:
Will come when somebody reviews my other open PRs, they are quite important as well :)

PR #2078, opened 2018-08-20
BUGFIX: Force re generation of copied node names

sbruggmann:
Addressing issue https://github.com/neos/neos-development-collection/issues/2089

### DIFF
--- a/Classes/Domain/Model/Changes/AbstractCopy.php
+++ b/Classes/Domain/Model/Changes/AbstractCopy.php
@@ -27,11 +27,12 @@ abstract class AbstractCopy extends AbstractStructuralChange
      * Generate a unique node name for the copied node
      *
      * @param NodeInterface $parentNode
+     * @param bool $forceRegeneration
      * @return string
      */
-    protected function generateUniqueNodeName(NodeInterface $parentNode)
+    protected function generateUniqueNodeName(NodeInterface $parentNode, $forceRegeneration = false)
     {
         return $this->contentRepositoryNodeService
-            ->generateUniqueNodeName($parentNode->getPath(), $this->getSubject()->getName());
+            ->generateUniqueNodeName($parentNode->getPath(), $forceRegeneration ? null : $this->getSubject()->getName());
     }
 }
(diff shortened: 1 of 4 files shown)


dimaip:
Nice one, thanks! Would love to have a review from somebody flow-y, like @kdambekalns @kitsunet

sbruggmann:
I was unsure about this too and decided to go with the definitely-not-breaking-way only because of that uncertainty

dimaip:
I have no clue either. Could you guys decide and merge, then I'd release, as it seems to be an important bugfix.

kdambekalns:
I'd keep the option (or rather, create two methods, `generateUniqueNodeName()` and `generateRandomNodeName()`) - but either way, imagine you have imported nodes, where the name has significance. In that case you want to keep the name when moving things around, and only adjust it if needed.

Then again… **why does this not work as it is?** Shouldn't `while ($this->nodePathExistsInAnyContext(NodePaths::addNodePathSegment($parentPath, $possibleNodeName))) {…}` work around exactly the issue of conflicting node names?

kdambekalns:
Well, these are really two distinct points. My main point is: why is this change needed at all, the CR code should avoid the name clash already. No?

sbruggmann:
`generateUniqueNodeName()` with the optional `$idealNodeName` only checks against the sibling nodes.

I guess that the wrapper method `generateUniqueNodeName()` in the Neos.UI is using this `$idealNodeName` for performance reasons.
But in my opinion it would be totally fine to use `generateUniqueNodeName()` without `$idealNodeName` and re generate always.

The `$forceRegeneration` argument only solves a hypothetical backwords compatibility problem.

kdambekalns:
> only checks against the sibling nodes.

Yes, but it should be using the siblings of the *target* when copying… no?

dimaip:
I would love to have this in today. Can I merge it as is, and later you go down to the root of this and create another PR for refactoring?

sbruggmann:
@kdambekalns yes. @dimaip it's not **that** important.
The problem only appears if a copied node get's pasted to a new parent and moved back to the original one. (https://github.com/neos/neos-development-collection/issues/2089)

@kitsunet & @dimaip This PR's solution is bugfix release ready, because it's backwards compatible.
But with the next minor or major version a cleaner "re generate always" should be merged.

And to be honest, this "Node with path "node-xyz" already exists." situation should never happen and maybe Neos.UI should ask for renaming the node name in such a case?

kitsunet:
I am not sure I would count this as breaking but good point anyways :)

PR #2079, opened 2018-08-20
BUGFIX: pin ckeditor5-engine to fixed version and fix eslint output along the way

dimaip:
There's a critical bugfix in CKE5 that we desperately need.
Let's pin it till they have a new release https://github.com/ckeditor/ckeditor5-engine/commit/16b0280258fb30a9e504136410f57914f2888a96

I also was hit by the upgrade of our eslint setup along the way (that I accidentally triggered by `yarn upgrade`), so I took the opportunity to clean console output.


.nvmrc
```diff
@@ -0,0 +1 @@
+8.11.0
```
mstruebing: Why do we need that?
https://github.com/neos/neos-ui/blob/master/package.json#L37

Makefile
```diff
@@ -46,6 +46,8 @@ check-requirements:
 		(echo yarn is not installed: https://github.com/yarnpkg/yarn && false)
 
 install:
+	nvm install || true
+	nvm use || true
```
mstruebing: Like above, is this really necessary?

PR #2081, opened 2018-08-20
BUGFIX: Integer validator with trailing chars

### LINKED ISSUE #2080: Input for property with IntegerValidator is accepted, even if it only starts with an integer
### Description

When creating a node type with a property using the integer-validator any value starting with an integer is accepted as valid:

image

### Steps to Reproduce

1. Create a node type with a property using the integer-validator, e.g.:

```
minValue:
  type: string
  defaultValue: null
  ui:
    label: someLabel
    inspector:
      group: someGroup
  validation:
    'Neos.Neos/Validation/IntegerValidator': []
```

2. Enter a value starting with an integer followed by any other character

#### Expected behavior

The value is marked as invalid.

#### Actual behavior

The value is accepted as valid.

### Affected Versions

Neos: 3.3.13 LTS
UI: 1.3.2

mstruebing:
closes #2080
So - the error should be somewhere else, not in the validator itself as the test is running through.

Edit:
There is a paradox test case, can someone explain this behaviour?

Ok found it, tests are running but doesn't think these are tests, tests-runner think's tests should be one level deeper:

```
$ yarn jest -- -w 2 --coverage                                                                                                                                                                                                               
$ NODE_ENV=test jest -w 2 --coverage                                                                                                                                                                                                         
No tests found                                                                                                                                                                                                                               
  32 files checked.                                                                                                                                                                                                                          
  roots: /home/maex/projects/contrib/neos-base-distribution/Packages/Application/Neos.Neos.Ui/packages/neos-ui-validators - 32 matches                                                                                                       
  testMatch: ./src/**/?(*.)spec.js - 0 matches                                                                                                                                                                                               
  testPathIgnorePatterns: /node_modules/ - 32 matches\
```

propbably some issue with:  https://github.com/neos/neos-ui/blob/master/packages/jest-preset-neos-ui/jest-preset.json#L18

### DIFF
--- a/packages/neos-ui-validators/src/Integer/index.js
+++ b/packages/neos-ui-validators/src/Integer/index.js
@@ -7,7 +7,13 @@ import I18n from '@neos-project/neos-ui-i18n';
 const Integer = value => {
     const number = parseInt(value, 10);
 
-    if (value !== null && value !== undefined && value.length !== 0 && !Number.isSafeInteger(number)) {
+    console.log('value', value, 'should be', /^[0-9-]+$/.test(value))
+
+    if (value !== null &&
+        value !== undefined &&
+        value.length !== 0 &&
+        !(/^[0-9-]+$/.test(value)) &&
+        !Number.isSafeInteger(number)) {
         return <I18n id="content.inspector.validators.integerValidator.aValidIntegerNumberIsExpected"/>;
     }
     return null;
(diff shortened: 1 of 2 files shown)


mstruebing:
The error was indeed from the `IntegerValidator` and should be fixed now by this PR @dennisirmscher

mstruebing:
I'm interpreting the :+1: from @dennisirmscher as a positive review and therefore merge.

PR #2084, opened 2018-08-21
BUGFIX: Make sure tree search/filter behaves gracefully

kitsunet:
We should check if a Node actually has a parent, which might
not be the case for a multitude of reasons, but we should catch
that case and not error out if the parentNode was NULL.


kitsunet:
Speaking of bandaids... I think this is impossible to solve in a much better way. We just need to accept that Nodes can not have a parent sometimes... Would like to get this into the next release as well.

kitsunet:
Mmm, that test error is not from this change.

Issue #1948, opened 2018-06-29
Render resource paths images in help tooltips


mstruebing:
Currently help tooltips are not able to show resource paths and only absolute image urls.

followup form #1694

PR #2087, opened 2018-08-22
TASK: Render resource paths images in help tooltips

h1Modeling:
fixes: #1948

**What I did**
I added a helper Method which checks if the thumbnail is on a resource path and in case resolves it.
In addition I added a class to the image with style max-width: 100%

**How I did it**

**How to verify it**
Create a property with a help thumbnail.
Absolute URI and resource path should work now

bildschirmfoto 2018-08-22 um 06 31 51

### DIFF
--- a/packages/neos-ui-editors/src/EditorEnvelope/index.js
+++ b/packages/neos-ui-editors/src/EditorEnvelope/index.js
@@ -109,15 +109,24 @@ export default class EditorEnvelope extends PureComponent {
         });
     };
 
+    getThumbnailSrc(thumbnail) {
+        if (thumbnail.substr(0, 11) === 'resource://') {
+            thumbnail = '/_Resources/Static/Packages/' + thumbnail.substr(11);
+        }
+
+        return thumbnail;
+    }
+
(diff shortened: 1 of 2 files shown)


h1Modeling:
I know - Node Creation Dialog is my next one.

packages/neos-ui-editors/src/EditorEnvelope/index.js
```diff
@@ -109,15 +109,24 @@ export default class EditorEnvelope extends PureComponent {
         });
     };
 
+    getThumbnailSrc(thumbnail) {
+        if (thumbnail.substr(0, 11) === 'resource://') {
```
dimaip: Shouldn't such stuff be done server side? /cc @kitsunet @mstruebing

daniellienert:
Also a hardcoded _Resources/... Path in JS isn’t  a good idea

kitsunet:
Eh, yes, definitely, that will fail in any environment where you don't have local resources or the configruation is not the default one. Also the client side should not know about these implmeentation details at all.

Issue #1995, opened 2018-07-10
NodeTreePrivilege not working since 1.2.0


vectoholic:
### Description

Since I got the latest changes right, the privilege for any nodes are checked now by clicking on the node in the node tree. This results in showing all nodes and it doens´t hides nodes anymore.  

### Steps to Reproduce

My Policy.yaml

```
'Neos\Neos\Security\Authorization\Privilege\NodeTreePrivilege':

    'Foo.Bar:BuzAllNodes':
      matcher: 'true'

    'Foo.Bar:BuzRestricted':
      matcher: >-
        isAncestorOrDescendantNodeOf("identifier")`

 'Foo.Bar:RestrictedUser':
    parentRoles: ['Neos.Neos:RestrictedEditor']
    privileges:
      -
        privilegeTarget: 'Foo.Bar:BuzRestricted'
        permission: GRANT
```

#### Expected behavior

See only the ancestor and descendent nodes of the chosen node (identifier)

#### Actual behavior
Shows all nodes

### Affected Versions

Neos: 3.3.12

UI: 1.2.0

skurfuerst:
@kitsunet @mstruebing this might be related to you policy changes on the sprint, no? :)

kitsunet:
Is this fixed with 1.2.3?

nlx-lars:
@kitsunet I am still facing this problem with 1.2.3.

vectoholic:
I tried also 1.2.3 and the problem still exists!

PR #2090, opened 2018-08-22
BUGFIX: enable NodeTreePrivilege checking

dimaip:
Fixes: #1995

This checks NodeTreePrivilege synchronously, as it would look strange, if the UI first loads the nodes, and then they disappear once the policy is loaded...

### DIFF
--- a/Classes/Fusion/Helper/NodeInfoHelper.php
+++ b/Classes/Fusion/Helper/NodeInfoHelper.php
@@ -22,12 +22,19 @@ use Neos\Neos\Service\LinkingService;
 use Neos\Neos\Service\Mapping\NodePropertyConverterService;
 use Neos\Neos\TypeConverter\EntityToIdentityConverter;
 use Neos\Neos\Ui\Domain\Service\UserLocaleService;
+use Neos\Neos\Ui\Service\NodePolicyService;
 
 /**
  * @Flow\Scope("singleton")
  */
 class NodeInfoHelper implements ProtectedContextAwareInterface
 {
+    /**
+     * @Flow\Inject
+     * @var NodePolicyService
+     */
+    protected $nodePolicyService;
+
     /**
      * @Flow\Inject
      * @var UserLocaleService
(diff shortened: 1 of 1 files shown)


kitsunet:
Can we avoid the null return type and just return an empty array instead?

I guess optional return type is not acceptable because that would rise the PHP requirement?

Apart from that we obviously need to carefully check how bad the performance is for that...

dimaip:
Empty array instead of null is a bit less clear, e.g. you can't just `array_filter` for it. As you see I removed the return type hint for now, till we drop 3.x support.

Regarding performance, I'm relying here on the experience from my previous measurements, where NodeTreePrivilege seemed to be the least slow of all, besides here it's only 1 check instead of 5, like we had before.

Finally, what could be an alternative to this?

Sebobo:
Thanks, I tested this and it works fine for simple usecases like hiding certain NodeTypes. With a more complex matcher like isDescendant or something I get errors like this:

```
Exception #1297759968 in line 210 of /MyExample/Data/Temporary/Development/SubContextLocal/Cache/Code/Flow_Object_Classes/Neos_Flow_Mvc_Controller_Argument.php: Could not convert target type "array<Neos\ContentRepository\Domain\Model\NodeInterface>", at property path "0": Only absolute paths are allowed for Context::getNode()


Previous exception: Exception #1284975105 in line 161 of /MyExample/Data/Temporary/Development/SubContextLocal/Cache/Code/Flow_Object_Classes/Neos_ContentRepository_TypeConverter_NodeConverter.php: Only absolute paths are allowed for Context::getNode()
    [array] => 
        previousException:
        [string] => Exception #1284975105 in line 161 of /MyExample/Data/Temporary/Development/SubContextLocal/Cache/Code/Flow_Object_Classes/Neos_ContentRepository_TypeConverter_NodeConverter.php: Only absolute paths are allowed for Context::getNode()
```

But I assume this is not related to this PR.

PR #2091, opened 2018-08-23
TASK: Force re generation of copied node names - Cleanup

sbruggmann:
Make https://github.com/neos/neos-ui/pull/2078 permanent but without backward compatibility.

This change is _hypothetically_ braking, but definitely not critical.


dimaip:
Merging, don't think it should have any consequences...

Issue #2096, opened 2018-08-24
Command make lint-editorconfig does not work on Mac 


markusguenther:
### Description

Like lerna in the past the command editorconfig-checker can not be found. Which leads to the problem that the command ```make lint-editorconfig``` does not work.

Had currently the problem that locally the listing was fine and on travis I had errors. The reason for that is the missing command ;)

### Steps to Reproduce

1. Go to neos.ui root folder
2. exec ```make lint-editorconfig```

#### Expected behavior

Checking the config reporting possible errors.

#### Actual behavior

Error message:
```
make: editorconfig-checker: No such file or directory
make: *** [lint-editorconfig] Error 1
```

<img width="588" alt="bildschirmfoto 2018-08-24 um 08 08 39" src="https://user-images.githubusercontent.com/1014126/44568162-df0d4080-a775-11e8-9c03-de48f621a6f7.png">

### Affected Versions

Neos: 4.x

UI: 1.3.x

PR #2097, opened 2018-08-24
BUGFIX: Execute make lint-editorconfig also on mac

markusguenther:
The command ```make lint-editorconfig``` throws an error:

```
make: editorconfig-checker: No such file or directory
make: *** [lint-editorconfig] Error 1
```

Had this problem already in the past with lerna. So I choose the same approach to fix it.

Fixes: #2096


mstruebing:
Can someone explain to me what's wrong with mac and make? 
`lerna` wasn't also in the $PATH ... but it should be

markusguenther:
@mstruebing I have no clue but I am not that deep into that dev ops stuff.

PR #2099, opened 2018-08-24
BUGFIX: Prevent backend inside of client frame in error case

kitsunet:
The old backend would show up inside the client frame in case a
rendering exception happened. The reason is the PageHandler in Neos
which just renders a (legacy) backend around the error message because
that was the only way the legacy backend could still work in case
of an error. This leads to the backend in backend problem and can be
avoided by a custom error handler for the new UI that replaces the
old one.

### DIFF
--- a/Classes/Fusion/ExceptionHandler/PageExceptionHandler.php
+++ b/Classes/Fusion/ExceptionHandler/PageExceptionHandler.php
@@ -4,6 +4,7 @@ namespace Neos\Neos\Ui\Fusion\ExceptionHandler;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Http\ContentStream;
 use Neos\Flow\Http\Response;
+use Neos\Flow\Mvc\View\ViewInterface;
 use Neos\Flow\Utility\Environment;
 use Neos\FluidAdaptor\View\StandaloneView;
 use Neos\Fusion\Core\ExceptionHandlers\AbstractRenderingExceptionHandler;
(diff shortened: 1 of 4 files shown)


Classes/Fusion/ExceptionHandler/PageExceptionHandler.php
```diff
@@ -0,0 +1,98 @@
...
+     * Handle an exception by displaying an error message inside the Neos backend, if logged in and not displaying the live workspace.
+     *
+     * @param string $fusionPath path causing the exception
+     * @param \Exception $exception exception to handle
+     * @param integer $referenceCode
+     * @return string
+     * @throws \Neos\Flow\Mvc\Exception\StopActionException
+     * @throws \Neos\Flow\Security\Exception
+     */
+    protected function handle($fusionPath, \Exception $exception, $referenceCode)
```
gerhard-boden: maybe add type hints?

Classes/Fusion/ExceptionHandler/PageExceptionHandler.php
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    /**
+     * Renders an actual HTTP response including the correct status and cache control header.
+     *
+     * @param \Exception the exception
+     * @param string $bodyContent
+     * @return string
+     */
+    protected function wrapHttpResponse(\Exception $exception, $bodyContent)
```
gerhard-boden: maybe here as well for the string and also the scalar return type

Classes/Fusion/ExceptionHandler/PageExceptionHandler.php
kitsunet: This is inherited from the abstract so I wouldn't change the signature here.

Classes/Fusion/ExceptionHandler/PageExceptionHandler.php
kitsunet: Right, I copied this from the Neos implemenation, btu we can definitely be more strict here.

kitsunet:
Add some typehints according to @gerhard-boden comments.

Issue #2103, opened 2018-08-25
Update react-css-themr


markusguenther:
### Description

The issue #2102 need the latest release of @friendsofreactjs/react-css-themr.

Issue #2102, opened 2018-08-24
TypeScript integration


dimaip:
This guide should help along the way: https://github.com/piotrwitek/react-redux-typescript-guide

TODO:
- [ ] https://github.com/FriendsOfReactJS/react-css-themr/pull/33
- [ ] setup tslint rules to follow our eslint style
- [ ] and typings to plow-js (@grebaldi could you help with this?)
- [ ] convert react-ui-components
- [ ] convert the state structure, reducers and sagas
- [ ] all other stuff

markusguenther:
Do we handle this in a different branch? Or will it be merged in master as well?

mstruebing:
master! :) - no risk no fun - even if this all in all will reduce risks ;)

markusguenther:
hehe ok ... I currently retry the Webpack update. After merging als is broken ;) So will try it again

mstruebing:
I think just updating the dependencies will not work in this case, this will need a bit more adjustments inside the configuration I think.

dimaip:
I'd do the TypeScript integration in small steps and in master.
E.g. some steps are quite painless, as e.g. converting leaf components, but converting the state should probably be done in one step, maybe during next sprint.

markusguenther:
Then I need to read bit more about TS and try something until the sprint ;)

dimaip:
That'd be awesome! E.g. this is example of a simple task: https://github.com/neos/neos-ui/pull/2106

markusguenther:
Can save time when we skip tslint stuff and use https://github.com/nickmccurdy/eslint-config-react-app-ts in the eslint configuration repository... what do you think?

dimaip:
I'm totally not opinionated about linters and stuff, the less I have to think about it the better. Though maybe somebody else can say something meaningful on the subject.

PR #2104, opened 2018-08-25
TASK: Update @friendsofreactjs/react-css-themr

markusguenther:
Now using version 3.3.2 for possible typescript usage.
The issue #2102 needs the TS compatible version of @friendsofreactjs/react-css-themr. So this updates to the latest version.

Resolves: #2103


dimaip:
Thank you!

Issue #2086, opened 2018-08-21
Is nodetype select dialog has scrolling, the help message is hidden below under scroll


dimaip:
![image](https://user-images.githubusercontent.com/837032/44404561-d17e6d80-a55f-11e8-9cfd-a2b38ca18b2a.png)

h1Modeling:
I can look into this (maybe this weekend). Just to clarify: What is the actual expected behavior? I expect the help message should be sticky at the bottom of the panel. Is this correct? Similar to the below image?

<img width="906" alt="bildschirmfoto 2018-08-23 um 20 48 51" src="https://user-images.githubusercontent.com/17990673/44545782-4abfc100-a716-11e8-86fe-3032021a728a.png">

dimaip:
@h1Modeling yep, that would be one way to do it.
In the old UI it was floating in the tooltip, but I never liked how it worked.

Issue #2088, opened 2018-08-22
Images in help tooltip from Node creation dialog


mstruebing:
Help tooltips in theNodeCreationDialog are not be able to render images currently (absolute path and resource uris).
As this was the case in the old ui this should be reimplemented.

followup from https://github.com/neos/neos-ui/issues/1694 and https://github.com/neos/neos-ui/issues/1948

markusguenther:
@mstruebing @dimaip Is this done or do we need to polish this at some point?

dimaip:
Uhm I'm not sure, probably #2092 is the way to go.

grebaldi:
Closing this. Help tooltips are well established now and I don't see what this issue would be waiting for. #2092 still needs to be fixed though.

PR #2105, opened 2018-08-25
TASK: Rework Help Message in SelectNodeType Dialog

h1Modeling:
**What I did**

- Added support for translation

- Moved the help message from the bottom of the dialog to the bottom of the NodeType Groups

- Added support for thumbnails (currently just absolute paths - resource path resolving has to be done inside the neos package)

- Removed the fixed height of the message to improve display of thumbnails

**How I did it**

- Moved the help message part from index.js to nodeTypeGroupPanel.js

- Added an additional class to the Dialog component to query select the dialog body
- modified help css

**How to verify it**
- add help properties to a nodeType in your yaml (message + thumbnail with absolute path)

ezgif com-video-to-gif-2

This fixes: #2086
and partly: #2088

### DIFF
--- a/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
+++ b/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
@@ -2,13 +2,10 @@ import React, {PureComponent} from 'react';
 import PropTypes from 'prop-types';
 import {connect} from 'react-redux';
 import {$get} from 'plow-js';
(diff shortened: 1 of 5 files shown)


dimaip:
@aertmann how do you feel about this UX-wise?

mstruebing:
Ping, what to do about this?

dimaip:
I'm not confident about the UX changes, but in the worst case we can just merge it as it, as it's definitely better than it was before.

PR #2108, opened 2018-08-27
EXPERIMENT TASK: convert System reducer to TS

dimaip:
This is my experiment for creating a strongly-typed reducer.
I wasn't able to make it work with immutablejs+plow-js (see https://stackoverflow.com/questions/52035047/typescript-deep-traversal-of-plain-and-immutablejs-objects), but seamless-immutable+plow-js seems to work.

This PR touches only one small part of the state. Eventually all reducers should be converted to TS and then we can automatically build out the type for the full state.

The next topic to figure out what be the runtime boundaries for json that we're getting on initialization and from AJAX requests.


kdambekalns:
I think this is way outdated…

PR #2112, opened 2018-08-30
FEATURE: Remember collapsed state of menu groups

aertmann:
**What I did**
Implemented state handling with local storage persisting for menu group collapsed states.

**How to verify it**
Check that the collapsed state of the menu groups is kept after re-opening the menu, even after reloading the UI.

### DIFF
--- a/packages/neos-ui-redux-store/src/UI/Drawer/index.js
+++ b/packages/neos-ui-redux-store/src/UI/Drawer/index.js
@@ -1,5 +1,5 @@
 import {createAction} from 'redux-actions';
-import Immutable from 'immutable';
+import {Map} from 'immutable';
 import {$toggle, $get, $set} from 'plow-js';
 
 import {handleActions} from '@neos-project/utils-redux';
@@ -8,13 +8,15 @@ import {actionTypes as system} from '../../System/index';
 
 const TOGGLE = '@neos/neos-ui/UI/Drawer/TOGGLE';
 const HIDE = '@neos/neos-ui/UI/Drawer/HIDE';
+const TOGGLE_MENU_GROUP = '@neos/neos-ui/UI/Drawer/TOGGLE_MENU_GROUP';
 
 //
 // Export the action types
 //
 export const actionTypes = {
     TOGGLE,
-    HIDE
+    HIDE,
+    TOGGLE_MENU_GROUP
 };
 
 /**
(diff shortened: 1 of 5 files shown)


packages/neos-ui-redux-store/src/UI/Drawer/index.js
```diff
@@ -27,24 +29,34 @@ const toggle = createAction(TOGGLE);
...
 // Export the reducer
 //
 export const reducer = handleActions({
-    [system.INIT]: payload => $set(
+    [system.INIT]: state => $set(
         'ui.drawer',
-        Immutable.fromJS($get('ui.drawer', payload) ? $get('ui.drawer', payload) : {isHidden: true})
+        new Map({
+            isHidden: $get('ui.drawer.isHidden', state) ? $get('ui.drawer.isHidden', state) : false,
+            collapsedMenuGroups: $get('ui.drawer.collapsedMenuGroups', state) ? $get('ui.drawer.collapsedMenuGroups', state) : ['content']
```
aertmann: content menu group collapsed by default, similar to the old UI

packages/neos-ui-redux-store/src/UI/Drawer/index.spec.js
```diff
@@ -68,3 +70,34 @@ test(`The "hide" action should set the "isHidden" key to "true".`, () => {
...
+    const nextState4 = reducer(nextState3, actions.toggleMenuGroup('management'));
+
+    expect(nextState1.get('ui').get('drawer').get('collapsedMenuGroups').includes('content')).toBe(true);
+    expect(nextState2.get('ui').get('drawer').get('collapsedMenuGroups').includes('content')).toBe(true);
+    expect(nextState2.get('ui').get('drawer').get('collapsedMenuGroups').includes('management')).toBe(true);
+    expect(nextState3.get('ui').get('drawer').get('collapsedMenuGroups').includes('content')).toBe(false);
+    expect(nextState3.get('ui').get('drawer').get('collapsedMenuGroups').includes('management')).toBe(true);
+    expect(nextState4.get('ui').get('drawer').get('collapsedMenuGroups').includes('content')).toBe(false);
+    expect(nextState4.get('ui').get('drawer').get('collapsedMenuGroups').includes('management')).toBe(false);
+});
```
aertmann: not sure if these actually pass, running `make test` passes but doesn't seem to run the actual tests?

mstruebing: There is a similar case where tests don't seem to run correctly.

dimaip: So, I guess we need to fix this before we can merge this PR, right? Any idea what might be the problem?

packages/neos-ui-redux-store/src/UI/Drawer/index.js
dimaip: But why did we do this before? Doesn't make a lot of sense to have Content collapsed by default to me. Also makes the code more ugly by having that hardcoded case for `content`.

aertmann: Two reasons. First because those links are not used very often and for single site installations, the link is simply a distraction. For multi site installations the site switching menu gets in the way of the more important and more frequently used menu items. E.g. a client of mine has 22 sites causing the menu to be very annoying to use.

aertmann: As for the hardcoding, that's definitely true, we make it configurable so it would be possible to define if a module group should be collapsed by default and then determine it in the `MenuHelper` instead. But don't see that as a big issue to be honest, it's already a hardcoded group in the `MenuHelper`.

packages/neos-ui-redux-store/src/UI/Drawer/index.spec.js
aertmann: No idea, never ran the tests previously so no idea how it's supposed to work.

aertmann:
@dimaip: answered your comments, will rebase after your response to them

packages/neos-ui-redux-store/src/UI/Drawer/index.js
dimaip: Makes sense. Could be left for further improvements ofc.

dimaip:
@aertmann @mstruebing if none of you has the time to investigate the test issue right now we could merge this PR and create a separate issue for fixing the tests.

aertmann:
Fixed merge conflict.

@dimaip: could look into it later this week, but have no clue how the tests are set up and supposed to look/run so would be quite inefficient for me to try to figure out why they're broken or when it happened

dimaip:
@aertmann well frankly I don't have any specific knowledge, I wasn't the one who set it up, but could try to look into it as well. Let's merge this one now.

aertmann:
Thanks @dimaip! Definitely let me know if these tests fail and I'll fix them right away when it's working.

PR #2115, opened 2018-09-03
BUGFIX: Prevent NodeToolbar from crashing if focused node not in DOM

suffle:
If the document node was changed and the old focused element was still in the store (it is a racing condition between set_src and set_context_path actions), the NodeToolbar crashed because it could not find the old elements dimensions.

I added conditions to the InlineUI to check if the focused element is in the GuestframeWindow and also to the NodeToolbar to check if the nodeElement is set. I also updated the forceUpdate- and setState-calls to be cancelled if the component was unmounted during the debounce-delay

### DIFF
--- a/packages/neos-ui-guest-frame/src/InlineUI/index.js
+++ b/packages/neos-ui-guest-frame/src/InlineUI/index.js
@@ -4,7 +4,6 @@ import {connect} from 'react-redux';
 import {$transform, $get, $contains} from 'plow-js';
 import {actions, selectors} from '@neos-project/neos-ui-redux-store';
 import {neos} from '@neos-project/neos-ui-decorators';
-import {findNodeInGuestFrame} from '@neos-project/neos-ui-guest-frame/src/dom';
 import NodeToolbar from './NodeToolbar/index';
 
 import style from './style.css';
@@ -49,7 +48,7 @@ export default class InlineUI extends PureComponent {
         const canBeEdited = $get('policy.canEdit', this.props.focusedNode) || false;
         const visibilityCanBeToggled = !$contains('_hidden', 'policy.disallowedProperties', this.props.focusedNode);
 
-        return (findNodeInGuestFrame(focusedNodeContextPath) &&
+        return (
             <div className={style.inlineUi} data-__neos__inline-ui="TRUE">
                 <NodeToolbar
                     shouldScrollIntoView={shouldScrollIntoView}
--- a/packages/neos-ui-guest-frame/src/InlineUI/index.js
+++ b/packages/neos-ui-guest-frame/src/InlineUI/index.js
@@ -4,6 +4,7 @@ import {connect} from 'react-redux';
 import {$transform, $get, $contains} from 'plow-js';
 import {actions, selectors} from '@neos-project/neos-ui-redux-store';
 import {neos} from '@neos-project/neos-ui-decorators';
+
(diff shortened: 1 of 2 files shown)


suffle:
I changed the implementation and also removed the eventListeners. Looks much cleaner like this but wasn't really sure if keeping the functions in a class property has an impact memory-wise

dimaip:
No no, it's totally fine and that's how we usually do it in general with debounced functions.
Great job!

packages/neos-ui-guest-frame/src/InlineUI/index.js
```diff
@@ -48,6 +48,8 @@ export default class InlineUI extends PureComponent {
         const canBeDeleted = $get('policy.canRemove', this.props.focusedNode) || false;
         const canBeEdited = $get('policy.canEdit', this.props.focusedNode) || false;
         const visibilityCanBeToggled = !$contains('_hidden', 'policy.disallowedProperties', this.props.focusedNode);
+        // Check if focusedNode is in Guest frame to prevent errors (e.g. if still in store)
+        const focusedNodeInGuestFrame = findNodeInGuestFrame(focusedNodeContextPath) ? focused : {};
```
dimaip: Just one last question. Shouldn't we just `return null` in case there's no `focusedNodeInGuestFrame`, instead of providing it with a dummy empty object instead of the node?

suffle: Good point, if we dont have the toolbar, we dont need the container

packages/neos-ui-guest-frame/src/InlineUI/index.js
```diff
@@ -48,10 +48,8 @@ export default class InlineUI extends PureComponent {
         const canBeDeleted = $get('policy.canRemove', this.props.focusedNode) || false;
         const canBeEdited = $get('policy.canEdit', this.props.focusedNode) || false;
         const visibilityCanBeToggled = !$contains('_hidden', 'policy.disallowedProperties', this.props.focusedNode);
-        // Check if focusedNode is in Guest frame to prevent errors (e.g. if still in store)
-        const focusedNodeInGuestFrame = findNodeInGuestFrame(focusedNodeContextPath) ? focused : {};
 
-        return (
+        return (findNodeInGuestFrame(focusedNodeContextPath) &&
```
dimaip: Hey, sorry for being picky, but isn't it enough to just check for presence of `focused`? Why do we need to find it in the guest frame?

suffle: Being picky is the way to a good product, so keep challenging me ;) The focussed element is reset in another saga as the change of the iframe src which can lead to the new content rendering without the focussed node being present thus leading to errors. This is also checked in the toolbar itself now so we could skip the check here but would end up with an empty container in the dom, since the toolbar returns null.

suffle: Removed that as well. I tried to figure out what the inlineUI-Component is for in difference to the Toolbar. I thought i figured it out seeing the Toolbar as a presentational component (only containing UI state) and the InlineUI-Component as a container with connection to the redux store...But if I'm right with that, shouldn't the decision, if the Toolbar is rendered, be made in the container-Component getting the relevant nodeElement there as well? Or did I find a pattern, that wasn't really intended?

dimaip:
Can I merge? All good?

PR #2116, opened 2018-09-05
BUGFIX: prevent selecting or uploading wrong image types in Image editor

dimaip:
There were two faulty scenarios:
- from some Image editor, go to media library and pick a .pdf (or other non-image file). A scary exception would be thrown.
- from some Image editor, click upload button, change allowed file extension in file picker to all files, pick pdf file: the spinner would be infinitely loading.

With this change:
- if you go to media library from Image editor, it would only be possible to pick images, and not just any assets.
- if you fool the file picker to picking a wrong filetype, it will be rejected correctly.


drlimbo:
It think the problem isn't fully fixed yet. 
Problem: I can choose non-image files (PDF) in Media Browser and get an exception
Neos: 4.1.7 / Neos-UI: v1.3.4

tell me if i can help you find the problem

`
teaserImage:
  ui:
    label: 'Teaser Image'
    inspector:
      group: 'settings'
      editorOptions:
        accept: 'image/png'
`

mediabrowser

Issue #1682, opened 2018-03-02
Dynamic configuration doesn't work for group or editor


kitsunet:
The feature #1618 does not work for setting the inspector group or editor (as the example in the PR shows) because that part of the configuration is extracted before the processing happens. It would be nice though to be able to do that.

kitsunet:
@dimaip This is it :)

dimaip:
I looked into this a while ago, and it's not easy to fix without huge render performance penalty.
I propose to not support moving a property dynamically between groups, but instead to add an additional option to hide and editor, e.g.
```
  ui:
    inspector:
      hidden: true
```

dennisirmscher:
@dimaip Would love to have something like this for dynamically show properties. I have to deal with circumstances with somehow depending properties quite often and it would be charming and less confusing for editors to display only the needed ones.

dimaip:
@dennisirmscher wouldn't this help https://github.com/neos/neos-ui/issues/1682#issuecomment-386269035?

dennisirmscher:
@dimaip It totally would! :)

dimaip:
@dennisirmscher I think it would be quite trivial to implement here: https://github.com/neos/neos-ui/blob/master/packages/neos-ui-editors/src/EditorEnvelope/index.js#L146
E.g. check for `hidden` and then just `return null`.

dennisirmscher:
All right, this is working so far. I added a property `hidden` to the inspector like this (shortened):

```
'foo':
  type: boolean
'image':
  ui:
    inspector:
      hidden: 'ClientEval:node.properties.foo ? false : true'
```

The property image now disappears if `foo` is set to false:

![dynamic configuration](https://user-images.githubusercontent.com/16836984/45105453-8f932100-b134-11e8-9f36-981d12943eb4.gif)

Of cause the value of the property `image` won't get removed as it only get's removed from the inspector. I don't think it would be a god idea to the delete the value but let the developer take care of dealing with the properties.

@dimaip wanted to discuss the naming of the property as `hidden` might not be the best option. Any other suggestions?

dimaip:
Yeah it definitely shouldn't mess with the value.
Next steps:

- [ ] create a PR
- [ ] create a documentation PR in https://github.com/neos/neos-development-collection/
- [ ] check if it's needed to adjust the NodeTypes schema for this property

dennisirmscher:
> The feature #1618 does not work for setting the inspector group or editor

[...]

PR #2118, opened 2018-09-06
FEATURE: Dynamically hide properties in the inspector

dennisirmscher:
Resolves: #1682

**What I did**

Added a property `hidden` to the inspector configuration for dynamically hiding editors in the inspector. See issue #1682.

**How to verify it**

For example add two properties to a Node Type, e.g.:

```
foo:
  type: boolean
  defaultValue: true
  ui:
    label: 'A totally unnecessary property'
    inspector:
      group: 'image'
image:
  ui:
    label: 'Image'
    inspector:
      group: 'image'
      hidden: 'ClientEval:node.properties.foo ? true : true'
```

The editor for the property image will only be shown if `foo` ist set to `true`:

45105453-8f932100-b134-11e8-9f36-981d12943eb4

### DIFF
--- a/packages/neos-ui/src/Containers/RightSideBar/Inspector/PropertyGroup/index.js
+++ b/packages/neos-ui/src/Containers/RightSideBar/Inspector/PropertyGroup/index.js
@@ -48,7 +48,7 @@ export default class PropertyGroup extends PureComponent {
                         const itemType = $get('type', item);
                         const label = $get('label', item) || '';
 
-                        if (itemType === 'editor') {
+                        if (itemType === 'editor' && $get('hidden', item) !== true) {
                             return (
                                 <InspectorEditorEnvelope
                                     key={$get('contextPath', node) + itemId}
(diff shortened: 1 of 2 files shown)


packages/neos-ui/src/Containers/RightSideBar/Inspector/PropertyGroup/index.js
```diff
@@ -48,7 +48,7 @@ export default class PropertyGroup extends PureComponent {
                         const itemType = $get('type', item);
                         const label = $get('label', item) || '';
 
-                        if (itemType === 'editor') {
+                        if (itemType === 'editor' && $get('hidden', item) !== true) {
```
dimaip: hmm I'd prefer if it worked not only with editors but also with views.
Alternative, add before this line: 
```
if ($get('hidden', item)) {
    return null;
}
```

dennisirmscher: All right, I'll take care of it.

dimaip:
Hey, I fixed a negated condition in your code, the way you had it would only show hidden peoperties.

dennisirmscher:
Thanks again for your awesome support!

dimaip:
Been my pleasure, thanks for the great contribution!

jonnitto:
Great feature, thanks! Is it also possible to add multiple properties?
`'ClientEval:node.properties.propertyone ? true : ClientEval: node.properties.propertytwo'`
or 
`'ClientEval:node.properties.propertyone || ClientEval: node.properties.propertytwo'` 
both seems not to work

jonnitto:
Would be nice to have this also in the groups. (`ui.inspector.groups`)

PR #2119, opened 2018-09-06
BUGFIX: fix focusing nodes in FF

dimaip:
For some weird reason FF returns empty composedPath for event which happened in the past and were stored in a closure.
Let's be safe and just use all parents, it's super quick anyways.


dimaip:
Merging without reviews as it's simple and rather urgent. Feel free to post-review.

PR #2120, opened 2018-09-07
TASK: add legacy aloha format migration and support both formats at the same time

dimaip:
This PR provides a core migration from old Aloha NodeTypes.yaml format to the new one, but also allows both formats to exist at the same time.

The old aloha format and the new format are supported at the same time, the new format takes precedence.


kdambekalns:
Moved from PR comment:

---

**How to verify**

The migration should convert the old format to the new one:

```bash
./flow flow:core:migrate Neos.NodeTypes.BaseMixins --version 20180907103800
```

PR #2123, opened 2018-09-08
BUGFIX: validate undefined properties and fix the unit tests setup

dimaip:
Fixes: #1651
Resolves: #1509

Currently we only validate those properties that are already set, while we should validate all properties.


JamesAlias:
If my opinion is wanted: I don't see any error.
Only the expected misbehaviour of some components that render `undefined` as className.
So, on the test side, everything seems to check out.

I am eager to have the testing working again, so I can test my PR as well 😉

I did not test your PR on my machine though.

dimaip:
Thanks! As this PR is huge we'd probably be able to detect more errors on master, so merging.

Issue #2127, opened 2018-09-10
Copying nodes acorss dimenions and websites


davidspiola:
### Description

This feature is about improving multi dimension and website support within the UI. It would be quite helpful for editors to have the same/similar tooling the have right now for a multi site environment as well. 

Since both websites share the same root path `sites` it would be great to use the existing implementation and extend it in a way, that you can past into other websites and dimensions. I had a short discussion with Christian Müller about it and he said, that is should work out from the CR perspective. 

Important to know is that the coping target **does't** have to be necessarily within the some dimension, and probably only coping of nodes does make sense IMHO.

dimaip:
Hey, so I had a quick look into the issue.
First of all, it is possible to copy nodes across dimensions on master. It's been implemented recently.
Now comes the problem of copying nodes across sites. It is easy to set current clipboard path in local storage, the way we do with other persisted settings... But the problem is that local storage is scoped to the current domain only, so you can't share it between sites.

Let's discuss possible solutions:

1. Persist some part of this state in user session data on PHP side. Would complicate things a little, but should be doable.
2. Use some hacky cross-domain cookie or local storage solution. Please let's not.
3. Any better ideas? Maybe @aertmann or @kitsunet have ideas?

dimaip:
I tried to do it the first way, see the PR

PR #2131, opened 2018-09-12
FEATURE: cross-site pasting of nodes

dimaip:
Resolves: #2127

This change stores clipboard path into Flow's session storage.
This allows pasting nodes across sites.

In addition to this, on paste the clipboard doesn't get cleared. IMO this is actually how it should have worked in the first place: when you copy the node, you should be able to paste it multiple times (the way it works in a text editor).


davidspiola:
Hi Dmitry, I checked the implementation and it basically works - awesome. But it might need some refinement.

- If you are logged in to site A and separately on site B you don't have the same session id and it does not work. On the other hand, if you use the site switcher is auth token authentication it works. 
- Once you have the same session an you copy something (even across dimensions) it works, but you have the refresh the target web site Ui once in order to paste it.

davidspiola:
One more thing. It looks like that the node name remains the same after copy. This could lead to problems. I guess it's related to https://github.com/neos/neos-ui/commit/167d39049e2d49fc88fe8986f46f6ed5cfe0de11

dimaip:
Yeah I kind of supposed that the editors would use the site menu when navigating across websites.
Alternatively we could store the state in the user's persistent record, but that would be a bit too much.
Do you think the current workflow of going through the site menu would be too limiting for editors? I'm just afraid that the other possible solutions are too monstrous technically...

davidspiola:
hmm.. I guess that users will be confused by the fact that they can't past it. Maybe it's somehow possible to end up with the same session id if you are already logged in with a other domain. Do you know who would be a good source to answer that question?

dimaip:
@davidspiola even if you end up in the same session id, users would still need to refresh the page in that scenario. Don't you think that'd be even more confusing?
I think the most obvious workflow is that you copy the node, navigate to a different website via the menu, paste it.
Regarding session ids, I guess anyone Flow-competent could answer that. Frankly I'm surprised that you get different sessions when logging into the same Neos instance, sounds more like a bug.

davidspiola:
@dimaip ok, I agree! I guess it's fine that you have to navigate within Neos to support the feature. Regarding the session Id thingy, I will check if it's a bug.  Beside that, I'd say that the feature works and could be merge after reviewing, do you agree?

davidspiola:
I'm a little bit lost now. I spoke with some editors about it, but I'm not sure if this proides me a complete picture. How would you get more input? The general feedback I have right now is that as long as the documentation is clear about it, it's fine. Having two browser windows open at the same time is simply not supported right now. What do you think?

dimaip:
@davidspiola I'd say let's keep it simple and do it least complicated way. If we'd get additional feedback, we can always improve this behavior later.

bwaidelich:
I'm a bit concerned with this change to be honest for a few reasons:
* Another generic `PersistedStateContainer` – it's not clear what it does and to what context it is bound (as you can tell from the discussions above)
* The state is an untyped array. This allows for unplanned extension but is very hard to maintain and to use if it grows
* It relies on the server session (which I personally try to avoid and I started to work on session less authentication for Neos).

Isn't there a way to store the node ids to be copied in the client somehow? I guess `sessionStorage` won't work due to cross domain restrictions, but maybe there's another way?

If not, I'd prefer a very strict API along the lines of:
```
POST /neos/ui-services/clipboard
Content-Type: application/json

{"nodeIdentifiers": ["<id1>", "<id2>"]}
```

dimaip:
Hi @bwaidelich! I'm actually quite happy to hear your concerns, because I have the same, just don't know how to do it better :)

> Isn't there a way to store the node ids to be copied in the client somehow?

Cross-domain local storage is a really messy topic. Basically we'd need an iframe embeded on every page that would communicate with the default domain and synchronize the storage... It felt like an overkill for such simple feature... On the other hand it would allow copy/pasting without reloading the page... Don't know if we should go that path.
How do you feel about that? Should we go the hacky client-side solution or this hacky session-based solution?

bwaidelich:
> Cross-domain local storage is a really messy topic

OK, I see, thanks for clarifying.

> Should we go the hacky client-side solution or this hacky session-based solution?

Is that a trick question? I still hope for a no-hack-solution :)

Maybe just a friday-evening-brain-hicup.. but what about using the [Clipboad API](https://developer.mozilla.org/en-US/docs/tag/Clipboard%20API)?

If there is no clean client-side solution we should at least make the API stricter to the usecase (copy/cut & paste nodes) so that we can reuse it in other places and replace the persistence mechanism more easily

bwaidelich:
> how would validate its contents across different installations?

What do you mean? The user has to have the privileges to add these nodes of course and we most probably need to store more than just their identifier (for example the site name, domain, ..) but apart from that it should be the same mechanism as with the server-site persistence

dimaip:
@bwaidelich Domain name won't do, as we don't have the exhaustive list of domain lists valid for this installation. If only we had a unique identifier for each installation...

dimaip:
> The state is an untyped array. This allows for unplanned extension but is very hard to maintain and to use if it grows

I did the refactoring to address that, please review.

Though in all honesty I had mixed feeling about this. My initial intuition was to decrease the coupling between JS and PHP code so that the backend would provide a generic persistence implementation without any domain knowledge, that could be used for other purposes later. Anyways, that's a long discussion, which is probably caused by my lack of software architecture skills, so I'd rather ask you personally about this when we'd have the time to talk about it together.

dimaip:
So yes, I don't think that clipboard api is a viable solution: you need to request permissions to access it, there's no proper way to identify the current Neos instance etc.

> It relies on the server session
I think when we'd really need to go session-less we could save the clipboard API into some other kind of persistence, e.g. either a DB table or a persistent cache backend.

bwaidelich:
> Domain name won't do

Right. Technically a list of node identifiers would work. If they don't exist in the current installation (or the user hasn't got the required privileges) they would not be pasted.
But to make it easier to debug, the site name of the nodes could be copied, too.

> I don't think that clipboard api is a viable solution: you need to request permissions to access it

right, that's probably a blocker for using the API directly right now.
draw.io managed to work around this limitation, copy&paste works even across browsers there. Probably a similar solution to the one from [lucidchart](https://www.lucidchart.com/techblog/2014/12/02/definitive-guide-copying-pasting-javascript/).

> there's no proper way to identify the current Neos instance

That's not an issue IMO.

Thank you for the new implementation. I can't look into it right now because there's too much on my table atm, but some comments regarding your mixed feelings:

> My initial intuition was to decrease the coupling between JS and PHP code so that the
> backend would provide a generic persistence implementation without any domain
> knowledge, that could be used for other purposes later

I can comprehend the intuition but I'm convinced that this kind of generic API leads to many of the issues we currently have. For example it's tempting to expose the FlowQuery API to JavaScript but with that step you lose all control – like you would when exposing a SQL API to JavaScript (fair enough, that example is a bit extreme ;)
It's easier to broaden an API lateron if we need more features. The other way is impossible without breaking changes.

dimaip:
Hey thanks for taking the time to reply, appreciate your attention to this topic, will be waiting for your code review!

We can continue the general discussion at the sprint, but can't resist to reply here.

> For example it's tempting to expose the FlowQuery API 
Hah, funny, that's exactly what we did :D Besides, it's somewhat similar to the idea of GraphQL and the like, in other words breaking the borders between FE and BE.

> It's easier to broaden an API lateron if we need more features
I often observe a different kind of threat, when new feature requests pop up, more and more ad-hoc solutions will be created, and often nobody takes the courage to do the refactoring to generalize the implementations.

Classes/Controller/BackendServiceController.php
```diff
@@ -327,6 +333,19 @@ public function changeBaseWorkspaceAction(string $targetWorkspaceName, NodeInter
...
     }
 
+    /**
+     * Persists the clipboard state
+     *
+     * @param string $clipboardSubject
+     * @param string $clipboardMode
+     * @return void
+     */
+    public function persistClipboardStateAction(string $clipboardSubject, string $clipboardMode)
```
bwaidelich: What are `$clipboardSubject` and `$clipboardMode`?

packages/react-ui-components/src/DateInput/__snapshots__/dateInput.spec.js.snap
```diff
@@ -47,6 +47,13 @@ exports[`<DateInput/> should render correctly. 1`] = `
       dateFormat={true}
       onChange={[Function]}
       open={true}
+      timeConstraints={
+        Object {
+          "minutes": Object {
+            "step": 5,
+          },
+        }
+      }
```
bwaidelich: How is this change related?

Classes/Controller/BackendServiceController.php
bwaidelich: After skimming over the JS code below I think `clipboardSubject ` refers to the `contextPath` of a single node in the clipboard, and `clipboardMode` is either the string `'COPY'` or `'CUT'`, correct?

Why not sth like
```php
class ClipboardState {
  public function copyNode(NodeInterface $node): void {}

  public function cutNode(NodeInterface $node): void {}

  public function pasteNode(): NodeInterface {}
}
```

Classes/Controller/BackendController.php
```diff
@@ -128,9 +135,14 @@ public function indexAction(NodeInterface $node = null)
...
         $siteNode = $node->getContext()->getCurrentSiteNode();
 
+        $clipboardSubject = $this->clipboardStateContainer->getClipboardSubject();
+        $clipboardMode = $this->clipboardStateContainer->getClipboardMode();
+
         $this->view->assign('user', $user);
         $this->view->assign('documentNode', $node);
         $this->view->assign('site', $siteNode);
+        $this->view->assign('clipboardSubject', $clipboardSubject);
+        $this->view->assign('clipboardMode', $clipboardMode);
```
bwaidelich: If we store the clipboard state on the server, that state needs to be flushed when the CB is pasted. So instead of pushing the clipboard content to the client for every request, I would suggest to do that with an explicit request (that also flushes the CB state)

packages/react-ui-components/src/DateInput/__snapshots__/dateInput.spec.js.snap
dimaip: It's a fix for broken tests, the tests were broken in master at that point. Now it's possible to just rebase.

Classes/Controller/BackendController.php
dimaip: Why? I think it makes sense to keep the last copied node in clipboard, so you could paste it multiple times. This is the way clipboards work in general, e.g. in operating systems.

Classes/Controller/BackendServiceController.php
dimaip: Sure, why not. Will take care.

bwaidelich: Only if you agree!?

dimaip: Hah, well, I think it's more about the preference on which spectrum of abstraction one likes to be. E.g. due to my temperament I prefer to start with the least amount of abstractions and write the imperative code without any abstractions, and then refactor to add more abstractions when I got the gut feeling telling me to do so.
I know that you come from the DDD background of architecting large domain-centric systems that are there to stay and evolve, so I understand where your suggestions are coming from, and I'm ready to adapt to this style of coding and learn a few things along the way.

dimaip:
@bwaidelich thanks for your words of wisdom, I try to let your vision into my head and blend in with my own ideas :) Would love to talk with you more on the subject in person!

davidspiola:
@dimaip I tested the implementation and it works as long as you have the same session and you (sometimes) refresh the target website to be able to past it. 

@bwaidelich I might get it wrong, but it sounds like that you are not 100% happy with server side clipboard storage, right?

bwaidelich:
on my way to vacations, but quickly:

> I might get it wrong, but it sounds like that you are not 100% happy with server side clipboard storage, right?

It's just: I'm concerned about the Neos API that is quite topsy-turvy already and not documented and tested very well. So I'm on the conservative side of things when it comes to new endpoints :)
IMO we should (in order of my personal preference):
# avoid a new endpoint (i.e. by using a client-side only approach)
# add a very dedicated slim endpoint (`POST` and `GET` to `/foo/clipboard` would be enough here IMO) and avoid that a session is required in the backend (by making the storage extensible or store in caches/db etc)
# use it as suggested, it's a great feature anyways :)

dimaip:
@bwaidelich I tried to do it exactly as you have suggested last time, what other things I could still improve?

Classes/Controller/BackendController.php
```diff
@@ -101,6 +101,13 @@ class BackendController extends ActionController
      */
     protected $styleAndJavascriptInclusionService;
 
+    /**
+     * @Flow\Inject
+     * @var ClipboardStateContainer
+     */
+    protected $clipboardStateContainer;
```
bwaidelich: It's just a (node) clipboard

    /**
     * @Flow\Inject
     * @var NodeClipboardInterface
     */
    protected $clipboard;

could work, no?

Classes/Controller/BackendServiceController.php
```diff
@@ -327,6 +333,24 @@ public function changeBaseWorkspaceAction(string $targetWorkspaceName, NodeInter
...
     }
 
+    /**
+     * Persists the clipboard state
+     *
+     * @param NodeInterface $clipboardSubject
+     * @param string $clipboardMode
+     * @return void
+     */
+    public function persistClipboardStateAction(NodeInterface $clipboardSubject, string $clipboardMode)
```
bwaidelich: I'd suggest to get rid of the magic string and create two separate endpoints:
```
public function copyNodeAction(NodeInterface $node)
{
    $this->clipboard->copyNode($node);
}

public function cutNodeAction(NodeInterface $node)
{
    $this->clipboard-> cutNode($node);
}
```

Classes/Controller/ClipboardStateContainer.php
```diff
@@ -0,0 +1,80 @@
...
+use Neos\Flow\Annotations as Flow;
+use Neos\Error\Messages\Message;
+use Neos\ContentRepository\Domain\Model\NodeInterface;
+
+/**
+ * This is a container for clipboard state that needs to be persisted server side
+ *
+ * @Flow\Scope("session")
+ */
+class ClipboardStateContainer
```
bwaidelich: As mentioned above, I think this could be renamed to (Node)Clipboard. And IMO it should implement an interface so we can easily replace it (via 3rd party packages even)

Classes/Controller/ClipboardStateContainer.php
```diff
@@ -0,0 +1,80 @@
...
+{
+    /**
+     * @var NodeInterface
+     */
+    protected $clipboardSubject = null;
```
bwaidelich: why `protected`?
```
/**
 * @var NodeInterface
 */
private $nodeInClipboard = null;
```

Classes/Controller/ClipboardStateContainer.php
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * @var string
+     */
+    protected $clipboardMode = '';
```
bwaidelich: What about introducing constants for the mode and tweak this to sth like:
```
/**
 * @var string one of the NodeClipboardInterface::MODE_*  constants
 */
private $mode;
```
to avoid the magic strings

Classes/Controller/ClipboardStateContainer.php
dimaip: Because the first thing I heard when meeting the Neos project is to never use `private`. I think the reasoning was that you might want to extend the class later. Frankly I never personally agreed with it, but just got used to doing it to follow the supposed convention. But if it's no longer the case I'd gladly do that.

dimaip:
Dear @bwaidelich, thanks a lot for your comments! I agree with them and will take care. Would love to learn more from you in the future :heart_eyes:

bwaidelich:
> Would love to learn more from you in the future

Likewise! :)

bwaidelich:
(sorry, didn't mean to close)

Classes/Controller/ClipboardStateContainer.php
dimaip: Oh gosh, it has to be protected in order for it to be persisted in the session! Waisted so much time debugging this...

dimaip:
So @bwaidelich, I tried to adjust the code according to your feedback, please give it another look when you got the time. Thanks a lot! :heart:

Classes/Controller/NodeClipboard.php
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * @var NodeInterface
+     */
+    protected $node = null;
+
+    /**
+     * @var string one of the NodeClipboardInterface::MODE_*  constants
+     */
+    protected $mode = '';
```
bwaidelich: nitpick: if you set this to `self::MODE_COPY` here (and in `reset()`) there is one less state to consider (mode is either copy or move then)

bwaidelich:
@dimaip Thanks a lot, I think this is much easier to comprehend & extend now. I hope you agree.
I just left a mini nitpick comment, but this can be safely ignored.
I can't test this right now because I don't have a multi site installation that shares the same node types

Classes/Controller/NodeClipboard.php
dimaip: Make sense, but I'd leave it as `''` for the sake of how this is later treated in the UI state: would be strange if the mode is set to Copy while there's no copied node, could cause some side-effects in some of the UI logic.

PR #2133, opened 2018-09-17
BUGFIX: Add timeConstraints to snapshot 

mstruebing:
Tests are currently broken becuase the jest snapshot wasn't updated after latest changes.


mstruebing:
Merging because of broken master and I only updated the snapshot

PR #2134, opened 2018-09-17
TASK: reduce linter noise

mstruebing:
This removes some warnings of the linter by adding aria-props.

There is one linter warning left: `warning  Method 'render' has a complexity of 25  complexity`
Which occurs in the file: `Neos.Neos.Ui/packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/index.js`.
However, this should be handled separately in my opinion.

### DIFF
--- a/packages/neos-ui-editors/src/Editors/Image/Components/PreviewScreen/index.js
+++ b/packages/neos-ui-editors/src/Editors/Image/Components/PreviewScreen/index.js
@@ -54,7 +54,7 @@ export default class PreviewScreen extends PureComponent {
                             className={(thumbnail ? style.cropArea__image : style['cropArea__image--placeholder'])}
                             src={thumbnail ? thumbnail.uri : '/_Resources/Static/Packages/Neos.Neos/Images/dummy-image.svg'}
                             style={thumbnail ? thumbnail.styles.thumbnail : {}}
-                            role="presentation"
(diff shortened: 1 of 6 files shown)


packages/neos-ui-editors/src/Editors/Image/Components/PreviewScreen/index.js
```diff
@@ -54,7 +54,7 @@ export default class PreviewScreen extends PureComponent {
                             className={(thumbnail ? style.cropArea__image : style['cropArea__image--placeholder'])}
                             src={thumbnail ? thumbnail.uri : '/_Resources/Static/Packages/Neos.Neos/Images/dummy-image.svg'}
                             style={thumbnail ? thumbnail.styles.thumbnail : {}}
-                            role="presentation"
+                            alt="thumbnail"
```
dimaip: This alt attribute would be used to display a tooltip on hover, and this tooltip would carry 0 relevant information, and for non-english speaking editors that'd be even annoying. I propose to use a `propertyName` for the alt attribute, at least it's slightly relevant to the information provided.

mstruebing: Thanks for that suggestion. Fixed.

PR #2135, opened 2018-09-17
BUGFIX: reset index after filter when checking for policies

dimaip:
The indexes of an array must be reset after filtering. This caused the result to be casted into object in JS, which broke the tree in effect.


dimaip:
Merging without reviews as it's a no-brainer and urgent.

PR #2136, opened 2018-09-17
FEATURE: mark shine-through nodes with orange outline

dimaip:
image

This PR mimics the behavior of the old UI: when node is a shine-through node, it will be marked with orange outline instead of blue.
The only difference to the old behavior is I chose to leave the orange background also when the node is focused, so not only on hover. IMO that will be more useful to editors, as the hover outline is easy to miss.


dimaip:
@mstruebing I doubt my PR could alter the position, isn't it the way it is now?
Regarding the dashed outline of the editable area, it didn't exist in the old UI and it only coincides with the position of the element outline. Actually I find that overlapping to be quite ugly, so I'd propose to decrease the margin on the editable outline. Then it could look somehow like this:
image
Wdyt?

PR #2137, opened 2018-09-18
BUGFIX: fix propertiesForValidationSelector

dimaip:
Transient values were not taken into account for validation, because they were not correctly set.


dimaip:
Merging as no-brainer, feel free to post-review

PR #2154, opened 2018-10-03
TASK: Update debug to 4.1

mstruebing:
This updates the debug dependency which was out of date.

This seems to need some investigation with our uglify/babel stuff.


mstruebing:
Cool that you are taking over :) :+

markusguenther:
> Cool that you are taking over :) 👍

Yeah but need to solve the ES6 issue within node_modules. Actually we exclude node_modules from the babel_loader.

PR #2156, opened 2018-10-03
TASK: Update fileloader

mstruebing:
I also moved the `file-loader` into the dev-dependencies as I think this is the better place since it's webpack related.

If I miss anything ping me and I will move it back.


mstruebing:
The error has $something to do with https://github.com/dimaip/ckeditor5-engine/tree/iframe

Maybe we fix this by updating your fork to latest master @dimaip ?

markusguenther:
We can close this PR. Since the webpack upgrade we already use the file-loader in version 3

https://github.com/neos/neos-ui/pull/2369

PR #2159, opened 2018-10-04
BUGFIX: Make make work on macOS

kdambekalns:
The original Makefile did not work for me, since webpack and cross-env
could not be found.

This change fixes it, calling those binaries like it is done for lerna.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,13 +29,12 @@
 ################################################################################
 
 
-# Add node_modules and composer binaries to $PATH
-export PATH := ./node_modules/.bin:./bin:$(PATH)
-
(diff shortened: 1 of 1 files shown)


kdambekalns:
The question of course is, why does this work for other macOS users. It could be broken since Mojave, or something is different, e.g. to the machine of @skurfuerst.

```
$ make -v
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0
```

PR #2161, opened 2018-10-05
FEATURE: Add support for asset sources in asset editor

robertlemke:
This change adds asset sources support to the Asset Editor, which
allows for searching and importing remote assets on the fly.

This change relies on the new Asset Proxies web service endpoint in Neos.Neos.

https://github.com/neos/neos-development-collection/issues/2198

### DIFF
--- a/packages/neos-ui-editors/src/Editors/AssetEditor/index.js
+++ b/packages/neos-ui-editors/src/Editors/AssetEditor/index.js
@@ -8,7 +8,7 @@ import {$get} from 'plow-js';
 import Controls from './Components/Controls/index';
 import AssetOption from '../../Library/AssetOption';
 import {AssetUpload} from '../../Library/index';
-import backend from "@neos-project/neos-ui-backend-connector";
+import backend from '@neos-project/neos-ui-backend-connector';
 
 const DEFAULT_FEATURES = {
     mediaBrowser: true,
--- a/packages/neos-ui-editors/src/Editors/AssetEditor/index.js
+++ b/packages/neos-ui-editors/src/Editors/AssetEditor/index.js
@@ -106,7 +106,7 @@ export default class AssetEditor extends PureComponent {
     handleSearchTermChange = searchTerm => {
         if (searchTerm) {
             this.setState({isLoading: true, searchOptions: []});
-            this.props.assetLookupDataLoader.search({}, searchTerm)
+            this.props.assetLookupDataLoader.search({assetsToExclude: this.getValues()}, searchTerm)
                 .then(searchOptions => {
                     this.setState({
                         isLoading: false,
--- a/packages/neos-ui-editors/src/Editors/AssetEditor/index.js
+++ b/packages/neos-ui-editors/src/Editors/AssetEditor/index.js
@@ -108,7 +108,6 @@ export default class AssetEditor extends PureComponent {
             this.setState({isLoading: true, searchOptions: []});
             this.props.assetLookupDataLoader.search({}, searchTerm)
                 .then(searchOptions => {
-                    console.log('searchOptions',searchOptions);
                     this.setState({
                         isLoading: false,
                         searchOptions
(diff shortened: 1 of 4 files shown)


robertlemke:
Not finished yet – I'd like to get rid of the duplicate import issue first.

packages/neos-ui-editors/src/Editors/AssetEditor/index.js
```diff
@@ -140,6 +160,7 @@ export default class AssetEditor extends PureComponent {
     }
 
     handleMediaSelected = assetIdentifier => {
+        console.log('handleMediaSelected', assetIdentifier, this.props);
```
daniellienert: I guess this console log needs to be removed

robertlemke: Yes indeed.

robertlemke:
I added a filter which removes those assets from a search result which are already selected in the asset editor.

Note that the asset source proxy implementation must be properly implemented so that _$proxy->getLocalAssetIdentifier()_ returns the Neos Media Asset identifier if the asset was already imported (via the _ImportedAssetRepository_).

dimaip:
Hey @robertlemke! What's the current state of this thing? Is it finished and ready for final review?

robertlemke:
Allright, I added some grouping and display of asset source labels, by suggestion of @bwaidelich (see Neos.Neos PR). I suggest that we get the Neos PR merged first, and then this here can be merged.

However, from my side, this should be fine already for a final review.

bwaidelich:
@robertlemke 
> I added some grouping and display of asset source labels
but did you _push_ it as well? :)

robertlemke:
@bwaidelich yes, I did: https://github.com/neos/neos-ui/pull/2161/commits/4033bcab511c4ba7ea665781ce395a7ea3c860fa

bwaidelich:
Ah apparently, Github is still not back to full speed. The commit just appeared now

kitsunet:
@robertlemke @bwaidelich @dimaip This change should be reverted immediately as it (unannouncedly) breaks compatiblity with Neos 3.3 for the UI. Neos 3.3 doesn't support asset proxies so the asset search is done on an empty url (so basically domain/?parameters) so it stays hidden and doesn't produce an obvious error. Asset usage is broken with this though.

dimaip:
Also see our quick discussion https://neos-project.slack.com/archives/C0U0KEGDQ/p1543414898215700

robertlemke:
I think that just reverting the feature in not the solution.

As @dimaip already wrote in Slack, the long-term solution is to make Neos UI a part of the Neos Development Collection. However, since we are not there yet, the logical solution in my opinion is, that we branch off the Neos UI repository in order to provide a backwards-compatible version for older Neos versions. That branch should only receive important bugfixes, all new features should go into master and be compatible with the latest / upcoming Neos.

I don't have insight into how Neos UI is released currently, but I guess it's rather simple, given that it is not part of the Development Collection yet? If so, let's create a branch, revert this feature in that branch and that should be it.

robertlemke:
I think that just reverting the feature in not the solution.

As @dimaip already wrote in Slack, the long-term solution is to make Neos UI a part of the Neos Development Collection. However, since we are not there yet, the logical solution in my opinion is, that we branch off the Neos UI repository in order to provide a backwards-compatible version for older Neos versions. That branch should only receive important bugfixes, all new features should go into master and be compatible with the latest / upcoming Neos.

I don't have insight into how Neos UI is released currently, but I guess it's rather simple, given that it is not part of the Development Collection yet? If so, let's create a branch, revert this feature in that branch and that should be it.

kitsunet:
Totally fine for me, but the branch needs to be maintained as long as 3.3 is maintained IMHO.

kitsunet:
Fact is, we cannot release a new version of the UI right now as that would be a breaking release and I now had to fork the UI for a 3.3 project to make sure we don't run into this.

dimaip:
True. It's one of the solutions. Will try to weigh-in all the pros and cons tomorrow at 13:00 in the hangout with Christian and all those who would want to join.

PR #2162, opened 2018-10-06
TASK: Update react-close-on-escape

mstruebing:


markusguenther:
Do we really need this dependency? We use it only in the dialog component and the package just checks if I press escape.

Couldn't we just implement this single method in the dialog component and get rid of the dependency?
Don't blame it, just asking?

dimaip:
Fine with me, makes total sense in spirit of the recent backlash against small packages :)

mstruebing:
Yep would absolutely agree on that.

markusguenther:
Should I adjust that on master or on typescript branch? Or do in in master and then adjust it for typescript branch? Don't know the time schedule for the typescript merge.

mstruebing:
I would say first do it on master, adjusting this for typescript shouldn't be that hard (maybe I'm wrong with this assumption^^)

markusguenther:
Did remove the package now with these PR https://github.com/neos/neos-ui/pull/2274

PR #2163, opened 2018-10-07
TASK: Make loading "spinner" a color-changing Neos logo

### LINKED ISSUE #1686: Make loading "spinner" a color-changing Neos logo
I imagine a Neos "N" going through several color variations as a loading animation. Would look so much cooler and more Neos than the "ring" we have now.

mstruebing:
Finally.

We can adjust the timings and colors as we like.

closes #1686

### DIFF
--- a/packages/neos-ui/src/index.js
+++ b/packages/neos-ui/src/index.js
@@ -22,6 +22,7 @@ import apiExposureMap from './apiExposureMap';
 import DelegatingReducer from './DelegatingReducer';
 
 import Icon from '@neos-project/react-ui-components/src/Icon/';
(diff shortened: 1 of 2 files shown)


aertmann:
Haha this is awesome! not easy to test though, disappears immediately.

Not sure with the rainbow party colors, maybe stick with the color palette?

Makes me think of what a cool animation it would be drawing the outlined logo in steps, like http://airbnb.io/lottie/
I know an app that does this for their loading animation, but can't remember right now. Next time I come across it I'll get back with it.

aertmann:
Anyway can confirm it works btw. 👍

PR #2164, opened 2018-10-08
FEATURE: node variant selector

dimaip:
This feature introduces a node variant selector which should help the editor to answer these questions:

1) Do I see a shine through node or a real thing? (the orange warning indicator)
2) Where does this node shine through from? (the first selectbox option, selected by default)
3) What other node variants are available? (the other selectbox options)
4) Quickly switch to those node variants (by selecting a node variant in selectbox)

TODO:

- [x] I18n
- [x] Adjust PropTypes

peek 2018-10-08 10-11

### DIFF
--- a/Classes/Fusion/Helper/NodeInfoHelper.php
+++ b/Classes/Fusion/Helper/NodeInfoHelper.php
@@ -18,6 +18,7 @@ use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Mvc\Controller\ControllerContext;
 use Neos\Flow\Persistence\PersistenceManagerInterface;
 use Neos\Neos\Domain\Service\ContentContext;
+use Neos\Neos\Domain\Service\ContentDimensionPresetSourceInterface;
 use Neos\Neos\Service\LinkingService;
 use Neos\Neos\Service\Mapping\NodePropertyConverterService;
 use Neos\Neos\TypeConverter\EntityToIdentityConverter;
@@ -90,6 +91,12 @@ class NodeInfoHelper implements ProtectedContextAwareInterface
     protected $ignoredNodeTypeRole;
 
     /**
+     * @Flow\Inject
+     * @var ContentDimensionPresetSourceInterface
+     */
+    protected $contentDimensionsPresetSource;
+
+    /**
      * @param NodeInterface $node
      * @param ControllerContext $controllerContext
      * @param bool $omitMostPropertiesForTreeState
@@ -159,6 +166,12 @@ class NodeInfoHelper implements ProtectedContextAwareInterface
         $nodeInfo['properties'] = $this->nodePropertyConverterService->getPropertiesArray($node);
         $nodeInfo['isFullyLoaded'] = true;
 
+        $otherNodeVariants = array_values(array_filter(array_map(function ($node) {
+            return $this->getCurrentDimensionPresetIdentifiersForNode($node);
+        }, $node->getOtherNodeVariants())));
+        $nodeInfo['dimensions'] = $this->getCurrentDimensionPresetIdentifiersForNode($node);
+        $nodeInfo['otherNodeVariants'] = $otherNodeVariants;
+
         if ($controllerContext !== null) {
             $nodeInfo = array_merge($nodeInfo, $this->getUriInformation($node, $controllerContext));
         }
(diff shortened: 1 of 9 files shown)


skurfuerst:
very nice :heart: - I did not check the code, but the feature is awesome :)

dfeyer:
Good move, but bad UX from my perspective.

UX wise, it's two place for the same action (yes with different information on each place ... but yheaa UX wise it's never a good idea to have two place for the same feature or variant of the same feature).

From my POV, I prefer to have the dimension selector on the inspector (top of of first tab, like in the screen shot, it's perfect), but we should remove it from the top bar.

I prefere the UI of the current dimension switcher (the one from the top bar), so the user can select dimension One, then an other, ... and the UI avoid impossible combination. With more than two dimension, the list will be long and hard to use.

There is too much information when the select box is closed, and most informations hidden behind the tooltip. When the select box is close, I will prefer to have kind of dashboard/table style UI, like:

- Line 1: Dimension Icon | Dimension Name | Dimension Value (this line can be duplicated for each dimensions
- Last line, a clear sentence to say if it's a shadow node, where it came, ...

But yeaa good move, but I prefere to polish more the UI before this feature go live.

dimaip:
Just to sum it up, I think there are two distinct questions:
1) Where to put the dimension switcher so it'd always be visible
2) How to display node variants information

I've already written above about 2), and regarding 1) there's really not much we can do. Here's my brute-force approach to solving that problem, here's what I think is the list of all possible places that comes to my mind where we could place the dimension switcher, we just got to decide:

image

dfeyer:
> I think you misunderstood what this PR is doing. It does not list all possible dimension configurations, it only shows already existing node variants. It is in no way trying to replace the dimension selector, but rather to compliment it.

I understand the technical subtilities ;) but I think editor will not get them, and need more guidance, because they will mix both concepts. We need to find a way to have a single place where information can be correctly prioritze, and hopeful in a understandable way.

I agree it's a big challenge ;)

I need to mockup some variants before our hangout. I'm in vacation between tomorrow and next Tuesday, so I think I can have something for the end of next week.

What we miss currently, it's an "always visible place", where we can communicate the "editing context", in editing context I include the following informations:

1. The currently selected Dimensions Preset (DP) (what the dimension switcher show when closed)
2. The current workspace (the current element can come from the live workspace, but also from a team workspace, ...)
2. The DP of the currently selected element (can be the document or an element, can be from the currently selected DP, or shine through from an othe DP.

Those informations can help use to not be lost ;)

Then we have the needs to be able to switch between thoses contexts

> First of all, we can not move the dimension switcher to the inspector, as a lot of editors prefer working having the inspector collapsed.

Yes, but we need to use the space that we have. 

In the current UI concept the Inspector is where metadata about the currently selected node needs to be. Other platform use the right inspector to show this kind of informations (available preview, translation, ...), so I think it's a good/common place.

To top bar (CK toolbar, link to preview, fullscreen, ...) is too small to accept a complicated UI like the current one.

During editing it's way more confortable to have the navigation tree closed, but keep the inpsector visible, you have to change some metadata quiet often.

My main concern is the full screen editing mode, if we loose the dimensions switcher in this mode it's pretty bad. Kind of Egg & Chicken issue. More I think about that, more I think it make sense to keep the inspector (closed by default) even in full screen mode. That can be a quiet powerful solution, and we keep access to the metadata even in fullscreen mode.

Need to think more about it ;)

dimaip:
Perhaps we could have a short live drawing session before you get deep into prototyping?
If not then let's wait till next week, though I really hate to loose the momentum on such complicated tasks...

davidspiola:
> Perhaps we could have a short live drawing session before you get deep into prototyping?
> If not then let's wait till next week, though I really hate to loose the momentum on such complicated tasks...

I'd like to participate when you guys do a hangout session

dimaip:
Yes, naturally, that was an open invitation for anyone else who wanted to join as well.

davidspiola:
> UX wise, it's two place for the same action (yes with different information on each place ... but yheaa UX wise it's never a good idea to have two place for the same feature or variant of the same feature). Switching dimensions and understanding node variant, content shine throught, ... it's the same feature, and MUST be collocated in the UI.

Is it really the same feature? IMHO the difference is that the dimension switcher is only for document nodes and not content nodes. The inspector seems to be the obvious choice. Showing document and content node information in on place could be confusing as well.

dimaip:
Extracted it here: https://github.com/Flowpack/Flowpack.NodeVariantSelector

PR #2165, opened 2018-10-08
BUGFIX: EditNodePrivilege combined with EditNodePropertyPrivilege

ComiR:
This PR contains multiple fixes to get to the point where the behavior of the UI corresponds to the behavior of the ContentRepository.

When you get an abstaining `EditNodePrivilege` (resulting in a "soft" DENY) in combination with an `EditNodePropertyPrivilege` GRANT, the UI ignores the latter and denies editing. The ContentRepository on the other hand (or rather Flows SecurityInterceptor) sees both privileges matching, one abstaining and the other granting, so it will allow access to the editProperty method.

The commit abc4c72310cadd5354a52f4a4114a892a7bffc0d isn't really connected that closely to this, but I stumbled upon it while fixing. If necessary, I'll extract it (or any other commit for that matter) into another PR.

### DIFF
--- a/Classes/Aspects/AugmentationAspect.php
+++ b/Classes/Aspects/AugmentationAspect.php
@@ -12,7 +12,6 @@ namespace Neos\Neos\Ui\Aspects;
  */
 
 use Neos\ContentRepository\Domain\Model\NodeInterface;
-use Neos\ContentRepository\Service\AuthorizationService;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Aop\JoinPointInterface;
 use Neos\Flow\Session\SessionInterface;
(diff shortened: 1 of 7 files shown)


dimaip:
Wow, looks cool! Would be great if someone with more PHP competence could look into this.

ComiR:
If you have a look at https://github.com/neos/neos-ui/blob/0d9bb0aaa50b9ef46d532fab55ab33fe3e59dbb1/packages/neos-ui-guest-frame/src/initializePropertyDomNode.js#L20, you can see that it's trying to filter by disallowed properties. But this doesn't work here because `nodes` in the guest frame isn't the same as in f.e. the inspector (i.e. there is no `policy`).

It would be great if someone had an idea how to resolve this. As the PR is right now, a disallowed property still gets its blue borders, but you can't edit it.

dimaip:
Could anyone please help with reviewing this?

PR #2170, opened 2018-10-12
BUGFIX: getAllowed(Grand)ChildNodeTypes won't return abstract NodeTypes anymore

ComiR:
**What I did**
If an abstract NodeType was allowed to be created as to another nodes constraints, but disallowed due to other policies (mentioned below), the "create node" buttons weren't deactivated.

**How I did it**
Add a check to `NodeTypesRegistry`s `getAllowedChildNodeTypes` and `getAllowedGrandChildNodeTypes` which checks if the NodeType is valid (i.e. not abstract) by looking it up in its own collection (the registry).

**How to verify it**
* `Node` with `NodeMixin`
* Document `ParentNode` with `constraints.nodetypes: [NodeMixin: true]`
* `CreateNodePrivilege` with `matcher: createdNodeIsOfType('Node')`, abstaining or denying

Now try to create a child node under the `ParentNode`. `Node` is theoretically allowed as child node due to being a NodeMixin, but blocked for the user due to the policy.
The create node button is not disabled due to it not knowing `NodeMixin` is abstract. But then the creation dialog does know it and won't display anyhing.

### DIFF
--- a/packages/neos-ui-contentrepository/src/registry/NodeTypesRegistry.js
+++ b/packages/neos-ui-contentrepository/src/registry/NodeTypesRegistry.js
@@ -55,13 +55,13 @@ export default class NodeTypesRegistry extends SynchronousRegistry {
     getAllowedChildNodeTypes(nodeTypeName) {
         const result = $get([nodeTypeName, 'nodeTypes'], this._constraints);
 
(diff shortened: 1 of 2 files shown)


ComiR:
Any objections to this? @dimaip maybe?

Issue #2141, opened 2018-09-25
SelectBox with validator validates only after apply


kitsunet:
If you have a property of type array (`array<Neos\Media\Domain\Model\Asset>` as in AssetList for example) - which results in a multi select box - and you add a validator (in my case custom validator that checks for minimum and maximum number of items) the validation happens only after pressing apply, but then the content is broken.

PR #2174, opened 2018-10-15
BUGFIX: make the count validator work with the references editor in inspector

kabarakh:
**What I did**
If you are using the count validator for an references editor, in the node creation dialog everything is fine, but in the inspector there's always an error. In fact this error is there because the count validator counts the Object-keys of the value to check, and for the inspector this is always the same number because an object of the class List (from immutable.js) is used which always has the same length.

**How I did it**
By calling the method asJS() we get the array of the immutable list and thus have the proper countable element which we pass to the original count function. 

Here's the question though: Should we use asObject or asArray to get an object to count, or should the conversion be earlier than in the validator itself? I didn't really know if there is a better way, this is what worked in my tests.

**How to verify it**
* Create a nodetype with a property of the type references
* add the count validator to it and use e.g. min 1 and max 3
* without the fix the inspector always displays "should be between 1 and 3", with the fix it displays a valid (green) formfield

### DIFF
--- a/packages/neos-ui-validators/src/Count/index.js
+++ b/packages/neos-ui-validators/src/Count/index.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import I18n from '@neos-project/neos-ui-i18n';
 import logger from '@neos-project/utils-logger';
+import {List} from 'immutable';
 
 /**
  * Checks if the given value is an object or array
@@ -20,6 +21,9 @@ const Count = (value, validatorOptions) => {
         return <I18n id="content.inspector.validators.countValidator.notCountable"/>;
     }
 
+    if (value instanceof List) {
+        value = value.toJS();
+    }
     const {length} = Object.keys(value);
 
     if (length < minimum || length > maximum) {


kabarakh:
I agree that usually a validator should get data in a specific format and validate that. I didn't find the part in the sourcecode where the data is passed to the validator ;)

Thing is that I found https://github.com/neos/neos-ui/issues/2141 open, where Christian had a simmilar problem (an immutable.js List in a validator) - I don't know if there are more places to look at, and my fix would only fix it for the CountValidator.

dimaip:
These are the two places where the validation happens: https://github.com/neos/neos-ui/search?q=validate+from+%27%40neos-project%2Fneos-ui-validators%2Fsrc%2Findex%27%3B&unscoped_q=validate+from+%27%40neos-project%2Fneos-ui-validators%2Fsrc%2Findex%27%3B

I guess the issue is rooted in the inspector, property values should be converted to JS datatypes before being passed to validation here: https://github.com/neos/neos-ui/blob/82de84fdde8f5b635037262e7cd4f6ed7b69348b/packages/neos-ui-redux-store/src/UI/Inspector/selectors.js#L49

Alternatively it might make sense to figure out why they end up being in immutablejs format at all and whether it's a correct behavior or not.

dimaip:
So this fix should no longer be relevant since we don't have immutablejs.

PR #2175, opened 2018-10-16
TASK: Remove duplicated code

mstruebing:
This removes logic which was done twice in the same file and extracts it into an own function, which is better imho.


bwaidelich:
FYI: with this change the `highlight` prop is no longer propagated to the wrapped component
```
const {highlight, validationErrors, ...rest} = this.props;
```
vs
```
const restProps = omit(this.props, ['validationErrors']);
```

dimaip:
That's why I hate passing all props via spread... You never really know what should have been there...

bwaidelich:
True, but frankly spread and `omit()` have the same issue in that regard and it was probably my mistake to consume a prop that was only there "by accident"

PR #2177, opened 2018-10-16
WIP (DO NOT MERGE) Event sourced

skurfuerst:
**What I did**

**How I did it**

**How to verify it**


Classes/Controller/BackendController.php
```diff
@@ -122,11 +154,17 @@ public function indexAction(NodeInterface $node = null)
...
-            $node = $this->findNodeToEdit();
+        if ($nodeAddress === null) {
+            $nodeAddress = $this->findNodeToEdit();
         }
 
-        $siteNode = $node->getContext()->getCurrentSiteNode();
+        $workspaceName = $this->userService->getPersonalWorkspaceName();
+        $workspace = $this->workspaceFinder->findOneByName(new WorkspaceName($workspaceName));
+        $subgraph = $this->contentGraph->getSubgraphByIdentifier($workspace->getCurrentContentStreamIdentifier(), $this->findDefaultDimensionSpacePoint());
+        $siteNode = $subgraph->findChildNodeConnectedThroughEdgeName($this->getRootNodeIdentifier(), new NodeName($this->siteFinder->findDefault()->nodeName));
```
hlubek: This seems wrong to me and only works when having a single site.

The actuals site node could be inferred from `$node` by traversing parents up to the static `sites` node.
Or we pass the site as an argument.
Or we include the site in something like a `NodeAddressWithSite` value object.

skurfuerst: totally true! will fix!

skurfuerst:
Removing, as superseded by https://github.com/neos/neos-ui/pull/2178

PR #2178, opened 2018-10-16
FEATURE: switch to backend-calculated parent node links (needed for Event Sourced CR)

skurfuerst:
This change does two main things:

# NeosUiDefaultNodesOperation should use TraversableNodeInterface

this is needed for Neos ES CR; and is a purely internal refactoring.

# Backend-Calculated Parent Node Links

(main part of this change)

## The Problem

There are a few places in the Neos UI where we rely on the fact that the `contextPath` of a node
is structured like `/a/b/c@..` - so we calculate the parent path in JavaScript.

In the Event Sourced CR, context paths are actually `NodeAddresses`, looking roughly like: `uuidOfNode__dimension__workspace`.

Thus, the user interface needs to treat the `contextPath` properties as opaque; and figure out
the parents in a different way.

## The Solution

1. We include a `parent` property as a reference to the parent node in the server-sent data.
2. We ensure this is used globally on the client side.


skurfuerst:
Actually this is going into the right direction now as soon as the core change is merged :)

composer.json
```diff
@@ -7,7 +7,7 @@
   ],
   "require": {
     "neos/neos": "^4.0 || dev-master",
-    "neos/neos-ui-compiled": "self.version"
+    "neos/neos-ui-compiled": "dev-master"
```
skurfuerst: TODO: before merging, this change must be reverted!

skurfuerst:
As soon as CircleCI is :+1:, we can merge this I think :)

Classes/Domain/Model/FeedbackCollection.php
```diff
@@ -75,4 +75,9 @@ public function jsonSerialize()
             'feedbacks' => $feedbacks
         ];
     }
+
+    public function reset()
+    {
+        $this->feedbacks = [];
+    }
```
skurfuerst: this does not really have anything to do with the core of the change; but we were missing a way to reset feedbacks beforehand.

Classes/FlowQueryOperations/NeosUiDefaultNodesOperation.php
```diff
@@ -11,6 +11,8 @@
  * source code.
  */
 
+use Neos\ContentRepository\Domain\NodeType\NodeTypeConstraintFactory;
+use Neos\ContentRepository\Domain\Projection\Content\TraversableNodeInterface;
```
skurfuerst: In this file, we implement everything against the TraversableNodeInterface instead of NodeInterface; so that we use the forward-compatible API already.

Classes/Fusion/Helper/NodeInfoHelper.php
```diff
@@ -212,6 +212,7 @@ protected function getBasicNodeInformation(NodeInterface $node): array
             'isAutoCreated' => $node->isAutoCreated(),
             'depth' => $node->getDepth(),
             'children' => [],
+            'parent' => $node->getParent()->getContextPath(),
```
skurfuerst: that's the server-side part of the "Parent"-change. Easy :D

packages/neos-ui-redux-store/src/CR/Nodes/helpers.ts
```diff
@@ -21,24 +21,6 @@ export const getAllowedNodeTypesTakingAutoCreatedIntoAccount = (baseNode: Node,
     return nodeTypesRegistry.getAllowedChildNodeTypes(baseNode.nodeType);
 };
 
-//
-// Helper function to get parent contextPath from current contextPath
-//
-export const parentNodeContextPath = (contextPath: NodeContextPath) => {
```
skurfuerst: basically, we remove this method and all transitive usages.

packages/neos-ui-sagas/src/CR/NodeOperations/helpers.js
```diff
@@ -1,15 +1,5 @@
 import {findNodeInGuestFrame, closestNodeInGuestFrame} from '@neos-project/neos-ui-guest-frame/src/dom';
 
-export const parentNodeContextPath = contextPath => {
```
skurfuerst: that's the 2nd occurence of client-side parent calculation which we need to get rid of

skurfuerst:
yeah yrah yeah :D

Issue #1471, opened 2017-12-21
Missing hover states


danielkestler:
A lot of hover states are missing, especially on buttons like "Apply" or "Publish". Maybe even more hover states should be introduced to enhance usability, e.g. on dropdowns, input fields or in the page tree.

mstruebing:
It's not missing as the old UI doesn't have this too.

But I agree that would be enhance the UX

kabarakh:
were you thinking about a mixin, an additional class, some scss-extendable-thingy with % or adding filter: brightness to every element with a hover?

mstruebing:
Yes this was my thought but I didn't know how it would work out if we do it.

dimaip:
I'd rather keep CSS as simple as possible and try to use SASS as less as possible. Instead I'd try to achieve reusability at React component level. In practice just using css variables is usually enough to get a DRY-enough code.

PR #2179, opened 2018-10-17
Add missing hover states (#1471)

danielkestler:
Added some important hover states for publish dropdown and publish button bottom right, see:
https://github.com/neos/neos-ui/issues/1471

bildschirmfoto 2018-10-17 um 12 09 07

### DIFF
--- a/packages/build-essentials/src/styles/styleConstants.js
+++ b/packages/build-essentials/src/styles/styleConstants.js
@@ -56,7 +56,9 @@ const config = {
     },
     colors: {
         primaryViolet: '#26224C',
+        primaryVioletHover: '#342f5f',
         primaryBlue: '#00ADEE',
+        primaryBlueHover: '#35c3f8',
         contrastDarkest: '#141414',
         contrastDarker: '#222',
         contrastDark: '#3f3f3f',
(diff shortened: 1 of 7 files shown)


dimaip:
Just curious, couldn't the same effect be achieved by `opacity` or by any other CSS property.
If not, then I think your approach is fine. It might just get a little hard to maintain once we have more colors.

danielkestler:
Well, you won't get the exact same color. Also not possible using lighten() or similar, because this changes perceived saturation (and does not work using --var-colors-).

I think my changes are just a "quick win". We should rethink hover states in general from a design perspective. For example, the hovering of "dark" elements (like the slide toggles) with the primary blue color is quite heavy – and my hover effects of this PR are much more decent. And what about all those other elements (nodes in the page tree, all those dropdowns and input fields ...)?

danielkestler:
Planning to contribute more in this direction. But I don‘t know where is the right place to discuss design changes – or better „just do it“ in a PR?

mstruebing:
A PR is always helpful because it already shows in detail what you mean - but if you have more general topics there is a `#guild-ux` channel on slack.

dimaip:
Hah, the best place to discuss changes is a code sprint! We got one coming up in November, so perhaps you'd want to join @danielkestler?
Other than that, as Max said: #guild-ux or discuss.neos.io. But in practice discussions around existing PRs seem to be the most concrete and fruitful.

PR #2180, opened 2018-10-17
BUGFIX: Use UTC date for dateOnly DateInput

### LINKED ISSUE #2147: The data picker saves date time of browser timezone but not the timezone of php configuration
### Description
If the browser timezone is different from the timezone of php, wenn a date and time is selected with date picker in the backend, the date and time is not the selected value. The value with time difference is saved in database.

### Steps to Reproduce
Background: 
- There is a filed names `startTime` using date picker in the page configuration.
- The browser client uses timezone `Central European Summer Time (UTC+2)` .
- The Timezone in PHP is `UTC`

1. Select a date time (2018-10-01 00:00:00) with date picker in the backend
backend-date-picker
2. Load the DateTime with `q(node).property('startTime')` In fusion (or php) 
3. The date time is shown in the timezone of UTC but not the value that is set
fusion-date

#### Expected behavior

The date time should be save with the same value that is selected in the date picker.

#### Actual behavior

Neos simply saves the value from browser without timezone checking

[What actually happened]

### Affected Versions

Neos:

UI: Latest

hlubek:
**What I did**

* Add utc prop to react-datetime if dateOnly is set
* Today button will use start of day in UTC if dateOnly is set

**How to verify it**

* Open storybook and test the DateInput stories
* Run tests

Relates #2147

### DIFF
--- a/packages/react-ui-components/src/DateInput/__snapshots__/dateInput.spec.js.snap
+++ b/packages/react-ui-components/src/DateInput/__snapshots__/dateInput.spec.js.snap
@@ -63,3 +63,68 @@ exports[`<DateInput/> should render correctly. 1`] = `
   </Component>
 </div>
 `;
+
+exports[`<DateInput/> should set "utc" on DatePickerComponent if "dateOnly" prop is set. 1`] = `
+<div
+  className=""
+>
+  <div
+    className="undefined"
+  >
+    <button
+      className=""
+      onClick={[Function]}
+    >
+      <Component
+        icon="far calendar-alt"
+      />
+    </button>
+    <div>
+      <div
+        className=""
+        onClick={[Function]}
+        role="presentation"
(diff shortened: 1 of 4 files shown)


dimaip:
Hey!

> Add utc prop to react-datetime if dateOnly is set

I don't seem to get the idea. Why configure utc? Why do in only in dateOnly mode?

As I understand, the desired behavior is that when user picks a date, it should be interpreted in server's locale, right?

dimaip:
@hlubek?

monofone:
👍 +1

daniellienert:
@dimaip: We have the same issue here. Any chance to get this reviewed and merged soon?

dimaip:
I must admin I still don't fully understand the issue, but will merge based on other people who tested the fix here.

PR #2181, opened 2018-10-18
TASK: Add title attributes to secondarytoolbar components

mstruebing:
There are no title attributes on the secondary toolbar buttons so no description on hover.


markusguenther:
Looks good to me :)

PR #2182, opened 2018-10-18
REFACTOR: [react-ui-components] migrate tests to typescript

JamesAlias:
**What I did**
I started to migrate test files (`*.spec.js`) to typescript (`*.spec.tsx`).
This commit only contains the changes to the Button component to keep this PR small and easy to understand.

**How I did it**
I modified the `jest-preset-neos-ui/jest-preset.json` to load the typescript test files.
I migrated the test files of the Button Component to typescript.

**How to verify it**
`make test`


packages/react-ui-components/src/Button/index.ts
```diff
@@ -1,6 +1,9 @@
 import {themr} from '@friendsofreactjs/react-css-themr';
 import identifiers from '../identifiers';
-import style from './style.css';
+
 import Button from './button';
 
+// tslint:disable-next-line:no-var-requires
+const style = require('./style.css');
```
dimaip: Why not import?

JamesAlias: TS wasn't able to resolve the module 'style.css' when using `import style from 'style.css'`.
The fix is to add a `global.d.ts` (for now to the `react-ui-components` directory - when needed it can be moved further up): 
```ts
declare module '*.css';
```

now TS won't complain :)
I guess there will be better ways to import css modules in the future but for now this is the only way that worked for me (without a considerable amount of work/pain).
There even is a webpack loader that creates TS definition files for each css module. But for now this would be overkill, imo.

I will update all `typescript-test*` PR's to adopt the change.

Thank's for putting a finger in that wound. It helped me to get my ass up an explore a better solution to the _import css modules_ problem than just using `require`. ❤️

PR #2185, opened 2018-10-18
REFACTOR: [react-ui-components] CheckBox component tests in typescript

JamesAlias:
> Note: This PR is a follow up from #2184.

**What I did**
I migrated the CheckBox component tests to TypeScript.

**How I did it**

**How to verify it**
`make test`


packages/react-ui-components/src/CheckBox/checkBox.spec.tsx
```diff
@@ -25,13 +28,6 @@ describe('<CheckBox/>', () => {
         expect(wrapper.prop('className')).toContain('fooClassName');
     });
 
-    it('should allow the propagation of additional props to the wrapper.', () => {
```
JamesAlias: `foo` is obviously not a valid prop for `CheckBox`. 
Is this test just wrong or is this actually a wanted functionality?

dimaip: see my comment on Headline

PR #2186, opened 2018-10-18
REFACTOR: [react-ui-components] Headline component tests in typescript

JamesAlias:
**What I did**
I migrated the Headline component tests to TypeScript.

**How I did it**
I had to make some changes to the test in the process. I will open a discussion to address these changes.

**How to verify it**
`make test`


packages/react-ui-components/src/Headline/headline.spec.tsx
```diff
@@ -25,12 +26,6 @@ describe('<Headline/>', () => {
         expect(wrapper.prop('className')).toContain('fooClassName');
     });
 
-    it('should allow the propagation of additional props to the wrapper.', () => {
```
JamesAlias: `foo` is obviously no valid prop for `Headline`. Is this actually a wanted functionality?

packages/react-ui-components/src/Headline/headline.tsx
```diff
@@ -7,10 +7,10 @@ type HeadlineType = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
 
 interface HeadlineTheme {
     readonly 'heading': string;
-    readonly 'heading--h1': string;
+    readonly 'heading--h1': string; // TODO ?
```
JamesAlias: I'm not sure what this does.
Does this have a purpose?

dimaip: Only @Inkdpixels would know for sure, but I think it was meant to be used to style different variants of a headling (see `style` prop)

packages/react-ui-components/src/Headline/headline.spec.tsx
dimaip: An html element may have so many possible attributes... I think this behavior was chosen in order to not have to enumerate all possible attributes.

JamesAlias: With TypeScript and React we can actually extend the interface (props) of a component with those of a HTML Element of our choice (eg. div).
See the Button component for an actual use case of this.

packages/react-ui-components/src/Headline/headline.tsx
JamesAlias: Ok, but there's only the `h1` one. I'm not even sure if it ever worked like this 😅.

I could either remove this or add theming for the other 5 HeadlineTypes and toggle them accordingly.

dimaip: I'd say that since for now we don't use it anywhere it'd be safe to remove it. If we'd ever need it, we could always re-introduce it properly.

PR #2188, opened 2018-10-19
WIP: FEATURE: add a splash-like loading experience & allow customization

JamesAlias:
# TODO

- [x] create default settings
- [x] allow customization 

**What I did**
While developing I noticed the rather stale loading experience of the Neos frontend.
I built a pretty simple animation for the Neos logo.
Also I noticed, that the user spends most of the time in front of a white screen while downloading, parsing and initially running the rather huge script files (I know, I am running the dev build, not production, but still..).

So I thought it would be nice to see this loading splash screen even before the scripts are being loaded.

The splash screen is a partial that is rendered inside the `app-container` wich will be overridden by React once the UI is ready to be rendered.
It is configurable in `Settings.yaml` of the Package:
```yaml
Neos:
    Neos:
        Ui:
            splashScreen:
                partial: 'SplashPartial'
```

**How I did it**
Added a splash screen to `index.html` that is the template for the page.
The splash and it's styles will be overwritten once React renders into the DOM.

We can also make this splash customizable through the `Settings.yaml`. (eg by setting the logo svg path or background color.. similar to the login screen customizations).

**How to verify it**
`flow server:run` -> open site with slow internet connection

neosloadingsplash_logo_only 2018-10-19 15_18_12

### DIFF
--- a/Classes/Controller/BackendController.php
+++ b/Classes/Controller/BackendController.php
@@ -133,6 +133,7 @@ class BackendController extends ActionController
         $this->view->assign('site', $siteNode);
         $this->view->assign('headScripts', $this->styleAndJavascriptInclusionService->getHeadScripts());
         $this->view->assign('headStylesheets', $this->styleAndJavascriptInclusionService->getHeadStylesheets());
+        $this->view->assign('splashScreenPartial', $this->settings['splashScreen']['partial']);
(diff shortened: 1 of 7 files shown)


kitsunet:
Seemed rather limited and worked on the spot, didn't see any necessity for WIP and so just merged it :)

PR #2190, opened 2018-10-19
REFACTOR: [react-ui-components] migrate Label component tests to TypeScript

JamesAlias:
> Note: This is a follow-up PR of #2189.

**What I did**
I migrated the Label component's tests to TypeScript.

**How I did it**

**How to verify it**
`make test`


dimaip:
Woow looks cool, it's amazing how productive you are!
However I'm personally not so productive and under really huge load before the sprint, so at worst case I'd be able to give your PRs a thorough look at the sprint, if nobody else does before.
Your contributions are super motivating btw.

PR #2191, opened 2018-10-21
TASK: extract Sagas into an external package and load all modals from the registry automatically

dimaip:
I've extracted the sagas into the separate package. It allows to easier re-use parts of the sagas when overriding them from 3rd-party plugins.
Also the modals are now automatically loaded from the registry, which allows to create custom modals from  3rd-party plugins.

Here are examples of how it may be used:

- Sagas: https://github.com/Flowpack/restrict-creation/blob/master/Resources/Private/RestrictCreation/src/sagas.js#L37
- Modals: https://github.com/Flowpack/restrict-creation/blob/master/Resources/Private/RestrictCreation/src/manifest.js#L17


packages/neos-ui-sagas/src/index.js
```diff
@@ -0,0 +1,29 @@
...
+    crNodeOperations,
+    crPolicies,
+    publish,
+    serverFeedback,
+    uiContentCanvas,
+    uiContentTree,
+    uiEditPreviewMode,
+    uiInspector,
+    uiPageTree,
+    uiHotkeys
```
skurfuerst: Hey Dmitri,

I somehow dislike that these names will then become "API" in some sense.

I do not yet understand why we need to do this; as one could do the following in a custom manifest:

```
watchPersist = sagasRegistry.get('neos-ui/Changes/watchPersist');
sagasRegistry.set('neos-ui/Changes/watchPersist', overrideWatchPersist(watchPersist);
```

This way the only API is the registry keys in the SagasRegistry. What do you think? :-) Then it does not matter at all if the sagas reside in a custom package or not.

dimaip: Oh, I didn't think about it, a very fair point.
However I'd still like to move Sagas out into a separate package, just not expose it via extensibility API, for the sake of consistency with other packages and just to structure the code a bit more clearly (e.g. highlight decoupled nature of sagas).

dimaip:
Thanks for the review!
I'll split this PR into two actually, was a bit lazy initially, but would be easier to resolve conflicts this way.

dimaip:
Thanks for the review!
I'll split this PR into two actually, was a bit lazy initially, but would be easier to resolve conflicts this way.

PR #2192, opened 2018-10-22
 TASK: load all modals from the registry automatically

dimaip:
The modals are now automatically loaded from the registry, which allows to create custom modals from 3rd-party plugins.

Can be used like this: https://github.com/Flowpack/restrict-creation/blob/master/Resources/Private/RestrictCreation/src/manifest.js#L17


dimaip:
Based on previous reviews.

PR #2193, opened 2018-10-22
TASK: move sagas out into a separate package

dimaip:
I've extracted the sagas into the separate package, for the sake of consistency with the rest of the app's structure.

In addition it exposes `nodeCreationWorkflow`, to allow easier overriding of the addNode saga. Usage example: https://github.com/Flowpack/restrict-creation/blob/master/Resources/Private/RestrictCreation/src/manifest.js#L11


dimaip:
Merge based on previous review.

Issue #2195, opened 2018-10-23
Apply button in right sidebar is always active


markusguenther:
### Description

Since the latest release the apply button is always active.

### Steps to Reproduce

1. Log in to backend
2. Click on a node 
3. Open the right side bar (if it is closed)

#### Expected behavior

By default the discard and apply button is disabled and 50% opaque

#### Actual behavior

Buttons are active

### Affected Versions

Neos: 4.1

UI: 1.4

markusguenther:
Ah and the discard button also ;)
But is fixed in the same way.

PR #2196, opened 2018-10-23
BUGFIX: Adjust button property

markusguenther:
The Button component has the property isDisabled and disabled
has been used.

Fix: #2195


dimaip:
Awesome, thanks for the fix!

PR #2205, opened 2018-10-26
TASK: Disable preview-button when documentnode is hidden

### LINKED ISSUE #2204: Disable PreviewButton when a DocumentNode is hidden
### Steps to Reproduce

1. Hide a DocumentNode
2. Click previewbutton

#### Expected behavior

The preview-button should be disabled because it doesn't make sense to view a page in the frontend which is hidden.

#### Actual behavior

The preview-button is clickable and the page which opens produces an error: `Validation failed while trying to call Neos\Neos\Ui\Controller\BackendController->redirectToAction(). `

### Affected Versions

Neos: 4.1

UI: 1.4.0

mstruebing:
closes #2204

### DIFF
--- a/packages/neos-ui/src/Containers/SecondaryToolbar/PreviewButton/index.js
+++ b/packages/neos-ui/src/Containers/SecondaryToolbar/PreviewButton/index.js
@@ -12,23 +12,35 @@ import {neos} from '@neos-project/neos-ui-decorators';
 }))
 
 @connect($transform({
-    previewUrl: $get('ui.contentCanvas.previewUrl')
+    previewUrl: $get('ui.contentCanvas.previewUrl'),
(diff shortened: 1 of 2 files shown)


markusguenther:
Disabling the Button works good, but the validation issue is still there.

Steps to reproduce:
* hide a page
* apply and publish
* show the page again and apply the change
* do not publish the change and click the preview button.

So when you have not published the change you still get the validation issue.

markusguenther:
Any progress here? Could somebody help out?

mstruebing:
You are right.
I don't know how to fix this atm as I dont have any property to know if  this is the case (hidden and not published) in the store. Maybe someone has an idea?

Sebobo:
As we now have the preview uri we don't need this anymore. It's perfecty fine to preview hidden nodes.

Issue #2130, opened 2018-09-12
Resetting Node Type filter or search expands all nodes


bwaidelich:
### Description

Solving #2069 introduced a small regression that leads to the whole document node tree to be expanded even if the `loadingDepth` should avoid that

### Steps to Reproduce

1. Configure the site as described in #2069
2. Configure the `Neos.Neos.userInterface.navigateComponent.nodeTree.loadingDepth` to be `1`
3. Enter something into the Node Tree search and/or set the Node Type filter
4. Reset the filter & search by clicking the `x`

#### Expected behavior

The Node Tree should be reset to the original state, with all nodes collapsed:

![image](https://user-images.githubusercontent.com/307571/45406210-0749da00-b666-11e8-87b7-cabce7700e8b.png)

#### Actual behavior

All Nodes are expanded:

![image](https://user-images.githubusercontent.com/307571/45406224-13359c00-b666-11e8-95a5-eb05679c414a.png)

### Affected Versions

UI: dev-master (after 9323ad3097f0acec8756cf3dfb0669af3e02d47d)

bwaidelich:
I could track down the issue to the [PageTree Saga](https://github.com/neos/neos-ui/blob/master/packages/neos-ui/src/Sagas/UI/PageTree/index.js#L42) that executes `neosUiFilteredChildren(baseNodeType).getForTree()` even if the filter is reset.
Instead it should probably execute the `neosUiDefaultNodes` operation for an empty filter & search, like we do at [reloadState](https://github.com/neos/neos-ui/blob/master/packages/neos-ui/src/Sagas/CR/NodeOperations/reloadState.js#L16)

bwaidelich:
I tried to rework `Sagas/UI/NodeTree.watchSearch` to switch if `(!searchQuery && !filterNodeType)` but it ended in a little mess. I guess some react pro can tackle this one easily

PR #2209, opened 2018-11-02
BUGFIX: Properly reload the page tree if filter or search is reset #2130

danielkestler:
Solves #2130: page tree is not properly reset after clearing filter or search.


bwaidelich:
I don't know why but it seems to solve #2130 indeed! :)

bwaidelich:
thank you, @danielkestler !

PR #2216, opened 2018-11-06
REFACTOR: [react-ui-components] migrate Frame to TypeScript

JamesAlias:
# TODO
- [ ] evaluate an alternative to `ReactDOM.unstable_renderSubtreeIntoContainer`.
Maybe a Portal (https://gist.github.com/robertgonzales/b1966af8d2a428a8299663b92fb2fe03)

**What I did**
Migrate Frame component and tests to TypeScript.
Changed the API just a liiiittle bit: `onLoad` property now has the same signature as the native HTML Iframe's `onLoad`.

**How I did it**

**How to verify it**
* `make test`
* `make build`
* Open neos backend in browser and login.


JamesAlias:
@dimaip I think we should just review and merge this and open an issue for finding an alternative the `ReactDOM.unstable_renderSubtreeIntoContainer`.

dimaip:
Yes, naturally, I only we can always do it later.
Since all your PRs are based one on top of another, I guess we should just review and merge the last one into the TS branch.
As a side note, it's not really convinient when PRs are sequentially dependant on each other, it could cause conflicts when merging PRs later on.
I'd either make independant PRs, or make just a single PR with multiple commits, so we could review them on commit-by-commit basis.

JamesAlias:
> As a side note, it's not really convinient when PRs are sequentially dependant on each other, it could cause conflicts when merging PRs later on.
> I'd either make independant PRs, or make just a single PR with multiple commits, so we could review them on commit-by-commit basis.

Yes. I'm thinking about a way to make reviewing a good experience. I'm not sure how to achieve that, though :)

But yeah, I can see the problems with the PRs depending on one another.

PR #2217, opened 2018-11-06
WIP TASK: convert User reducer to typescript

dimaip:


packages/neos-ui-redux-store/src/User/Name/index.spec.js
```diff
@@ -1,6 +1,6 @@
 import {Map} from 'immutable';
 
-import {reducer} from './index.js';
+import {reducer} from './index';
```
JamesAlias: can even be just `import {reducer} from '.'`. 😄

JamesAlias: but nevermind that^^

GREAT WORK!
👍

Issue #1214, opened 2017-11-04
Options of the select box editor in the node creation dialog should visible


daniellienert:
(Needs to be said first: The adjustable node creation dialog is a really great feature of the new UI!)

If the selectbox editor is expanded, the options are hidden in the frame - and one need to scroll down. This needs a better solution.

![image](https://user-images.githubusercontent.com/642226/32407613-14f7f1ac-c18c-11e7-91ab-04dae0ca7e73.png)

dimaip:
Yep, been hit by it too.
But in general, I have no idea of how the looks of this node creation dialog could be improved. E.g. now it's obviously too wide. Make it more narrow? Allow to layouts elements inside it in a grid? And do all of this while not over-engineering...

Perhaps good first step would be to make it more narrow and fix this z-index issue.

dimaip:
It's easy to make selectbox jump out of the modal, but then it may be a problem that tall selectboxes would be chopped of by the screen height. So I have no idea how to solve it so far.

markusguenther:
After reverting the last change we need another try ;)

Nikdro:
Reverted?! What was the problem?

markusguenther:
I changed the overflow and that lead to an issue with lot of node-types in the dialog

PR #2221, opened 2018-11-06
Allow select box in create dialog to flow out of the dialog body

markusguenther:
If you configure a create dialog with a select box. The select box options are cutted and you need to find out that you are able to scroll down.

This PR will define a new configuration to limit the visible items in the dropdown content section and let the dropdown overflow the dialog body.

example:
```
'Neos.Neos:Document':
  ui:
    group: 'general'
    creationDialog:
      elements:
        title:
          type: string
          ui:
            label: i18n
            editor: 'Neos.Neos/Inspector/Editors/TextFieldEditor'
          validation:
        country:
          type: string
          ui:
            label: 'Country'
            editorOptions:
              limitRenderedOptions: 5
              values:
                italy:
                  label: 'Italy'
                  group: 'Southern Europe'
                austria:
                  label: 'Austria'
                  group: 'Central Europe'
                germany:
                  label: 'Germany'
                  group: 'Central Europe'
                poland:
                  label: 'Poland'
                  group: 'Central Europe'
                netherlands:
                  label: 'Netherlands'
                  group: 'Central Europe'
                denmark:
                  label: 'Denmark'
                  group: 'Central Europe'
                sweden:
                  label: 'Sweden'
                  group: 'Central Europe'
                norway:
                  label: 'Norway'
                  group: 'Central Europe'
                switzerland:
                  label: 'Swiss'
                  group: 'Central Europe'
                finnland:
                  label: 'Finnland'
                  group: 'Central Europe'
          validation:
```

bildschirmfoto 2018-11-06 um 22 11 40

Closes: #1214


packages/react-ui-components/src/SelectBox/selectBox.js
```diff
@@ -5,12 +5,14 @@ import {$get} from 'plow-js';
...
 import mergeClassNames from 'classnames';
 import isEqual from 'lodash.isequal';
+import {styleConstants} from '@neos-project/build-essentials/src/styles';
 
 // TODO: document component usage && check code in detail
 export default class SelectBox extends PureComponent {
     static defaultProps = {
         options: [],
         optionValueField: 'value',
+        limitRenderedOptions: 0,
```
dimaip: Sorry for the late review, but is there no way to do that without introducing yet one more API prop to SelectBox? It's a beast the way it is now anyways, and I would love to not further complicate it.

markusguenther: Can define a value or height, but thought at the Codesprint that it would be nice to be able to adjust that. No one had objections.

markusguenther: Should we abandon the change or define a fix value?

dimaip: I'd set it to some sane default value and not make it configurable. But does it really need to be limited in height at all?

dimaip: Thinking more about it, as an editor I definitely wouldn't like SelectBox to be limited in height, it's actually nice that you see a lot of stuff:

image

jonnitto: But in the creationDialog in can make sense

markusguenther: Therefore we decided more or less in salzburg that a configuration would be a nice way. But I can also understand @dimaip who wants to keep the complexity as low as possible.

To be honest I don't know another way yet to make it dynamically work without extra configuration.

markusguenther: Should we probably discuss that in the guildUX or directly on discuss. Or do you think that not many will join the discussion?

dimaip: You can give it a try, good idea to ask

markusguenther:
Created a thread on discuss. So hope we get some insights and ideas of the other users.

https://discuss.neos.io/t/selectbox-size-configurable-static-or-something-else/4120

packages/react-ui-components/src/SelectBox/selectBox.js
jonnitto: I know, it is one thing more to configure. *But:* From the UI/UX point of view it makes sense. Especially in the creation dialog. The current behavior isn't very user-friendly. If we don't want to have configurable, it would make sense, to have at least in the creation dialog a maximum height. @danielkestler @dfeyer: What do you think about that?

dimaip: @jonnitto what do you mean by current behavior: that it doesn't flow out or that it becomes too long if it does?
While naturally it should flow out of the dialog body, I'm not really sure if that would be a better UX for the selectbox to be limited in height. How does it help the editor? I personally would love to see as much content as possible.

danielkestler: @jonnitto Yes, absolutely makes sense in the creation dialog. I don't know if we need a setting to limit the items – maybe it should depend on your browser size?

jonnitto: @dimaip Main issue is that it doesn't flow out.

dimaip: @jonnitto well, that's not somethings that we are discussing, it definitely should fly out :) I just don't think we need any special setting for it.

danielkestler: @jonnitto I think the easiest way to solve the problem is to set the height of the box to a sane default like in your screenshot (around 200px height). If you can set it to „dynamically“ adapt to the viewport height, fine, even better. A configuration option should be avoided if possible.

dimaip: Yep, exactly what I was trying to say :)

PR #2223, opened 2018-11-07
FEATURE: Skip 'new element' dialog if there is only one

sbruggmann:
Skip "new element" dialog if there is only one (allowed) Node Type
Issue https://github.com/neos/neos-development-collection/issues/2174

### DIFF
--- a/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
+++ b/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
@@ -167,11 +167,11 @@ export default class SelectNodeType extends PureComponent {
 
     componentDidUpdate() {
         const {insertMode} = this.state;
-        if (insertMode === 'into'
-            && this.getAllowedNodeTypesByCurrentInsertMode().length === 1
-            && this.getAllowedNodeTypesByCurrentInsertMode()[0]['nodeTypes'].length === 1 ) {
+        if (insertMode === 'into' &&
+            this.getAllowedNodeTypesByCurrentInsertMode().length === 1 &&
+            this.getAllowedNodeTypesByCurrentInsertMode()[0].nodeTypes.length === 1) {
             setTimeout(() => {
-                this.handleApply(this.getAllowedNodeTypesByCurrentInsertMode()[0]['nodeTypes'][0]['name']);
+                this.handleApply(this.getAllowedNodeTypesByCurrentInsertMode()[0].nodeTypes[0].name);
             }, 1);
         }
     }
--- a/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
+++ b/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
@@ -165,6 +165,17 @@ export default class SelectNodeType extends PureComponent {
 
     handleNodeTypeFilterChange = filterSearchTerm => this.setState({filterSearchTerm});
 
+    componentDidUpdate() {
+        const {insertMode} = this.state;
+        if (insertMode === 'into'
+            && this.getAllowedNodeTypesByCurrentInsertMode().length === 1
+            && this.getAllowedNodeTypesByCurrentInsertMode()[0]['nodeTypes'].length === 1 ) {
+            setTimeout(() => {
+                this.handleApply(this.getAllowedNodeTypesByCurrentInsertMode()[0]['nodeTypes'][0]['name']);
+            }, 1);
+        }
+    }
+
     render() {
         const {isOpen} = this.props;
 
--- a/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
+++ b/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
@@ -165,21 +165,22 @@ export default class SelectNodeType extends PureComponent {
 
     handleNodeTypeFilterChange = filterSearchTerm => this.setState({filterSearchTerm});
 
-    componentDidUpdate() {
+    skipNodeTypeDialogIfPossible() {
         const {insertMode} = this.state;
         if (insertMode === 'into' &&
             this.getAllowedNodeTypesByCurrentInsertMode().length === 1 &&
             this.getAllowedNodeTypesByCurrentInsertMode()[0].nodeTypes.length === 1) {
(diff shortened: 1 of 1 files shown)


bwaidelich:
@sbruggmann only found this now, you're awesome :)

kitsunet:
Nice!

kitsunet:
And obviously I only wanted to comment, not to close^^

bwaidelich:
I can't judge whether the implementation might have issues but my first tests were quite successful.
To test this, I adjusted the `Neos.Demo:Carousel` node type so that it only allows `Neos.NodeTypes:Image` child nodes.
If I extend the `Neos.NodeTypes:Image` node type to use a `creationDialog` that is displayed directly.

A little drop of bitterness though:
Currently it only works if the carousel is empty. When the first image was added, you no longer select the collection but the image so the insert mode is "after" by default.
So afterwards you get the dialog again even though there is no other option either (except for changing the insert mode to "before" and that's why we can't skip it here..)

packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
```diff
@@ -165,6 +165,17 @@ export default class SelectNodeType extends PureComponent {
 
     handleNodeTypeFilterChange = filterSearchTerm => this.setState({filterSearchTerm});
 
+    componentDidUpdate() {
+        const {insertMode} = this.state;
+        if (insertMode === 'into' &&
+            this.getAllowedNodeTypesByCurrentInsertMode().length === 1 &&
+            this.getAllowedNodeTypesByCurrentInsertMode()[0].nodeTypes.length === 1) {
```
dimaip: I guess this would result in error if there's no allowed nodetype. Usually shouldn't happen, but an extra safety wouldn't hurt, at least till we have TS.

packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
```diff
@@ -165,6 +165,17 @@ export default class SelectNodeType extends PureComponent {
 
     handleNodeTypeFilterChange = filterSearchTerm => this.setState({filterSearchTerm});
 
+    componentDidUpdate() {
```
dimaip: Why `componentDidUpdate` and not `componentDidMount`?

packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
```diff
@@ -165,6 +165,17 @@ export default class SelectNodeType extends PureComponent {
...
+        const {insertMode} = this.state;
+        if (insertMode === 'into' &&
+            this.getAllowedNodeTypesByCurrentInsertMode().length === 1 &&
+            this.getAllowedNodeTypesByCurrentInsertMode()[0].nodeTypes.length === 1) {
+            setTimeout(() => {
```
dimaip: If timeout actually needed here?

packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
sbruggmann: On `componentDidMount` I was missing `this.state.insertMode` and/or the getAllowedNodeTypesByCurrentInsertMode result, I think.

packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
sbruggmann: Then `insertMode` should be blank.

packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
sbruggmann: Without overruling the Modal did not work.
Hacky I know.. but in the worst case (super slow Computer) nothing should happen and the normal Modal is opening.

sbruggmann:
@bwaidelich true, it only applies if it's triggerd on a ContentCollection.
In every other case, you **maybe** want to insert it above the current selection.

In my personal usecases.. inserting could be always set to "below".

packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
dimaip: That's super strange actually... Would be cool to figure out why it happened, but I guess it's ok to leave it like this if you got no time.

packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
dimaip: Ah, makes sense

packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
dimaip: Ok, let's hope that it doesn't accidentally break, but I guess TS would force a check here.

bwaidelich:
@sbruggmann @dimaip @kitsunet Sorry for pinging, but it would be super nice to get this one in the next release IMO. Are there still blockers in your opinion?

dimaip:
I would disagree to only limit it to below, inserting above is also perfectly legal...
But I think that could be left for further improvements, the current functionality is already a big improvement.

bwaidelich:
> But I think that could be left for further improvements, the current functionality is already a big improvement.

Yes, I agree. Limiting to below is not a general option! But apart from that, can this be merged?

packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
dimaip: Could you at least leave a comment here to justify the usage of setTimeout?

sbruggmann: @dimaip okay I checked it again. It was a problem with the asynchronous setState.
This is solved now in the latest commit by executing the check in the render function itself.

sbruggmann:
Please check the latest commit.
I moved the former `componentDidUpdate` actions directly into the render function now.

dimaip:
Thanks for this wonderful contribution, I'm sure it would safe a lot of clicks to the poor editor folk!

Issue #2199, opened 2018-10-24
Validation logic doesn't work the same as old UI and Flow


kitsunet:
In Flow and the old UI a validator could declare to accept empty values and many validators did. That allowed to validate IF a value was entered but still have the field optional (you could pair with an NotEmptyValidator to prevent empty values then). The react UI doesn't seem to respect that. A StringLengthValidator on a node property with a minimum > 0 will error even if the field was empty (but wouldn't error in the old UI).

markusguenther:
Thanks to @sbruggmann for taking care.

bwaidelich:
Related: #2552

PR #2228, opened 2018-11-07
BUGFIX: Don't validate, if the value is empty

sbruggmann:
Solves Issue #2199


bwaidelich:
This one introduced a regression unfortunately: #2552

PR #2235, opened 2018-11-08
TASK: remove Images reducer as it doesn't seem to be used anywhere

dimaip:


dimaip:
@grebaldi why is it there, doesn't seem to be doing anything, no?

PR #2226, opened 2018-11-07
WIP:  REFACTOR: [react-ui-components] migrate TextArea to TypeScript

kayliz:

PR #2238, opened 2018-11-09
REFACTOR: [react-ui-components] migrate TextArea to TypeScript

JamesAlias:
This is the reviewed and updated version of #2226


JamesAlias:
Strangely the `shallow` rendering of enzyme did not work as expected and didn't render the `<TextareaAutosize>` component, but only the `<TextArea>` component.
So I had to use `mount` instead of `shallow`.

JamesAlias:
Re-trigger CI

PR #2242, opened 2018-11-10
TASK: Use correct json indentation in `react-ui-components`

mstruebing:
The components have their own `.ediorconfig` which sets everything to 4
spaces by default but the `package.json` uses two spaces indentation.
The fallback to `[*]` overwrites the global one two space indentation
for `*.json`-files.

**What I did**

Added the `json`-glob to the `.editorconfig` for the `react-ui-components`-package.

**How I did it**

With style.

**How to verify it**

Setup your editor to read the `.editorconfig` correctly and try to indent something in the `package.json` from the `react-ui-components`-package.

### DIFF
--- a/packages/react-ui-components/.editorconfig
+++ b/packages/react-ui-components/.editorconfig
@@ -6,7 +6,7 @@ trim_trailing_whitespace = true
 indent_style = space
 indent_size = 4
 
-[*.{yml,yaml}]
+[*.{yml,yaml,json}]
 indent_size = 2
 
 [*.{md}]


markusguenther:
Looks good to me, but when you have other JSON files inside the package you will also get 2 spaces as rule. Actually we only have coverage reports and the don't use it at all but if you need fixture data or something else as JSON you will have the 2 spaces rule there as well. But that are future problems ;)

markusguenther:
I am fine with that. Just searched for other files with JSON to check if it can break something.
But as I mentioned ... more future issues :)

PR #2244, opened 2018-11-12
REFACTOR: [react-ui-components] migrate DateInput to TypeScript

JamesAlias:
# TODO

- [x] check #1874


package.json
```diff
@@ -143,7 +144,7 @@
     "react-click-outside": "^3.0.1",
     "react-close-on-escape": "^2.0.0",
     "react-codemirror2": "^5.0.1",
-    "react-collapse": "^2.4.1",
+    "react-collapse": "^4.0.3",
```
dimaip: I think this update causes some bug in FF, could you check if it's not the case? https://github.com/neos/neos-ui/pull/1874

JamesAlias: Will do, thanks :)

JamesAlias: reverted it.

PR #2245, opened 2018-11-12
BUGFIX: remove defer from core JS assets

dimaip:
We can't defer those, since currently our extensibility layer relies on JS code being loaded in the right order


dimaip:
Merging as no-brainer, feel free to contribute a better fix.

PR #2248, opened 2018-11-14
TASK: convert CR/Nodes reducer to TS

dimaip:


packages/neos-ui-redux-store/src/CR/Nodes/index.ts
```diff
@@ -0,0 +1,474 @@
...
+import {action as createAction, ActionType} from 'typesafe-actions';
+import {actionTypes as system, InitAction, GlobalState} from '@neos-project/neos-ui-redux-store/src/System';
+
+import * as selectors from './selectors';
+import {parentNodeContextPath} from './helpers';
+
+export type NodeContextPath = string;
+export type NodeTypeName = string;
+
+export interface Node {
```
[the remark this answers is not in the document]
dimaip: No, good catch, totally forgot about it!

tslint.json
```diff
@@ -13,6 +13,7 @@
     "semicolon": [true, "always"],
     "typedef": [true, "call-signature"],
     "variable-name": false,
+    "no-shadowed-variable": false,
```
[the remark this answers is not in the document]
dimaip: There were a few cases where an action was named the same way as an argument of a different action I think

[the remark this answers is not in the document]
dimaip: I guess I'll just add an ignore comment there

packages/neos-ui-redux-store/src/CR/Nodes/selectors.ts
```diff
@@ -0,0 +1,481 @@
+import {$get} from 'plow-js';
+import {createSelector, defaultMemoize} from 'reselect';
+import {GlobalState} from '@neos-project/neos-ui-redux-store/src/System';
+import {parentNodeContextPath} from './helpers';
+import {NodeContextPath, NodeMap, Node, NodeTypeName, ClipboardMode} from '@neos-project/neos-ui-redux-store/src/CR/Nodes';
+
+export interface NodeType {
```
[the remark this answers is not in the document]
dimaip: Yes, I had the same idea actually. It's a kind of WIP state currently.

packages/neos-ui-redux-store/src/CR/Nodes/selectors.ts
```diff
@@ -0,0 +1,481 @@
...
+        };
+        validation?: {
+            [propName: string]: {
+                [propName: string]: any;
+            }
+        };
+    };
+}
+
+// TODO: move to nodetypesregistry itself
```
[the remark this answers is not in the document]
dimaip: Yep, once it is converted to TS I will

packages/neos-ui-redux-store/src/CR/Nodes/index.ts
```diff
@@ -0,0 +1,474 @@
...
+ *
+ * @param {String} contextPath The context path of the node to be removed
+ */
+const remove = (contextPath: NodeContextPath) => createAction(actionTypes.REMOVE, contextPath);
+
+/**
+ * Set CR state on page load or after dimensions or workspaces switch
+ */
+const setState = (
+    {siteNodeContextPath, documentNodeContextPath, nodes, merge}: {
```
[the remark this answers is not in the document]
dimaip: Why not? We always do it everywhere. If feels like a named argument then, e.g. you don't have to memorize the arguments order etc. Anyways, since those action creators are semi-api, so we can't change them easily.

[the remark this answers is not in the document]
JamesAlias: No Problem, just asking for the intention 😉

packages/neos-ui-redux-store/src/CR/Nodes/index.ts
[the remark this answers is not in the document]
dimaip: @JamesAlias I tried making it Readonly, but I got all sort of errors, I think caused by immer.
Somehow similar to this I guess https://stackoverflow.com/questions/51786842/typescript-immer-argument-of-type-is-not-assignable-to-parameter-of-type-draftar/51793994#51793994
I'll try to polish this PR in other directions for now, and get back to making it readonly later (unless you know a quick fix)

packages/neos-ts-interfaces/src/index.ts
```diff
@@ -0,0 +1,141 @@
+export type NodeContextPath = string;
+export type NodeTypeName = string;
+
+// TODO: for some reason (probably due to immer) I can not use ReadonlyArray here
```
JamesAlias: Maybe use `type` here:
```
type NodeChild = {...};
type NodeChildren = ReadonlyArray<NodeChild>;
```
So if you iterate over `NodeChildren` an element has the type `NodeChild` instead of `{....}`. It's a little bit more expressive.

I don't think it will fix the immer issue, though. Let's have a look at this later :)

packages/neos-ts-interfaces/src/index.ts
```diff
@@ -0,0 +1,141 @@
...
+export interface NodeChildren extends Array<{
+    contextPath: NodeContextPath;
+    nodeType: NodeTypeName;
+}> {}
+
+export interface NodePolicy extends Readonly<{
+    disallowedNodeTypes: NodeTypeName[];
```
JamesAlias: Is `ReadonlyArray` not usable here because of `immer`, too?

packages/neos-ts-interfaces/package.json
```diff
@@ -0,0 +1,28 @@
+{
+    "name": "@neos-project/neos-ts-interfaces",
+    "version": "1.4.1",
+    "description": "Neos domain-related TypeScript interfaces",
```
JamesAlias: Good Idea, we can even run test for the `typings` too :)

packages/neos-ts-interfaces/src/index.ts
dimaip: Will give it a try. Perhaps we could even have a pair coding hangout to finalize this PR. I won't have the time today though.

packages/neos-ts-interfaces/src/index.ts
dimaip: Will check.

packages/neos-ts-interfaces/package.json
dimaip: Tests? How? What sort of tests?

packages/neos-ts-interfaces/src/index.ts
dimaip: Yes, it gets me `Type 'NodeMap' is not assignable to type '{ [x: string]: any; }'.
  Index signatures are incompatible.
    Type 'Node' is not assignable to type '{ contextPath: string; name: string; identifier: string; nodeType: string; label: string; isAutoCreated: boolean; depth: number; children: any[]; matchesCurrentDimensions: boolean; properties: any; isFullyLoaded: boolean; uri: string; policy?: any | undefined; }'.
      Types of property 'policy' are incompatible.
        Type 'NodePolicy | undefined' is not assignable to type '{ disallowedNodeTypes: string[]; canRemove: boolean; canEdit: boolean; disallowedProperties: string[]; } | undefined'.
          Type 'NodePolicy' is not assignable to type '{ disallowedNodeTypes: string[]; canRemove: boolean; canEdit: boolean; disallowedProperties: string[]; }'.
            Types of property 'disallowedNodeTypes' are incompatible.
              Type 'ReadonlyArray<string>' is not assignable to type 'string[]'.
                Property 'pop' is missing in type 'ReadonlyArray<string>'.`

packages/neos-ts-interfaces/src/index.ts
dimaip: tslint tells me `Use an interface instead of a type literal. [interface-over-type-literal]`

JamesAlias: We could do 
```
interface NodeChild {...}
type NodeChildren = ReadonlyArray<NodeChild>;
```
but it does not feel consistent somehow..

PR #2251, opened 2018-11-16
REFACTOR: [react-ui-components] migrate ToggablePanel to TypeScript

gradinarufelix:


markusguenther:
Hey, we should use the typescript branch as base. The Panel is not migrated yet so if you have some sparetime to polish this it would be awesome. Otherwise I could hand over and create a new PR based on your commits. So you work will be used in the end :)

kdambekalns:
I think this is way outdated…

PR #2256, opened 2018-11-26
FEATURE: Allow setting startingPoint for LinkInput

nlx-lars:
**What I did**
Made starting point for LinkInput node search configurable.
**How I did it**
I stole the code that determines the context node from the referenceDataLoader: https://github.com/neos/neos-ui/blob/9fa571adba1fb5631e5398886932634a1d043bc1/packages/neos-ui-editors/src/Editors/Reference/referenceDataLoader.js#L109
**How to verify it**
Create a nodetype that uses the new LinkEditor and search for nodes outsite of the current site
```
'Something.Test:NodeThatCanLinkToOtherSites':
  superTypes:
    'Neos.Neos:Content': true
    'Neos.NodeTypes.BaseMixins:LinkMixin': true

  ui:
    label: i18n
    icon: 'link'
    group: 'general'
    inspector:
      groups:
        link:
          label: i18n
          icon: 'link'
          position: 5

  properties:
    link:
      ui:
        inspector:
          group: 'link'
          editorOptions:
            startingPoint: '/sites'

```


packages/neos-ui-editors/src/Library/LinkInput.js
```diff
@@ -80,11 +81,16 @@ export default class LinkInput extends PureComponent {
     };
 
     getDataLoaderOptions() {
+        const contextNode = this.props.options.startingPoint ?
```
dimaip:
```suggestion
        const contextNode = $get('options.startingPoint', this.props) ?
```

dimaip: This should be safer, in case there are no options

nlx-lars: done

markusguenther:
Looks good to me but @dimaip need to verify it and you could probably close and reopen the PR to trigger travis again. Then your tests can pass 👍

Issue #2250, opened 2018-11-15
Extensibility: neos-ui-editors not correctly exported for extensions


kitsunet:
The mappings for extensibility contain

`'@neos-project/neos-ui-editors': '@neos-project/neos-ui-extensibility/src/shims/neosProjectPackages/neos-ui-editors/index'`

but in fact there is no index.js in the neos-ui-editors (/src). So nothing is actually exported. We should fix that to allow embedding existing editors (or parts of them, like secondary inspectors) to be reused.

dimaip:
Yep, the cleanest way is to just split index and manifest files, like in all other extensions.
And add tests for this :)

PR #2257, opened 2018-11-27
BUGFIX: Correctly export all editors for extensibility

kitsunet:
Makes sure the editors and their parts are exported to be used by
UI extensions.

Fixes: #2250

### DIFF
--- a/packages/neos-ui-editors/src/index.js
+++ b/packages/neos-ui-editors/src/index.js
@@ -1,0 +1,96 @@
+import EditorEnvelope from './EditorEnvelope/index';
+
+import {
+    TextField,
+    TextArea,
+    Boolean,
+    DateTime,
+    Image,
+    SelectBox,
+    Link,
+    Reference,
+    References,
+    NodeType,
+    CodeMirror,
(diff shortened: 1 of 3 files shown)


packages/neos-ui-editors/src/index.js
```diff
@@ -0,0 +1,96 @@
...
+    MediaSelectionScreen,
+
+    AssetUpload,
+    AssetOption,
+    LinkInput,
+    LinkOption,
+    NodeOption
+};
+
+export default {
```
dimaip: Why does need to be exported twice? Is there no way to avoid it somehow? Doesn't look DRY...

kitsunet: I think I asked that question about a year ago or so and didn't get a good answer, since then I am just doing it that way (as in other parts of the projects). Back then I didn't find a way to make it work with just one.

dimaip: Duh ok if it's this way everywhere could also be fixed later...

dimaip:
And yes, ideally we should have created tests for this...

Issue #2258, opened 2018-11-28
Moving Documents in and out of subfolders without publishing can detach childNodes since 1.3


mficzel:
When a document is accidentally moved below another document and instantly moved one level up again it looses its childNodes after publishing. This problem usually occurs very easily when editors try to reorder documents. 

### Steps to Reproduce

1. Take a Document with content and child-documents
2. Move it below one of its siblings (like an error during sorting)
3. Move it out again
4. Publish

#### Expected behavior

This should simply work :-)

#### Actual behavior

The site shows and error that no main collection is found and all sub-nodes (contens and documents) are gone aswell but can be found in the database under the original nodePath.

### Affected Versions

Neos: 4.1 (probably 4.*)
UI: 1.3, 1.4

dimaip:
So is this issue resolved now or not? /cc @sbruggmann

mficzel:
It is solved from a ui perspective as soon as it is released. The underlying cr issue is still there but it will hardly be triggered.

dimaip:
Ok, cool, then I'll close this

PR #2259, opened 2018-11-29
BUGFIX: Don't re-generate nodeNames on move if not needed.

sbruggmann:
Partially rolling back changes from #2098 because of issues like #2258, neos/neos-development-collection#2284 and neos/neos-development-collection#2282.

### DIFF
--- a/Classes/Domain/Model/Changes/AbstractMove.php
+++ b/Classes/Domain/Model/Changes/AbstractMove.php
@@ -73,12 +73,12 @@ abstract class AbstractMove extends AbstractStructuralChange
      *
      * @param NodeInterface $parentNode
      * @param NodeInterface $node
-     * @return string
+     * @return string|null
      */
     protected function generateUniqueNodeNameIfNeeded(NodeInterface $parentNode, NodeInterface $node)
     {
         if ($this->contentRepositoryNodeService->nodePathAvailableForNode($parentNode->getPath() . $node->getName(), $node)) {
-            return $node->getName();
+            return null;
         }
 
         return $this->generateUniqueNodeName($parentNode);
--- a/Classes/Domain/Model/Changes/AbstractMove.php
+++ b/Classes/Domain/Model/Changes/AbstractMove.php
@@ -73,7 +73,7 @@ abstract class AbstractMove extends AbstractStructuralChange
      *
      * @param NodeInterface $parentNode
      * @param NodeInterface $node
-     * @return string|null
+     * @return string|null If the current name can be used, null is returned
      */
     protected function generateUniqueNodeNameIfNeeded(NodeInterface $parentNode, NodeInterface $node)
     {
--- a/Classes/Domain/Model/Changes/AbstractMove.php
+++ b/Classes/Domain/Model/Changes/AbstractMove.php
@@ -69,7 +69,7 @@ abstract class AbstractMove extends AbstractStructuralChange
     }
 
     /**
-     * Generate a unique node name for the copied node if needed
+     * Generate a (new) unique node name for the copied node if the current name is in use already
      *
      * @param NodeInterface $parentNode
      * @param NodeInterface $node
--- a/Classes/Domain/Model/Changes/AbstractMove.php
+++ b/Classes/Domain/Model/Changes/AbstractMove.php
@@ -67,4 +67,20 @@ abstract class AbstractMove extends AbstractStructuralChange
         return $this->contentRepositoryNodeService
             ->generateUniqueNodeName($parentNode->getPath());
     }
+
+    /**
+     * Generate a unique node name for the copied node if needed
+     *
+     * @param NodeInterface $parentNode
+     * @param NodeInterface $node
+     * @return string
+     */
+    protected function generateUniqueNodeNameIfNeeded(NodeInterface $parentNode, NodeInterface $node)
+    {
+        if ($this->contentRepositoryNodeService->nodePathAvailableForNode($parentNode->getPath() . $node->getName(), $node)) {
+            return $node->getName();
+        }
+
+        return $this->generateUniqueNodeName($parentNode);
+    }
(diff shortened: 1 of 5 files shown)


dimaip:
Sorry, I can not provide a review since I don't have a slightest clue about this part of the code. /cc @kitsunet @kdambekalns @bwaidelich and all ya clever CR people :)

Classes/Domain/Model/Changes/AbstractMove.php
```diff
@@ -67,4 +67,20 @@ protected function generateUniqueNodeName(NodeInterface $parentNode)
...
+     *
+     * @param NodeInterface $parentNode
+     * @param NodeInterface $node
+     * @return string
+     */
+    protected function generateUniqueNodeNameIfNeeded(NodeInterface $parentNode, NodeInterface $node)
+    {
+        if ($this->contentRepositoryNodeService->nodePathAvailableForNode($parentNode->getPath() . $node->getName(), $node)) {
+            return $node->getName();
+        }
```
mficzel: Should'nt this simple return 'null' if no new name is needed.

mficzel: Other than that the PR works for the cases i see.

sbruggmann: `null` ends with the same result but is better more transparent :+1:

Classes/Domain/Model/Changes/AbstractMove.php
```diff
@@ -67,4 +67,20 @@ protected function generateUniqueNodeName(NodeInterface $parentNode)
         return $this->contentRepositoryNodeService
             ->generateUniqueNodeName($parentNode->getPath());
     }
+
+    /**
+     * Generate a unique node name for the copied node if needed
+     *
+     * @param NodeInterface $parentNode
+     * @param NodeInterface $node
+     * @return string|null
```
bwaidelich: I'm always very skeptical when it comes to mixed return types. It probably makes sense here, but we should document what `null` means exactly

bwaidelich: ..and what "if needed" means in this case

bwaidelich:
I left some (not very helpful) comments but just from reading the code it is not clear to me what this actually fixes. Can't we add a test that breaks without the fix?

Classes/Domain/Model/Changes/AbstractMove.php
```diff
@@ -67,4 +67,20 @@ protected function generateUniqueNodeName(NodeInterface $parentNode)
         return $this->contentRepositoryNodeService
             ->generateUniqueNodeName($parentNode->getPath());
     }
+
+    /**
+     * Generate a unique node name for the copied node if needed
```
kdambekalns:
```suggestion
     * Generate a (new) unique node name for the copied node if the current name is in use already
```

Classes/Domain/Model/Changes/AbstractMove.php
kdambekalns:
```suggestion
     * @return string|null If the current name can be used, null is returned
```

Classes/Domain/Model/Changes/MoveAfter.php
```diff
@@ -43,7 +43,7 @@ public function apply()
             $before = self::cloneNodeWithNodeData($this->getSubject());
             $parent = $before->getParent();
 
-            $nodeName = $this->generateUniqueNodeName($this->getSiblingNode()->getParent());
+            $nodeName = $this->generateUniqueNodeNameIfNeeded($this->getSiblingNode()->getParent(), $this->getSubject());
```
bwaidelich: What about moving the logic here to make this a little more explicit? So along the lines of:
```
if ($this->requiresNewNodeName(...)) {
  $nodeName = $this->generateUniqueNodeName($this->getSiblingNode()->getParent());
  $this->getSubject()->moveAfter($this->getSiblingNode(), $nodeName);
} else {
  $this->getSubject()->moveAfter($this->getSiblingNode());
}
```

kdambekalns:
Maybe @nlx-lars can test this as well!?

Classes/Domain/Model/Changes/MoveAfter.php
kdambekalns: I like!

bwaidelich:
Apart from the nitpicks above I wonder if this truely fixes the bug: If you moved a node forth and back and during that a rename _was_ required (because a name was already taken) we'd still end up in an invalid state, right? But maybe there's no easy way around that (except for event sourcing *g)

sbruggmann:
@bwaidelich Correct! **If** two nodeNames collide, or if it's a copy instead of a move, the basic bug still applies.

This is only a fix that the problem in the CR is less critical. But it's still there.

nlx-lars:
> Maybe @nlx-lars can test this as well!?

This does fix the problem for me as well.

mficzel:
@bwaidelich if you trigger a renaming because you had a collision in the node name (13 char randomness) AND move the node back to the old parent node afterward you will still trigger the faulty cr behavior. Currently this is triggered whenever a node is moved to another parent and back before publishing. So this does not fix the cr bug, but will not expose it that much.

bwaidelich:
@sbruggmann @mficzel thanks for clarifying :)

Classes/Domain/Model/Changes/MoveAfter.php
sbruggmann: Sure I can do that if you wish, I see the benefits of the explicit style.
But this needs to be done three times and then it's just duplicated code.

kdambekalns:
That build failure should be solveable by `yarn update` to get rid of that disappeared dependency. YMMV. :)

kdambekalns:
Ok then…

PR #2268, opened 2018-12-07
WIP REFACTOR: [react-ui-components] migrate SelectBox to TypeScript

JamesAlias:
Master PR for everything SelectBox related.


JamesAlias:
@dimaip I'm not sure how to wait for the select box options to appear before counting them. I can figure it out myself but then again, I don't have much time these days.
The `selector-timeout` is too short for (uncached) select box options to be loaded.
How do I listen to the loading state of the select box options?
Can I listen to Redux actions or do I somehow observe the Redux state?

I wouldn't be too sad if you would take over here and we do a final review (maybe even with Sebastian) when all tests are done. I am pretty busy right now and need too much time writing those e2e tests.

packages/react-ui-components/src/SelectBox_Header/selectBox_Header.tsx
```diff
@@ -82,6 +82,8 @@ export default class SelectBox_Header extends PureComponent<SelectBox_Header_Pro
 
         return (
             <span>
+                // TODO: themr looses default props type information of components
```
JamesAlias: TODO: When exporting the component enhanced with themr, just reset the type to the original component type with `export default themr(...) as Component`.

dimaip:
@kdambekalns Why close?

kdambekalns:
Let me know if you need help reopening this and/or resurrecting the branch. 😬

dimaip:
Would be nice to do something about it so we don't loose valuable work by @JamesAlias...

markusguenther:
Hmm, that was quite a lot of work, but over the time it is hard to refactor this.

JamesAlias:
I propose the following approach: 

I'll close the PR, but I'll keep the branches.

I really, really want to get back to Neos ASAP, but the corona situation just didn't allow it over the past years and still continues to do so (sandstorm and family have my highest prio).

When I eventually do get the time to work on Neos UI again, I will use my current branches as guidelines but will start the TS conversion from scratch. This way I can start with a fresh fork and won't unintentionally revert current fixes & features.

Also I got so much better at TS & React and both tools greatly evolved since I wrote that code. This years old PR is just sooo outdated 🙈 😅

PR #2270, opened 2018-12-16
BUGFIX: fix activePresetsSelector

dimaip:


dimaip:
Merging as no-brainer

PR #2272, opened 2018-12-19
BUGFIX: fix a few inspector regressions

dimaip:


packages/neos-ui/src/Containers/Modals/UnappliedChangesDialog/index.js
```diff
@@ -87,13 +88,15 @@ export default class UnappliedChangesDialog extends PureComponent {
...
                     <Button
+                        id="neos-UnappliedChangesDialog-resume"
                         style="lighter"
                         onClick={this.handleResume}
                         className={`${style.button} ${style.resumeButton}`}
                         >
                         <I18n id="Neos.Neos:Main:content.inspector.unappliedChangesDialog.button.default"/>
                     </Button>
                     <Button
+                        id="neos-UnappliedChangesDialog-apply"
```
markusguenther: It is more a nitpick but neos-unappliedChangesDialog-apply looks better or is the uppercase U important?

markusguenther: Ok then it makes sense. Just looked a bit weird.

Issue #2273, opened 2018-12-19
Get rid of react-close-on-escape


markusguenther:
### Description

We have the outdated dependency with react-close-on-escape, but only use this in the dialog component. So we discussed in the upgrade PR that it would be better to remove the dependency.

#### Expected behavior

Dialogs also closes on escape without the package.

#### Actual behavior

Wrapper component handles the close on escape by the react-close-on-escape package.

Neos:

4.2

UI:

1.4.1.

PR #2274, opened 2018-12-19
TASK: Replace react close on escape

markusguenther:
**What I did**

Removed the dependency react-close-on-escape and implemented the behavior in the dialog component itself.

**How to verify it**

Open a dialog (create dialog for instance) and close via escape still works.

Closes: #2273


markusguenther:
Thanks for merging @dimaip.
We can close this PR as well ;) https://github.com/neos/neos-ui/pull/2162

PR #2279, opened 2019-01-02
BUGFIX: fix a number of smaller regressions

dimaip:


packages/neos-ui-redux-store/src/UI/Inspector/index.ts
```diff
@@ -98,7 +98,8 @@ export const reducer = (state: State = defaultState, action: InitAction | Action
             }
             const focusedValues = draft.valuesByNodePath[focusedNodePath];
             if (focusedValues) { // dummy type guard
-                if (transientValueDiffers) {
+                // if hooks are defined for given property, the value should not be cleared even if it doesn't differ
+                if (transientValueDiffers || hooks) {
```
dimaip: The reason why it worked in the old UI was that the value was an immutablejs object, so the condition would always pass

Issue #2296, opened 2019-01-24
Dynamically hide properties in the inspector doesn't work anymore


jonnitto:
### Description

Dynamically hide properties in the inspector doesn't work anymore

### Steps to Reproduce

Create a node property described here:
https://github.com/neos/neos-ui/pull/2118

#### Expected behavior

The property should show/hide depends on the other property

#### Actual behavior

Stays hidden

### Affected Versions

Neos: `4.2.4`

UI: `dev-master`

dimaip:
Figured it out, will fix

jonnitto:
You can give a report when it is done, I'll test that with joy :-D

PR #2298, opened 2019-01-25
BUGFIX: fix ClientEval

dimaip:
Fixes: #2296


jonnitto:
Did you already make an upmerge? Because I'm testing on master (Neos 4.2)

jonnitto:
I've tested it on Neos 4.2.4, works like a charm ❤️

PR #2300, opened 2019-01-25
BUGFIX: prevent &nbsp; on empty content from CKE5

dimaip:


jonnitto:
Tested it on 4.2.4: Works on auto paragraph and plain text inputs

pmaas:
Maybe related to this: if you add a whitespace to the end of an inline editable property, a `&nbsp;` is visible inside inspector text inputs.

bildschirmfoto 2019-02-08 um 23 16 34

dimaip:
No, it's not related, always has been the case. Don't know what would be correct way to fix it though.

PR #2302, opened 2019-01-25
BUGFIX: Disallowed NodeTypes should be delivered as array

kitsunet:
The UI expects the ``disallowedNodeTypes`` in policy information
as a simple array of strings. Due to the way `array_map` works in
PHP the keys of the array could be strings which would result in
a JSON object representation, breaking the UI.


dimaip:
Classic :)

PR #2307, opened 2019-01-26
TASK: Remove legacy backend switch leftovers

daniellienert:
At the beginning, switching between old and new UI was possible
using a button in the backend.
Since Version 1.0 this button is gone and new UI is used when it is installed. 
With this PR the backend code for this feature is removed.


dimaip:
Wait, but isn't it still possible to switch to the old UI without uninstalling the new UI completely?

daniellienert:
@dimaip: That is still possible. As I wrote in the description: Some time ago, there was a button in the user menu to switch back from new to old ui. This button is gone, so also the used PHP code can be removed. 
As the change is only made in the UI package, it doesn't change anything if the UI package is not installed (obviously :) ) and you still get the old UI then.

dimaip:
@daniellienert I was asking if it was possible to switch **without** uninstalling the package, i.e. via the switch. Totally forgot that we removed it completely...

Classes/Aspects/NodeTypeSchemaCacheHeaderAspect.php
```diff
@@ -1,49 +0,0 @@
-<?php
```
dimaip: Wait, why is this removed? As I see, it's currently fixed in Neos 4.0, so Neos 3.3. would still be affected by the lack of caching, no?

daniellienert: Well, but this is targeted to master, which requires Neos 4.0. Do I miss something?

dimaip: I'm blind

Issue #2328, opened 2019-02-01
Inspector TextAreaEditor doesn't show stored content


sbruggmann:
### Description

Even if the property has a value, the `Neos.Neos/Inspector/Editors/TextAreaEditor` is shown as an empty TextArea.

### Affected Versions

Neos: 3.3 +?
UI: 2.0 +?

### Last working Version

Neos: 3.3
UI: 1.4.1

PR #2329, opened 2019-02-02
BUGFIX: fix value in TextArea element

dimaip:
Fixes: #2328


sbruggmann:
Thank you for fixing this.
Works now as expected.

dimaip:
Yup, no problem!

Issue #2313, opened 2019-01-28
Icon overlay in Image editor is too small


jonnitto:
The icons in the image inspector are too small:

### Version 1.4.1
<img width="301" alt="image" src="https://user-images.githubusercontent.com/4510166/51873329-0be50580-235d-11e9-8adc-39048505a709.png">

### Version  3.0.0

<img width="302" alt="image" src="https://user-images.githubusercontent.com/4510166/51873256-c1638900-235c-11e9-8b4a-8b5091d67fe6.png">
<img width="295" alt="image" src="https://user-images.githubusercontent.com/4510166/51873372-2fa84b80-235d-11e9-82cd-5a649cfa200d.png">

### Affected Versions

Neos: `4.2.4`

UI: `3.0.0`

jonnitto:
Did not test on 2.0.0. Have to search for a Neos 3.3

dimaip:
Thanks! No need to test 2.0 separately, they are almost identical.

markusguenther:
The size attribute for the icons doesn't seem to work anymore. Therefore the icon is not 5x like intended. Try to solve it.

markusguenther:
Couldn't we close the issue now? Or what needs to be achieved for closing an issue with the new branching?

PR #2331, opened 2019-02-02
Bugfix: Icon overlay in Image editor is too small

markusguenther:
The image editor icon was too small and the css class for the bigger icon (5x) did not work.
This patch extracts the font-size from the combined font style and make it work.

Fixes: #2313

### DIFF
--- a/packages/react-ui-components/src/Icon/style.css
+++ b/packages/react-ui-components/src/Icon/style.css
@@ -1,8 +1,8 @@
 .icon {
     composes: reset from './../reset.css';
     display: inline-block;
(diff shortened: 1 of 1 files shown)


dimaip:
Thank you!

markusguenther:
Ah ok. Then I will create a new PR for 2.x later this day. Just need to verify it first, but I bet that it is also broken in 2.x

dimaip:
Ohh I missed that it was opened against master. PR template explicitly stated that it was to be opened against 2.x...

dimaip:
2.x and master are almost identical, no need to verify it, it's broken for sure

PR #2340, opened 2019-02-08
BUGFIX: MediaDetailsScreen shows original asset

ComiR:
In the media browser `ImageVariant`s are hidden, so we won't show them here either (but the original `Image`).

This may have a side effect on custom `AssetVariant`s. They will be turned into their original asset counterpart too. But since it is just for the `MediaDetailsScreen`, I think this won't be an issue for nearly everyone.


dimaip:
Looks good for me, but could be great to get a review from somebody who actually understands assets like @kitsunet

kitsunet:
I would rather not do that now. Robert is working on something in that regard and IMHO the asset detail view is bad there anyway as IMHO it has no real use. what do you do if you clicked an image and end in the detail view?

dimaip:
So what do we do about this?

ComiR:
Well, although @robertlemke may be working on an overhaul, for the current version, this is probably more in line with the expectations.
Up until the update, I'd like to see this merged.

dimaip:
So what do we do about this one @kitsunet?

kdambekalns:
It actually fixes a bug, so good to have that in.

PR #2343, opened 2019-02-10
TASK: Restyle modal dialogs, especially select node type dialog

danielkestler:
This is probably a controversial change... I restyled the select node type dialog to look like a grid of nodetypes, and to match the collapsible heading stylings of the inspector (border on the top, not on the bottom).

Maybe it needs more iteration, but what do you think about the general idea?

Before:
bildschirmfoto 2019-02-10 um 15 21 10

After:
bildschirmfoto 2019-02-10 um 15 20 40


jonnitto:
I personally like this approach.

danielkestler:
Okay, well, found out there is probably more to do. I added contextual variants for the dialogs in general, and added a heading to the create dialog (was missing until now).

I know the changes are currently not passing the tests, maybe because of position changes? Will take care later on.

bildschirmfoto 2019-02-11 um 09 02 32

bildschirmfoto 2019-02-11 um 09 02 42

bildschirmfoto 2019-02-11 um 09 02 51

dimaip:
I like the general direction of this!
Things to keep in mind:
- help messages
- responsive behavior, e.g. how it'd look on mobile

I'm no design expert, but I'm somehow irritated by how the "close" icon on modals is getting a red border from only two sides. Maybe it could be padded? Or just somehow completely reworked, dunno.
image

The same goes for various borders and shadows around the secondary buttons: 
image

danielkestler:
@dimaip Yes, there's indeed some strange things going on. Will have a look at it.

Responsive behaviour is a bigger topic. Currently, the backend is not really responsive in general, at least not down until mobile. There's an open issue floating around regarding this: https://github.com/neos/neos-ui/issues/1716 Perhaps this should be tackled in a separate PR.

dimaip:
Just for reference, how other projects do the same task:

Wordpress:
image

Squarespace:
image

Tilda.cc:
image

Wix:
image

Webflow:
image

daniellienert:
Hey guys,
I really like the new style of the node type selection. Looks much cleaner this way. I absolutely prefer the new arangement of the "insert into / untder / above" selector, which is overseen so often by editors.  According to @dimaips researches, seems like the blocks are common sense :) I guess the help messages could work as before?

Regarding the close button - I would remove the "X" from most/all of the dialogs anyway. It is kind of a legacy Windows behaviour to have a "I don't want to read the message - just close" button. We have meaningful buttons for that.

dimaip:
@danielkestler I think while we are at it, we should at least make sure this dialog degrades properly for mobile, e.g. the number of grid columns could progressively lower.

I reflected a bit more on the design and realized that it feels a bit "stripy" to me, e.g. there are so many horizontal lines in the current design, would have been cool to make it look lighter, more like Wordpress or Squarespace. Wdyt?
image

dimaip:
Also what I like about WP's solution is that they center the filter and make it focused by default, visually prompting that "hey, you probably want to filter in most cases, don't waste time visually searching"

danielkestler:
Thank you all! @dimaip Nice overview, thanks. Seems we are on the right track this way ... I also like your idea with the filter and maybe removing some lines or make it look less cluttered. Will think about a way to achieve this without compromising on usability.

danielkestler:
I like the Webflow tiles very much, they look much more detailed than our Font Awesome icons, which are a bit too abstract. But if we wanted to go a more illustrative route, we need a way to define this for the integrators (nodetypes yaml file) – which leads to a big burden for them having to create assets like that.

danielkestler:
Cleaned things up a bit. Cannot get the filter input to autofocus, I try setting setFocus={true} but it doesn't work.

bildschirmfoto 2019-02-11 um 14 10 55

dimaip:
Nice, looks cleaner indeed.
But moving insert position to the right is definitely a no-go, actually it's the most important piece of information in the whole dialog.
If you don't have sudden new ideas regarding placement of filter, perhaps it's better to leave it where it was before...

danielkestler:
You're right. Changed it back.

Restyled and refactored the "unapplied changes" dialog:
bildschirmfoto 2019-02-11 um 15 34 24

It's better now, but needs a proper title. Where do the language labels come from?

danielkestler:
I‘m struggling to get the integration tests passing. It has something to do with onRequestClose and the removal of the Close-Buttons... but I only removed the assignment by props, because it is not needed in the render()-method and linting fails if I leave it in.

Anyone has an idea? 😉

danielkestler:
BTW, we should really think about all the labels in our UI. See this screenshot: "Confirm" is not a good button label, it should be "Discard all" or similar, so that the user really triggers the action he wants.

bildschirmfoto 2019-02-12 um 08 26 38

danielkestler:
Removing the icon color for now for consistency reasons, as suggested by Jens Hoffmann on Slack. Needs some time to get used to it, but absolutely makes sense.

bildschirmfoto 2019-02-12 um 08 40 40


Still failing to build ... this time:
Error in fixture.afterEach hook -
AssertionError: expected 'Warning: Each child in an array or iterator
should have a unique "key" prop.

iseries:
I just have a question: What happend if there are help informations (`ui->help->message` or `ui->help->thumbnail`) for a NodeType? Where is the help dialog shown now and where to "trigger" the help information?

btw. I like your restyles so much! Keep up the good work.

danielkestler:
@iseries I will work on it, currently I have not yet tested the help messages ;-)

danielkestler:
Okay, here you go. Can probably be further improved by highlighting which element the help message is for ...

bildschirmfoto 2019-02-12 um 11 30 03

dimaip:
Awesome, looks much cleaner without blue icons!

daniellienert:
@danielkestler. I would like it more, if the help message is closer to the element where it beongs to. Would something like that be possible / better?

image
(Rough Photoshop Mockup)

danielkestler:
@daniellienert This would be the best way to do it ... I will have a look, as I currently don't know, where/when/how this help box is generated.

dimaip:
Before we used to show it in a popup tooltip.
It's rendered here https://github.com/neos/neos-ui/blob/master/packages/neos-ui/src/Containers/Modals/SelectNodeType/nodeTypeGroupPanel.js#L101

danielkestler:
@daniellienert Sadly I currently see no possibility to do this without it getting "hacky". But I'm open for ideas and maybe we could just do it afterwards in a new PR (as the current state of this PR is a lot better than master).

I don't get what the integration test error means. Perhaps somebody could help me on this one? Thanks :-) 

Here's a screenshot with the change a few commits ago for the unapplied changes dialog:
bildschirmfoto 2019-02-18 um 10 52 10

packages/react-ui-components/src/Dialog/style.css
```diff
@@ -63,6 +63,9 @@
 .dialog__actions > * {
     margin-left: 2px;
 }
+.dialog__closeBtn {
```
dimaip: Hey, just noticed this, please don't do such hacks.
Also, you are remove X close icon on all dialogs, not only on the insert node dialog. Is it intended?
I'm no UX expert, but I think some people might be used to it.

danielkestler: Thank you, will have a look!

dimaip:
Regarding the failing E2E tests, I think the errors reported are rather clear: there are proptype warnings in console;

This code might need adjusting, because you removed `<span>` from buttons: `.click(ReactSelector('NodeTypeItem').find('button>span').withText('Headline'));`

No idea why this doesn't work, try to run the tests locally and see if the element is really there: `.click(Selector('#neos-ContentTree-AddNode'))`

dimaip:
If my explanation about TestCafe is not enough, we could have a call and I could introduce you to it a bit more.

dimaip:
There's a gap of 1px between buttons, but there's no gap between buttons and the dialog border.
I'd either remove all the gaps, or add the same 1px gap between the buttons and the dialog border.
Wdyt?

image

danielkestler:
Thank you @dimaip. Will try to resolve the test issues. You are right, the one you mentioned is indeed quite obvious, but I don't know where those "Error in fixture.afterEach hook" errors came from.

Added the close buttons back in for now, as they do no harm. Also removed the button gaps.

packages/react-ui-components/src/Dialog/style.css
danielkestler: Bring back the close button ;-) Maybe we can remove it later on if we really don't need it, but it does no harm if it is there.

danielkestler:
Okay, fixed the errors you mentioned, but the other ones remain.

dimaip:
As said, our TestCafe setup is done in the way that it would fail on any `console.error` message, including React warnings. That catches quite some regressions.

See:
image

danielkestler:
I am stupid. Found it. Accidentially passed the "Create new" title as array.

danielkestler:
Okay, I'm done. Feel free to merge at any time if you like.

danielkestler:
Here are the final screenshots:

bildschirmfoto 2019-02-19 um 16 13 11

bildschirmfoto 2019-02-19 um 16 13 30

bildschirmfoto 2019-02-19 um 16 13 38

bildschirmfoto 2019-02-19 um 16 13 44

johannessteu:
Once again, great change!! 

Just some ideas:
1) What about `Create new Document` and `Create new content` as a title for the modal?
bildschirmfoto 2019-02-22 um 12 15 10

2) Personaly i'd use the gap between the buttons. Probably also raise it to 2px so it's more clear that there are several click events / buttons. 

3) Isn't the red border on delete modals "too much"? Shouldn't the red button to confirm etc. be enough?

What do you think about a switch to toggle the element sice between this card layout and a list layout to get everything more compact? It could be handy if you're working on a notebook/small screens

dimaip:
1. More like `Create new document node`/`Create new content node`. Or is it `element` and not `node` in end user terminology?

I personally think a switch with different layouts would be too much. Actually the current layout takes up almost the same amount of space, as it's narrower than before.

danielkestler:
@johannessteu Thank you. Regarding your points:

1. We could change it, but the labels are provided by Neos.Neos-package, right? And currently, the two node selection dialogs are basically the same (but there must be a distinction somewhere, because they show different nodetypes of course).

2. You mean a gap around the buttons, so that a gap between the dialog border and the button appears? IMO the clou of the current proposed design is that the primary button of a dialog blends with the border, and the border highlights the primary action.

3. see 2 ... but yes, if you want, I could change it back.

dimaip:
Regarding 2 and 3, I think we shouldn't pressure you too much here, and I don't have anything against it. I'd say do as you feel it appropriate.

danielkestler:
Okay, I propose to merge it as it is now (I have at least changed to label from "confirm" to "discard" in the discard dialog), and do the other language-related ideas in a next PR (and a PR in the Neos.Neos-package – or we add it to the Neos.Neos.Ui-package language files ...).

PR #2344, opened 2019-02-12
BUGFIX: Actually import and expose all editors for extensibility

kitsunet:
In the bugfix PR #2257 it was planned to expose all editors to the
extensibility layer as was planned originally. Unfortunately the
import path for components was not changed resulting in a totally
broken state at the moment. This change tries to rectify the situation
by adjusting the path. It has been tried in a project that uses the
``LinkOption`` component in a custom editor, which would fail without
this change.


kitsunet:
Given that the current state of export is just broken this shoudl be merged and released ASAP I think.

kitsunet:
<3 Thanks

PR #2347, opened 2019-02-13
TASK: Style scrollbars

danielkestler:
Styled the scrollbars in a slim, unobtrusive way. Hard to capture in a screenshot, but makes everything look a bit more elegant.

Tried the primary blue before, and while it certainly looked cool, it made them to catchy visually.

bildschirmfoto 2019-02-13 um 15 42 41

On the left: the small bar is part of the scrollbar of the content canvas
On the right: scrollbar of the inspector panel

### DIFF
--- a/packages/neos-ui-guest-frame/src/style.css
+++ b/packages/neos-ui-guest-frame/src/style.css
@@ -46,3 +46,7 @@
     padding: 5px;
     border: 1px solid gray;
 }
+
+:global body::-webkit-scrollbar-track {
+    background-color: var(--colors-ContrastDarkest) !important;
+}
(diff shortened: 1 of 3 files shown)


dimaip:
I like that it has more contrast now, always found it to be so annoying that we had such low contrast there.
But is the main scrollbar wide enough to be usable?

danielkestler:
Ah, you mean with the mouse? ;-) Well ... I always thought this scrollbar is more a "visual indication" than an actual element to click and drag. But if you want, we could make it bigger for the content canvas.

dimaip:
@danielkestler yeah I actually know people who drag it...
And even myself when I'm working with a mouse and need to quickly scroll back and forth a long page I use a scrollbar, because it works quicker and gives more control.

I'd make content canvas scrollbar bigger, and leave other scrollbars small.

kitsunet:
Same here, I actually use scroll bars ;)

Issue #2348, opened 2019-02-13
"Apply changes" dialog should respect inspector validation


maya-shines:
If you leave the URL path segment of a page empty, a validator will pop up, informing you that "this property is required". If you're stubborn enough and click "Apply" anyways, the url path segment will be set to an empty string. You are then immediately redirected to the parent page and can't change it anymore in the backend.

danielkestler:
Similar issue occurs if you manually change the slug to be identical to a sibling page. Then you can forever only load one of the two pages. But this may be a Neos core issue.

dimaip:
Can't verify on latest 2.x

kdambekalns:
With UI 1.3.4 and a field that has validation, neither "Apply" in the inspector, nor "Apply Changes" in the dialog are enabled. This seems to affect only versions > 1.x…

PR #2350, opened 2019-02-14
BUGFIX: fix UnappliedChangesDialog disabled states

dimaip:
Fixes: #2348


dimaip:
Wait, I don't think it's the right way to fix it

Issue #2349, opened 2019-02-14
Properties should be backend-validated


daniellienert:
Currently defined property validators are only evaluated in the frontend and can be circumvented there.

PR #2351, opened 2019-02-14
BUGFIX: Use backend validators if available

daniellienert:
This change resolves the according flow validator to a configured
property validator. If it exists, it validates the result.
This is non braking, as it returns true, if the validator is not
found so custom validators are not affected.

Resolves: #2349, #2348

### DIFF
--- a/Classes/Service/NodePropertyValidationService.php
+++ b/Classes/Service/NodePropertyValidationService.php
@@ -55,7 +55,7 @@ class NodePropertyValidationService
     protected function resolveValidator(string $validatorName, array $validatorConfiguration)
     {
         $nameParts = explode('/', $validatorName);
-        if (!$nameParts[0] === 'Neos.Neos') {
+        if ($nameParts[0] !== 'Neos.Neos') {
             $this->logger->log(sprintf('The custom frontend property validator %s" is used. This property is not validated in the backend.', $validatorName), LOG_INFO);
             return null;
         }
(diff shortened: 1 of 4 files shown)


Classes/Service/NodePropertyValidationService.php
```diff
@@ -0,0 +1,63 @@
...
+class NodePropertyValidationService
+{
+
+    /**
+     * @param $value
+     * @param string $validatorName
+     * @param array $validatorConfiguration
+     * @return bool
+     */
+    public function validate($value, string $validatorName, array $validatorConfiguration): bool
```
dimaip: Is this compatible with all the PHP versions that Neos 3.3 supports?

daniellienert: Thx for the hint, this one is, but https://github.com/neos/neos-ui/pull/2351/files#diff-b1f15cbe77fdc0aceaae2512fd4a3c12R48 isn't due the nullable return type.

Tests/Functional/Service/NodePropertyValidationServiceTest.php
```diff
@@ -0,0 +1,57 @@
...
+    public function resolveValidator()
+    {
+        $validator = $this->nodePropertyValidationService->_call('resolveValidator', 'Neos.Neos/Validation/NotEmptyValidator', []);
+        $this->assertInstanceOf(NotEmptyValidator::class, $validator);
+    }
+
+    /**
+     * @test
+     */
+    public function validate()
```
dimaip: I'd also add a test case for validator which is missing on the server side. Would help to understand the required behavior during future refactorings. But fine to merge as is ofc.

daniellienert: Thanks. Done and fixed a bug on the way.

PR #2352, opened 2019-02-14
BUGFIX: deprecate isDisabled button state

dimaip:
With this change we deprecate the usage of `isDisabled` prop and fix `disabled` prop. `isDisabled` would continue to work, but would show a console warning.

A little background. `isDisabled` used to be the official way to get the disabled state, but we still passed all the props down to the `<button>` tag, which also made `disabled` prop work, and actually that's what we used ourselves in a lot of the places.
All other components use `disabled` as a prop name in similar scenario. Also it's how it's called in html. So I propose to rename it, which would be much easier than renaming `disabled` => `isDisabled` in all other places.

Fixes: #2348


dimaip:
Come on guys, I need help with this one, it's quite a critical decisions to merge it blindly.

skurfuerst:
@dimaip from what I understand, totally makes sense <3 Go ahead!

dimaip:
Thanks for the timely review! :heart:

PR #2364, opened 2019-02-18
TASK: Restyle main modules, move selected node dropdown out of the tabs area

danielkestler:
I know this is probably ultra controversial, but I just had to do it because it currently makes no sense: moved the selected node dropdown out of the tab area and combine it with the inspector toggling. Could of course be further improved, but as a first draft this solves some UX-issues regarding this dropdown.

(Gotta do something similar on the left hand side to get rid of the "navigate" button which also does not make sense ...?)

bildschirmfoto 2019-02-18 um 13 11 21


daniellienert:
Thanks! Absolutely makes sense to me! Never noticed, that this selector is rendered to every tab.
image

jonnitto:
Another benefit is that the collapse inspector button and the tabs get separated.

danielkestler:
Okay guys, this could be getting bigger :-D 
Adding more consistency overall and making room for the Neos logo next to the menu icon.

Marking this as WIP, as there is more to do.

bildschirmfoto 2019-02-18 um 14 52 29

bildschirmfoto 2019-02-18 um 14 52 33

danielkestler:
That's it for today, hope you like it and get an idea of the general direction. Needs proper labels etc., but that's easy.

bildschirmfoto 2019-02-18 um 16 58 36

bildschirmfoto 2019-02-18 um 16 58 39

dimaip:
Wow, really cool how you come to the same ideas as we've had here for ages.

1) Move selected element out of tabs -- 100% support this.
2) trees -- I understand your idea, but there are a few downsides to it: a lot of wasted vertical space and also again this "stripy" feeling, currently counted 8 horizontal rows in this design.

Btw, another often requested feature is to be able to maximize the content tree to the full height, e.g. when editors are working for a long time on the same page with a lot of content.

Another problem I've often noticed: having two identical toolbars editors often click on the wrong toolbar. E.g. an editor wants to delete a content node but intuitively clicks without thinking on to the delete button from the top toolbar and deletes the whole document node.

danielkestler:
Thank you @dimaip! Tried streamlining it a bit and it now feels very natural and elegant to me. 

Regarding your points:
1. The trees ... shouldn't feel as stripy now. Maximizing the content tree would be a cool option, should add that in a later PR.
2. Confusing toolbars ... Hmm, never happened to me or my editors. But I think as there are now headings to the sections ("Navigate" and "Content tree"), it should get more clear, wdyt?

bildschirmfoto 2019-02-19 um 08 51 39

bildschirmfoto 2019-02-19 um 08 51 42

dimaip:
Wow! You are really genius at making these little tweaks!

1) The "strippiness" problem is totally gone! It looks really awesome how you put all those widgets like inspector and the trees on the black background.
My brain stumbles upon the logical mismatch between labels: "Toggle content tree", "Navigate" and "Selected element". They look visually identical, but do completely different things.

"Toggle content tree" acts as a label to the toggle button.
"Navigate" looks like a caption to the page tree, though it toggles the whole left sidebar.
"Selected element" happens to just be on the same line with the toggle (and in itself it's totally fine).
One way to fix it could be to write "Document tree" and "Content tree".

And still there remains a problem of wasted vertical space, especially on smaller screens (some of our editors work on *768 laptop screens...). If the main reason why you introduced the "Toggle content tree" panel was to separate the trees visually, another way to do it could be to make the tree button toolbars black, and leave the content tree toggle where it was. But really, I'm totally ignorant in visual area, so feel free to ignore my remarks here.

danielkestler:
@dimaip Thanks for your comments, highly appreciated. 

Yeah, we absolutely have to change the labels there. Ideally, the labels would be part of the toggle button. What do you think about the following labels:

1. "Toggle content tree" => "Content Tree"
2. "Navigate" => "Document Tree"
3. "Selected element" => "Element Inspector"

How do we change these labels? As they come from the Neos.Neos package ... I will have a look if there are any appropriate labels already there, but in the long term it would be better to not depend on the other package for language labels, right?

dimaip:
> it would be better to not depend on the other package for language labels, right?

It was this way before, and it was a mess in terms of translation workflow (we translate labels in https://crowdin.com/). So unless somebody comes up with a really good plan I'd leave things as they are for now.

> I saw that there are also translation files in the UI-package, so I will just add it there

I think those are just leftover and we currently use the ones from the core, no?

> What do you think about the following labels
:+1:

danielkestler:
> I think those are just leftover and we currently use the ones from the core, no?

Yes, realised it after commenting. Hmm.

package.json
```diff
@@ -101,9 +101,9 @@
     "@ckeditor/ckeditor5-paragraph": "^10.0.3",
     "@ckeditor/ckeditor5-table": "^11.0.0",
     "@fortawesome/fontawesome-svg-core": "^1.2.2",
-    "@fortawesome/free-brands-svg-icons": "^5.6.0",
-    "@fortawesome/free-regular-svg-icons": "^5.6.0",
-    "@fortawesome/free-solid-svg-icons": "^5.6.0",
+    "@fortawesome/free-brands-svg-icons": "^5.2.0",
```
dimaip: btw, why does it downgrade fontawesome?

jonnitto: We could even upgrade it to 5.7

danielkestler: No it does not exactly downgrade, it was like this before. After upgrading, it threw errors.

danielkestler: But I'd love to update to 5.7

johannessteu:
Hey @danielkestler ! Such a great effort! Just checked it out local. Looks great! I really like the direction of this PR! 

The only think i'd probably like to discuss the the toggle for the left and right sidebar. It's kinda behaves like it is right now. But what about some more infos next to the arrow?

bildschirmfoto 2019-02-22 um 12 03 59

If we would keep the `Element Inspector` next to the arrow? Could this help the user for a better understanding what he actually toggles?

bildschirmfoto 2019-02-22 um 12 03 55

dimaip:
Perhaps a tooltip is enough for this case?
IMO `Element inspector` is more like a caption for the panel below, and not for the toggle button itself (in that case it would have been more like `Toggle element inspector`).

danielkestler:
@johannessteu @dimaip Thank you. The caption can be both – labeling the panels, and also labeling what will be toggled. I could try to leave the labels in if collapsed ...

danielkestler:
What can we do to get this merged? Any changes required?

Sebobo:
I'm testing it now and think overall it's really really cool change!

Two things I personally think need improvement:

* We already have little space in the inspector and many view are dense. With this change we loose another few pixels because of the additional padding on the left (right for the navigation panel). Therefore I would suggest to remove the padding there to give more space to the panel content again.
* Give the structure tree 50% of the height again. It's rather small and has the additional header bar.

I couldn't find any errors so far 👍

dimaip:
> With this change we loose another few pixels because of the additional padding on the left

I agree with Sebastian, but on the other hand I see what you are trying to achieve visually, so if you see no other way to solve it, I guess we could sacrifice those few pixels (or maybe even make the sidebar a few pixels wider).

On the other hand, I'm still much more sad to see the vertical space wasted for the content tree title, on smaller screens it becomes quite critical. E.g. 1366 x 768 screens are still quite popular among editors...
But yet again, if you see no other way around it, let's go for it. Just test at 1366 x 768 one more time before merging :)

danielkestler:
Hi, thanks for your feedback. I will try to implement the changes by @Sebobo.

@dimaip I don't think the content tree title as a "wasted space", as it is such a big UX improvement, especially for Neos-newcomers, that it outweighs the cons. But I will have a look at the whole picture again to see if we can do something to improve it ;-)

danielkestler:
Here you go.

Sebobo:
Love it!

dimaip:
@danielkestler could you please try to rebase on top of 2.x?

Resources/Private/Translations/en/Main.xlf
```diff
@@ -191,6 +191,15 @@
 			<trans-unit id="searchBoxLeftToType" xml:space="preserve">
 				<source>Please enter ###CHARACTERS### more character</source>
 			</trans-unit>
+            <trans-unit id="Containers.LeftSideBar.DocumentTree.heading" xml:space="preserve">
```
dimaip: I don't think we use labels from here, instead they need to be put into Neos.Neos, as that is translateable through crowdin.
Alternatively we could move all content module labels from Neos.Neos into Neos.Neos.Ui and setup the Crowdin integration.

danielkestler:
> @danielkestler could you please try to rebase on top of 2.x?

Okay, now at least the error makes sense. One has to toggle the search bar first ... I fixed this in the old e2e tests but it maybe got lost during the circle CI transition.

Resources/Private/Translations/en/Main.xlf
danielkestler: Okay, moved them out of the UI package ;-)

dimaip:
I see some strange commits in this PR, but I guess we could just squash-merge it once the tests pass.
Could you also please create the final screenshot and post it to #guild-ux for the very final approval (since it's a big visual change)?

danielkestler:
Here is the final screenshot: 
Bildschirmfoto 2019-03-26 um 08 53 39

Resources/Private/Translations/en/Main.xlf
dimaip: Btw, if by any chance you (or somebody else reading this) wants to take care of sorting the Translations mess, that would be cool. Things to do:
1) Discuss with the team how we want to proceed: 
  a) delete everything from Neos.Neos.Ui and leave everything in Neos.Neos (easy, just `rm -rf Resources/Private/Translations`)
  b) move all content module labels to Neos.Neos.Ui and setup crowdin integration for Neos.Neos.Ui (hard)
2) do it

johannessteu:
i like it!

PR #2372, opened 2019-02-19
BUGFIX: Same image cannot be selected twice

### LINKED ISSUE #2367: Same image cannot be selected twice
### Steps to Reproduce

1. Select an image
2. Select the image again

#### Expected behavior

The image is selefcted

#### Actual behavior

Spinner spins forever

Neos: 4.2

UI: 3.x

reopening same image fails

danielkestler:
Just close the secondary inspector if you select the same image twice, closes #2367.

### DIFF
--- a/packages/neos-ui-editors/src/Editors/Image/index.js
+++ b/packages/neos-ui-editors/src/Editors/Image/index.js
@@ -171,10 +171,16 @@ export default class ImageEditor extends Component {
     }
 
     handleMediaSelected = assetIdentifier => {
-        const {commit} = this.props;
+        const {commit, value} = this.props;
         const newAsset = {
             __identity: assetIdentifier
         };
+
+        // Same image selected as before?
+        if (value && value.__identity === assetIdentifier) {
+            this.handleCloseSecondaryScreen();
+            return;
+        }
 
         this.setState({
             image: null,


markusguenther:
Looks good to me and fixes the issue. Sad that we have no tests for this component.

dimaip:
Yup, good point, would be nice to add e2e tests for this.
Also I wonder if there's less hackish fix possible, but don't have the time to dive deep into it.

danielkestler:
> Also I wonder if there's less hackish fix possible, but don't have the time to dive deep into it.

Hey ;-) While not exactly hackish, it indeed works around the actual problem by "not triggering it". Don't know why there is a problem to set the image twice. Could have another look ...

dimaip:
@danielkestler sorry didn't mean to sound rude... I meant that it's a bit scary to merge it as is without understanding why this workaround is needed. Could backfire in the future somehow.
But if you feel confident about it and see it as a legit fix we can also merge it as is.

markusguenther:
Hope that I am not too tired in the evening hours and look into it... there has to be an backend reason for this. So if I find it we maybe find a better solution. But after a day on the construction I am sometimes  completely exhausted.

johannessteu:
I think it's due to the cwrp here:
https://github.com/neos/neos-ui/blob/2.x/packages/neos-ui-editors/src/Editors/Image/index.js#L88

This line will evaluate false if the same image is selected. And then we're missing `this.handleCloseSecondaryScreen();` We could adjust the cwrp function, but i think the pr is good to go and probably the correct place to fix this!

johannessteu:
Also the select media selects the image in state to null:
https://github.com/neos/neos-ui/blob/2.x/packages/neos-ui-editors/src/Editors/Image/index.js#L179

And since the image state is not set within the cwrp we're running in this problem!

dimaip:
Cool, thanks for looking into it!

Anyone wanting to contribute an E2E test? Otherwise will merge.

markusguenther:
@dimaip lets create an issue for the e2e test. Only team members can append commits to the PR.
So even if I want to add some tests I need to create a new PR on top of this PR.

dimaip:
Sure, no problem

PR #2375, opened 2019-02-20
TASK: lazily expose node variants information

dimaip:
Since #2164 got blocked, I extract the logic for loading node variant information into this PR, so it would be possible to experiment with node variants selector in a separate package.

I refactor `getPolicyInformationAction` into `getLazyNodeDataAction`, a place where we'd be able to fetch all node information, that is expensive to calculate and load, I'm sure it'd be useful in the future.

### DIFF
--- a/Classes/Controller/BackendServiceController.php
+++ b/Classes/Controller/BackendServiceController.php
@@ -21,6 +21,7 @@ use Neos\Flow\Mvc\ResponseInterface;
 use Neos\Flow\Mvc\View\JsonView;
 use Neos\Flow\Persistence\PersistenceManagerInterface;
 use Neos\Neos\Domain\Service\ContentContextFactory;
+use Neos\Neos\Domain\Service\ContentDimensionPresetSourceInterface;
 use Neos\Neos\Service\PublishingService;
 use Neos\Neos\Service\UserService;
 use Neos\Neos\Ui\ContentRepository\Service\NodeService;
@@ -112,6 +113,12 @@ class BackendServiceController extends ActionController
      * @var NodeClipboard
      */
     protected $clipboard;
+
+    /**
+     * @Flow\Inject
+     * @var ContentDimensionPresetSourceInterface
+     */
+    protected $contentDimensionsPresetSource;
 
     /**
      * Set the controller context on the feedback collection after the controller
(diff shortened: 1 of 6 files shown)


danielkestler:
Are you reading minds? I think this is exactly what @kitsunet, @davidspiola and me were talking about in our call this Monday. Great, from this point on we can experiment with or sketch different solutions for the problem.

dimaip:
@danielkestler hah no, it was our plan with @davidspiola since ages :)

Configuration/Routes.Service.yaml
```diff
@@ -79,8 +79,8 @@
   httpMethods: ['GET']
 -
   name: 'Get Policy Info'
-  uriPattern: 'get-policy-info'
+  uriPattern: 'get-lazy-node-data'
```
kitsunet: I feel the naming is off here, we should probably try and find a better wording.

`get-extended-node-information` or `get-additional-node-metadata` or such?

kitsunet: Same goes for all other isntances of `lazy` ofc.

Classes/Controller/BackendServiceController.php
```diff
@@ -393,25 +400,51 @@ public function loadTreeAction(NodeTreeBuilder $nodeTreeArguments, $includeRoot
...
         /** @var NodeInterface $node */
         foreach ($nodes as $node) {
-            $result[$node->getContextPath()] = ['policy' => $this->nodePolicyService->getNodePolicyInformation($node)];
+            $otherNodeVariants = array_values(array_filter(array_map(function ($node) {
+                return $this->getCurrentDimensionPresetIdentifiersForNode($node);
+            }, $node->getOtherNodeVariants())));
+            $result[$node->getContextPath()] = [
+                'policy' => $this->nodePolicyService->getNodePolicyInformation($node),
+                'dimensions' => $this->getCurrentDimensionPresetIdentifiersForNode($node),
+                'otherNodeVariants' => $otherNodeVariants
```
kitsunet: Wait so this contains only dimension presets in which variants exist? not even information about those variants. It seems to me super expensive to do this for every node just to get the "filled presets".

kitsunet:
I am still very vary about this, especially combining it with policies. We need the policies quickly, adding more stuff makes the whole thing slower. IMHO the variants are convenience nothing more. As we discussed monday I would really like to defer loading that information until you want to look at the variant relations for a single node and then loading it for that node.

Configuration/Routes.Service.yaml
dimaip: Yeah, I could come up with a good name quickly... `get-additional-node-metadata` sounds OK.

Classes/Controller/BackendServiceController.php
dimaip: What kind of information could we load here? We already have dimension presets information in the UI. That's really all we need.
Is there cheaper way to get this info?

dimaip:
Hi, @kitsunet! Thanks for the review!

> We need the policies quickly, adding more stuff makes the whole thing slower.
Lazy-loading it after focusing the node would kill the UX, according to our discussion with @davidspiola this information is quite critical for editors in  multi-dimensional sites.

Policy information is extremely expensive, I'm sure just getting node variants would be nothing in comparison to this. But I can do some measurements. What would you consider a good testcase for this assumption? How many node variants should we test for?

dimaip:
@kitsunet I fixed the naming.

Regarding performance, I did some measurements. On the demo site it takes 0.04s to calculate all this information in development context. The request in total takes 1.42s. That's about 3% slowdown.
I'd propose to not do premature optimization further.
We really need this data to be available in the UI to do further work in #guild-ux, delaying it may kill the momentum.

In the future if we'd want to optimize the performance further, I think it should be done on the CR API side: it should provide a cheap way to know all node variants of a node.

dimaip:
@kitsunet thanks a lot for taking the time to dig into the subject and for your thorough review :heart:

PR #2380, opened 2019-03-03
Data source optional caching

mireo91:
**What I did**
Added option to disable caching of data source.
**How I did it**
By looking for key 'disableCaching' in 'options.dataSourceAdditionalData'. Only when set to false or not existing key the cache should be set.
**How to verify it**
- for SelectBox editor use dataSource as source of options
- to disable default caching set:
              dataSourceAdditionalData:
              &nbsp;&nbsp;disableCaching: TRUE


Sebobo:
Hi @mireo91,

thanks for your PR!

It looks you created the PR from the 3.x branch which contains many more changes than you actually want to add. Can you recreate or change the PR on the 2.x branch? 

Thanks,
Sebastian

PR #2381, opened 2019-03-06
BUGFIX: disregard baseNodeType for search without preset

dimaip:
This PR fixes regression caused by https://github.com/neos/neos-ui/commit/9323ad3097f0acec8756cf3dfb0669af3e02d47d

Neos used to be able to search for all document nodes, disregarding `baseNodeType` setting. It makes a lot of sense, e.g. in projects where there are too many nodes to be shown in the tree at the same time, but yet you'd want to make them searchable.

An alternative could be to introduce a separate setting like `searchBaseNodeType` for tree presets, but I'd propose to start with restoring the original behavior, and introduce yet another setting if it would really be demanded by real projects.

### DIFF
--- a/packages/neos-ui-sagas/src/UI/PageTree/index.js
+++ b/packages/neos-ui-sagas/src/UI/PageTree/index.js
@@ -140,7 +140,6 @@ export function * watchSearch({configuration}) {
         yield put(actions.UI.PageTree.setSearchResult(result));
 
         const {contextPath, query: searchQuery, filterNodeType} = action.payload;
-        const effectiveFilterNodeType = filterNodeType || configuration.nodeTree.presets.default.baseNodeType;
 
         yield put(actions.UI.PageTree.setAsLoading(contextPath));
         let matchingNodes = [];
@@ -150,7 +149,7 @@ export function * watchSearch({configuration}) {
             const query = q(contextPath);
 
             if (filterNodeType || searchQuery) {
-                matchingNodes = yield query.search(searchQuery, effectiveFilterNodeType).getForTreeWithParents();
+                matchingNodes = yield query.search(searchQuery, filterNodeType).getForTreeWithParents();
             } else {
                 const clipboardNodeContextPath = yield select($get('cr.nodes.clipboard'));
                 const toggledNodes = yield select($get('ui.pageTree.toggled'));


bwaidelich:
Thanks for fixing the regression.
As mentioned on slack already, I have some reservations to just reverting it though because it would lead to a worse experience in large projects that would have to search through _all_ document nodes now. Also the search result might yield only results that the user is not interested in.
E.g. we have something like 50k imported data that contains all kind of contents but we use dedicated presets to filter/find those.

Having said that, of course I won't block this one (especially since it was me who broke the behavior to begin with)

dimaip:
Hey @bwaidelich, thanks for your feedback again!

> but we use dedicated presets to filter/find those.

The problem with this, if you use a filter, it would load _ALL_ those 50000 nodes before you are able to search for them. How does it work for you?

I'm in no way insisting on this particular way to fix the problem, but so far I only see two alternatives:
1) restore the old behavior
2) add an extra `searchBaseNodeType` setting

Any other ideas?

bwaidelich:
> How does it work for you?

We don't define a filter that shows the imported data because (IMO) the tree is not suited to display such huge amounts (and it killed Neos when users accidentally switched to the wrong filter).
Instead we created flat list for those (based on your awesome [FlatNav](https://github.com/psmb/Psmb.FlatNav and extended with a simple search).

I understand that this can't be the default solution, especially because the flat navigation is no core feature (yet).

> so far I only see two alternatives [...]

Yes, I can't think of another one as well and I would prefer the second solution (extra setting) because of the mentioned issues with the default behavior.

Maybe we should start a discuss post for others to share their opinion on this one?

dimaip:
@bwaidelich I could try to implement that extra setting, though I cry every time I add yet another setting to the configuration :)

I'm lazy to create discuss posts, I'll just drop a link to this discussion in #teams now :D

dimaip:
> and extended with a simple search

Btw, maybe you'd like to contribute the search functionality back to FlatNav? That would be quite cool!

kdambekalns:
Huh. 🤷‍♂️

- I'd expect to exclude only stuff I never want to see in the tree, so… I'm fine with the way it is now.
- I can understand such a behavior might not fit others, so I'd vote for the setting, even though adding settings sucks.

bwaidelich:
> I cry every time I add yet another setting to the configuration

yes, I know.. It sucks. maybe there are other options (e.g. allow filter and search term to be specified at once or some kind of pagination for many nodes on the same level – so that it doesn't load the whole universe into the tree)

> maybe you'd like to contribute the search functionality back to FlatNav?

@kitsunet implemented that and we would probably need to ask the customer for permission, but this shouldn't be an issue.

dimaip:
So there are possible fixes to this:

1) EASY (and ulgy). Add the setting.
2) MEDIUM. Add search to FlatNav and add FlatNav into the core. That would solve the problem.
3) HARD. Develop a super cool tree component with pagination and all bells and whistles.

bwaidelich:
> Add search to FlatNav and add FlatNav into the core

I'm happy to help there and IMO it doesn't even _have_ to be a core package. IIRC our fork just needs a little bit of fine tuning, styling tweaks and some documentation. I hope that we can soon share it back upstream.

> Develop a super cool tree component with pagination and all bells and whistles

Right.. That's something we have on the plate for quite a while already (https://github.com/neos/neos-development-collection/issues/1058, https://github.com/neos/neos-development-collection/issues/1099, ..)

PS: I'm on vacation now for 2 weeks but I'm happy to help resolving this one afterwards!

dimaip:
Ok, so I'll close this for now

PR #2382, opened 2019-03-08
BUGFIX: Add missing legacy formatting options

daniellienert:
Feature switches for formating options del / u were not converted to their new names.


dimaip:
Anyways, I'll merge it for now, but the code migration would have been nice as well.

daniellienert:
@dimaip I’m just on it!!

PR #2383, opened 2019-03-08
FEATURE: Remember active tab in inspector

Sebobo:
With this change the sidebar will remember
which inspector tab was selected last and
restore the correct tab when switching between
nodes. Even if the order of the tabs is different.

**What I did**

Instead of storing the active tabs index I store its id which is unique.

**How I did it**

Just store the different property.

**How to verify it**

Select the document node and the SEO tab, then switch between content nodes and the document node and see that the tab will be restored.


Sebobo:
I could modify it to insert the index as id if no id is set. Then it would behave the same. And make the id optional therefore.

Sebobo:
After some thought, I'm not sure if it's sensible to support both behaviors. 
The code needed is a bit more complex than I thought as supporting the index AND id creates special cases in several areas. Like what happens if the id is not set everywhere etc...

Big question is: Is anyone really using this as standalone right now?

dimaip:
> Is anyone really using this as standalone right now?

Hey, I really doubt that, but who knows?
Perhaps we could permit ourselves to slip in this breaking change into a minor release then?

Sebobo:
@dimaip I changed the code based on our discussion. Have a look.

packages/react-ui-components/src/Tabs/tabs.tsx
```diff
@@ -106,8 +107,8 @@ export default class Tabs extends PureComponent<TabsProps> {
 
         return (
             <div className={theme!.tabs__content}>
-                {React.Children.map(children, (panel, index) => {
-                    const isActive = activeTab === index;
+                {children.map((panel, index) => {
```
dimaip: What's the difference between `React.Children.map` and `children.map` I keep on forgetting?

Sebobo: I forgot again, but I had an issue with React.Children, I think it couldn't access something from the outside scope.

PR #2385, opened 2019-03-12
FEATURE: Add support for asset sources in asset editor

kitsunet:
This change adds asset sources support to the Asset Editor, which
allows for searching and importing remote assets on the fly.

This change relies on the new Asset Proxies web service endpoint in Neos.Neos.

It will not allow to link to (non imported) assets from other sources as that doesn't really work well with the current linking concept.


kitsunet:
Fixing the TS errors....

dimaip:
Makes sense!
The code looks good too, at least at first sight.

kitsunet:
The test failure seems unrelated, probably the branch too far behind or so

dimaip:
Oh yes it surely needs upmerging!

kitsunet:
I did merge master into it 2 days ago. it should be fine to merge really. I don't get why the push check fails but the pr check does not...

dimaip:
Did you upmerge 2.x into master before doing it?

robertlemke:
@dimaip shouldn't 2.x upmerges happen separately from new features? I'm not familiar with the current workflow for the Neos UI.

Anyway, I tried an upmerge, but there are too many conflicts which I wouldn't dare to solve, because I didn't follow what was happening in 2.x recently.

image

robertlemke:
To be honest, someone who is more familiar with what recently happened in Neos UI should solve the conflicts for the upmerge. For example, I'm not sure how to proceed with this one here:

image

dimaip:
> shouldn't 2.x upmerges happen separately from new features?

It's exactly as it is in neos-development-collection. I usually do upmerges before doing releases. I'll try to look into merge conflicts on Monday.

dimaip:
FYI: I've merged 2.x into master recently, don't know if it helps here.

kitsunet:
This is finally green after flushing the travis caches. seems it was some old package versions that didn't upgrade easily....

PR #2393, opened 2019-03-18
FEATURE: Allow redux store extensibility

Sebobo:
Currently the combineReducers function in `neos-ui-redux-store` 
will reset the structure of the global state after each action.

With this change extensions to the ui can store
their own values in `plugins.myExtension.myValue`
without it getting lost after any action.

**What I did**

Add a separate `plugins` reducer that just returns itself without
checking or changing anything.

The naming can be changed of course and it will need documentation.

**How to verify it**

Add another reducer in your extensions manifest which writes a value that does not exist yet inside the `plugins` namespace.
Trigger any action in the Neos backend and check if the value is still there.

### DIFF
--- a/packages/neos-ui-redux-store/src/index.ts
+++ b/packages/neos-ui-redux-store/src/index.ts
@@ -27,7 +27,9 @@ export const reducer = combineReducers({
     cr: CR.reducer,
     system: System.reducer,
     ui: UI.reducer,
-    user: User.reducer
+    user: User.reducer,
+    // NOTE: The plugins reducer is UNPLANNED EXTENSIBILITY, do not modify unless you know what you are doing!
+    plugins: (state) => state || {}
 });
 
 //


Sebobo:
Hi @skurfuerst,

I stumbled upon this problem while trying to integrate Yoast with the Neos UI. 
I need to store several objects which currently get lost as soon as I trigger any action on the redux store.

Dmitri and I discussed this during the Sprint in Lucerne and came up with this approach which works for my use case.
We now would like your feedback to this issue before I continue putting more effort into it :)

Thanks!

dimaip:
I think the solution that we came up so far is not very clean, but we couldn't think of anything better.

skurfuerst:
sorry, I don't get yet how you use this or how this can work? Could you add a small example? :)

Sebobo:
@skurfuerst here is my actual implementation https://github.com/Sebobo/Shel.Neos.YoastSeo/blob/master/Resources/Private/Scripts/YoastInfoView/src/actions/index.js

From how I understood the unplanned extensibility I should be able to store something in the global state. This currently does not work, when it's not a preexisting property of the Neos core.

In my code I currently store it additionally in a closure and refill the global state after every action to not loose my properties.

skurfuerst:
Now I got it. Yeah, totally makes sense <3 :) :) Thanks for explaining!

PR #2399, opened 2019-03-20
TASK: Rollback the FEATURE: Skip 'new element' dialog if there is only one

sbruggmann:
The recently introduced Feature https://github.com/neos/neos-ui/pull/2223/files led to irritations on editors side. Let's remove this and someday replace it with other optimizations like: https://github.com/neos/guild-ux/issues/10

### DIFF
--- a/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
+++ b/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
@@ -172,23 +172,11 @@ export default class SelectNodeType extends PureComponent {
 
     handleNodeTypeFilterChange = filterSearchTerm => this.setState({filterSearchTerm});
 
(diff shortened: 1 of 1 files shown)


rolandschuetz:
Really? I likes this feature a lot.

Maybe just show some kind of toast "Creating new X node" and show it for 3 seconds, so that it's clearer for users?

sbruggmann:
I think it's better to normalize the behaviour again, without bloating the code base for such a small feature.
It would be better to invest more effort into general optimizations like https://github.com/neos/guild-ux/issues/10

What do you think? @rolandschuetz

dimaip:
Yeah I agree. The same button shouldn't just suddenly change the behavior.

rolandschuetz:
I understand your reasoning. I feel that this small change made working way faster with things like galleries. But both are fine for me.

PR #2403, opened 2019-03-21
TASK: refactor e2e testing setup

dimaip:
TODO:

- [x] migrate the rest of the tests
- [ ] don't flush all caches


johannessteu:
What about the circle ci stuff btw?

dimaip:
@johannessteu on my list for tomorrow, this took a bit longer then expected...

.circleci/config.yml
```diff
@@ -53,15 +53,16 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - full-app-cache-{{ checksum "Build/CircleCi/composer.json" }}
+            - full-app-cache-{{ checksum "Tests/IntegrationTests/TestDistribution/composer.json" }}-{{ checksum "Tests/IntegrationTests/TestDistribution/Configuration/Settings.yaml" }}
```
johannessteu: Why do we need the Settings.yaml here? The cache will only affect packages installed via composer and we overrite the settings.yaml with the one from the repo anyways, right?

dimaip: but what if we change the Settings.yaml e.g. as I did for Flow  fix?

johannessteu: Hmm, would need to test actually. But i thought the change should be there as it's stored in git. Or maybe we would have to change the order? first restore cache and then run the checkout command. Anyway - doesn't hurt at all to get the checkusm

Issue #2366, opened 2019-02-18
boolean editor closes creation dialog


jonnitto:
**I've created a repo with this issue to reproduce this bug: https://github.com/jonnitto/Neos.Issues**

### Description

If I have a `BooleanEditor` in a `creationDialog` and a user click on it, the creation dialog gets closed immediately.

![2366](https://user-images.githubusercontent.com/4510166/53009734-6e04c800-343c-11e9-9a83-797e3cb3e496.gif)

### Steps to Reproduce

1. Create a `creationDialog` with a `BooleanEditor`
2. Create the element 
3. Click on the checkbox

#### Expected behavior

The creation dialog stays open

#### Actual behavior

The creation dialog gets closed immediately.

### Affected Versions

Neos: 4.28
UI: 3.1

danielkestler:
I don't know why, all checkboxes close their parent dialog or dropdown if clicked. Same in publish dropdown.

dimaip:
Probably smth to do with clickoutside behavior...

markusguenther:
![neos-create-dialog](https://user-images.githubusercontent.com/1014126/54710508-2448f380-4b48-11e9-8831-bdbc3cd9d1ae.gif)

I tried to reproduce that for document types and content but does not made it to reproduce it. Maybe it has been fixed while overhauling the UX.

I used latest Neos with UI version on 2.x branch.

markusguenther:
Would be nice if @jonnitto can try it again.

jonnitto:
Yes, I will

jonnitto:
@markusguenther I've updated the repo https://github.com/jonnitto/Neos.Issues to dev-master of neos-ui. Is dev-master on the same level as the 2.x branch?

But the bug is still here: 
![ScreenFlow](https://user-images.githubusercontent.com/4510166/54718388-af32e980-4b5a-11e9-9660-d61b312e36ca.gif)

markusguenther:
Thanks for the testing distribution. That helped a lot. Just a small explanation what happens.
Actually the BooleanEditor always use the identifier as ID and when you have the same field in the dialog and the inspector you will control the checkbox in the inspector and not in the dialog.

Only when you click directly on the checkbox and not on the label it will work in the dialog.
The PR is on the way and I hope that solves the issue :)

PR #2406, opened 2019-03-21
BUGFIX: Prevent closing of the node create dialog

markusguenther:
Adding a random suffix to the checkbox identifier to
prevent to close the node identifier dialog.

The problem is that the BooleanEditor don´t add a random string to the identifier of the checkbox. When you now have the right inspector bar open and the create dialog, you will have two checkboxes with the same identifier. The selectbox in the right inspector wins because this box was rendered first and therefore the dialog gets closed.

You don´t change the checkbox in the dialog, you change the checkbox in the inspector.

Fixes: #2366


markusguenther:
Thanks to @jonnitto ... your test distribution helped a lot.

jonnitto:
Very nice, thank you!

dimaip:
Ohh that's a nasty bug, good catch!
Is there no way to go without `htmlFor`, e.g. just by wrapping a <label> around?

markusguenther:
The BooleanEditor uses this property but in the end the id will be the same in the dom and therefore it will occur. Or I don’t understand you.

dimaip:
I just mean we could get rid of ids completely.

markusguenther:
Ah so Surround the checkbox by the label and then get rid of the id and for in the rendered markup. Get your point ... will change it later today :)

markusguenther:
@dimaip Adjust that... thanks for the hint. Is much clearer and easier now.

PR #2409, opened 2019-03-21
TASK: Add circle ci config

dimaip:


.circleci/config.yml
```diff
@@ -0,0 +1,118 @@
...
+          key: full-app-cache2-{{ checksum "Build/CircleCi/composer.json" }}
+          paths:
+            -  /home/circleci/app
+
+  e2e:
+    environment:
+      FLOW_CONTEXT=Production
+    docker:
+      - image: quay.io/yeebase/ci-build:7.2
+      - image: circleci/mariadb:10.2
```
dimaip: What does it me that there are two images here @johannessteu?

johannessteu: It's running 2 docker containers. The first one is the container where the steps are executed and the second is a mariadb that is running and can be used

dimaip:
Thank you @johannessteu! :heart_eyes:

PR #2419, opened 2019-03-25
TASK: add more browsers to e2e setup

dimaip:


markusguenther:
For firefox we need also adjustments I guess.

kdambekalns:
This can be revisited if needed, but I'll close this for now.

PR #2426, opened 2019-03-27
BUGFIX: Fix broken extensibility webpack config

johannessteu:


dimaip:
We should create e2e tests for extensibility ASAP!

PR #2430, opened 2019-03-27
BUGFIX: Make tabs work without setting ids

Sebobo:
Keeps API compatibility with previous 2.x and 3.x versions of the UI.


Sebobo:
Yes that's why I didn't want to do it in the beginning...

PR #2432, opened 2019-03-27
TASK: Upgrade d3-scale to version 3

markusguenther:
Like some other packages d3-scale is outdated and we are actually using version 2 instead of 3.
We actually only use it in the TimeSeriesView but to be honest I don´t know if it would be easily possible to replace this by an own implementation.

Guess it is easier to maintain the dependency then rebuilding it to get rid of the package.


markusguenther:
It seems that we have an issue with es6 code from the node_modules...
Also had issues with upgrading debug that is now using es6 instead of es5 and d3-scale also drop es5.

markusguenther:
As we have the issue with es6 packages and the uglify-es is abandoned in uglifyjs-webpack-plugin  (see since version 2 https://github.com/webpack-contrib/uglifyjs-webpack-plugin/releases) we should use the terser package which is a fork of the discontinued uglify-es.

PR #2434, opened 2019-03-27
TASK: Added 'tabs__panel' class to TabsTheme

pmaas:
In order to fix https://github.com/psmb/Psmb.FlatNav/issues/2, we need to add styles to `<div role="tabpanel" …>`.

### DIFF
--- a/packages/react-ui-components/src/Tabs/tabs.spec.tsx
+++ b/packages/react-ui-components/src/Tabs/tabs.spec.tsx
@@ -16,8 +16,8 @@ describe('<Tabs/>', () => {
         ...tabsDefaultProps,
         theme: {
             'tabs': 'baseTabsClassName',
-            'tabs__content': 'basTabsContentsClassName',
-            'tabs__panel': 'basTabsPanelsClassName',
+            'tabs__content': 'baseTabsContentsClassName',
+            'tabs__panel': 'baseTabsPanelsClassName',
             'tabNavigation': 'baseTabsNavigationClassName',
             'tabNavigation__item': 'baseTabsNavigationItemClassName',
             'tabNavigation__item--isActive': 'activeTabsNavigationItemClassName',
--- a/packages/react-ui-components/src/Tabs/tabs.spec.tsx
+++ b/packages/react-ui-components/src/Tabs/tabs.spec.tsx
@@ -17,6 +17,7 @@ describe('<Tabs/>', () => {
         theme: {
             'tabs': 'baseTabsClassName',
             'tabs__content': 'basTabsContentsClassName',
+            'tabs__panel': 'basTabsPanelsClassName',
             'tabNavigation': 'baseTabsNavigationClassName',
             'tabNavigation__item': 'baseTabsNavigationItemClassName',
             'tabNavigation__item--isActive': 'activeTabsNavigationItemClassName',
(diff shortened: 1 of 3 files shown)


markusguenther:
Hi,

I bet you need to update the snapshots for the unit tests as well to fix the tests.
In this case you can go to packages/react-ui-components and execute jest --updateSnapshots

Thanks for submitting the PR. And glad to have you as first time contributor

👍

pmaas:
Thanks for the hint @markusguenther!

packages/react-ui-components/src/Tabs/tabs.spec.tsx
```diff
@@ -17,6 +17,7 @@ describe('<Tabs/>', () => {
         theme: {
             'tabs': 'baseTabsClassName',
             'tabs__content': 'basTabsContentsClassName',
+            'tabs__panel': 'basTabsPanelsClassName',
```
dimaip: typo `basT`

packages/react-ui-components/src/Tabs/tabs.spec.tsx
```diff
@@ -17,6 +17,7 @@ describe('<Tabs/>', () => {
         theme: {
             'tabs': 'baseTabsClassName',
             'tabs__content': 'basTabsContentsClassName',
```
dimaip: also another typo in existing code

packages/react-ui-components/src/Tabs/tabs.spec.tsx
pmaas: Sorry, copy pasted from the line above..

packages/react-ui-components/src/Tabs/tabs.tsx
```diff
@@ -120,6 +121,7 @@ export default class Tabs extends PureComponent<TabsProps> {
 
                     return (
                         <div
+                            className={theme!.tabs__panel}
```
dimaip:
```suggestion
                            className={theme.tabs__panel}
```
since this `tabs_panel` is already in TabsTheme, `!` shouldn't be necessary here, right?

pmaas: In this case I can't build the package:
```
ERROR in /Packages/Application/Neos.Neos.Ui/packages/react-ui-components/src/Tabs/tabs.tsx(124,40)
TS2532: Object is possibly 'undefined'.
```

dimaip: Wait a minute, but it IS possibly undefined! The theme object is marked as optional here: https://github.com/neos/neos-ui/blob/973d575154b48c8b0f5593abd0010abd656aafdc/packages/react-ui-components/src/Tabs/tabs.tsx#L27

But since we export the Tabs component wrapped in `themr`, the `theme` prop would always be there: https://github.com/neos/neos-ui/blob/973d575154b48c8b0f5593abd0010abd656aafdc/packages/react-ui-components/src/Tabs/index.ts#L6

So I'd say just mark the `theme` object as required.

pmaas: Done. Thanks for the help :)

markusguenther:
Welcome to the neos ui contributors list @pmaas 🎉
Would be awesome to see more PR´s in the future!

PR #2447, opened 2019-04-09
FEATURE: Responsive Backend

danielkestler:
Make the backend responsive – finally. Starting by supporting tablets like the iPad.

Bildschirmfoto 2019-04-09 um 13 57 51

### DIFF
--- a/packages/neos-ui/src/Containers/LeftSideBar/style.css
+++ b/packages/neos-ui/src/Containers/LeftSideBar/style.css
@@ -16,7 +16,7 @@
         clear: both;
     }
 
-    @media(max-width: 540px) {
+    @media(max-width: 576px) {
         width: 100%;
     }
 }
@@ -24,7 +24,7 @@
 .leftSideBar--isHidden {
     transform: translateX(-var(--size-SidebarWidth));
 
-    @media(max-width: 540px) {
+    @media(max-width: 576px) {
         transform: translateX(-540px);
     }
 }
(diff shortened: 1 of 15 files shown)


dimaip:
So currently only drag-n-drop wouldn't work, right?
Still a good start!

danielkestler:
Thanks, it's WIP ;-) Drag'n'drop works on the iPad ... in the page tree.

packages/neos-ui/src/Containers/LeftSideBar/style.css
```diff
@@ -15,10 +15,18 @@
     &:after {
         clear: both;
     }
+
+    @media(max-width: 540px) {
```
dimaip: Ideally we try to extract all constants, I think it definitely makese sense to do the same for breakpoints as well, so it would be easier to manage them.

danielkestler: Yes, it does. I did not manage to get the constants working inside the media query – is there a trick?

dimaip: Since we are using postcss, we could use this package: https://www.npmjs.com/package/postcss-media-variables

danielkestler: Does not go well with our PostCSS-configuration sadly ... https://github.com/WolfgangKluge/postcss-media-variables/issues/8

dimaip: Ah, too bad

danielkestler:
We could merge this at any time – it is already way better than before. Have some editors here waiting for this change ;-) But it should be well-tested, because this PR moves the basic panels from position: absolute; to position: fixed;

danielkestler:
Okay, tested it in the evergreen browsers – everything working as expected. We could merge at any time.

PR #2451, opened 2019-04-09
BUGFIX: fix JSON encoding of empty `allowedPresets`

ComiR:
For now only used for `contentDimensions.allowedPresets`.
It is probably needed in other places too.

This fixes a problem when the backend outputs an empty array and the frontend expects it to be an object. This is due to PHP arrays being converted by `json_encode()` to JS objects when they are associative and to arrays when not. Empty arrays are interpreted as non-associative (unless `JSON_FORCE_OBJECT` is used).
In many cases we don't want the whole array/object which gets json_encoded to be forced to an object, so we must catch this edge case sooner.

### DIFF
--- a/Classes/Fusion/Helper/ApiHelper.php
+++ b/Classes/Fusion/Helper/ApiHelper.php
@@ -19,11 +19,13 @@ class ApiHelper implements ProtectedContextAwareInterface
      * Converts an empty array to an empty object. Does nothing if array is not empty.
      *
      * Use this helper to prevent associative arrays from being converted to non-associative arrays by json_encode.
+     * This is an internal helper and might change without further notice
+     * FIXME: Probably better to produce objects in the first place "upstream".
      *
      * @param array $array Associative array which may be empty
      * @return array|\stdClass Non-empty associative array or empty object
      */
-    public function keepAssociative(array $array)
+    public function emptyArrayToObject(array $array)
     {
         return $array === [] ? new \stdClass() : $array;
     }
--- a/Classes/Fusion/Helper/ApiHelper.php
+++ b/Classes/Fusion/Helper/ApiHelper.php
@@ -1,0 +1,39 @@
+<?php
+namespace Neos\Neos\Ui\Fusion\Helper;
+
+/*
+ * This file is part of the Neos.Neos.Ui package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+use Neos\Eel\ProtectedContextAwareInterface;
+
+class ApiHelper implements ProtectedContextAwareInterface
+{
+    /**
+     * Converts an empty array to an empty object. Does nothing if array is not empty.
+     *
+     * Use this helper to prevent associative arrays from being converted to non-associative arrays by json_encode.
+     *
+     * @param array $array Associative array which may be empty
+     * @return array|\stdClass Non-empty associative array or empty object
+     */
+    public function keepAssociative(array $array)
+    {
+        return $array === [] ? new \stdClass() : $array;
+    }
+
+    /**
+     * @param string $methodName
+     * @return bool
+     */
+    public function allowsCallOfMethod($methodName) : bool
+    {
+        return true;
+    }
+}
--- a/Configuration/Settings.yaml
+++ b/Configuration/Settings.yaml
@@ -109,6 +109,7 @@ Neos:
         Math: Neos\Eel\Helper\MathHelper
         Json: Neos\Eel\Helper\JsonHelper
         I18n: Neos\Flow\I18n\EelHelper\TranslationHelper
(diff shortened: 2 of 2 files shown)


kitsunet:
I am not quite convinced of the naming really (Apart from the general fact that this is more of a hack than a solid solution).

`keepAssociative` is quite deceptive if you don't look at the code and use case, the array is clearly not kept associative, because it's turned into a stdClass. And from the JS perspective this makes no sense at all. Effectively this Helper is pointless if you don't turn the result into JSON, so it feels to me it shouldn't stand on it's own. I would kind of prefer a Ui helper that has `JsonStringifyEmptyArrayToObject` and does both steps in one, given the hack-ish nature of this.

ComiR:
We probably should deconstruct the variables in the following two places (+ others where needed) before converting them to JSON:
https://github.com/neos/neos-ui/blob/2.x/Resources/Private/Fusion/Prototypes/Page.fusion#L33
https://github.com/neos/neos-ui/blob/2.x/Resources/Private/Fusion/Backend/Root.fusion#L292
Then we can `JSON_FORCE_OBJECT` only the parts where we need it.

dimaip:
Hi! So what do we do about this?
I think it's quite critical to fix it. If we can't find a cleaner solution, I'd just merge the existing fix for now...

Classes/Fusion/Helper/ApiHelper.php
```diff
@@ -0,0 +1,39 @@
...
+{
+    /**
+     * Converts an empty array to an empty object. Does nothing if array is not empty.
+     *
+     * Use this helper to prevent associative arrays from being converted to non-associative arrays by json_encode.
+     *
+     * @param array $array Associative array which may be empty
+     * @return array|\stdClass Non-empty associative array or empty object
+     */
+    public function keepAssociative(array $array)
```
kitsunet:
```suggestion
    public function emptyArrayToObject(array $array)
```

Configuration/Settings.yaml
```diff
@@ -121,7 +122,7 @@ Neos:
           previewUrl: '${Neos.Ui.NodeInfo.createRedirectToNode(controllerContext, q(documentNode).context({workspaceName: documentNode.context.workspace.baseWorkspace.name}).get(0))}'
           contentDimensions:
             active: '${documentNode.context.dimensions}'
-            allowedPresets: '${Neos.Ui.ContentDimensions.allowedPresetsByName(documentNode.context.dimensions)}'
+            allowedPresets: '${Neos.Ui.Api.keepAssociative(Neos.Ui.ContentDimensions.allowedPresetsByName(documentNode.context.dimensions))}'
```
kitsunet:
```suggestion
            allowedPresets: '${Neos.Ui.Api.emptyArrayToObject(Neos.Ui.ContentDimensions.allowedPresetsByName(documentNode.context.dimensions))}'
```

Configuration/Settings.yaml
```diff
@@ -139,7 +140,7 @@ Neos:
           contentDimensions:
             byName: '${Neos.Ui.ContentDimensions.contentDimensionsByName()}'
             active: '${documentNode.context.dimensions}'
-            allowedPresets: '${Neos.Ui.ContentDimensions.allowedPresetsByName(documentNode.context.dimensions)}'
+            allowedPresets: '${Neos.Ui.Api.keepAssociative(Neos.Ui.ContentDimensions.allowedPresetsByName(documentNode.context.dimensions))}'
```
kitsunet:
```suggestion
            allowedPresets: '${Neos.Ui.Api.emptyArrayToObject(Neos.Ui.ContentDimensions.allowedPresetsByName(documentNode.context.dimensions))}'
```

Classes/Fusion/Helper/ApiHelper.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+use Neos\Eel\ProtectedContextAwareInterface;
+
+class ApiHelper implements ProtectedContextAwareInterface
+{
+    /**
+     * Converts an empty array to an empty object. Does nothing if array is not empty.
+     *
+     * Use this helper to prevent associative arrays from being converted to non-associative arrays by json_encode.
```
kitsunet:
```suggestion
     * Use this helper to prevent associative arrays from being converted to non-associative arrays by json_encode.
     * This is an internal VH and might change without further notice
     * FIXME: Probably better to produce objects in the first place "upstream".
```

kitsunet:
I made suggestions to make the naming and internal nature clearer, but I cannot push into @ComiR branch, so would be nice if you accepted the changes, then I would be good with merging.

dimaip:
Btw, could you also check a similar case when you have no content dimensions defined? It leads to console proptype warnings being thrown, and has the same nature as this issue. Would be sweet to have it fixed as well.

danielkestler:
The exact same works for contentDimensions.byName, too:

```
          contentDimensions:
            byName: '${Neos.Ui.Api.emptyArrayToObject(Neos.Ui.ContentDimensions.contentDimensionsByName())}'
```

Issue #2452, opened 2019-04-10
Validators fails and input field is uneditable


jonnitto:
If you have a [validator](https://flowframework.readthedocs.io/en/stable/TheDefinitiveGuide/PartV/ValidatorReference.html) on a property, the backend throws an error: 
```
Argument 3 passed to Neos\Neos\Ui\Service\NodePropertyValidationService_Original::validate() must be of the type array, boolean given
```

### Steps to Reproduce

Create a property with a validator: 
```yaml
  properties:
    title:
      type: string
      validation:
        'Neos.Neos/Validation/LabelValidator': true
```

Try to edit the value => You will get the error

<img width="492" alt="Bildschirmfoto 2019-04-10 um 15 05 49" src="https://user-images.githubusercontent.com/4510166/55881032-2da30a00-5ba2-11e9-9ad4-c9c23bd4977d.png">

### Affected Versions

Neos: 4.3-dev // 4.2

UI: dev-master // 3.2

jonnitto:
If you set the validator like this: 
```yaml
properties:
    title:
      type: string
      validation:
        'Neos.Neos/Validation/LabelValidator': []
```
everything works. Still, it is a breaky behavior, because before also boolean values worked

markusguenther:
This occurs because of this commit https://github.com/neos/neos-ui/commit/87b33af7ec48975a75325af97172b84dafebcbc1#diff-b1f15cbe77fdc0aceaae2512fd4a3c12

So the documentation seems to be wrong at this place or we need to adjust the piece of code. So that it is possible to have not validation configuration like in the CountValidator with the properties minimum and maximum.

To be honest does not find an example in the docs where a boolean is set.

markusguenther:
So from my point of view we first need to clarify which notation is correct and then adjust the docs and the behavior in the neos-ui.

What do you think @jonnitto @daniellienert @dimaip ???

PR #2455, opened 2019-04-15
BUGFIX: Check for validators w/o configuration

markusguenther:
It can happen that validators does not have additional configuration
like a maximum or minimum.

The exception is kind of eval that you get for the label validator. So we should check this here and prevent the hard exception.

Fixes: #2452


markusguenther:
Merci @daniellienert :)

PR #2457, opened 2019-04-16
TASK: Add Neos logo and favicons to the backend

danielkestler:
Last time this lead to some discussions, so we removed this change from the UI restyling process – here it is again, to be discussed.

Bildschirmfoto 2019-04-16 um 08 53 40
Bildschirmfoto 2019-04-16 um 08 53 46


packages/neos-ui/src/Containers/PrimaryToolbar/MenuToggler/style.css
```diff
@@ -1,16 +1,31 @@
...
     color: var(--colors-PrimaryBlue);
+    background: var(--colors-PrimaryBlue);
+    color: #fff;
+
+    .menuToggler__icon {
+        color: #fff;
+
+        &:hover,
+        &:focus {
+            color: #fff;
```
markusguenther: You can use the constants ;) and the new color makes no sense when it is the same.

color: var(--contrastBrightest);

packages/neos-ui/src/Containers/PrimaryToolbar/MenuToggler/style.css
```diff
@@ -42,3 +57,11 @@
         color: var(--colors-PrimaryBlue);
     }
 }
+
+.menuToggler__logo {
+    position: absolute;
+    top: 8px;
```
markusguenther: Same here we can use the constants and calculations of it.
e.g. var(--spacing-Half) for 8 pixels.

Would it help when I create a cheat sheet with all constants and values?

markusguenther:
But besides the code stuff I like the idea to put the Neos logo to the header :)

packages/neos-ui/src/Containers/PrimaryToolbar/MenuToggler/style.css
danielkestler: Sorry, obviously missed that (happened during merge ...). Will fix of course.

packages/neos-ui/src/Containers/PrimaryToolbar/MenuToggler/style.css
danielkestler: Thank you, I know there is a file with all the constants 😉 But as e.g. the spacing constants are not semantic in any way, I am careful in using them – because it might break later on if someone decides "half" to be 12px instead of 8px.

Might make sense to introduce semantic constants, like "componentPadding", "toolbarPadding" or similar.

dimaip: @danielkestler that's a very good point!
Though `toolbarPadding` might be a bit too specific, but I like the general idea.

Anyways, I'm not super keen on using spacing constants either, but either way we should try to be consistent: if we don't want them, let's just get rid of them. Or use your suggestion.

In anycase let's not overengineer, do what feels best to you :)

dimaip:
I'm neutral on this change, let's ping #guld-ux for another round of opinions? :)


Btw, I've created an extension that allows to create bookmarklets for website editors by dragging the logo: https://github.com/psmb/Psmb.Bookmarklet
Our editors really love it. Maybe makes sense to have smth like this in core?

dimaip:
But yeah, it seems most CMS do put a logo in the toolbar: WP, Joomla, TYPO3...

jonnitto:
With the new layout, I like this now ❤️

packages/neos-ui/src/Containers/PrimaryToolbar/MenuToggler/style.css
markusguenther: I don't know if we have a proper grid system. If so it would be nice to have all values we need and use them everywhere. With a robust system you can change the look and feel of neos backend with just a couple values. But therefore the whole code base have to use the constants and variantions of them. So eliminating static values would be nice from my point of view.

But maybe discuss is a better place to talk about that then a pull request :D

johannessteu:
Can't we use the monochrome version? https://github.com/neos/brand/blob/master/logos/Neos/neos_monochrome.png

The UI is made to focus the content, but the colored logo would somehow get too much attraction in my opinion.

dimaip:
@danielkestler what's your personal opinion on this?

danielkestler:
@dimaip I like it visually, but it is a bit hard to "read" or recognize. Would rather go back to the colored one ... Or we may have to create a logo variant for small display with thicker lines.

KPeschke:
I like the colored variant more.
How about the "neos_negative_light.svg", but could be only allowed for light backgrounds by brand guidelines.

Why is the Neos logo on hover connected to the menu button? We could instead use the logo to show a dropdown menu with usefull links, like the neos-top-navigation on discuss.neos.io.

cvette:
My favorite is the very first colored variant. It draws some more attention to the menu, which is rather hard to notice at the moment.

Does the Logo act as a button for the menu?

jonnitto:
It looks like we have to make a poll ;-)

From my point of view, I like the monochrome variant the most, and yes the lines need to be thicker, but this should be no problem. What do you think about this, @aertmann, @Inkdpixels, @skurfuerst and @dfeyer?

danielkestler:
Yes, at least in this PR, logo and menu button are unified. This way, the logo alignes nicely with the text in the document tree toggler and the items in the actual menu.

We could of course change that.

dfeyer:
I think we can use this place for a more useful UI ;)

https://github.com/neos/neos-ui/pull/2462

It's not because other CMS love to put their brand everywhere, that we need to do that. I think we can introduce the logo in the bottom of the left Drawer (the hiddeable module navigation), here it can be bigger, more elegant... 

I really appreciate our "brandless" CMS ;)

dfeyer:
> From my point of view, I like the monochrome variant the most

I don't remember exactly, but I think the wireframe logo should be reserved for technical users, but not sure if it's part of the brand guide.

danielkestler:
Disagree here. Branding and UI design should go hand in hand (see Design Systems etc.) and the logo really is missing here. But the best way to integrate is open for discussion 😉

dfeyer:
> Disagree here. Branding and UI design should go hand in hand (see Design Systems etc.) and the logo really is missing here. But the best way to integrate is open for discussion wink

You branding team ;) I'm fine with having the logo, but please don't couple it with with the hamburger icon, so it can be replace easily

packages/neos-ui/src/Containers/PrimaryToolbar/MenuToggler/index.js
```diff
@@ -8,6 +8,8 @@ import Button from '@neos-project/react-ui-components/src/Button/';
 import {actions} from '@neos-project/neos-ui-redux-store';
 import {neos} from '@neos-project/neos-ui-decorators';
 
+import logo from './logo.svg';
```
dfeyer: I will not couple to logo with the Menu toggler, but introduce a static component that is integrated via the registry. Like this is can be replaced with our extendability features

danielkestler: Okay, good idea. There are lots of components that should be refactored ... ;)

dfeyer:
image

About the logo integration in the bottom of the Drawer, I think it's a better place to have our branding. We have humility in our project value, so having the logo a bit hidden is not a bad idea. My integration is really fat, but why not, we have room here. And we can have links to our website, support platform, ...

If this area is fully customizable (logo and links) I think it's a nice place for Agency to put their own support platform, customer logo or some "per project" branding.

danielkestler:
Uhhh ... maybe let's integrate the logo in a more decent way? 😳

up-to-date
update-available

dimaip:
@danielkestler I didn't quite follow the disucssion, did we already give up on the idea to have the Neos logo always visible? I understand @dfeyer's reasoning, just want to make sure everyone else agrees with it.

As said, personally I'm fine with both options and I see nothing wrong with always showing the logo.

danielkestler:
@dimaip Well, no, I don't agree with it ;-) I would rather make the logo sticky on the top left. But if we decide against it, then the place in the menu also makes sense.

markusguenther:
> Uhhh ... maybe let's integrate the logo in a more decent way? 😳
> 
> up-to-date
> update-available

Hey,

like the decent idea in the drawer. But the logo with the information afterwards looks strage.
What do you think about the Logo and then just Version 4.3.x instead of Neos 4.3.x again. So we have no double of Neos.

Maybe also put the version and the state information right of the logo and align it right and the logo left.

dimaip:
@danielkestler then I would not decide against it so quickly, without having consensus on the topic...
Perhaps a compromise solution could be to put into the registry, so it would be possible to override it for those who want to keep the unbranded experience?

danielkestler:
@dimaip Of course, I'm fine ;-) That's just my opinion, had no intention to outvote other ideas. We could also make it a setting in Neos.Ui-Configuration.yaml.

@markusguenther Yes, we could also just write "Version" – it's more about the visual appearance here. The "state" information is not even a feature that exists currently.

dimaip:
@danielkestler no no I meant the contrary, that we should dismiss your initial idea so quickly.
I'd try to avoid yet another config in Settings.yaml if possible, better to keep the higher-level API clean from such things (IMO).

KPeschke:
I like the idea of branding the backend. But can also understand that so many users already love the clean look. 

How about a favicon for the backend browser? Our installations don't have a favicon for the backend, frontends have the own favicon. It's easier for editors to find the tab for the backend, for pinned tabs too. If the backend is ready for use on mobile we need touch icons. Currently, the Neos login page doesn't have a good looking icon on home screens.

Now we have some ideas to vote on?
1. don't brand
2. logo top left
3. logo in the drawer
4. favicon/touch icon for backend tabs
5. combine 3. and 4.
... all in variations of colored and monochrome

dfeyer:
> Uhhh ... maybe let's integrate the logo in a more decent way? flushed

I'm fully indecent ;) and fully agree

danielkestler:
> 1. don't brand
> 2. logo top left
> 3. logo in the drawer
> 4. favicon/touch icon for backend tabs
> 5. combine 3. and 4.
>    ... all in variations of colored and monochrome

2. and 4. ;-)

danielkestler:
Okay, now I separated the logo and the menu icon and put the logo in its own component and register it via containerRegistry. This way, you can easily change or remove the logo if you don't want to have it.

(Sorry, something got a little wrong when rebasing, but the changed files are correct and we could squash merge.)

danielkestler:
@dimaip Thanks, but tbh., I did not know where exactly to put the components first. So I put the "pure" logo into the react-components folder and the other one in the containers folder. Hope, that's okay.

dimaip:
react-ui-components is the place to put generally-useful components, that can potentially be used outside the content-module UI (e.g. when building backend modules).
I'm not sure if having a Neos logo there is all that useful (as it's quite trivial and there are many different variants of the logo out there), but I trust your judgement here.

packages/neos-ui/src/Containers/PrimaryToolbar/MenuToggler/index.js
danielkestler: Done.

packages/neos-ui/src/Containers/PrimaryToolbar/MenuToggler/style.css
danielkestler: No longer applies ... ;-)

danielkestler:
Feature-complete for now, could merge at any time.

Sebobo:
@dfeyer you requested changes, are they now done?

@danielkestler is the PR title still matching the change? Seems to be a bit more than adding a logo ;)

Sebobo:
Should we merge this then? Anyone against this?

PR #2460, opened 2019-04-23
FEATURE: Allow adding and overriding fusion code for Neos UI backend in all packages

dlubitz:
The `@package` placeholder gets replaced by the package name, so you are able to add your own Fusion code for Neos UI backend in your own packages in path `/Private/Fusion/Backend/`.

### DIFF
--- a/Classes/View/BackendFusionView.php
+++ b/Classes/View/BackendFusionView.php
@@ -18,6 +18,6 @@ class BackendFusionView extends FusionView
     public function __construct(array $options = [])
     {
         parent::__construct($options);
-        $this->setFusionPathPattern('resource://@package/Private/Fusion/Backend/');
+        $this->setFusionPathPattern('resource://@package/Private/Fusion/Backend');
     }
 }
--- a/Classes/View/BackendFusionView.php
+++ b/Classes/View/BackendFusionView.php
@@ -18,6 +18,6 @@ class BackendFusionView extends FusionView
     public function __construct(array $options = [])
     {
         parent::__construct($options);
-        $this->setFusionPathPattern('resource://Neos.Neos.Ui/Private/Fusion/Backend');
+        $this->setFusionPathPattern('resource://@package/Private/Fusion/Backend/');
     }
 }


dimaip:
Could you please add a little example how to use this?
Also I think it would make sense to add it to the docs.

dlubitz:
Sure, just put a `Root.fusion` in your package in path `VIVOMEDIA.MyPAckage/Resources/Private/Fusion/Backend`. There you are able to add other URI endpoints for example like:

```
backend {
	routes {
		vivomedia = Neos.Fusion:RawArray {
			newsletter = Neos.Fusion:RawArray {
				sendingEndpointUri = VIVOMEDIA.Newsletter:SendingUri {
					arguments.node = ${documentNode}
				}
                       }
		}
	}
}
```

to use them in for example in editors:

```
url: $get('routes.vivomedia.newsletter.testingEndpointUri', neos)
```

I would also add some documentation, but AFAIS the docs for UI extension are in the neos repo. So I'd like to wait to create them, until this has been approved and merged ;)

markusguenther:
Thanks again @dlubitz for the contribution.
Would be nice if we also add bit documentation for that.

DrillSergeant:
Yes, documentation would be great. Right now I have no clue, what I can do with this.

dlubitz:
Sure, I'm struggling a bit where to put it.
Here: https://neos.readthedocs.io/en/stable/ExtendingNeos/UiExtensibility.html ?

I would add another section at the bottom with "Backend Fusion". Other ideas?

markusguenther:
Hi,

we have the shiny new site https://docs.neos.io/cms/manual/extending-the-user-interface.
It is a neos site and we actually have now contribution workflow.
But if you want we can ask @rolandschuetz for a backend user. Otherwise you can write something and we put this to the site :)

dlubitz:
As I figured out now, this change doesn't work. IDK what happend that day I created the PR ... 

Falsly I assumed the placeholder will be replaced by all packages. But actually it just get replaced by the current package. In this case `Neos.Neos.Ui`. So at the end, the same as before, just more confusing. 

(!) So I would recommand to revert the commit. Any thougths?

A proper solution for the issue is much more complex than expected and needs to be tackled later. :/

PR #2462, opened 2019-04-23
TASK: Move the DimensionSwitcher to the PrimaryToolbar

dfeyer:
This change make the DimensionSelector always visible in the PrimaryToolbar. This avoid a lots of confusion for the Editor cause by the lack of information about the current dimension preset when editing content (CK toolbar replace the DimensionSelector)

image


danielkestler:
I am basically absolutely in favor of this change, as we discussed earlier. But please don't put it on the top left, as this gives it too much weight IMO – perhaps just move it to the right, next to the Edit-Toggle?

dfeyer:
> I am basically absolutely in favor of this change, as we discussed earlier. But please don't put it on the top left, as this gives it too much weight IMO – perhaps just move it to the right, next to the Edit-Toggle?

From my POV, the current Dimension preset, is the current editing context of the UI, so at least as important as the edit / preview. So maybe we can group both item, aligned on the top/left of the content frame. That can also solve the fact that the edit/preview is not really related to the user profile.

Current we have only left/right section in the primary toolbar, I'm in favor to introduce left/middle/right, and use the right area for Dimension Switcher and Edit/Preview

@danielkestler What do you think ?

dfeyer:
The new version looks nicer

image

packages/neos-ui/src/manifest.containers.js
```diff
@@ -14,6 +14,7 @@ import KeyboardShortcutModal from './Containers/Modals/KeyboardShortcutModal/ind
 import UnappliedChangesDialog from './Containers/Modals/UnappliedChangesDialog/index';
 
 import PrimaryToolbar from './Containers/PrimaryToolbar/index';
+import DimensionSwitcher from './Containers/PrimaryToolbar/DimensionSwitcher/index';
```
dfeyer: Not sure if we consider this as a breaking change

packages/neos-ui/src/manifest.containers.js
```diff
@@ -58,14 +58,14 @@ manifest('main.containers', {}, globalRegistry => {
 
     containerRegistry.set('PrimaryToolbar', PrimaryToolbar);
     containerRegistry.set('PrimaryToolbar/Left/MenuToggler', MenuToggler);
-    containerRegistry.set('PrimaryToolbar/Right/EditModePanelToggler', EditModePanelToggler);
+    containerRegistry.set('PrimaryToolbar/Middle/DimensionSwitcher', DimensionSwitcher);
```
dfeyer: Not sure if we consider this as a breaking change

packages/neos-ui/src/Containers/PrimaryToolbar/style.css
```diff
@@ -1,20 +1,19 @@
 .primaryToolbar {
-    &:before,
-    &:after {
-        content: '';
-        display: table;
-    }
-    &:after {
-        clear: both;
-    }
+    display: flex;
```
dfeyer: The CSS now use Flex, way more easy than the table and all absolute position.

danielkestler:
@dfeyer Yes, good idea. Here is a rough sketch (including the Neos logo 😅):

56616926-dc256100-661e-11e9-877a-aa9f84097a80-1

dfeyer:
> @dfeyer Yes, good idea. Here is a rough sketch (including the Neos logo sweat_smile):

OK so look like my new commit implement your sketch nicely ;)

So this change is ready for review.

> including the Neos logo sweat_smile

Does it give user any value ? No ... so should be somewhere else ;) I hate brand (as you see)

DrillSergeant:
Hi guys.

Thanks for working on the UI, I like the improvements a lot!

About the logo:
I agree with @dfeyer , the logo does not add any value for the user.  I like more the version without logo at all, not even somewhere else.
Maybe the (monocrome) logo could be visible in the bottom of the module navigation, when that is opened via the burger-icon.

dimaip:
Yay, thanks for bringing up this topic, @dfeyer!

1. Show dimension switcher at all times -- yes, please, that's an absolute must
2. Moving it out from the lower bar -- that's not so good, as we initially intended the lower bar to be directly related to the content canvas content, while the upper bar was for system-wide context.
3. Alternatively we could move the dimension switcher to the upper part of the inspector, and when the inspector is collapsed still show it int the right part of the toolbar? That would allow to make a more advanced version of the dimension switcher there, because it would have more space. But maybe there are better ideas.
4. And please, don't put anything in the center, IMO it looks really bad to have stuff center-aligned there.

For Neos-logo discussion there's a separate PR to discuss that (And yes, it does provide a value to the editor, at lease they'd know the name of a system they are using and that's what all other CMS' do).

dfeyer:
Please discuss the logo stuff in the related the PR ;)



> Show dimension switcher at all times -- yes, please, that's an absolute must

We are on the same track here



> Moving it out from the lower bar -- that's not so good, as we initially intended the lower bar to be directly related to the content canvas content, while the upper bar was for system-wide context.

Dimension is a system wide context, it affect the document tree also. So I think it's a good candidate for the top bar



> Alternatively we could move the dimension switcher to the upper part of the inspector, and when the inspector is collapsed still show it int the right part of the toolbar? That would allow to make a more advanced version of the dimension switcher there, because it would have more space. But maybe there are better ideas.

I agree that we need a more advanced UI for the Dimensions in the inspector can be a view in the "gear" tab, but that's an other topic. And that can be a "information only" view (explain fallback, shine trought contents...)

Moving the DimensionSwitcher to the inspector, also make it less visible. For project with multiple dimension this information need to be visible MOST of the time if not ALL the time. It's really easy to be lost in Dimensions. The inspector can have tab, ... so the current Dimension Preset is not visible all the time.

> And please, don't put anything in the center, IMO it looks really bad to have stuff center-aligned there.

It's not "center aligned", it's aligned on the right side of the content area. I agree with the comment from @danielkestler if the Dimenson Switcher if aligned on the full right, it make it too important ... and a bit lost in the UI.

Grouping the Dimension Switcher and the Edit/Preview make sense (at least more sense than grouping the Edit/Preview with the user profile).

dfeyer:
And this change is not radical for our user base, no need to explain where the feature is, it's just 40px on top of the old position. This is also something that we need to take into account when moving feature around.

dimaip:
> The inspector can have tab, ... so the current Dimension Preset is not visible all the time.

No no, I didn't mean to put it inside inspector, but on top of it, so it'd be visible all the time, the way currently selected node selector works now. But yeah, was just an idea, would definitely need a lot of work.

>  it's aligned on the right side of the content area

Ah, didn't get the logic initially. You meant the left side?
In that case I'd join up the border to form a single visual group: 
image
Wdyt?

dfeyer:
> Ah, didn't get the logic initially. You meant the left side?

Yes left ;) ... by brain is wrongly connected with left/right



>  In that case I'd join up the border to form a single visual group:

Yes that can be tweaked a bit

dfeyer:
> No no, I didn't mean to put it inside inspector, but on top of it, so it'd be visible all the time, the way currently selected node selector works now. But yeah, was just an idea, would definitely need a lot of work.

Screen get wider but not taller, I will avoid to limit the height of the inspector as much as possible. And it's a more radical change for our user base.

mficzel:
@dimaip @dfeyer I think moving the dimension switcher to the inspector is not a good idea as we encourage editors to close the inspector for simple tasks  like writing. In that case the dimension switcher would not be accessible directly.

Since the current position is not a good choice either i see no a real alternative to the suggestion of @dfeyer. Also the argument that the main bar shall not modify the preview is partly invalid as it already contains the "edit/preview" button which has a huge impact on the preview. Imho it would be quite consistent to move the dimension-selector beside the "edit/preview" button.

Side Note: For additional consistence "edit/preview" could become a dropdown with two optgroups. Would make much more sense then the current two swiping bars.

markusguenther:
Updated this is bit but still need to tackle the e2e test.

dimaip:
@markusguenther do you need help with e2e tests? If you want we could have a hangout and fix the tests together?

dimaip:
Also should target master.

dimaip:
Would be great to have it for 5.1!

davidspiola:
Hi Folks, do you think you gonna make it for the 5.1 feature freeze?

markusguenther:
Plan to work next week full time on neos-ui... hope that it is enough time to finish my emberjs stuff and this.

markusguenther:
Closes in favor of the updated PR that works on master.
https://github.com/neos/neos-ui/pull/2587

Thanks for the foundation @dfeyer ... will finish it in the new PR und hopefully make it to neos 5.1 🎉

PR #2464, opened 2019-04-24
BUGFX: Fix vertical alignement bug in Firefox in Drawer

dfeyer:


dfeyer:
Before: 

image

After:

image

dfeyer:
Tested on Chrome too, works

PR #2468, opened 2019-04-29
Enhance circleci configuration to use better caching and workspace

johannessteu:


johannessteu:
Any idea why the e2e actually doesn't start? My change _shouldn't_ be related at all as the souce code is still copied and all files are there. Otherwise there would be something like testcafe binary could not be found etc.

johannessteu:
Yeah, git it figured out. I had the fork setup in my namespace. CircleCI run the job in my namespace instead of the neos org. So some credentials were missing. I deleted the project in my personal namespace and now it's looking good!

So i think this change will - again - improve the pipeline speed/duration and also adds some more readability to the yaml file!

markusguenther:
Thank you for taking care 🙏

PR #2470, opened 2019-05-06
Restrict style for td's in guest frame to CKE content

dlubitz:
The global `td` style is not resticted to any subject within the guest frame. Moreover it affects the styles of users.

### DIFF
--- a/packages/neos-ui-guest-frame/src/style.css
+++ b/packages/neos-ui-guest-frame/src/style.css
@@ -42,7 +42,7 @@
     background-image: linear-gradient(to bottom, rgba(50, 50, 50, .1), rgba(50, 50, 50, .2), rgba(50, 50, 50, .1));
 }
 
-.ck-content .table table td {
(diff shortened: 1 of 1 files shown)


dlubitz:
I see. Will change it.

packages/neos-ui-guest-frame/src/style.css
```diff
@@ -42,7 +42,7 @@
     background-image: linear-gradient(to bottom, rgba(50, 50, 50, .1), rgba(50, 50, 50, .2), rgba(50, 50, 50, .1));
 }
 
-:global(td) {
+.ck-content .table table td {
```
dimaip:
```suggestion
:global(.ck-content .table table td) {
```
But it still has to be global, right?

dlubitz: Ahh, I see. Had to read docs twice to get the idea of `:global`.

Issue #2379, opened 2019-02-28
Order of positional sorting is inconsistent with Flow


PRGfx:
### Description

The positional array sorter claims to sort by _weight_, when using a keyword like `end`, `start` etc. implying items weighted higher to appear before items with a lower weight.

### Steps to Reproduce

1. set node-types' `ui.position` to e.g. `start 10` and `start 20`
2. see new node dialog

#### Expected behavior

When sorting by weight, the items should be sorted in descending order. Compare the Flow test-case:
https://github.com/neos/flow-development-collection/blob/master/Neos.Utility.Arrays/Tests/Unit/PositionalArraySorterTest.php#L79

#### Actual behavior

Items with weight are sorted according to ascending weight as per the sorter's test:
https://github.com/neos/neos-ui/blob/643338e7e08ca128319dc0100a6e73325db5ea8e/packages/positional-array-sorter/src/positionalArraySorter.spec.js

### Affected Versions
UI 1.4.1; quoted code is from current master

dimaip:
Oh, could catch!

PRGfx:
After I had a more thorough look at the current implementation, I got curious about the `positionValue.includes('start')` [parts](https://github.com/neos/neos-ui/blob/8c81506206fafa446357882bae8ce598112d6733/packages/positional-array-sorter/src/positionalArraySorter.ts#L60). Are there real-life problems that led to `includes()` over `startsWith()` (or `trim().startsWith()`)? The current implementation would mess up sorting with keys using these "keywords", e.g.
```
[
  {position: 'after start'},
  {position: 'start'},
  {position: 'end'},
  {key: 'start'}
]
```
`position: after start` contains `start` and precedes `position: start` so it would go first rather than after the "middle key" `start`.

Other than that I would amend my initial issue by noting that sorting by weight would need to happen ascending or descending depending on the positioning being `before` or `end` and `after` or `start` respectively.

Maybe I get to create an alternative implementation in the next days.

dimaip:
Thanks for your thorough research, the fix would indeed be very welcome.

> Are there real-life problems that led to includes() over startsWith() (or trim().startsWith())? 

Can't answer that at the moment. In any case, the Flow implementation should be the reference I'd say.

PRGfx:
It took some time but now I've completely rewritten the thing according to my prior re-implementation of the php version in C# (that's why I was confident, I would have gotten here sooner...).
[...]

PR #2474, opened 2019-05-12
TASK: rewrite positional array sorter to behave like the php implementation

PRGfx:
**What I did**
Adjust the ordering in the positional-array-sorter to behave more like the implementation in neos/utility-arrays (#2379): `before` and `after` positions may now be weighted, `start` and `after` positions will now be sorted appropriately according to descending weight. Positions like `after start` are no longer matched as `start` key.

**How I did it**
Mending the existing implementation did not prove easy as features like weighted `before` or `after` positions were not implemented, so I implemented the function anew.

**How to verify it**
I developed against the test cases (mostly ported from the php version). As a lot of parts in the UI use this functionality, I can't really tell how to best verify this in the actual UI.

### DIFF
--- a/packages/positional-array-sorter/src/positionalArraySorter.spec.js
+++ b/packages/positional-array-sorter/src/positionalArraySorter.spec.js
@@ -217,6 +217,23 @@ test('Handles non-word keys', () => {
     ];
     expect(positionalArraySorter(source)).toEqual(result);
 });
+test('Handle invalid position strings', () => {
+    const source = [
+        {position: 10},
+        {position: 'after'},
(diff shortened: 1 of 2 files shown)


markusguenther:
@dimaip as you are requested for that PR do find the time for the review?
Otherwise I would dive into that piece of code and check it.

markusguenther:
Thank you very much @PRGfx 
Tried that with neos 3.3 and it worked well.

I am glad to see that you now also contribute to the neos-ui for the first time.  🎉 🎉 🎉

Issue #2472, opened 2019-05-08
Property editor array<Assets> not finding any assets


simonedelmann:
### Description

The search functionality within the property editor for array<Assets> does not work at all. Assets could never be found, although they're there.
(Extracted from https://github.com/neos/neos-ui/issues/2471)

#### Steps to Reproduce

1. Fresh install of Neos + Neos.Demo site
2. Add an Neos.NodeTypes.AssetList somewhere.
3. Try to add an asset via the search field. (Enter correct filename...)

#### Expected behavior

In Step 3 there should be search results.

#### Actual behavior

"No matches found", although the asset does exist and filename was correct.

### Affected Versions

Neos: v4.3.0
UI: v3.3.0

Tried also with Bitnami Neos stack 4.3.1.

beardcoder:
Oki this problem is only in the master and not in the 2.x Branch

markusguenther:
This issue has been resolved in the PR (https://github.com/neos/neos-ui/pull/2478) 
We will release a new UI version soonish 🎉

Issue #2471, opened 2019-05-08
Property editor array<Assets> not working as expexted/at all


simonedelmann:
### Description

The property editor for array<Asset> seems to be broken somehow. If the array is saved and non-empty, clicking on an asset in the media browser opens the asset's properties (title, description, ...) instead of adding the asset to the array.

Also asset search is not working entirely. Assets are never be found.

To aviod environmental influence, I could reproduce both bugs within a fresh Bitnami Neos stack (with Neos.Demo installed).

#### Steps to Reproduce

1. Fresh install of Neos + Neos.Demo site
2. Add an Neos.NodeTypes.AssetList somewhere.
3. Add one or multiple elements from media browser to the list and apply changes.
4. Try to add another element from media browser.

#### Expected behavior

In Step 4 another element should be added to the list.

#### Actual behavior

A spinning wheel appears within the property editor. 
The media browser opens the detail view, where I could edit the asset's properties (title, description, ...).

### Bug 2: Adding assets via search field

Edit: Bug 2 moved to separate issue. (https://github.com/neos/neos-ui/issues/2472)

### Affected Versions

Neos: v4.3.0
UI: v3.3.0

Tried also with Bitnami Neos stack 4.3.1.

danielkestler:
Yes, confirmed, see my PR: https://github.com/neos/neos-ui/pull/2466

simonedelmann:
Thank you! I will wait for your PR getting merged into the core.
I will edit my issue and move bug 2 to another issue.

danielkestler:
Closing as the PRs are merged.

PR #2478, opened 2019-05-13
FIX: Correct dataLoader and add new asset

beardcoder:
Resolves: #2472, #2471


markusguenther:
Trigger CI again

kitsunet:
Note, this change introduced a big regression by limiting the asset editor to only assets that exist already in neos. The loader was definitely the right one before. I will test this and try to provide a fix for the regression. The AssetEditor should search through all asset sources.

PR #2488, opened 2019-05-15
TASK: Remove deprecated legacy logger

kdambekalns:


daniellienert:
Looks good!
@markusguenther - are these upmerges still missing to get the tests green?

kdambekalns:
Ok, I am clueless regarding the CI failure:

- Does the master branch build currently? Where can I see that?
- If it does, how can PHP changes cause a merge conflict in TS files?
- It it does not, can someone please fix it? :)

kdambekalns:
Ah, I played around with CircleCI a bit, and found out: master is indeed broken: https://circleci.com/gh/neos/neos-ui/907

I merged 2.x and ran `yarn upgrade` afterwards, but it still fails: https://circleci.com/gh/kdambekalns/neos-ui/2

danielkestler:
Don‘t do `yarn upgrade`, I think that also broke it for me one time today. There is a .lock file that should be fine, just `make clean` and `yarn install`.

markusguenther:
Trigger new CI build

markusguenther:
Hmm it seems that circle ci caches a lot and I updated the master but the CI still uses the old state of the master and not the updated on. Will try to fix that.

Sorry for that inconvenience.

markusguenther:
Ok checkout works now but the end to end test still need tome love in master. But guess that is ok because you change is not related. Will take a look into that in a separate PR.

kdambekalns:
> Don‘t do `yarn upgrade`, I think that also broke it for me one time today. There is a .lock file that should be fine, just `make clean` and `yarn install`.

I know… but that told me about an outdated package:

    Browserslist: caniuse-lite is outdated. Please run next command
    npm update caniuse-lite browserslist

but the suggested solution from the error message didn't work  (see https://github.com/postcss/autoprefixer/issues/1184 and others having that issue.) So I tried the full upgrade, after which the build ran fine locally. But not on master.

Anyway, I'll rest my case, as it's been merged. Thanks. :)

PR #2492, opened 2019-05-16
TASK: Remove close button for relogin dialog

maya-shines:
Former behavior:
When the backend session expires, the relogin dialog forces the user to login again.
This relogin dialog has an "x"-Button to close it, but the button does nothing.

image

New behavior:
The relogin dialog still opens, but now no longer has the useless "x"-Button:

image

Sidenote: The dialog.tsx already checks, if the "onRequestClose" is set to something. However, "{emptyFn}" seems to evaluate as not empty, thus rendering the button, despite it not having a function. All I did was to replace the empty function with an empty string.

### DIFF
--- a/packages/neos-ui/src/Containers/Modals/ReloginDialog/style.css
+++ b/packages/neos-ui/src/Containers/Modals/ReloginDialog/style.css
@@ -1,7 +1,7 @@
 .modalContents {
     padding: var(--spacing-Full);
 }
-.inputField {
+.inputFieldWrapper {
     margin-bottom: 15px;
 }
 .loginButton {
--- a/packages/neos-ui/src/Containers/Modals/ReloginDialog/style.css
+++ b/packages/neos-ui/src/Containers/Modals/ReloginDialog/style.css
@@ -2,7 +2,7 @@
     padding: var(--spacing-Full);
 }
 .inputFieldWrapper {
-    margin-bottom: 15px;
+    margin-bottom: var(--spacing-Half);
 }
 .loginButton {
     width: 100%;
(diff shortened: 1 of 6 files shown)


danielkestler:
Awesome, thanks. While you're over it, would you mind adding some margin after the input fields?

maya-shines:
On it!

danielkestler:
A while ago, we discussed getting rid of the close button altogether – so if this is easier, maybe we can just remove it from all dialogs?

daniellienert:
👍🏼

markusguenther:
Marked it as work in progress. Maya told me that she plans additional work on it.

maya-shines:
Hey you guys, I added some changes. The close button is now removed from all modals. I could only find one modal that doesn't have a cancel button to act as an exit strategy (other than pressing the esc key or clicking elsewhere). That was the keyboard shortcut modal. So I added a little close button there. Last change was adding some margin to the input fields in the relogin dialog:
image


I hope this will do!

packages/neos-ui/src/Containers/Modals/ReloginDialog/style.css
```diff
@@ -1,7 +1,7 @@
 .modalContents {
     padding: var(--spacing-Full);
 }
-.inputField {
+.inputFieldWrapper {
     margin-bottom: 15px;
```
danielkestler: Maybe use spacingHalf constant here? Just a nitpick...

maya-shines: done ;)

Sebobo:
@maya-shines cool thanks! 
So it's no WIP anymore?

markusguenther:
I put the WIP on the PR after Maya mentioned that she wants to continue this tasks after the sprint.

PR #2494, opened 2019-05-16
BUGFIX: Fix missing q helper on latest master in backend

mficzel:
The neos master registers `q` as a static function which is not compatible with later versions of neos where `q` is a magic name. Since some of the services in the ui rely on flowQuery and use a custom defaultConfiguration this causes trouble when running on latest master.

As backwards compatible solution this change includes the eel default context from `Neos.Fusion.defaultContext` and merge those with `Neos.Neos.Ui.configurationDefaultEelContext` before evaluating the expressions.

This also allows to remove the default Eel helpers from the configuration and only register the
additional helpers of the ui itself needs.

This is related to: https://github.com/neos/flow-development-collection/pull/1580

### DIFF
--- a/Classes/Domain/Service/ConfigurationRenderingService.php
+++ b/Classes/Domain/Service/ConfigurationRenderingService.php
@@ -28,10 +28,16 @@ class ConfigurationRenderingService
     protected $eelEvaluator;
 
     /**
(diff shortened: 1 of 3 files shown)


markusguenther:
Awesome that you took over this adjustment.

Issue #2444, opened 2019-04-05
Clicking on the wrong copy button leads to an exception


daniellienert:
![Capture 2019-04-05 at 14 34 01](https://user-images.githubusercontent.com/642226/55631069-7714e300-57b7-11e9-91d3-a00c3365f531.png)

### Description

When selecting a document and clicking the copy button above the content tree, an exception is thrown.

### Steps to Reproduce

1. Klick on a document
2. Klick on the wrong copy button

#### Expected behavior

The button is deactivated, when a document is selected

#### Actual behavior

Exception:

```
Neos\Flow\Mvc\Exception\RequiredArgumentMissingException: Required argument "node" is not set.
#25 Data/Temporary/Production/Cache/Code/Flow_Object_Classes/Neos_Flow_Mvc_Controller_AbstractController.php(371): mapRequestArgumentsToControllerArguments
#24 Data/Temporary/Production/Cache/Code/Flow_Object_Classes/Neos_Neos_Ui_Controller_BackendServiceController.php(2312): mapRequestArgumentsToControllerArguments
#23 Data/Temporary/Production/Cache/Code/Flow_Object_Classes/Neos_Neos_Ui_Controller_BackendServiceController.php(0): call_user_func_array
#22 Packages/Framework/Neos.Flow/Classes/Aop/AdvicesTrait.php(56): Flow_Aop_Proxy_invokeJoinPoint
#21 Packages/Framework/Neos.Flow/Classes/Aop/Advice/AdviceChain.php(59): proceed
#20 Data/Temporary/Production/Cache/Code/Flow_Object_Classes/Neos_Flow_Security_Aspect_PolicyEnforcementAspect.php(65): enforcePolicy
#19 Packages/Framework/Neos.Flow/Classes/Aop/Advice/AroundAdvice.php(35): invoke
#18 Packages/Framework/Neos.Flow/Classes/Aop/Advice/AdviceChain.php(57): proceed
#17 Data/Temporary/Production/Cache/Code/Flow_Object_Classes/Neos_Neos_Ui_Controller_BackendServiceController.php(2324): mapRequestArgumentsToControllerArguments
#16 Data/Temporary/Production/Cache/Code/Flow_Object_Classes/Neos_Flow_Mvc_Controller_ActionController.php(204): processRequest
#15 Data/Temporary/Production/Cache/Code/Flow_Object_Classes/Neos_Neos_Ui_Controller_BackendServiceController.php(1465): processRequest
[...]

PR #2498, opened 2019-05-20
TASK: Disable copy button in content tree if selected node is in page tree 

maya-shines:
fixes https://github.com/neos/neos-ui/issues/2444


markusguenther:
Ah Sebastian was faster ... guess woud be nice to have a E2E test for that.
That come to my mind yesterday in the evening. But we can also add this as separate PR.

PR #1929, opened 2018-06-27
WIP: TASK: Show deleted Nodes until they’re really deleted

johannessteu:

PR #2502, opened 2019-05-21
WIP: TASK: Show deleted nodes in trees

danielkestler:
I completely messed up and I am deeply sorry about it. This PR is originally by @johannessteu (#1929) and my intention was just to rebase it on the 2.x branch, as long time nothing happened regarding this topic.

So here it is again, with only minor visual changes and still WIP.

### DIFF
--- a/Classes/Controller/BackendServiceController.php
+++ b/Classes/Controller/BackendServiceController.php
@@ -458,7 +458,7 @@ class BackendServiceController extends ActionController
 
         $flowQuery = new FlowQuery(array_map(
             function ($envelope) {
-                return $this->nodeService->getNodeFromContextPath($envelope['$node']);
+                return $this->nodeService->getNodeFromContextPath($envelope['$node'], null, null, true);
             },
             $createContext['payload']
         ));
@@ -472,13 +472,13 @@ class BackendServiceController extends ActionController
         switch ($finisher['type']) {
             case 'get':
                 $result = $nodeInfoHelper->renderNodes($flowQuery->get(), $this->getControllerContext());
-            break;
+                break;
             case 'getForTree':
                 $result = $nodeInfoHelper->renderNodes($flowQuery->get(), $this->getControllerContext(), true);
-            break;
+                break;
             case 'getForTreeWithParents':
                 $result = $nodeInfoHelper->renderNodesWithParents($flowQuery->get(), $this->getControllerContext());
-            break;
+                break;
         }
 
         return json_encode($result);
--- a/Classes/Fusion/Helper/NodeInfoHelper.php
+++ b/Classes/Fusion/Helper/NodeInfoHelper.php
@@ -119,6 +119,7 @@ class NodeInfoHelper implements ProtectedContextAwareInterface
         $this->userLocaleService->switchToUILocale();
 
         $nodeInfo = $this->getBasicNodeInformation($node);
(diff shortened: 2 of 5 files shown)


mficzel:
@danielkestler We all mess up sometimes. Thanks a lot for pushing this topic. I think this is the biggest current usability problem we have.

danielkestler:
@mficzel Okay, thanks.

I am facing a problem ... The removed nodes get loaded correctly for the tree, but then disappear after a short blink – can anyone guide me to what is happening here? I guess it could have something to do with getAdditionalNodeMetadata?

johannessteu:
If i remember right `make node not disappear directly after removing` is the hard part at all. The merge action needs the nodes  to be inline in the document. As they are removed they are not there and that's why they disappear. We need to adjust the server side to also render meta data of deleted nodes as well. But that had other downsides

danielkestler:
Do you know where the removed nodes are filtered out? Is this happening while PropertyMapping or is there some kind of Doctrine query constraint set for nodes?

johannessteu:
@danielkestler in general it's when the context is created. You can pass some options to the `ContextFactory` like show removed/hidden nodes, workspace etc. I'll try to find the spot and post it here later!

johannessteu:
So, if i remember correct the merge action merges the current tree state with the inline meta data. 

This data is set here via Fusion:
https://github.com/neos/neos-ui/blob/2.x/Resources/Private/Fusion/Prototypes/Page.fusion#L111

wich actually calls the AutmentationAspect here
https://github.com/neos/neos-ui/blob/2.x/Classes/Aspects/AugmentationAspect.php#L262

So it uses the current documentNode / node to render the metadata. And that node already has a specific content context. We _could_ add `removedContentShown` to the default context over here
https://github.com/neos/neos-ui/blob/2.x/Classes/Controller/BackendController.php#L220
But this might have an impact on other things as well. I'm also not sure if this is the right place to do it. Need to dig deeper into that topic again. 

This is btw the interface for the content context options: 
https://github.com/neos/neos-development-collection/blob/master/Neos.ContentRepository/Classes/Domain/Service/ContextFactory.php#L70

danielkestler:
Wow, thank you @johannessteu. Will try it!

danielkestler:
Well, this was not the exact place, but thank you very much for guiding me to the (hopefully) right direction – context was the problem, see my last commit.

danielkestler:
Here you go, now the nodes also remain in the tree directly after removing them. Basically, this is now "feature-complete", but should of course be heavily tested in different and complex scenarios (node dimensions etc.).

danielkestler:
Struggling to get the e2e-tests passing, so if anybody has an idea ... fixed some of them (see above), but strangely cannot get the most obvious ones working.

danielkestler:
Still help needed regarding e2e-tests, otherwise this works great ;-)

mficzel:
@dimaip @markusguenther can you give advice here to get the e2e tests running. This would fix one of the worst ux problems we still have.

dimaip:
@mficzel thanks for the ping! I have a really crazy week, but I'll _try_ to find the time over weekend to help with this.
Or if you have a concrete question about TestCafe I can try to answer.

markusguenther:
Sorry at the moment I am in Denmark for vacation and don’t answer that fast. Just try to code stuff in the evenings 

Did not work out each day ;)

PR #2517, opened 2019-06-17
TASK: Adds filter to trigger circle ci build

markusguenther:
The circle ci tests are running only on branch changes by default.
So when a new tag has been pushed the build chain will not run.
This is now enabled for tags that follow the scheme x.x.x

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -125,7 +125,14 @@ workflows:
   version: 2
   build_and_test:
     jobs:
-      - checkout
+      - checkout:
+          filters:
+            tags:
+              only:
+                - /^\d\.\d\.\d/
+            branches:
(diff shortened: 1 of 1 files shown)


markusguenther:
With these PR we try to prevent issues like in the 3.6.0 release that we release a new version based on the wrong branch. The setup failed but we did not recognized it because the tags did not trigger circleci workflows.

markusguenther:
Bildschirmfoto 2019-06-18 um 07 49 27

It is hard to test here because you need to create a tag for that. So I have a small screenshot of my testing fork :)

The documentation related to that topic can be found here https://circleci.com/docs/2.0/workflows/#using-contexts-and-filtering-in-your-workflows

PR #2523, opened 2019-06-28
FEATURE: extract testing distribution

markusguenther:
**What I did**

I extracted the TestDistribution to a separate repository.
github.com:neos/neos-ui-testing-distribution

It is an uncomfortable to test e2e locally. With the separate distribution it is a bit easier 
to handle.


markusguenther:
One todo is definitely left ... need to find a way to cache the distribution.

markusguenther:
Ok Finally done, but we should squash all the experiments ;)

dimaip:
We addressed the problem with https://github.com/neos/neos-ui/pull/2564

PR #2524, opened 2019-07-01
FEATURE: Expose CK widget and highlight plugins

Sebobo:
This allows using them in custom packages without getting
errors and warning for duplicate plugins & modules.

The widget plugin allows creating inline widgets for
placeholders, icons etc…

The highlight plugins allows marking text with custom
classes and styles.

The class list is reformatted to see future diffs better
and get a faster overview of available classes.

These additional classes have to be imported like this:

`import {toWidget, Widget, HighlightEditing} from 'ckeditor5-exports';`


skurfuerst:
:+1:

PR #2529, opened 2019-07-11
BUGFIX: DataSource SelectBoxEditor must re-load data if options change

skurfuerst:
This is needed when using e.g. ClientEval, to dynamically evaluate DataSource options.


pmaas:
Hey @skurfuerst, thanks for the PR. Works as expected for me <3

PR #2531, opened 2019-07-15
FEATURE: Make Inline UI Buttons configurable by placing them in manifest

skurfuerst:
With this change, the Inline UI buttons are configurable; so that they can be extended properly.


hlubek:
Looks good to me! Would be nice to have this to extend the Node Toolbar with custom buttons.

PR #2534, opened 2019-07-19
FEATURE: make LinkInput options extendable

dimaip:
This PR makes LinkInput options extendable by putting them into a registry.

This is useful when you'd want to add some custom options there, e.g.:
image

Usage example: https://github.com/di-unternehmer/DIU.Neos.Linkeditor.Extension/blob/master/Resources/Private/LinkEditor/src/manifest.js#L24

PR #2537, opened 2019-07-31
BUGFIX: make LinkInputOptions work after refactoring

dimaip:
This commit broke the standard editor options of the link input: https://github.com/neos/neos-ui/pull/2534/commits/f8436887e147fdff7d9f5537e3bd529f379aa077
Here's the fix.

### DIFF
--- a/packages/neos-ui-editors/src/Library/LinkInputOptions.js
+++ b/packages/neos-ui-editors/src/Library/LinkInputOptions.js
@@ -8,7 +8,7 @@ import style from './LinkInput.css';
 const LinkInputOptions = ({
     i18nRegistry,
     onLinkChange,
-    options,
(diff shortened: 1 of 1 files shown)


akappler:
I tested it for inline Links with Neos.Demo on a text node type and on a document node type (Chapter) string property which is inline editable.

Following browser tested:
Firefox 68, Chrome 75, Safari 12.1, Edge 18 with default Neos.Demo Package

Also with Neos.Demo and a custom package https://github.com/di-unternehmer/DIU.Neos.Linkeditor.Extension to extend the Neos Link Editor with custom values. 
Tested with Firefox 68, Chrome 75 and Edge 18

dimaip:
Thanks for the extensive testing!

PR #2538, opened 2019-08-05
FEATURE: validate inline properties

dimaip:
Takes existing property validators configuration (`properties.propertyName.validation`) and applies it inline.
If property value is not valid, it will not be persisted until it becomes valid again.

TODO:
- [x] fix the tests
- [x] E2E tests
- [x] update the docs

valid

### DIFF
--- a/Tests/IntegrationTests/Fixtures/1Dimension/createNewNodes.e2e.js
+++ b/Tests/IntegrationTests/Fixtures/1Dimension/createNewNodes.e2e.js
@@ -1,13 +1,16 @@
-import {Selector} from 'testcafe';
+import {Selector, RequestLogger} from 'testcafe';
 import {ReactSelector} from 'testcafe-react-selectors';
 import {beforeEach, subSection, checkPropTypes} from './../../utils.js';
 import {Page} from './../../pageModel';
 
 /* global fixture:true */
 
+const changeRequestLogger = RequestLogger(request => request.url.endsWith('/neos/ui-services/change') && request.method === 'post' && request.isAjax);
+
 fixture`Create new nodes`
     .beforeEach(beforeEach)
-    .afterEach(() => checkPropTypes());
+    .afterEach(() => checkPropTypes())
+    .requestHooks(changeRequestLogger);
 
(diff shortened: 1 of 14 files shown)


dimaip:
I have written extensive e2e tests and also updated the docs https://github.com/neos/neos-development-collection/pull/2626

Please review!

lorenzulrich:
Great change :-)

bwaidelich:
Is there a specific reason why this only made it into 5.0.* 4.0.* and 2.9.* but not into any 3.* branch?

markusguenther:
@bwaidelich Hi, 

the 3.x branch is deprecated and the people should switch in the composer.json to 4.0.
Therefore this change has now been merged into this branch.

*We actually have:*
 * 2.9.x for Neos 3.3
 * 4.x for Neos 4.x
 * 5.x for Neos 5.x

bwaidelich:
@markusguenther Thanks for clarifying! Shouldn't we deprecate those versions on packagist.org, too?

markusguenther:
Hmm good point. Know that for whole packages, need to check how to do that for just a branch.

bwaidelich:
> need to check how to do that for just a branch.
I think we can just deprecate/remove the affected tags:

image

But I just stumbled upon a different issue:
When using string based validators, HTML tags are considered.
This wasn't an issue with the plaintext props in the inspector, but for inline editing this leads to unexpected effects:

image

Not sure how to deal with that..

markusguenther:
Ah that leads to the discussion in the #neos-ui channel I guess. Thats not that cool. Definite a bug and we should file an issue for that.

bwaidelich:
> Definite a bug and we should file an issue for that

OK. I think it can be fixed in the JS StringLengthValidator. I'll check and file a bug report / PR

kitsunet:
> OK. I think it can be fixed in the JS StringLengthValidator. I'll check and file a bug report / PR

Careful though, because in other places you might actually not want / need that, eg. StringLengthValidator on the page title or so.

dimaip:
Yup, just wanted to write that. Maybe add an option to it or smth?

PR #2543, opened 2019-08-14
FEATURE: Add EditPreviewDropdown to the Primary toolbar instead of the EditPreviewPanelToggler

mficzel:
With this change the editPreviewModes can be switched directly from the dropdown which saves lots of space and is way shorter than opening a panel. Since the change uses the spot of the exiting EditPreviewPanelToggler no additional place in the primary toolbar is used. Instead the now the title and edit/prebiew icon of the current mode are visible directly so editors actually habe a chance to spot in which mode they currently are.

An additional benefit is that the current editing mode become visible in screenshots of non-working backends which in the past often were caused by editors who accidentally switched into the preview mode.

This change allowed to remove the EditPreviewPanelToggler and EditPreviewPanel.


mficzel:
DevH-Hotels_com_-_Offizielle_Website_der_Hyperion__H4__H2___H__Hotels

dimaip:
I like the idea!

dimaip:
I agree, I wonder if all the things that this PR uses have been converted to TS already. We should definitely spend some effort to finish the TS integration!

mficzel:
@dimaip @markusguenther since the code this pr replaces is no ts either. That is why i wonder wether this is worth waiting. I am definetly not able to update this pr to ts.

dimaip:
I'd say unless you explicitly want to spend the time on this then no, it's definitely not a blocker

markusguenther:
Second that. Would not wait for the TS migration as I also have not the time right now :(

danielkestler:
@dimaip It might not be perfectly polished, but we could easily live with it right now and maybe get back to design refinement later on. So green light from my side.

PR #2547, opened 2019-08-30
BUGFIX: make Aloha config migration handle multiple nodetypes per file correctly

dimaip:
Thanks @baschny for reporting!


dimaip:
Don't do nodetypes :-)

Issue #107, opened 2016-04-04
Make default behavior of "Publish"/"Publish all" buttons configurable


grebaldi:
As described by @baschny in #guild-ux:

> There are some flaws in the publishing process that lead to frustration. One particular irritation is that not „everything“ is published, because the editors don’t perceive the difference between the „publish“ (big orange) and the „publish all“ (other pages also affected = small orange arrow). For my editors I would propose to switch the meaning: big button = „Publish all“ because in 90% the cases thats what you want to do. And hide the „Publish only this page“ inside the arrow dropdown menu. Doesn’t anyone of you have the same „problem“?

> for example just today: editor creates a „folder“ node and inside of the folder he creates his article.. he writes it up and hits „publish“.. but nothing is shown because its parent isn’t published.. I end up with buggy entries in ES, the list view shows the article but the link is incorrect (parentPath does not exist) etc..

@kitsunet's suggestion, how to proceed with this:

> the reasonable way is: configurable -> deprecate old behavior -> change behavior

This still needs discussion and probably won't be part of the current rewrite but could be considered for future development. Thus this issue is kept as a reminder.

danielkestler:
Fixed in my PR.

PR #2549, opened 2019-08-30
FEATURE: Make default behavior of "Publish"/"Publish all" buttons configurable

danielkestler:
Add a user preference ```publishingModeAll``` (true/false) to configure the publishing behavior, see: https://github.com/neos/neos-ui/issues/107

If set to true, the main publish dropdown button will publish all dirty nodes, if set to false, only the current document node.

Also fixes a bug (?): the auto-publishing settings have not been saved into user preferences until now, this is also fixed.

bildschirmfoto 2018-10-26 um 15 05 22

### DIFF
--- a/Configuration/Settings.yaml
+++ b/Configuration/Settings.yaml
@@ -184,7 +184,8 @@ Neos:
           preferences:
             interfaceLanguage: '${q(user).property(''preferences.interfaceLanguage'')}'
           settings:
-            isAutoPublishingEnabled: false
+            isAutoPublishingEnabled: '${q(user).property("preferences.preferences")["isAutoPublishingEnabled"]}'
+            isPublishingModeAll: '${q(user).property("preferences.preferences")["isPublishingModeAll"]}'
             targetWorkspace: 'live'
       changes:
         types:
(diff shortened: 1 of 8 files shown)


dimaip:
I'm still not sure such setting belongs to the UI, and not in Settings.yaml. If the main motivation is to make the publishing process simpler for novice editors, wouldn't presenting them with yet another setting would give them another opportunity to make a mistake?
Anyways, would love to chat with you about this feature, maybe I just miss something.

jonnitto:
I've talked with some of our editors and they would like to see this kind of feature in the UI. Perhaps it would make sense to make a role where I can enable or disable this certain setting for a user.

danielkestler:
Thanks for your feedback, @jonnitto and @dimaip ... I understand both points of view and maybe it would be best to gather some more feedback.

danielkestler:
I have added the "Publish all to" to the main publish dropdown button, if this mode is active. This is a new LL, so we have to make sure it exists in the Neos.Neos package.

davidspiola:
Hi Folks, is this a feature for the 5.1 release?

markusguenther:
@danielkestler It's been a while, do you want to continue the PR or shall we close it down and start again if necessary.

danielkestler:
I think we should close it and start again if this feature is still desired ;)

PR #2555, opened 2019-09-19
TASK: Pull in ^5.0 only

kdambekalns:
This pulls in only `neos/neos-ui` as `^5.0`. That in turn pulls in
`neos/neos` as `^5.0`, which in turn requires it's own dependencies.
That _should_ suffice. If it needs to be more pinned down, we could
add `"neos/neos": "~5.0.0"` as well, to limit it to 5.0.x.

### DIFF
--- a/Tests/IntegrationTests/TestDistribution/composer.json
+++ b/Tests/IntegrationTests/TestDistribution/composer.json
@@ -6,21 +6,7 @@
         "bin-dir": "bin"
     },
     "require": {
-        "neos/neos": "dev-master",
-        "neos/flow": "dev-master",
-        "neos/eel": "dev-master",
-        "neos/fluid-adaptor": "dev-master",
-        "neos/neos-ui": "dev-master",
-        "neos/neos-ui-compiled": "dev-master",
-        "neos/fusion-afx": "dev-master",
-        "neos/media": "dev-master",
-        "neos/content-repository": "dev-master",
-        "neos/fusion": "dev-master",
-        "neos/media-browser": "dev-master",
-        "neos/http-factories": "dev-master",
-        "neos/form": "dev-master",
-        "neos/neos-setup": "dev-master",
-        "neos/setup": "dev-master",
+        "neos/neos-ui": "^5.0",
         "neos/test-site": "@dev",
         "neos/test-nodetypes": "@dev"
     },
--- a/Tests/IntegrationTests/TestDistribution/composer.json
+++ b/Tests/IntegrationTests/TestDistribution/composer.json
@@ -7,7 +7,6 @@
     },
     "require": {
         "neos/neos-ui": "^5.0.0",
-        "neos/fusion-afx": "*",
         "neos/test-site": "@dev",
         "neos/test-nodetypes": "@dev"
     },
--- a/Tests/IntegrationTests/TestDistribution/composer.json
+++ b/Tests/IntegrationTests/TestDistribution/composer.json
@@ -7,6 +7,7 @@
     },
     "require": {
         "neos/neos-ui": "^5.0.0",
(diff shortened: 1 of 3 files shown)


dimaip:
We definitely need to have it limited to "^5.0.0", 5.1 will have a separate branch.

But why pull neos-ui? Shouldn't we pull neos/neos, and that would pull the UI?

dimaip:
Wait, why doesn't neos/neos require neos/neos-ui? https://github.com/neos/neos/blob/5.0/composer.json

I've thought we had changed it.

kdambekalns:
No, because `neos/neos` does not depend on the UI. But the other way around, it does. See https://packagist.org/packages/neos/neos

kdambekalns:
Why should `neos/neos` depend on the UI? What if I don't even need a UI? But the base dist does: https://packagist.org/packages/neos/neos-base-distribution

dimaip:
But why? That feels weird. neos/neos won't work without the UI.

kdambekalns:
Of course, if I just import things and use the rendering stack… think "headless CMS"

dimaip:
Yeah, perhaps it makes sense for the future. But as of now I guess the backend won't even boot without the UI.

Tests/IntegrationTests/TestDistribution/composer.json
```diff
@@ -6,7 +6,7 @@
         "bin-dir": "bin"
     },
     "require": {
-        "neos/neos-ui": "^5.0",
+        "neos/neos-ui": "^5.0.0",
```
kdambekalns: Now it limits it to *UI* 5.0.x - if that was intended, fine. So, it is planned to have releases for every branch in sync? I.e. 5.1.x, 5.2.x…

dimaip: yes, absolutely. See https://github.com/neos/neos-ui#versioning

kdambekalns:
> But as of now I guess the backend won't even boot without the UI.

Well, in "headless" I don't even try to log in. But yes, unless you change the default after login from the content module to something else, it won't work. But if you do, it should.

Tests/IntegrationTests/TestDistribution/composer.json
dimaip: the plan is to release UI and Neos exactly in sync, except for the bugfixes

dimaip:
Ugh
> The fusion dsl-object for the key afx was not configured

https://2023-47460181-gh.circle-artifacts.com/0/home/circleci/app/Data/Logs/Exceptions/20190919130849b56d3f.txt

dimaip:
Sorry for the messy commits, editing from the Github UI, will squash them

Tests/IntegrationTests/TestDistribution/composer.json
```diff
@@ -6,21 +6,8 @@
...
-        "neos/media": "dev-master",
-        "neos/content-repository": "dev-master",
-        "neos/fusion": "dev-master",
-        "neos/media-browser": "dev-master",
-        "neos/http-factories": "dev-master",
-        "neos/form": "dev-master",
-        "neos/neos-setup": "dev-master",
-        "neos/setup": "dev-master",
+        "neos/neos-ui": "^5.0.0",
+        "neos/fusion-afx": "*",
```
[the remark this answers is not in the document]
kdambekalns: They do! But… sometimes composer picks them up only reluctantly. E.g. you need to remove a symlink to a local package for `composer update` to pick up new/changed dependencies…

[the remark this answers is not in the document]
dimaip: Thanks for clarifying! For now I guess it would be simpler to just add it to main composer.json.

PR #2556, opened 2019-09-22
BUGFIX: Restore compatibility with IE11

### LINKED ISSUE #2508: Login with IE11 on 3.2.x + does not work
After login, the "N"-screen is waiting forever.

### Steps to Reproduce

1. Login with to the backend Neos UI 3.2.0 or newer

#### Expected behavior

The backend appears

### Affected Versions

Neos: Tested with 4.2 & 4.3

UI: Tested with 3.2 & 3.3

It worked with the Neos UI version 3.1

zickgraf:
This fixes #2508.

Notes:
1. I could not reproduce the original regular expressions using the website mentioned in the code (https://mothereff.in/regexpu). I assume the website has been updated and now uses a newer Unicode spec.
2. I could not get the `transpileNodeModuleRegex` to work. Thus, I included all three packages which should be transpiled separately. I think this has the additional benefit of being more explicit while not requiring more code.
3. There might be more issues hidden since I did not do extensive testing. However, at least editing pages (using CKEditor 4) works for me.


dimaip:
Thank you so much! Congrats on the first contribution!

PR #2568, opened 2019-10-17
FEATURE: batch operations on tree nodes

dimaip:
Resolves: #2599

Support multiple-selection on trees via these keyboard shortcuts:
- hold CMD/CTRL and click: select multiple nodes
- hold SHIFT and click: select a range of nodes (only on one level though)

When multiple nodes are selected it's possible to do operations on all of them at once: move them (via dragging or via cut/paste buttons), copy, hide, delete.

multi-demo

TODO:

- [x] hide
- [x] delete
- [x] copy
- [x] move
- [x] dnd
- [x] inspector summary
- [x] single move of unselected nodes
- [x] don't focus inline on batch content focus clicks
- [x] refactor selectors
- [ ] tooltip to make the feature discoverable
- [x] i18n
- [x] support Shift key for selecting ranges
- [ ] improve DND visualization
- [x] adjust the tests
- [ ] write e2e tests

In the future:
- [ ] batch editing in inspector


bwaidelich:
I added an issue for this in order to simplify release management and to collect higher-level details for the release notes: #2599

davidspiola:
@markusguenther and @jonnitto do you guys have time for a review? @dimaip said it's ready for merging

markusguenther:
Right now we have advent celebration with kids and grandma but can check the notebook later :)

Classes/Controller/BackendServiceController.php
```diff
@@ -348,12 +355,16 @@ public function changeBaseWorkspaceAction(string $targetWorkspaceName, NodeInter
...
      *
-     * @param NodeInterface $node
+     * @param array $nodes
      * @return void
      */
-    public function copyNodeAction(NodeInterface $node)
+    public function copyNodesAction(array $nodes)
     {
-        $this->clipboard->copyNode($node);
+        // TODO @christianm want's to have a property mapper for this
```
markusguenther: Is this Todo just a relict as you use the propertyMapper?

Classes/Controller/BackendServiceController.php
```diff
@@ -369,12 +380,16 @@ public function clearClipboardAction()
...
      *
-     * @param NodeInterface $node
+     * @param array $nodes
      * @return void
      */
-    public function cutNodeAction(NodeInterface $node)
+    public function cutNodesAction(array $nodes)
     {
-        $this->clipboard->cutNode($node);
+        // TODO @christianm want's to have a property mapper for this
```
markusguenther: Same as in copy nodes action

Configuration/Routes.Service.yaml
```diff
@@ -31,19 +31,19 @@
   httpMethods: ['POST']
 
 -
-  name: 'Copy node to clipboard'
-  uriPattern: 'copy-node'
+  name: 'Copy nodes to clipboard'
+  uriPattern: 'copy-nodes'
```
markusguenther: Isn`t that kind of breaking when we change the end-point URL, even it is makes sense?

packages/neos-ui/src/Containers/Modals/DeleteNode/index.js
```diff
@@ -47,20 +47,30 @@ export default class DeleteNodeModal extends PureComponent {
...
         return (
             <div>
                 <Icon icon="exclamation-triangle"/>
                 <span className={style.modalTitle}>
-                    <I18n id="Neos.Neos:Main:delete" fallback="Delete"/>
-                    &nbsp;
-                    <I18n id={nodeTypeLabel} fallback="Node"/>
-                    &nbsp;
-                    "{$get('label', node)}"
+                    Delete {nodesToBeDeletedContextPaths.length} nodes
```
markusguenther: We should use labels here.

Classes/Controller/BackendServiceController.php
dimaip: Oh no, @kitsunet said that in the future he wants to have all of this stuff to happen automatically on the property mapper level, so the comment should stay to keep reminding us of it :)

packages/neos-ui/src/Containers/Modals/DeleteNode/index.js
dimaip: Ouuuch, good catch, I thought I had replaced all of those, will fix

Configuration/Routes.Service.yaml
dimaip: Those endpoints are not part of our public API, it's merely an implementation detail to be able to keep the copied nodes between different sites/domains.

dimaip:
Hey @markusguenther, thanks a lot for thorough review!
I've fixed I18n stuff.

I think I saw your remark about not being able to delete nodes or something, but I can no longer find that comment... And deleting works just fine for me :)

markusguenther:
Yes in the first testing round I tried to move nodes and then delete them. But after the move the delete button was deactivated. So i thought it is a bug but maybe just the parent was also selected then. 

Because could not reproduce it anymore. Except you select nodes with their parent node. But guess that it is intended.

dimaip:
Yup, deleting nodes with parents doesn't make much sense and could lead to errors.
Theoretically we could automatically filter out all the children and permit delete in such situations, I don't know how much UX value would actually come from this...

dimaip:
Would be great to get more reviews on this, it's a huge feature after all!
I'd wait with merging till the end of the day, if no reviews come in I'd merge.

dimaip:
And yes, if somebody would be merging it without me this PR should definitely be squashed.

kdambekalns:
> this PR should definitely be squashed

Why? I firmly believe there is a lot of value in being able to see the development of a feature… And it doesn't add confusion, since you can always just look at the merge to see "the full change at once…"

dimaip:
@kdambekalns you can still see the original commits in the PR history, e.g. https://github.com/neos/neos-ui/pull/2589/commits
Having dozens of commits per each change is really annoying e.g. when doing git-bisect and in general when navigating history.
I'm fine with PRs to be merged if they have nice edited history, but in this particular case with all those "fix linter" commits, I would like them to end up in the codebase. Neither do I care about them enough to actually do `git rebase -i`...
But that's just my take on it, I'm open for doing things differently.

kdambekalns:
I was rather just interested in the reasoning… for me in this case the ration of useful/useless ("real change" vs "stupid fix") is quite good… :)

> you can still see the original commits in the PR history

But that assumes GH is used, I more often work with the history locally, using my favourite Git client. And there squash commits don't really work well, especially since the original branch shows up as unmerged (really confusing IMHO), and disappears when deleted.

Anyway, no blocker for anything, as long as I get my "proper merge" for my stuff, at least. ;)

dimaip:
I've merged it as is with just one review since I intend to keep on improving this feature (dnd + e2e), and I would be able to fix it future bugfix releases if something wrong is discovered.

PR #2571, opened 2019-10-20
BUGFIX: Do not force update within render method

gerhard-boden:
There was a call within the render method that contained `this.forceUpdate()`, which led to a warning in the console only visible for dev builds.

Moving the call to the `componentDidUpdate` lifecycle function resolves the issue while keeping the same functionality (`ClientEval` annotations are still evaluated).

### DIFF
--- a/packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
+++ b/packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
@@ -91,7 +91,7 @@ export default class Inspector extends PureComponent {
             );
 
             const processedViewConfiguration = this.preprocessViewConfiguration(
-                {node: nodeForContext}, [], {...originalViewConfiguration}, {...originalViewConfiguration}
+                {node: nodeForContext}, [], originalViewConfiguration, originalViewConfiguration
             );
 
             this.state.viewConfiguration = processedViewConfiguration || {...originalViewConfiguration};
@@ -128,7 +128,6 @@ export default class Inspector extends PureComponent {
     //
     preprocessViewConfiguration = (context = {}, path = [], viewConfiguration, originalViewConfiguration) => {
         const currentLevel = path.length === 0 ? viewConfiguration : $get(path, viewConfiguration);
-
         Object.keys(currentLevel).forEach(propertyName => {
             const propertyValue = currentLevel[propertyName];
             const newPath = path.slice();
--- a/packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
+++ b/packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
@@ -120,12 +120,14 @@ export default class Inspector extends PureComponent {
     // Update viewConfiguration, while keeping originalViewConfiguration to read original property values from it
     //
     preprocessViewConfiguration = (context = {}, path = []) => {
-        const currentLevel = path.length === 0 ? this.state.viewConfiguration : $get(path, this.state.viewConfiguration);
+        const viewConfiguration = {...this.state.viewConfiguration};
+        const originalViewConfiguration = {...this.state.originalViewConfiguration};
+        const currentLevel = path.length === 0 ? viewConfiguration : $get(path, viewConfiguration);
         Object.keys(currentLevel).forEach(propertyName => {
             const propertyValue = currentLevel[propertyName];
             const newPath = path.slice();
             newPath.push(propertyName);
-            const originalPropertyValue = $get(newPath, this.state.originalViewConfiguration);
+            const originalPropertyValue = $get(newPath, originalViewConfiguration);
 
             if (propertyValue !== null && typeof propertyValue === 'object') {
                 this.preprocessViewConfiguration(context, newPath);
@@ -133,11 +135,11 @@ export default class Inspector extends PureComponent {
                 const {node} = context; // eslint-disable-line
                 const evaluatedValue = eval(originalPropertyValue.replace('ClientEval:', '')); // eslint-disable-line
                 if (evaluatedValue !== propertyValue) {
-                    this.setState(prevState => ({
-                        viewConfiguration: produce(prevState.viewConfiguration, draft => {
+                    this.setState({
+                        viewConfiguration: produce(viewConfiguration, draft => {
                             setIn(draft, newPath, evaluatedValue);
                         })
-                    }));
+                    });
                 }
             }
         });
--- a/packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
+++ b/packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
@@ -101,7 +101,7 @@ export default class Inspector extends PureComponent {
         }
     }
 
-    componentDidMount() {
+    componentWillMount() {
         this.preprocessViewConfigurationDebounced();
     }
 
--- a/packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
+++ b/packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
@@ -88,12 +88,10 @@ export default class Inspector extends PureComponent {
 
     componentWillReceiveProps(newProps) {
         if (newProps.focusedNode !== this.props.focusedNode) {
-            this.setState(prevState => ({
+            this.setState({
                 viewConfiguration: newProps.nodeTypesRegistry.getInspectorViewConfigurationFor($get('nodeType', newProps.focusedNode)),
-                originalViewConfiguration: {...prevState.viewConfiguration}
-            },
-            () => this.preprocessViewConfigurationDebounced()
-            ));
+                originalViewConfiguration: newProps.nodeTypesRegistry.getInspectorViewConfigurationFor($get('nodeType', newProps.focusedNode))
+            });
         }
         if (!newProps.shouldShowSecondaryInspector) {
             this.setState({
--- a/packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
+++ b/packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
@@ -93,6 +93,10 @@ export default class Inspector extends PureComponent {
         }
     }
 
+    componentDidUpdate() {
+        this.preprocessViewConfigurationDebounced();
+    }
+
     componentWillUnmount() {
         // Abort any debounced calls
         this.preprocessViewConfigurationDebounced.cancel();
@@ -241,7 +245,6 @@ export default class Inspector extends PureComponent {
             return this.renderFallback();
         }
 
-        this.preprocessViewConfigurationDebounced();
         const {viewConfiguration} = this;
 
         if (!$get('tabs', viewConfiguration)) {
--- a/packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
+++ b/packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
@@ -101,6 +101,10 @@ export default class Inspector extends PureComponent {
         }
     }
 
+    componentDidMount() {
+        this.preprocessViewConfigurationDebounced();
+    }
+
     componentDidUpdate() {
         this.preprocessViewConfigurationDebounced();
     }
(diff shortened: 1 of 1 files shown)


gerhard-boden:
Hi dmitiri 😄 

Thanks for having a look at this!

Sorry, not too familiar how this evaluation works in detail so excuse my stupid questions 🙈 :
Why the first render is important for the editors. When field testing the change I noticed that multiple re-renderings take place (with this change even two less  which I consider a good thing).  So the `ClientEval` code I tested worked correctly in my test case. 🤷‍♂ 

We could probably save another render call if we do the evaluation in `componentDidMount` like you suggested,  but I think I had concerns about calling `this.forceUpdate()` so early (not too familiar with `forceUpdate()` in general, always tried to avoid it best I could).

> Or we can use the `useState` hook as well.

We can't use hooks inside class components if I understood you correctly.

> How about `getDerivedStateFromProps` then

This could work, so this is the improved version of `componentWillReceiveProps`. 🤔
But can we call `this.forceUpdate()` from there?!

dimaip:
Hey!

> Why is the first render so important for the inspector editors?

Because some editor could break if it got passed the original uneval'd string. E.g. imagine some editor depending on its input to be a date, but it gets string instead.

> When field testing the change I noticed that multiple re-renderings take place

Yeah that's currently true, but I don't feel safe relying on this side-effect. E.g. some day we may actually fix it, and then there'd be a bug here.

> We can't use hooks inside class components if I understood you correctly.

True, forgot about that. I mean if writing it from scratch I'd probably just not use a class, but no sense in re-writing it now.

> But can we call this.forceUpdate() from there?!

Frankly I have no idea why we used `forceUpdate` at all. Why couldn't we put `this.viewConfiguration` into state instead? So basically at this line we'd do smth like
`this.setState({viewConfiguration: this.viewConfiguration})`: https://github.com/neos/neos-ui/blob/a21afa5d9df368b73cb52bc81d8a009c4623800f/packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js#L156

While writing this I realised that still we would be able to use `getDerivedStateFromProps` because we calculate viewConfiguration async.

======================

So, bottom line, at the very least we should call `this.preprocessViewConfigurationDebounced` also from `componentDidMount`, just to make it a tad more robust.

In addition to that, we may want to call `this.preprocessViewConfiguration` (not debounced) in the constructor, so at initial render all values would be eval'd.
 
Does my reasoning make sense or am I overcomplicating things?

dimaip:
Wdyt, @gerhard-boden?

gerhard-boden:
Thanks for the detailed answers @dimaip!

I decided to call ``preprocessViewConfigurationDebounced`` in ``componentDidMount`` and everything was fine.

But I also tried to be smart and moved the viewConfiguration class properties into state (mostly to get rid of the forceUpdate alltogether). 

But now the inspector is not updated correctly anymore when switching to a different node.. 😢 
I guess something bad goes wrong around here:
https://github.com/neos/neos-ui/pull/2571/files#diff-a7ae28cc7ab235343fc04fe7203f0b82R91-R96

Any tips what I could try to make things right? Or should I just move the configuration from state back into class properties?

dimaip:
Hey, it's totally a good idea to move `viewConfiguration` into the state.
Did you debug that condition on line 90, does it get through on node update?
Anyways, don't have any quick tips, the code looks fine at first sight, needs debugging.

packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
```diff
@@ -73,17 +73,27 @@ export default class Inspector extends PureComponent {
...
     }
 
     componentWillReceiveProps(newProps) {
         if (newProps.focusedNode !== this.props.focusedNode) {
-            this.cloneViewConfiguration(newProps);
+            this.setState(prevState => ({
+                viewConfiguration: newProps.nodeTypesRegistry.getInspectorViewConfigurationFor($get('nodeType', newProps.focusedNode)),
+                originalViewConfiguration: {...prevState.viewConfiguration}
+            },
+            () => this.preprocessViewConfigurationDebounced()
```
dimaip: why is this call necessary? We already have one in `componentDidUpdate`, right?

gerhard-boden: Without this callback the view configuration processing was off (for example a conditional field in the inspector would vanish after applying changes to the node). Of course this should not be necessary since we do the processing after any update and trigger a re-rendering. But I think this bug is connected to the general current update issue.

gerhard-boden:
Thanks for the feedback and reassurance that I'm still on the right track 😃  Hopefully will find time to work on this later this evening or tomorrow. Will do some debugging, might be some sort of race condition where the processing is using stale data.

And yes, the condition is fulfilled on node change, but need to do more research where things go south.

gerhard-boden:
Found the problem 🎉 
https://github.com/neos/neos-ui/pull/2571/commits/f9b2c6b811fdd47ce010a54408004177e1bae8de
For some reason I read from``prevState`` when setting the ``originalViewConfiguration``, which is of course not correct in many cases. 

image
The OCD person in me is pleased how this turned out 😂

gerhard-boden:
Another thing about `componentDidMount`: The first render will still have "unevaluated" view configuration. So ``componentWillMount`` is what we actually want (although it has been deprecated for newer versions of React).

dimaip:
What if we do it in constructor, like I've suggested above?

gerhard-boden:
I tried to avoid that because I was afraid things would get bloated.
See we would have to do this in the constructor as well, correct? 
https://github.com/neos/neos-ui/blob/26525f644515d2cc1cb0eec5e995608e3951daec/packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js#L147-L155

Should we outsource that to another method? How would we name it? That's basically the pre-processing for the pre-processing 😂  But srsly, should we name it `generatePreprocessingContext`?

dimaip:
Exactly.
Name it whatever you feel like, I'm won't be the one nitpicking on it ;)

gerhard-boden:
I tried moving the pre-processing to the constructor, but it brought implications that made things more complicated than they already are.

I published this commit just to demonstrate what I mean:
https://github.com/gerhard-boden/neos-ui/commit/563d2647ff1180539201dee1e88ad4b19b64a1d5

We can not call a method that uses ``setState`` from constructor, so we'd have to move the state update out of ``preprocessViewConfiguration`` and with that a couple of more changes. This lead to more overhead and complicated the code in my eyes.

So I moved the pre-processing from ``componentDidMount`` to ``componentWillMount``, because this also guarantees preprocessed view configuration on first render. 

If you insist on doing it in the constructor I can have another look at it. The commit linked above is not working correctly currently, but I aborted because it felt like investing time in making things worse 🙈

dimaip:
Hey, thanks again for the extensive research!

`componentWillMount` is deprecated, we should not use it.

> This lead to more overhead and complicated the code in my eyes.

Why? I think the code actually improved. E.g. before we would do multiple setState calls per each `clientEval:` string, and now we'd do it only once.

gerhard-boden:
Thanks for the super fast feedback 😊 
Left some answers to your code comments.

gerhard-boden:
I sat down again and made it work to preprocess the configuration inside the constructor, before setting the intial state.

Thanks for your advice in the code comments!

Do we have a helper function to compare objects?
https://github.com/neos/neos-ui/pull/2571/files#diff-a7ae28cc7ab235343fc04fe7203f0b82R165

Just using `===` won't work because objects are reference types and it will never refer to the same object in this case.

dimaip:
Oh, we are cloning the viewConfiguration (with spread), I haven't noticed that. I that case plain equality check won't work.
Let's have a flag to detect if there were any changes, ok? It would be way faster than doing deep equality check.

gerhard-boden:
> Let's have a flag to detect if there were any changes, ok? It would be way faster than doing deep equality check.

`JSON.stringify` is pretty fast when it comes to comparing objects, but it comes with limitations (if the keys are not in the same order it will still consider the objects to be different).

I reintroduced the flag very pretty similar to the way it was in the original code.

packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
```diff
@@ -73,17 +73,38 @@ export default class Inspector extends PureComponent {
...
+
+        if (props.focusedNode) {
+            const originalViewConfiguration = props.nodeTypesRegistry.getInspectorViewConfigurationFor($get('nodeType', props.focusedNode));
+            const nodeForContext = this.generateNodeForContext(
+                props.focusedNode,
+                props.transientValues
+            );
+
+            const processedViewConfiguration = this.preprocessViewConfiguration(
+                {node: nodeForContext}, [], {...originalViewConfiguration}, {...originalViewConfiguration}
```
dimaip: Please pardon my ignorance, but why is it necessary to clone (shallowly) the originalViewConfiguration here?
We are never mutating this object, we are using `produce`.
Besides, if we needed to clone it, why are we doing it shallowly? It won't save from deep state mutations.

And I'm sorry for missing your changes and taking so long to review them :(

gerhard-boden: Hi @dimaip , don't worry about it! Just wanted to get it in as long as I remember a wee bit what I was doing here 🙈  
Maybe I don't  understand how `immer.js` works, but aren't we changing the reference here https://github.com/neos/neos-ui/pull/2571/files#diff-a7ae28cc7ab235343fc04fe7203f0b82R139 and here https://github.com/neos/neos-ui/pull/2571/files#diff-a7ae28cc7ab235343fc04fe7203f0b82R145

I removed the swallow copy in your marked line, because this not state (yet), but I feel safer leaving it in here: https://github.com/neos/neos-ui/pull/2571/files#diff-a7ae28cc7ab235343fc04fe7203f0b82R165-R166

I think a deep copy is not necessary since `produce` is it in place and we're only changing the reference.

gerhard-boden:
> See comment

Left a commit and an answer 😊

packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
dimaip: @gerhard-boden no, I don't think we are mutating anything. We are assigning a totally new object to the variable, that's not mutating the old object.
E.g. consider this code:
```
let a = {
first: 'object'
}
let b = a
a = {
  totally: 'different object'
}
console.log(b)
// {first: "object"}
```

dimaip:
Replied. If you like we could have a call and pair-code this PR to finish it.

packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
```diff
@@ -93,65 +114,74 @@ export default class Inspector extends PureComponent {
...
+        const nodeForContext = this.generateNodeForContext(
+            this.props.focusedNode,
+            this.props.transientValues
+        );
+
+        this.configurationIsProcessed = false;
+        const processedViewConfiguration = this.preprocessViewConfiguration(
+            {node: nodeForContext},
+            [],
+            {...this.state.viewConfiguration},
```
dimaip:
```suggestion
            this.state.viewConfiguration,
```

packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
```diff
@@ -93,65 +114,74 @@ export default class Inspector extends PureComponent {
...
+            {...this.state.originalViewConfiguration}
```
dimaip:
```suggestion
            this.state.originalViewConfiguration
```

gerhard-boden:
> Replied. If you like we could have a call and pair-code this PR to finish it.

Sure why not, would accelerate this subject matter quit a bit, otherwise it's just back and forth over weeks. Will contact you on Slack so we can find a fitting date. 😊

gerhard-boden:
image

Whenever some faulty evaluation would break the backend, we now show this warning in the console.
(this should also make it easier to find the culprit)

dimaip:
Could you fix  the style please?

dimaip:
Yay! Thanks for your patience!

gerhard-boden:
Thanks for your support! Really happy we made it to the merge 🤗 🎉

markusguenther:
💙

Issue #2579, opened 2019-11-18
Allow StringLengthValidator to ignore HTML tags


bwaidelich:
# Description

The `StringLengthValidator` counts the characters of the given string including all HTML tags.
This is consistent with the corresponding Flow validator and makes sense for regular strings.
Since #2538 we can use the validators for Inline Editing too and the current behavior is quite confusing because an RTE text contains a lot of (hidden) tags:

![image](https://user-images.githubusercontent.com/307571/69053612-18307100-0a0a-11ea-8509-f752973e24ff.png)

We should extend the validator so that it can (optionally) strip HTML tags before counting the characters.

PR #2580, opened 2019-11-18
FEATURE: Allow StringLengthValidator to ignore HTML tags

bwaidelich:
Adds an option `ignoreHtml` to the StringLengthValidator

Usage:

      validation:
        'Neos.Neos/Validation/StringLengthValidator':
          maximum: 10
          ignoreHtml: true

Resolves: #2579

### DIFF
--- a/packages/neos-ui-validators/src/StringLength/index.tsx
+++ b/packages/neos-ui-validators/src/StringLength/index.tsx
@@ -32,8 +32,8 @@ const StringLength = (value: any, validatorOptions: StringLengthOptions) => {
 
     let castedValue = value.toString ? value.toString() : '';
     if (validatorOptions.ignoreHtml) {
-        const document = new DOMParser().parseFromString(castedValue, 'text/html');
-        castedValue = document.body.textContent ? document.body.textContent : castedValue;
+        const doc = new DOMParser().parseFromString(castedValue, 'text/html');
+        castedValue = doc.body.textContent ? document.body.textContent : castedValue;
     }
     const stringLength = castedValue.length;
     if (stringLength < minimum || stringLength > maximum) {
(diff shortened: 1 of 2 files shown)


packages/neos-ui-validators/src/StringLength/index.tsx
```diff
@@ -28,7 +30,12 @@ const StringLength = (value: any, validatorOptions: StringLengthOptions) => {
         return 'The minimum StringLength can not be less than zero';
     }
 
-    const stringLength = (value !== undefined && value !== null && value.toString) ? value.toString().length : 0;
+    let castedValue = value.toString ? value.toString() : '';
+    if (validatorOptions.ignoreHtml) {
+        const document = new DOMParser().parseFromString(castedValue, 'text/html');
```
markusguenther: The code itself will work but we need to rename the variable because document is a reserved word. Like class or so.

bwaidelich:
*NOTE:* The corresponding Flow change is required for this to work without errors: https://github.com/neos/flow-development-collection/pull/1849

packages/neos-ui-validators/src/StringLength/index.tsx
kdambekalns: and isn't `castValue` grammatically the same as `castedValue` and the latter wrong (language-wise?) Never stop learning: https://grammarist.com/usage/casted/ – so, it's ok, a nice contrast to our modern TS codebase. ;)

bwaidelich: Haha, right, maybe we should call it `castleValue`? :)
Feel free to change both names.

markusguenther:
Tested and looks good. Just added some tests.

PR #2584, opened 2019-11-21
[TASK] Update install command for dev-master

johachi:
**What I did**
Updated to correct install command for `dev-master`.


markusguenther:
Thanks again @johachi  🙇 
You are right, when you`re not using the development distribution of neos it will fail because the compiled and regular ui always be a twin couple.

Issue #2377, opened 2019-02-28
Make user aware of validation failure in other tabs


lorenzulrich:
### Description

In the inspector, if property validation fails in another tab that you're currently on, you are not aware about this. You just cannot apply the changes and see an OK button with an exclamation mark in front:

![image](https://user-images.githubusercontent.com/1816023/53559309-23750100-3b4a-11e9-8915-94853f44acf2.png)

Only by checking all tabs, you will find out that you have a validation error:

![image](https://user-images.githubusercontent.com/1816023/53559347-2f60c300-3b4a-11e9-844c-53d16f46d234.png)

#### Expected behavior

It would be good to somehow mark tabs with validation errors in a way the user knows where the validation error occurs:

![image](https://user-images.githubusercontent.com/1816023/53559555-a4cc9380-3b4a-11e9-8a5b-ce95b2202d77.png)

### Affected Versions

UI: tested only in 1.x

dimaip:
A very important suggestion! You may also drop it into #guild-ux

dfeyer:
If we can make it even more visible, with a red counter of the amout of error in each tag, it can be nice, but the red line is a good start, and way better than nothing.

dfeyer:
Or maybe the full tab icon can be red when containing error, should be as easy to implement than the slim red bottom border

wrybit:
Nice Idea!

I've got a tiny question:
Will the red bar be stacked on-top of the blue (active) bar, 
if the error happens in the active tab – or not visible?
If not visible .. this inconsistent could cause some confusion.
If stacked .. this could become an issue for colorblind persons.
.. or it's placed on top of a tab, not as sexy but quite functional.

danielkestler:
I‘d make the top border (the blue one if active) red. Or the whole tab nav item.

lorenzulrich:
Just to clarify, my screen was not meant as a design suggestion. I think the bottom line generally is the indicator of the selected tab, so I'm not sure it's clever to abuse it for the "validation failed" warning.

mficzel:
Since this is a very important information i prefer the suggestion of @dfeyer (a red counter beside the tab icon). This cannot be confused with active state and with enough contrast it should also be visible for all audiences.

lorenzulrich:
So it would be something like this?

![image](https://user-images.githubusercontent.com/1816023/53634745-46bbb100-3c1b-11e9-8ba5-cc00219c47bd.png)

[...]

PR #2397, opened 2019-03-19
TASK: Display number of validation errors for each tab in inspector

lorissikora:
**What I did**
Completed issue #2377. The number of errors now gets displayed as a badge at the top right corner. This helps to show, why the "Apply" button is disabled when errors are in other tabs and thus not visible.

**How I did it**
The "validationError" object was already available in the redux store. I used it to summarize a count for each tab by testing how many of the field ids, whose values are faulty, are contained inside a tab group. Finally, I created a notification badge for the "Tabs" component, which is displayed if the count is higher than 0.

**How to verify it**
Delete the value of for instance a mandatory field and check if a badge with the number 1 will appear.

Bildschirmfoto 2019-03-19 um 16 53 20

PR #2589, opened 2019-11-26
FEATURE: Display number of validation errors for each tab in inspector

markusguenther:
What I did
Completed issue #2377. The number of errors now gets displayed as a badge at the top right corner. This helps to show, why the "Apply" button is disabled when errors are in other tabs and thus not visible.

How I did it
The "validationError" object was already available in the redux store. I used it to summarize a count for each tab by testing how many of the field ids, whose values are faulty, are contained inside a tab group. Finally, I created a notification badge for the "Tabs" component, which is displayed if the count is higher than 0.

How to verify it
Delete the value of for instance a mandatory field and check if a badge with the number 1 will appear.

Resolves:  #2377


markusguenther:
As replacement for https://github.com/neos/neos-ui/pull/2397

Was not able to change the PR anymore. So I created a new one and used @lorissikora changes.
Thanks for the Idea and basic work Loris.

Hope it's in your interest.

markusguenther:
Thanks to @lorissikora for bringing that topic on the table. This is a really nice feature.
Would be so happy if you find time to give neos a second try ;)

dimaip:
Could you please attach a screenshot @markusguenther

markusguenther:
Bildschirmfoto 2019-11-27 um 14 17 41
Little Screenshot :)

packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
```diff
@@ -277,12 +281,21 @@ export default class Inspector extends PureComponent {
...
+                                }, 0) : 0;
+
                             return (
                                 <TabPanel
                                     key={$get('id', tab)}
                                     id={$get('id', tab)}
                                     icon={$get('icon', tab)}
                                     groups={$get('groups', tab)}
+                                    notifications={notifications}
+                                    title={Boolean(notifications) && <Badge className={style.tabs__notificationBadge} label={String(notifications)}/>}
```
dimaip: Could we maybe have a more informative label for notifications? E.g. "3 properties in this tab are failing validation". That would require I18n ofc.

dimaip:
I like the looks @markusguenther! Left one suggestion.

packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
markusguenther: Sure will take care

markusguenther:
@dimaip Did you mean something like that?
Bildschirmfoto 2019-11-27 um 15 16 21

dimaip:
Yes, so such tooltip would be shown on hover

markusguenther:
Bildschirmfoto 2019-11-27 um 16 29 25

Added a tool tip and found sadly an UI bug.

Sebobo:
Bit long label. Especially when we translate it to German.

Maybe just `General - 2 validation issues`?

dimaip:
Typography nazi: also it should be a endash, not a minus sign :)

markusguenther:
Changed the labels and added the special dash ;)
The plural groups are obviously not supported in the UI.

Try to create an PR for that and as the handler takes the first one it works for german and danish.
Just in english it is bit weird to have 2 issue and not issues.

bwaidelich:
I really like this, looking at the screenshots you provided. But I failed to test this locally..
What I did:
```
git checkout -b markusguenther-lorissikora-2.x master
git pull https://github.com/markusguenther/neos-ui.git lorissikora-2.x
make setup
make build
```

(and set `frontendDevelopmentMode: true` ofc)

but no badges for me..

markusguenther:
Thanks @bwaidelich ... bit late today (came back from fire brigade) but will check that tomorrow.

markusguenther:
Bildschirmfoto 2019-11-30 um 13 34 14

Tested this on a fresh installation and had the same issue.
Reason was the error detection. So I now extracted that and now it works  in different setups.

@bwaidelich hope it now also works for you :)

jonnitto:
thank you @markusguenther! This one is a very very important feature! Thank you for taking care! ❤️

packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
```diff
@@ -219,10 +222,24 @@ export default class Inspector extends PureComponent {
         this.setState({toggledPanels: newState});
     };
 
+    getAmountOfValidationErrors = (tab, validationErrors) => {
+        let errors = 0;
```
dimaip: Btw, this could also be a `reduce` (no need to change anything now, just saying)

PR #2591, opened 2019-11-27
FEATURE: Parse backend errors instead of showing raw code

Sebobo:
**What I did**

Show human readable output instead of code in the error flash messages.

**How I did it**

Check what kind of text we have in a backend flash message error and parse it.

**How to verify it**

Go to the Forms example in the Demo site and then:

Force JSON error: Rename the form nodetype datasource in yaml to something else.

Force 500 HTML error: break PHP code in form datasource. F.e. by adding `$foo = $forms * $formDefinitions;`

JSON error before:

Bildschirmfoto 2019-11-27 um 12 04 02

JSON error after:

Bildschirmfoto 2019-11-27 um 12 03 34

HTML error before:

Bildschirmfoto 2019-11-27 um 12 02 20

HTML error after:

Bildschirmfoto 2019-11-27 um 12 01 39

### DIFF
--- a/packages/neos-ui/src/index.js
+++ b/packages/neos-ui/src/index.js
@@ -152,6 +152,30 @@ function * application() {
     });
 
     fetchWithErrorHandling.registerGeneralErrorHandler((message = 'unknown error') => {
+        // Check if the message is a Flow exception response
+        if (message.indexOf('Flow-Debug-Exception-Header') >= 0) {
+            const htmlContainer = document.createElement('div');
+            htmlContainer.innerHTML = message;
+            const exceptionHeader = htmlContainer.querySelector('.Flow-Debug-Exception-Header');
+            if (exceptionHeader && exceptionHeader.textContent) {
+                const exceptionSubject = exceptionHeader.querySelector('.ExceptionSubject');
+                message = exceptionSubject.textContent + ' - Check the logs for details';
+            } else {
+                message = 'Unknown error from unexpected HTML response. Check the logs for details.';
+            }
+        }
+        // Check if the message is a JSON string
+        else if (message.indexOf('{') >= 0) {
+            try {
+                const jsonMessage = JSON.parse(message);
+                if (jsonMessage.error && jsonMessage.error.message) {
+                    message = jsonMessage.error.message + ' - Reference code "' + jsonMessage.error.referenceCode + '"';
+                } else {
+                    message = 'Unknown error from unexpected JSON response. Check the logs for details.';
+                }
+            } catch (e) {}
+        }
+
         store.dispatch(actions.UI.FlashMessages.add('fetch error', message, 'error'));
     });
 
--- a/packages/neos-ui/src/index.js
+++ b/packages/neos-ui/src/index.js
@@ -165,7 +165,7 @@ function * application() {
             }
         }
         // Check if the message is a JSON string
-        else if (message.indexOf('{') >= 0) {
+        else if (message.trim()[0] === '{') {
             try {
                 const jsonMessage = JSON.parse(message);
                 if (jsonMessage.error && jsonMessage.error.message) {
--- a/packages/neos-ui/src/index.js
+++ b/packages/neos-ui/src/index.js
@@ -163,9 +163,8 @@ function * application() {
             } else {
                 message = 'Unknown error from unexpected HTML response. Check the logs for details.';
             }
-        }
-        // Check if the message is a JSON string
-        else if (message.trim()[0] === '{') {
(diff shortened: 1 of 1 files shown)


Sebobo:
Note: I'm not super happy with the implementation right now. 
This should be wrapped somewhere into a helper or something like that. So I'm very open for feedback.

kdambekalns:
To me this looks very slim and easy, in other words: maintainable. Wrapping that into something adds abstraction that's not needed…

packages/neos-ui/src/index.js
```diff
@@ -152,6 +152,32 @@ function * application() {
     });
 
     fetchWithErrorHandling.registerGeneralErrorHandler((message = 'unknown error') => {
+        // Check if the message is a JSON string
+        if (message.indexOf('{') >= 0) {
```
dimaip: usually people write `!== -1` or even `message.includes('{')`, but nevermind

dimaip: What if the Flow's html error message would contain `{` somewhere in the output?

I would first test for `message.includes('Flow-Debug-Exception-Header')`, then in the else block (since there's no need to test at the same time for Flow's error message and json error message at the same time) try to convert it from JSON and in case if it fails then output this fallback `Unknown error message thing`.

packages/neos-ui/src/index.js
```diff
@@ -152,6 +152,32 @@ function * application() {
...
+                }
+            } catch (e) {}
+        }
+
+        // Check if the message is a Flow exception response
+        if (message.indexOf('Flow-Debug-Exception-Header') >= 0) {
+            const htmlContainer = document.createElement('div');
+            htmlContainer.innerHTML = message;
+            const exceptionHeader = htmlContainer.querySelector('.Flow-Debug-Exception-Header');
+            debugger;
```
dimaip: Boom!

.nvmrc
```diff
@@ -1 +1 @@
-8.11.0
+8.16
```
dimaip: I'd rather have this as a separate PR, if you want to update node.

dimaip:
> Note: I'm not super happy with the implementation right now.
> This should be wrapped somewhere into a helper or something like that. So I'm very open for feedback.

I agree with @kdambekalns, if the thing is used only in one place and very unlikely to ever be needed, please leave it in place.
Abstracting it away in the future when needed is trivial, having an abstraction without cause adds a mental complexity with 0 benefit.

packages/neos-ui/src/index.js
```diff
@@ -152,6 +152,30 @@ function * application() {
...
+            const exceptionHeader = htmlContainer.querySelector('.Flow-Debug-Exception-Header');
+            if (exceptionHeader && exceptionHeader.textContent) {
+                const exceptionSubject = exceptionHeader.querySelector('.ExceptionSubject');
+                message = exceptionSubject.textContent + ' - Check the logs for details';
+            } else {
+                message = 'Unknown error from unexpected HTML response. Check the logs for details.';
+            }
+        }
+        // Check if the message is a JSON string
+        else if (message.indexOf('{') >= 0) {
```
dimaip: Why have this condition at all here? Why not just `else {`?
Also if you insist on having it, you could at least harden it to this: `message.trim()[0] === '{'` (every JSON begins with "{", right?).
But I would not do it. The best way to detect if it's JSON or not is to actually parse it.

Sebobo: I prefer to keep the condition to skip the parse when not needed. But I can adjust it as you said.

dimaip: Alright, so you are trying to optimise performance of it? I guess it should almost be identical, as the parser would tumble upon any first invalid character, but I don't mind.

packages/neos-ui/src/index.js
```diff
@@ -152,6 +152,29 @@ function * application() {
...
+                message = exceptionSubject.textContent + ' - Check the logs for details';
+            } else {
+                message = 'Unknown error from unexpected HTML response. Check the logs for details.';
+            }
+        } else if (message.trim()[0] === '{') {
+            // Check if the message is a JSON string
+            try {
+                const jsonMessage = JSON.parse(message);
+                if (jsonMessage.error && jsonMessage.error.message) {
+                    message = jsonMessage.error.message + ' - Reference code "' + jsonMessage.error.referenceCode + '"';
```
dimaip: could have used a template literal here btw, but I don't mind to have it as is

Sebobo: I'm not used to them and always forget...

PR #2594, opened 2019-11-27
BUGFIX: Concat the label parts of the param replacement

markusguenther:
The method substitutePlaceholders replaces the markers in language
labels with the given variables. For that all substrings will be saved in an
array and the array will be returned.

This patch concats the array items to a string. Because the translate method returns
strings in general and not arrays.

**What I did**
Just concat the array items to a string so that the return of the translate is consistent.

**How to verify it**

Use a label in JS like that.
```javascript
const label = i18nRegistry.translate(
    'labelId',
    'Default',
    {
        paramA: foo,
        paramB: bar
    },
    'Neos.Neos.Ui'
);
```

Create your label like
```xml
<trans-unit id="labelId" xml:space="preserve">
    <source>{paramA}: Don't be a fool, {paramB}</source>
</trans-unit>
```

Without the params the translate method will just return `{paramA}: Don't be a fool, {paramB}` as string.
But when you submit params you will get an array like that:
```javascript
[
    "",
    "foo",
    ": Don't be a fool, ",
    "bar"
]
```

So this patch will join that and you have your regular label as string with params.

### DIFF
--- a/packages/neos-ui-i18n/src/registry/I18nRegistry.js
+++ b/packages/neos-ui-i18n/src/registry/I18nRegistry.js
@@ -56,7 +56,7 @@ const substitutePlaceholders = function (textWithPlaceholders, parameters) {
 
     result.push(textWithPlaceholders.substr(offset));
 
-    return result;
+    return result.join('');
 };
 
 export default class I18nRegistry extends SynchronousRegistry {


markusguenther:
As this touches really old code in the react ui and it is just copied from the emberjs  ui it would be great if some of the elders take a look ;)

kitsunet:
Is master right though? Shouldn't this bugfix go into previous branches?

markusguenther:
I asked you because I am not sure if it can break something. If there is a usage I did not find and the people work around that like me in my other PR, than it breaks their code.

But I did not find something, but don’t know everything.
If it is fine for you we can pin it to 2.x because the code is there from the very first beginning of the ui.

kitsunet:
I just think we don't use placeholders much and this hasn't been noticed since ever. I would never expect translate to return anything else than a string.

markusguenther:
Ok then I will adjust that tomorrow and use 2.x as base branch.

markusguenther:
Thank for your inputs 🙏

Issue #2500, opened 2019-05-21
Preview button leads to validation error for hidden pages


Sebobo:
### Description

When opening the preview for a hidden page a validation error shows up instead of handling the error or showing the preview.

### Steps to Reproduce

1. Hide page in backend
2. Click on preview button

#### Expected behavior

Either a meaningful message or the actual preview.

If it shouldn't be possible to see the preview for hidden pages, the editor shouldn't get the option to open it. 
If it should be possible, then we need to show the rendered page.

#### Actual behavior

The following error appears: 

```Validation failed while trying to call Neos\Neos\Ui\Controller\BackendController->redirectToAction()```

### Affected Versions

Neos: 4.3

UI: 3.4

Sebobo:
Older versions of the UI might be affected too, but didn't have time yet to check.

kdambekalns:
The same holds true for "hidden before/after" and it worked with the legacy UI, according to a customer.

kdambekalns:
So it seems quite some people are triping over this and the "ugly URLs" as well as hidden things being shown unexpectedly.

Maybe a solution would be to add the parameter only when previewing a document that is itself hidden. That would avoid the problem, but keep the behaviour for other documents?!

dimaip:
I would expect such feature to work this way: hidden status on the document node would be ignored, but all other nodes within it would respect their hidden status.

dlubitz:
@kdambekalns At the moment this featue leads to some (hopefully unexpected) bahavior. If I have an hidden content node and click the preview button, I see the hidden content in the preview. (see screenshots). AFAIS the feature should just show hidden document nodes, but not content nodes.

(Neos UI 4.0.1, Neos 4.3.7)

![image](https://user-images.githubusercontent.com/13046100/69666947-650ceb00-108d-11ea-8f03-5db850ff8152.png)

![image](https://user-images.githubusercontent.com/13046100/69666962-6f2ee980-108d-11ea-81a8-409af1b68e6c.png)

kdambekalns:
Ok, I'll give that a go…

bwaidelich:
@kdambekalns This is affected by https://github.com/neos/neos-development-collection/issues/2653 as well (which we plan to include into 5.1). Would be great if we could join forces here!

kdambekalns:
[...]

PR #2545, opened 2019-08-23
FEATURE: Custom Route for node preview

bwaidelich:
Depends on https://github.com/neos/neos-development-collection/pull/2654

Related: https://github.com/neos/neos-development-collection/issues/2653

PR #2597, opened 2019-11-29
TASK: Append showInvisible only if needed

kdambekalns:
This appends the `showInvisible` query parameter only of the node to be
previewed is actually hidden.

See #2500


kdambekalns:
With #2545 this stuff should be removed completely again. But since that will make it into 5.1 (or later) only, this needs to be fixed anyway.

kdambekalns:
See #2598 for a better "solution"!?

dimaip:
@kdambekalns so this can be closed, now that #2598 is merged?

PR #2602, opened 2019-12-02
FEATURE: Prevent page reload when editing fixed childnodes

Sebobo:
**What I did**

Instead of telling the UI to force a reload on the whole document when a node is changed that is not child of a `ContentCollection` now the server looks for a node that can be safely reloaded in the list of parents of the node.

**How I did it**

Traverse the parents and fusion path of the node until we find a good spot to reload.

**How to verify it**

Add a node like the following to the demo site and edit the TextWithImage node.
Without this change the document reloads. With the change only the `FixedChildren` node.

```
'My.Site:Content.FixedChildren':
  superTypes:
    'Neos.Neos:Content': true
  ui:
    label: 'Node with fixed children'
    icon: columns
  childNodes:
    text:
      type: 'Neos.Demo:Content.Text'
    textWithImage:
      type: 'Neos.Demo:Content.TextWithImage'
```

```
prototype(My.Site:Content.FixedChildren) < prototype(Neos.Neos:ContentComponent) {

    text = ${q(node).children('text').get(0)}
    textWithImage = ${q(node).children('textWithImage').get(0)}

    renderer = afx`
        <div>
            <header>
                Der Header:
                <Neos.Demo:Content.Text @context.node={props.text} />
            </header>
            <section>
                Ein Bild mit Text:
                <Neos.Demo:Content.TextWithImage @context.node={props.textWithImage} />
            </section>
        </div>
    `
}
```

### DIFF
--- a/Classes/Domain/Model/Changes/Property.php
+++ b/Classes/Domain/Model/Changes/Property.php
@@ -239,7 +239,9 @@ class Property extends AbstractChange
                     // which would allows us to reload its children. Then we request a reload on the child that is
                     // a parent of our modified node.
                     $closestCollectionChildNode = $node;
-                    while (!($closestCollectionChildNode->getParent()->getNodeType()->isOfType('Neos.Neos:ContentCollection') || $closestCollectionChildNode->getParent()->getNodeType()->isOfType('Neos.Neos:Document'))) {
+                    while ($closestCollectionChildNode->getParent()
+                        && !($closestCollectionChildNode->getParent()->getNodeType()->isOfType('Neos.Neos:ContentCollection')
+                            || $closestCollectionChildNode->getParent()->getNodeType()->isOfType('Neos.Neos:Document'))) {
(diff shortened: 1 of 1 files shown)


Sebobo:
Checking now what's up with the e2e tests.

dimaip:
@Sebobo check this out, seems like a legit exception https://2820-47460181-gh.circle-artifacts.com/0/home/circleci/app/Data/Logs/Exceptions/20191202102423c3e63d.txt

That's why I like our E2E setup, I can almost blindly press merge and hope that the tests pickup the regressions!

Sebobo:
@dimaip ah thx. Yes absolutely! 
The bug was quick to fix but it took me some time again to have the tests run locally to verify my fix.

dimaip:
@markusguenther btw we've discussed this already a few times, including with @kdambekalns recently, I personally prefer to squash-merge such PRs where the commit history does not make much sense for the future. It's much easier to do git-bisect and in general overview the history of the project. How do you feel about this?

PR #2604, opened 2019-12-02
FEATURE: Custom Route for node preview

bwaidelich:
Depends on neos/neos-development-collection#2654

Related: neos/neos-development-collection#2653


markusguenther:
Bildschirmfoto 2019-12-02 um 19 18 12
See URL bottom left.

PR #2605, opened 2019-12-02
FEATURE: Save searchbar state in persitedState

markusguenther:
**What I did**
Replaced the local state of the component with a redux persited state.
So that the state of the searchbar is stored in the local storage like the left sidebar or right sidebar.
So the users don`t need to open it over and over again.

**How to verify it**
Open or close the search bar in the left sidebar and reload. The state should be the same.

Bildschirmfoto 2019-12-02 um 18 25 01

Resolves: #2606

### DIFF
--- a/Configuration/Settings.yaml
+++ b/Configuration/Settings.yaml
@@ -146,6 +146,8 @@ Neos:
             isHidden: false
             contentTree:
               isHidden: true
+            searchBar:
+              searchToggled: true
           rightSideBar:
             isHidden: false
           addNodeModal:
--- a/Configuration/Settings.yaml
+++ b/Configuration/Settings.yaml
@@ -63,6 +63,7 @@ Neos:
           'UI.FullScreen.toggle': 't f'
           'UI.LeftSideBar.toggle': 't l'
           'UI.LeftSideBar.toggleContentTree': 't c'
+          'UI.LeftSideBar.toggleSearchBar': 't s'
 
           'UI.AddNodeModal.close': 'c m'
 
--- a/Configuration/Settings.yaml
+++ b/Configuration/Settings.yaml
@@ -147,7 +147,7 @@ Neos:
             contentTree:
               isHidden: true
             searchBar:
-              searchToggled: true
+              searchToggled: false
           rightSideBar:
             isHidden: false
           addNodeModal:
--- a/Configuration/Settings.yaml
+++ b/Configuration/Settings.yaml
@@ -148,7 +148,7 @@ Neos:
             contentTree:
               isHidden: true
             searchBar:
-              searchToggled: false
+              isVisible: false
           rightSideBar:
             isHidden: false
           addNodeModal:
--- a/Tests/IntegrationTests/Fixtures/1Dimension/treeSearch.e2e.js
+++ b/Tests/IntegrationTests/Fixtures/1Dimension/treeSearch.e2e.js
@@ -84,3 +84,14 @@ test('Pagetree search field state is stored', async t => {
         .expect(nodeTreeSearchInput.exists)
         .ok();
 });
+
+test('Pagetree search field toggles on hotkey', async t => {
+    subSection('Search is initially hidden and we open it with "t s"');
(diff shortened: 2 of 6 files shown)


packages/neos-ui/src/Containers/LeftSideBar/NodeTreeSearchBar/index.js
```diff
@@ -72,19 +79,18 @@ class NodeTreeSearchBar extends PureComponent {
...
-        const {i18nRegistry} = this.props;
-        const {searchToggled, searchValue, searchFocused, filterNodeType} = this.state;
+        const {i18nRegistry, isHiddenSearchBar} = this.props;
+        const {searchValue, searchFocused, filterNodeType} = this.state;
         const searchLabel = i18nRegistry.translate('search', 'Search', {}, 'Neos.Neos', 'Main');
 
         const searchToggleClassName = mergeClassNames({
             [style.searchToggleButton]: true,
-            [style['searchToggleButton--active']]: searchToggled
+            [style['searchToggleButton--active']]: !isHiddenSearchBar
```
dimaip: Why did you reverse it? I like the old name better :)

markusguenther: Can also name it like this.

Configuration/Settings.yaml
```diff
@@ -146,6 +146,8 @@ Neos:
             isHidden: false
             contentTree:
               isHidden: true
+            searchBar:
+              searchToggled: true
```
dimaip: Wait, but shouldn't it be hidden by default?

Tests/IntegrationTests/Fixtures/1Dimension/treeSearch.e2e.js
```diff
@@ -16,12 +16,10 @@ test('PageTree search and filter', async t => {
     const searchmePage = 'Searchme page';
     const searchmeShortcut = 'Searchme shortcut';
 
-    const nodeTreeSearchToggler = ReactSelector('NodeTreeSearchBar IconButton');
```
dimaip: The old behaviour was actually correct. But to test that the toggles state sticks in the local storage we can actually reload amid the test run and assert that the collapsed state stayed

markusguenther: OK changed that :)

markusguenther:
Thanks @dimaip for the feedback.

packages/neos-ui-redux-store/src/UI/LeftSideBar/index.ts
```diff
@@ -8,13 +8,19 @@ export interface State extends Readonly<{
...
     };
 }> {}
 
 export const defaultState: State = {
     isHidden: false,
     contentTree: {
         isHidden: false
+    },
+    searchBar: {
+        searchToggled: false
```
Sebobo: Wouldn't `isHidden` fit better with the other state variables and be more explicit.
From `searchToggled` I wouldn't easily know which state means what.

Sebobo:
It's cool we have this feature soon. I wished for it some days ago ❤️

packages/neos-ui-redux-store/src/UI/LeftSideBar/index.ts
markusguenther: Had this before ;)

dimaip: The thing is that it's visible by default. Maybe 'searchShown' or 'searchVisible'?

dimaip: I dunno, maybe 'isHidden' would also work

Sebobo: `isVisible` is also fine.

markusguenther:
And again renamed the property 🎉

Issue #2603, opened 2019-12-02
Search term in document tree search should be trimmed


lorenzulrich:
### Description

If you accidentially (e.g. through copy & paste) have a space after or before a search term, you won't have a result.

### Steps to Reproduce

1. Open the search field
2. Enter the name of an existing page
3. The page is found
4. Add a space to the search term
5. The page is not found anymore

#### Expected behavior

It is quite usual to trim white space before and after search terms.

#### Actual behavior

The white space is not trimmed.

### Affected Versions

Neos: 4.3+
UI: 4.0.1

markusguenther:
That was luckily an easy fix. Thanks for reporting @lorenzulrich ❤️

lorenzulrich:
Thanks for taking care!

markusguenther:
We should first close it when we merged the change. Github will take care for that ;)
Today I was thinking to may be test it also on the 2.x branch. I bet that is also an issue for neos 3.3.

PR #2608, opened 2019-12-04
BUGFIX: Trim search value for document search

markusguenther:
Fixes: #2603


markusguenther:
I trimmed it not in state because then you are not able to type searches with whitespaces without going backward after a letter.

markusguenther:
Changed the base to 2.x ... the issue also occurs in neos 3.3

PR #2609, opened 2019-12-11
FEATURE: Allow inserting nodes into any position inside an element

Sebobo:
Before this change a new node was always inserted as direct child.
With this change there can be any number of dom nodes and one
can be defnied as anchor for insertion.
When multiple anchors exist, the first one will be used.

The primary use case is for me was to have a ContentCollection that is also a content element itself. Usually the content collection has its own wrapper and all the children are in there.
But when it's also a content with some custom markup it can have any structure and therefore we cannot just assume inserting the new node into the content nodes main tag is the right position.

**What I did**

Allow defining a subnode in the template of a content collection to be the parent of new nodes.

**How I did it**

A tag can have a new attribute which marks it as an anchor.
When a new node is inserted a check is made whether there is an anchortag inside the new nodes parent and if yes then this tag will be used to append the new child.

**How to verify it**

See included e2e test

### DIFF
--- a/Tests/IntegrationTests/Fixtures/1Dimension/createNewNodeInContainer.e2e.js
+++ b/Tests/IntegrationTests/Fixtures/1Dimension/createNewNodeInContainer.e2e.js
@@ -1,0 +1,35 @@
+import {Selector} from 'testcafe';
+import {ReactSelector} from 'testcafe-react-selectors';
+import {beforeEach, subSection, checkPropTypes} from './../../utils.js';
+import {Page} from './../../pageModel';
+
+/* global fixture:true */
+
+fixture`Create new nodes`
+    .beforeEach(beforeEach)
+    .afterEach(() => checkPropTypes());
+
+test('Create a text node in a new container element at the correct position', async t => {
+    await t.switchToMainWindow();
+
+    subSection('Create content collection node');
+    await t
+        .click(Selector('#neos-ContentTree-ToggleContentTree'))
+        .click(Page.treeNode.withText('Content Collection (main)'))
+        .click(Selector('#neos-ContentTree-AddNode'))
+        .click(ReactSelector('NodeTypeItem').find('button>span>span').withText('Container'));
+    await Page.waitForIframeLoading(t);
+
+    subSection('Create text node in container');
+    await t
+        .click(Page.treeNode.withText('Container'))
+        .click(Selector('#neos-ContentTree-AddNode'))
+        .click(Selector('#into'))
+        .click(ReactSelector('NodeTypeItem').find('button>span>span').withText('Text'));
+    await Page.waitForIframeLoading(t);
+
+    await t.switchToIframe('[name="neos-content-main"]');
+
+    const textIsInWrap = Selector('.test-container .test-text').parent().hasClass('test-container__inner-wrap');
+    await t.expect(textIsInWrap).ok();
+});
--- a/packages/neos-ui/src/manifest.js
+++ b/packages/neos-ui/src/manifest.js
@@ -11,6 +11,7 @@ import {SynchronousRegistry, SynchronousMetaRegistry} from '@neos-project/neos-u
 import {
     getGuestFrameDocument,
     findNodeInGuestFrame,
+    closestNodeInGuestFrame,
     findAllOccurrencesOfNodeInGuestFrame,
     createEmptyContentCollectionPlaceholderIfMissing,
     findAllChildNodes,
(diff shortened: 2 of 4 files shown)


Sebobo:
@markusguenther with your commit the parent is now determined every time instead of just when doing the „into“ insertion. 
The „before“ and „after“ don‘t really need it.

Sebobo:
We could make the search for the insertion point a bit easier if the data attribute holds the parents node context too. Then we need a bit more elaborate helper in Fusion to add the necessary data attribute to a tag.

markusguenther:
@Sebobo I am sorry ... just had the intention to fix the linting issue.

Sebobo:
@dimaip can you give me a hint on how to write a good test for this change?
I'm not sure right now where to start.

dimaip:
Hmm an intersting feature. Would be great if more people from the team had a look at this approach.

As for testing, it's quite easy:
1. create some nodetype + fusion to test this feature here: https://github.com/neos/neos-ui/tree/master/Tests/IntegrationTests/SharedNodeTypesPackage/Configuration
2. then try to imitate all the steps you'd normally take to test this feature, create a node of the nodetype, create a new node inside it, assert that it appeared in the right place in the DOM. Similar to this: https://github.com/neos/neos-ui/blob/master/Tests/IntegrationTests/Fixtures/1Dimension/createNewNodes.e2e.js

packages/neos-ui/src/manifest.js
```diff
@@ -334,6 +335,19 @@ manifest('main', {}, globalRegistry => {
         }
 
         const fusionPath = contentElement.dataset.__neosFusionPath;
+        // Check if an insertion anchor is defined and use this one for appending the childnode
+        const findInsertionParentByAnchor = () => {
+            let insertionParent = parentElement;
+            const insertionAnchors = parentElement.querySelectorAll('[data-__neos-insertion-anchor]');
```
dimaip: I'm bad at thinking out names, but maybe somebody has an idea for something better instead of `inserition-anchor`? Or do you all like it?
Since it's a public API that we'd have to maintain forever we really need to think twice. Maybe also think of some additional requirements that could come in handy here.

Sebobo: Yes absolutely. That's why the PR is still a draft.

The current code is a proof of concept what was required to make it work.
I played with a few names and then decided to rather get outside input instead of breaking my head.

Tests/IntegrationTests/Fixtures/1Dimension/createNewNodeInContainer.e2e.js
```diff
@@ -0,0 +1,38 @@
+import {Selector, RequestLogger} from 'testcafe';
+import {ReactSelector} from 'testcafe-react-selectors';
+import {beforeEach, subSection, checkPropTypes} from './../../utils.js';
+import {Page} from './../../pageModel';
+
+/* global fixture:true */
+
+const changeRequestLogger = RequestLogger(request => request.url.endsWith('/neos/ui-services/change') && request.method === 'post' && request.isAjax);
```
dimaip: Why are you adding a request logger if you are not using it anywhere?

Sebobo: Copy & paste, I'll remove it.
Btw. thanks to your hint the test was done quite quickly.

skurfuerst:
Hey everyone,

Awesome Change 👍 would you please update docs.neos.io when this is merged?

... BTW - should we maybe provide a fusion object and viewhelper in Neos.Neos for generating this data attribute? IMHO this would be good, as it makes the API way more explicit and we could change it later on, as it makes the data attribute an implementation detail.

All the best,
Sebastian

dimaip:
@skurfuerst totally agree

Sebobo:
@skurfuerst @dimaip yes sure. Had already a discussion with @mficzel last Friday regarding the Fusion part.

kdambekalns:
Great to see that!

But… this is one of those features that are almost a bugfix. I have that issue in a Neos 4.3 project, and that way it will never be fixed there. Looking at the code, this is backwards-compatible and poses next to no risk, IMHO.

kdambekalns:
Also, there is zero documentation in this PR… 😬 Let me try:

---

To sepcify a point for inserting children of a contentcollectiomn node in your markup, you can use an attribute on the container element: `data-__neos-insertion-anchor`. If found, that container will be used for inserting new items, otherwise the parent element (as usual).

Sebobo:
As I mentioned on discuss see also the two mentioned PRs #2667 and #2891 in Neos.Neos. Those make the usage much better for most use cases .

kdambekalns:
Original PR description comment moved here

---


**What I did**

Allow defining a subnode in the template of a content collection to be the parent of new nodes.

**How I did it**

A tag can have a new attribute which marks it as an anchor.
When a new node is inserted a check is made whether there is an anchortag inside the new nodes parent and if yes then this tag will be used to append the new child.

**How to verify it**

See included e2e test

mhsdesign:
Just stumbled onto this, i didnt know that `data-__neos-insertion-anchor` is api - there should be a fusion object for that ...

> ... BTW - should we maybe provide a fusion object and viewhelper in Neos.Neos for generating this data attribute? IMHO this would be good, as it makes the API way more explicit and we could change it later on, as it makes the data attribute an implementation detail.

PR #2578, opened 2019-11-17
Removing dependency Ramda

johachi:
**What I did**
Removed the dependency `ramda` and so we only use `lamda` as PR #2522 

**How to verify it**
Please read the file changes.

**Other**
This will be my first contribution to this project. Please feel free to give me any feedback, and let me know if I have made any mistakes.

This will close #2522

PR #2611, opened 2019-12-12
BUGFIX: Reolve mistake while upmerging

markusguenther:
We have removed ramda as dependency and while doing
upmerges the changes collides somehow.

The commit:
https://github.com/neos/neos-ui/commit/c138b961444e9d80ba2afb56b840f0d50eb4f33d#diff-a7ae28cc7ab235343fc04fe7203f0b82R101

and the PR for removing ramda lead to that issue.
https://github.com/neos/neos-ui/pull/2578/files#

I compared the changes and fixed that somehow merge conflict.


markusguenther:
Ok checked the tests locally and it works... the saucelabs tests show that we still have the create dialog issue there. This is already fixed in neos.neos.

https://app.saucelabs.com/tests/c453954f946a4311a2795c2c8392c2f7#1

Issue #2586, opened 2019-11-22
AssetEditor hides preview if "upload" feature is disabled


bwaidelich:
### Description

Slightly related to #2585 the `AssetEditor` does not really work with `features.upload = false`.
Reason: The `AssetUpload` component is hidden in this case, but that also renders the SelectBox with the asset preview(s):

![image](https://user-images.githubusercontent.com/307571/69419672-37d2dc80-0d1d-11ea-8e2d-ee1682414ac5.png)

Without that, there is no way to tell what file has been selected (or if any at all) and existing files can't be removed.

markusguenther:
Thanks for reporting that ❤️

markusguenther:
Tried it and it is easy to reproduce.
<img width="917" alt="Bildschirmfoto 2019-12-13 um 12 17 26" src="https://user-images.githubusercontent.com/1014126/70796753-56b81200-1da3-11ea-9f2f-695d9d2972c1.png">
<img width="916" alt="Bildschirmfoto 2019-12-13 um 12 17 56" src="https://user-images.githubusercontent.com/1014126/70796754-5750a880-1da3-11ea-85f4-171e51a540f9.png">

markusguenther:
Will be fixed in the upcoming 4.0 release.

skurfuerst:
@markusguenther is this fixed already? If so, do you have a commit link for me? :)

markusguenther:
> @markusguenther is this fixed already? If so, do you have a commit link for me? :)

This should be the PR. Tried it that time and it worked for me :)

markusguenther:
<img width="946" alt="Screenshot 2021-10-12 at 23 33 32" src="https://user-images.githubusercontent.com/1014126/137032163-800de467-ea90-4ae9-8f60-a555b18de922.png">

Is fixed thanks to @kitsunet

PR #2612, opened 2019-12-12
BUGFIX: AssetEditor selects are rendered without upload enabled

kitsunet:
Fixes: #2586


packages/neos-ui-editors/src/Editors/AssetEditor/index.js
```diff
@@ -192,85 +192,92 @@ export default class AssetEditor extends PureComponent {
         );
     }
 
-    renderAssetUpload() {
-        if (!this.isFeatureEnabled('upload')) {
```
kitsunet: This is the root of the problem, so no upload, editor is basically empty. You cannot remove an asset ever again.

PR #2613, opened 2019-12-12
BUGFIX: AssetEditor works correctly with asset proxies

kitsunet:


kitsunet:
This was broken  from the beginning then fixed ot be even more borken. This version seems to work for me, the conundrum with detail endpoints needs to be figured out but it works like this.

PR #2614, opened 2019-12-12
BUGFIX: Focus node in tree after switching dimensions

nlx-lars:
Previously, the focused node was not updated after switching,
leading to unexpected behaviour when hiding/deleting nodes from the
NodeTreeToolbar.

**What I did**
Focus new target node after switching dimensions from the DimensionSwitcher.

**How to verify it**
1. Select page in tree
2. Switch content dimensions
3. Hide or delete node from the toolbar

Old behaviour:
The node from the source dimension would be hidden/deleted.

New behaviour:
The node in the new dimension is hidden/deleted.

### DIFF
--- a/packages/neos-ui-sagas/src/CR/ContentDimensions/index.js
+++ b/packages/neos-ui-sagas/src/CR/ContentDimensions/index.js
@@ -35,8 +35,9 @@ export function * watchSelectPreset() {
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


nlx-lars:
@dimaip i'm not sure which branch to target this against. is 2.x still maintained or should this go into 4.x?

dimaip:
I think until Neos 5.1 we should still target 2.x.

Issue #2618, opened 2019-12-14
Content-tree icon does not toggle state anymore


markusguenther:
### Description

When you close and open the content tree in version 5.1 the icon does not toggle anymore.

### Steps to Reproduce

1. Open 5.1 backend and open content-tree
2. close content tree

#### Expected behavior

The icon of the toggle button changes

#### Actual behavior

Icon is always the same.

<img width="330" alt="Bildschirmfoto 2019-12-14 um 18 41 04" src="https://user-images.githubusercontent.com/1014126/70852403-556d0f00-1ea1-11ea-8d45-3e989acaa7fb.png">

<img width="323" alt="Bildschirmfoto 2019-12-14 um 18 41 10" src="https://user-images.githubusercontent.com/1014126/70852402-54d47880-1ea1-11ea-97c6-35479155f669.png">

### Affected Versions

Neos: 5.1

UI:5.1.0

You can get this information by running `composer show` or using the package management module within Neos.
The UI version can also be found inside the left sidebar after version 1.0.3.

skurfuerst:
fixed in 5.1.1

PR #2619, opened 2019-12-14
BUGFIX: Read the is hidden flag for content tree

markusguenther:
Fixes: #2618


dimaip:
Nice catch!

skurfuerst:
fixed in 5.1.1

PR #2624, opened 2019-12-25
!!! TASK: major Christmas upgrade: react, redux, ckeditor & co

dimaip:
Ckeditor upgrade is technically breaking, but since we don't declare it as a public API it's fine to go in a minor version. Besides there aren't many breaking changes.

Also I had to switch to Neos 5.1 for E2E tests as relying on master is just too unstable.


markusguenther:
Thanks Dmitri. By reading it looks pretty good. Need to test it locally.

PR #2636, opened 2020-01-06
TASK: Upgrade typesafe-actions to version 5.1

markusguenther:
Normally this upgrade has breaking changes but we don't use this functions.
It is by the way a bit confusion that we ```import {action as createAction, ActionType} from 'typesafe-actions';``` because createAction is also part of the general API of `typesafe-actions`.

It is not a problem, more a naming thing because when you want to use the createAction you need to rename the import.

### DIFF
--- a/package.json
+++ b/package.json
@@ -179,7 +179,7 @@
     "regenerator": "^0.8.46",
     "reselect": "^3.0.1",
     "terser-webpack-plugin": "^2.3.1",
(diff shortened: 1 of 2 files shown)


dimaip:
I think we use them like this because we used to have a custom function `createAction` and when integrating TS it was the quickest way to integrate it.

PR #2638, opened 2020-01-07
TASK: Removes npm-cli-login

markusguenther:
Removes outdated and unused npm-cli-login


markusguenther:
@dimaip would be good if you can double check that. I guess we dont need it and the cli tools can use npm itself. Maybe a legacy relict.

markusguenther:
@dimaip we could replace it with ```npx npm-cli-login``` at this point or not?

dimaip:
Yes, probably.

Sebobo:
Updated it in #2679 to solve the release errors we currently have

markusguenther:
@Sebobo Why do you mention the Jenkins. Checked the config and the command is not used in general there. Just via our provided shell scripts. Or did I miss something?

Sebobo:
Maybe I meant that. Was 4 months ago...

markusguenther:
Close this now, as we need to change that caused by the 2FA anyway

markusguenther:
we removed that now in commit https://github.com/neos/neos-ui/commit/eef47db5a220cc008a16184afb502773bf0292e5

As the tool did not work anymore and we just use the NPM token now and lerna publish

Issue #2642, opened 2020-01-08
AssetEditor in single mode broken


kitsunet:
The AssetEditor in single mode (eg. property with type `Neos\Media\Domain\Model\Asset` is broken as uploads do not work at all. The reason is the missing object type check as is done in the "multiple" code path. Incidentally I think my PR for asset proxies broke it... I will ty to fix it but should go in and be released ASAP.

PR #2643, opened 2020-01-08
BUGFIX: Fix single AssetEditor uploads #2642

kitsunet:
The AssetEditor changes to fix functionality for asset proxies sadly contained some code erorrs in the path for single usage editors. This breaks uploads with type errors. The change fixes that by adapting to similar code as is used in the multiple code path.

Fixes: #2642


kitsunet:
Agreed :)

Issue #2644, opened 2020-01-08
Login screen keeps flashing in guest frame


gerhard-boden:
When authentication authentication has failed (or the session timed out) the login page is flashing in an endless loop behind the login overlay.

![flickering_login](https://user-images.githubusercontent.com/10533739/72000900-978d3480-3244-11ea-8ccc-64c5bd70bae0.gif)

Sebobo:
Is this still the case for you? If yes, which version?

For me there is no flashing in the background.

Sebobo:
And duplicate of #2590

Issue #2590, opened 2019-11-27
Infinite reload while relogin dialog is shown


markusguenther:
Was reported by @creative-resort in the neos distribution.

### Description

The re-login form is re-loaded in an endless loop

### Steps to Reproduce

1. Log in to a demo site and click on a page, e.g. "Features" in In-Place Edit Mode
2. Let the session's login duration expire.
3. In the navigation tree, click on a different page (e.g. "Shortcuts"), so a re-login is initiated.

#### Expected behavior

Normal re-login screen.

#### Actual behavior

Here is a video of how the re-login form keeps being reloaded in a loop:

![neos login error](https://user-images.githubusercontent.com/12367437/69498750-aa91b280-0eeb-11ea-989b-5281452b35d1.gif)

As the form is constantly re-loaded, logging in is impossible.
Only after canceling the page loading via browser button "cancel", the reloading stops.

### Affected Versions

Neos: 5.0.4
Flow: 6.0.5

PHP 7.3

Browser: latest Google Chrome Browser

markusguenther:
Thought we had something like that in the past, but did not find the issue yet. So need to investigate that later. As we have feature freeze on 1st of december will wait with bugfixes until the freeze ;)

Sebobo:
Is this still happening?

I cannot reproduce it currently.

grebaldi:
As per this description:
> The re-login form is re-loaded in an endless loop

This is not happening anymore.

However, there still is a slight problem: The loading indicator runs on a loop when the Login Dialog appears. Similar behavior has been described elsewhere:

- https://github.com/neos/neos-ui/issues/1718
- https://github.com/neos/neos-ui/issues/1664

I think it's best to have a separate issue to address the loading indicator behavior.

I'll close this one.

durimjusaj:
This issue still occurs in Neos 9.1.5.

PR #2645, opened 2020-01-08
BUGFIX: Do not show login page in guest frame

gerhard-boden:
We should now show (the ugly flashing) login page when an authentication error occurs, since we already handle the login within our login popup overlay.

Keeping the current page open and retrying after successful login is the more logical approach.

resolves: https://github.com/neos/neos-ui/issues/2644

### DIFF
--- a/packages/neos-ui-backend-connector/src/FetchWithErrorHandling/index.ts
+++ b/packages/neos-ui-backend-connector/src/FetchWithErrorHandling/index.ts
@@ -95,7 +95,10 @@ class FetchWithErrorHandling {
                 if (response.ok) {
                     // CASE: all good; no errors!
                     resolve(response);
-                } else if (response.status === 401) {
+                } else if (
+                    response.status === 401 ||
+                    response.headers.get('X-Authentication-Required')
+                ) {
                     // CASE: Unauthorized!
                     // - all following requests have to fail; thus we enqueue them.
                     // - we enqueue our current request; so that it is re-run after successful re-login.
--- a/packages/neos-ui-guest-frame/src/initializeGuestFrame.js
+++ b/packages/neos-ui-guest-frame/src/initializeGuestFrame.js
@@ -43,13 +43,14 @@ export default ({globalRegistry, store}) => function * initializeGuestFrame() {
     const inlineEditorRegistry = globalRegistry.get('inlineEditors');
     const guestFrameWindow = getGuestFrameWindow();
     const documentInformation = Object.assign({}, guestFrameWindow['@Neos.Neos.Ui:DocumentInformation']);
+    const authenticationTimeout = yield select(selectors.System.authenticationTimeout);
 
     // The user may have navigated by clicking an inline link - that's why we need to update the contentCanvas URL to be in sync with the shown content.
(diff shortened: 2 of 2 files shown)


gerhard-boden:
This is almost, finished, but I feel like I'm too late for the party here:
https://github.com/neos/neos-ui/pull/2645/files#diff-2fc0ad8a6fad158667c1dc4ebaa995feR46

I do not want to load anything within the guest frame until authentication is done.

packages/neos-ui-backend-connector/src/FetchWithErrorHandling/index.ts
```diff
@@ -95,7 +95,10 @@ class FetchWithErrorHandling {
                 if (response.ok) {
                     // CASE: all good; no errors!
                     resolve(response);
-                } else if (response.status === 401) {
+                } else if (
+                    response.status === 401 ||
+                    response.headers.get('X-Authentication-Required')
```
markusguenther: Are you sure that this is always with upper-camel case? Believe me to remember that some browsers also use lower case, but I am not 100% sure. And HTTP header names are case-insensitive, according to RFC 2616.

gerhard-boden: thanks for the hint! 
Foll disclosure: I'm not even 100% sure if this part is needed. This came in with a new Neos bugfix for Chrome 79 (see: https://github.com/neos/neos-development-collection/issues/2845)

The newest fix works also without this adjustment (the login prompt shows up), but just wanted to make sure to break nothing I do not understand / see. Maybe someone else has some insights.

gerhard-boden:
> By reading it looks good. Did not test it yet!

This is not fully finished yet, see https://github.com/neos/neos-ui/pull/2645#issuecomment-572174204

Added new "work in progress" label

markusguenther:
@gerhard-boden Hi anything new regarding this? How could we help you?

gerhard-boden:
> @gerhard-boden Hi anything new regarding this? How could we help you?

Thanks for checking in @markusguenther, haven't touched anything since January here, if I look up the comment I made it looks like I had trouble reading the `authenticationTimeout` flag so we would stop trying to initialize the guest frame (which causes the flickering issue)

kdambekalns:
Ping?

crydotsnake:
Ping

packages/neos-ui-backend-connector/src/FetchWithErrorHandling/index.ts
kdambekalns: I'd keep that in place, and the case-sensitivity is no problem - otherwise that bugfix would not havwe worked ;)

kdambekalns (approved):
Cannot say much about the change and currently lack the time to test. But I'd trust you with this…

gerhard-boden:
Sorry, guys, for being unresponsive and digging this up! It's been 4 years, and I THINK the problem has been fixed elsewhere in the meantime  https://github.com/neos/neos-ui/issues/2590#issuecomment-1531490615

Gonna close this for now so nobody wastes any more time on this 😊.

mhsdesign:
Haha thanks @gerhard-boden ^^ i stumbled a few times over this pr and tried to make sense of the code ^^ glad to see it can now rest in peace :D

PR #2649, opened 2020-01-15
FEATURE: publish transpiled (ES-Modules) version of react-ui-components

skurfuerst:
To make react-ui-components usable standalone, we
should publish a transpiled-to-es-modules version
of the code, in the same manner as many other projects
publish their library code.

This way, the react-ui-components are embeddable
e.g. using parcel.js, or via webpack; without setting
up typescript compilation of node_modules (which is
discouraged).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,8 +77,15 @@ setup: check-requirements install build
 build-subpackages:
 	$(lerna) run build --concurrency 1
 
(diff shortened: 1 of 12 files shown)


skurfuerst:
**Note: we need to ensure that this does not have any negative side-effects on our normal build; I wasn't able to test this yet.**

Issue #2654, opened 2020-01-21
make build fails because of node version mismatch


Sebobo:
Lately a lot of dependencies are being updated but those updates ignored our defined node version in our .nvmrc.
Currently make install fails as some dependencies.

I assume this happens as we don't have a node version defined in our package.json.
Should we simply add it there, or is there a reason why we don't have it?

Sebobo:
This already happened before btw. At the time I adjusted the .nvmrc

dimaip:
I'm sorry, I don't understand the issue. Where does it fail? Locally? In CI?

Sebobo:
Locally. `make install` fails when you use `nvm` or have some other global node version that doesn't match.

PR #2664, opened 2020-01-27
FEATURE: Add new Icons from fontawesome

markusguenther:
Upgrading from version 5.8.0 to 5.12.0.
This adds lots of new icons.


Sebobo:
Cannot test it because of https://github.com/neos/neos-ui/issues/2654

Sebobo:
Do those icons btw. also work in the backend modules properly?

markusguenther:
@Sebobo In the react based paert you can use the new Icons with this PR and for the non react Modules I have the other PR (https://github.com/neos/neos-development-collection/pull/2881)

Would be nice to have only one source at some time but it is not that easy.

markusguenther:
@Sebobo Thanks for the hint. Already updated it :)

Issue #2671, opened 2020-02-12
Backend always sends two requests to the preview url


Sebobo:
### Description

Every page view in the backend send two requests to the preview url.

### Steps to Reproduce

Open the backend and look into the inspector network tab.
The first request is sent and immediately canceled and the second one is then sent successfully.

#### Expected behavior

Only one request is sent

#### Actual behavior

Two requests appear:

![Bildschirmfoto 2020-02-12 um 15 34 30](https://user-images.githubusercontent.com/596967/74346830-70c89d80-4db0-11ea-8ea3-bf040dcc58ed.png)

### Affected Versions

Neos: 4.3+
UI: 4.0+

PR #2673, opened 2020-02-13
BUGFIX: Prevent second page load

Sebobo:
Most of the time `win.location.href` is not yet set and therefore
`componentDidUpdate` and `componentDidMount` both trigger
a load of the new page in the iframe.

Resolves: #2671

### DIFF
--- a/packages/react-ui-components/src/Frame/frame.tsx
+++ b/packages/react-ui-components/src/Frame/frame.tsx
@@ -13,10 +13,12 @@ export interface FrameProps extends React.IframeHTMLAttributes<HTMLIFrameElement
 
 interface FrameState {
     readonly transitioning: boolean;
+    readonly location: string;
 }
 
 const initialState: FrameState = {
-    transitioning: false
+    transitioning: false,
+    location: '',
 };
 export default class Frame extends PureComponent<FrameProps> {
     // tslint:disable-next-line:readonly-keyword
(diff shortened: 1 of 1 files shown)


Sebobo:
Contains the fix from #2672 as I couldn't build without it.
So #2672 should be merged first :)

packages/react-ui-components/package.json
```diff
@@ -6,7 +6,7 @@
   "main": "./src/index",
   "scripts": {
     "build": "exit 0",
-    "build-standalone-esm": "rm -Rf .postcssrc lib-esm && tsc --build tsconfig.esm.json && tsc --build tsconfig.esmtypes.json && rsync --prune-empty-dirs  -r --include='*/' --include='*.css' --exclude='*' src/ lib-esm -v && node scripts/generate-standalone-postcssrc.js",
+    "build-standalone-esm": "rm -Rf .postcssrc lib-esm && tsc --build tsconfig.esm.json && tsc --build tsconfig.esmtypes.json && rsync --prune-empty-dirs  -r --include='*/' --include='*.css' --include='*.svg' --exclude='*' src/ lib-esm -v && node scripts/generate-standalone-postcssrc.js",
```
dimaip: This should be in a the other PR, right? I guess we can just merge that one and rebase.

Sebobo: Yes see my comment. This commit will just disappear then.

Sebobo:
@dimaip yes I first simply checked if the frame is already `transitioning` but I was scared that this is not reliable enough even though it seemed to work.

Issue #2421, opened 2019-03-26
Figure out where to put UI translations


dimaip:
Currently we use translations from Neos.Neos, but there's still some leftover unused translations lying around here in Neos.Neos.Ui.

What we can do:
  a) delete everything from Neos.Neos.Ui and leave everything in Neos.Neos (easy, just `rm -rf Resources/Private/Translations`)
  b) move all content module labels to Neos.Neos.Ui and setup crowdin integration for Neos.Neos.Ui (hard)

danielkestler:
b) might be harder, but be the better way in the end – because of the tight dependencies between UI code and the language labels, it makes sense to put it in one package.

But it also depends on the future of the Neos UI in general (the other modules, like the media management) ... will they also be separate packages for each module? Because even then, they will share some code (e.g. the main menu and all the labels), which could/should belong to the Neos.Neos package.

dimaip:
I agree.

> will they also be separate packages for each module? 

I guess the first candidate is the Neos.Media module, I don't think others will be re-written so soon.

kdambekalns:
I am currently evaluating Weblate as a replacement for CrowdIn, it looks really promising and would make adding new packages to the translation process way easier. Stay tuned!

kdambekalns:
As of today, Neos.Ui is available in Weblate for translations, branches 5.3 to master:

- https://hosted.weblate.org/projects/neos/neos-ui-main-master/
- https://hosted.weblate.org/projects/neos/neos-ui-main-7-2/
- https://hosted.weblate.org/projects/neos/neos-ui-main-7-1/
- https://hosted.weblate.org/projects/neos/neos-ui-main-7-0/
- https://hosted.weblate.org/projects/neos/neos-ui-main-5-3/

When translating, identical translations are propagated to other branches. Changes are pushed to the repository once every 24 hours.

So at least actually translating things for the UI is possible/easier now.

mhsdesign:
As of Neos 9.2 the following (few) translations still reside in Neos.Neos that are used by Neos.Ui

<details><summary>The Neos 9.1 state</summary>
<p>


```
Neos.Neos:Main:above
Neos.Neos:Main:activateFullscreen
Neos.Neos:Main:apply
Neos.Neos:Main:authenticating
Neos.Neos:Main:back
Neos.Neos:Main:below
Neos.Neos:Main:cancel
Neos.Neos:Main:choose
Neos.Neos:Main:close
[...]

PR #2674, opened 2020-02-24
BUGFIX: Localize publish actions

Sebobo:
**What I did**

Use translator for UI messages on workspace actions.

And use the workspace title instead of its technical name
which can be misleading when the title of a workspace
is changed, as the name is not changed in this case.

**How I did it**

Use translator and also support plural/singular forms.

**How to verify it**

Publish & discard changes in the UI

publish-text


Sebobo:
I had a problem that in German it doesn't properly show the translated version.
I'm not sure why the locale is not applied.

Maybe @kdambekalns or @kitsunet have an idea?

Sebobo:
Ah this issue might be related: https://github.com/neos/neos-ui/issues/2421

Sebobo:
Ah misinterpreted things. They stay in Neos.Ui. But I still have to check why the German translation doesn't show up.

Sebobo:
Now it works. The controller wasn't aware of the users locale and had to set it myself.

PR #2690, opened 2020-03-23
BUGFIX: Allow ts imports in plugins

Sebobo:
Because of the way the webpack configs are merged the extensions list defined in build-essentials was overriden and later replaced by the webpack default which only contains normal js extensions.


Sebobo:
@dimaip we currently have no deep merge there which is a bit of an issue. But I didn't want to change all that now for a bugfix that includes 4.x.
We need a new configuration that would also allow to merge stuff from the plugin itself to make the final web pack config really adjustable. Maybe something for 5.3

PR #2704, opened 2020-04-06
BUGFIX: Evaluate ClientEval also on load

### LINKED ISSUE #2703: The reference editor should evaluate ClientEval on load
If I use

```
startingPoint: 'ClientEval: node.properties._path'
```

in a reference editor, it uses the current node as the starting point for the search. That works fine, but when loading the document, the editor tries to initialize using the literal string ClientEval: node.properties._path as context node, which of course does not work.

jonnitto:
This fixes #2703


markusguenther:
Thanks for the contribution ❤️

Issue #2705, opened 2020-04-07
Views are no longer rendered in Inspector


bwaidelich:
### Description

With #2648 the behavior of the inspector was changed leading to Views no longer being displayed for nodes that are covered by a `EditNodePrivilege`.

### Steps to Reproduce

(taken from #2648)

1. as an example, in the Demo page, use the following `Policy.yaml`:

  ```yaml
  privilegeTargets:
    'Neos\ContentRepository\Security\Authorization\Privilege\Node\EditNodePrivilege':
      'Neos.Demo:EditPageNodes':
        matcher: 'nodeIsOfType("Neos.Demo:Document.Chapter")'
      'Neos.Demo:EditTextNodes':
        matcher: 'nodeIsOfType("Neos.Demo:Content.Text")'
  ```

  This policy forbids editing `Document.Chapter` nodes and `Text` nodes for *everybody*.

2. Log into the backend.
3. Select one of the following nodes:
   - In the page tree, select a chapter node
   - On a page, select a `Text` node through the Structure tree

#### Expected behavior

- The Inspector should not contain editable components, but Views (and potentially components in readonly mode)

![image](https://user-images.githubusercontent.com/307571/78673076-e7ca7000-78e1-11ea-8dfc-1d3783673181.png)

#### Actual behavior

- The inspector stays completely empty

![image](https://user-images.githubusercontent.com/307571/78673163-092b5c00-78e2-11ea-8a24-fa98ddbb3a46.png)

This is especially problematic for custom views that have been created for read-only documents/nodes

### Affected Versions

UI: 4.0.4+, 5.0.4+, 5.1.3+

PR #2706, opened 2020-04-07
BUGFIX: Don't disable views for read only nodes

bwaidelich:
This partly reverts #2648 that completely disabled
the inspector for read-only nodes only removing
editor-properties.

Fixes: #2705


skurfuerst:
looks great - good work :+1: feel free to merge

Issue #2696, opened 2020-03-31
Focus first input element / the title element, when a new document is created


daniellienert:
### Description

![image](https://user-images.githubusercontent.com/642226/78010366-9438af00-7342-11ea-9cb5-a62ebc4231dc.png)

When I add a new docuentNode, I have to click the title element manually, even if it is the only element. Which is unexpected and really anoying when creating a lot of pages.

Neos UI: 5.1.x / master

markusguenther:
As it was intended but just did not work, I created a BUGFIX PR against 4.0 and this should work then in all versions :)

I bet that we broke it with the TS adjustment.

daniellienert:
Wooohooo!! Thanks a lot!

PR #2708, opened 2020-04-13
BUGFIX: Focus first form element in node creation dialog

markusguenther:
**What I did**
The NodeCreationDialog always throws errors because the options are passed thru as array instead of an object. So I modified the data type here.

The already set autoFocus property did not work for the first element of the Creation Dialog.
So this patch also fixes the setting of the autofocus for the editor fields.

Another issue was that the dialog always gets the focus by default and therefore the first element was not focused.

**How to verify it**

Create a new page in the backend and the title field will be focused.

Fixes: #2696

### DIFF
--- a/packages/react-ui-components/src/Dialog/dialog.tsx
+++ b/packages/react-ui-components/src/Dialog/dialog.tsx
@@ -59,6 +59,11 @@ export interface DialogProps {
     readonly actions: ReadonlyArray<ReactNode>;
 
     /**
+     * This prop controls the focus state of the Dialog.
+     */
+    readonly autoFocus: boolean;
+
+    /**
      * An optional `className` to attach to the wrapper.
      */
     readonly className?: string;
@@ -129,8 +134,8 @@ export class DialogWithoutEscape extends PureComponent<DialogProps> {
 
     public readonly componentDidMount = (): void => {
         document.addEventListener('keydown', (event : KeyboardEvent) => this.handleKeyPress(event));
-
-        if (this.ref) {
+        const {autoFocus} = this.props;
+        if (this.ref && autoFocus) {
             this.ref.focus();
         }
     }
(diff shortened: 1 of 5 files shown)


packages/react-ui-components/src/Dialog/dialog.tsx
```diff
@@ -58,6 +58,11 @@ export interface DialogProps {
      */
     readonly actions: ReadonlyArray<ReactNode>;
 
+    /**
+     * This prop controls the focus state of the Dialog.
+     */
+    readonly autoFocus: boolean;
```
dimaip, line 64: How does this differ from `setFocus` prop that we already have?

dimaip: nevermind, was looking at the wrong component....

packages/react-ui-components/src/TextInput/textInput.tsx
```diff
@@ -52,11 +52,16 @@ export interface TextInputProps extends Omit<React.InputHTMLAttributes<HTMLInput
...
+
+    constructor(props: any) {
+        super(props);
+        this.ref = React.createRef();
+    }
 
     public readonly componentDidMount = (): void => {
-        if (this.ref && this.props.setFocus) {
-            this.ref.focus();
+        if (this.ref && this.ref.current && (this.props.setFocus || this.props.autoFocus)) {
```
dimaip, line 63: Where is `autoFocus` prop coming from? It's not defined in proptypes. How does it differ from `setFocus`?

markusguenther: The NodeCreationDialog sets the property since some years ;)

markusguenther: TextInput also has the prop for some time.

dimaip: Ah, you mean to say that it's is just passed down to the DOM element as a DOM attribute here? https://github.com/neos/neos-ui/blob/72caa0aa8a97ed79ab37a9c701136e8b6405f125/packages/react-ui-components/src/TextInput/textInput.tsx#L107
Didn't think about it at first, good point then.

PR #2711, opened 2020-04-23
BUGFIX: Avoid URLs to be rendered twice in link- and reference editors

bwaidelich:
This removes some ancient hack that tried to determine the "uriInLiveWorkspace".

Background:

The code extracted the context path that was previously specified via
`@<context>` in the "backend URLs". Due to the dedicated preview routes
that were introduced with Neos 5.1 this no longer worked and led to the
preview URL rendered twice instead..

Related: neos/neos-development-collection#2965

### DIFF
--- a/packages/neos-ui-backend-connector/src/Endpoints/index.ts
+++ b/packages/neos-ui-backend-connector/src/Endpoints/index.ts
@@ -437,14 +437,14 @@ export default (routes: Routes) => {
                 if (!nodeType) {
                     throw new Error('.node-type not found in result');
                 }
-                const uri = uriElement.innerText.trim();
                 return {
                     dataType: 'Neos.ContentRepository:Node',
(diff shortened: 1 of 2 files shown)


bwaidelich:
Without the fix:
image

With the fix:
image

With the fix and neos/neos-development-collection#2966 :

image

kitsunet:
I think this bugfix should've gone into the 4.0 branch, given that that is still a supported version 🤷

bwaidelich:
> I think this bugfix should've gone into the 4.0 branch, given that that is still a supported version?

But only 5.1+ is affected, isn't it?

kdambekalns:
> But only 5.1+ is affected, isn't it?

I think not, see https://github.com/neos/neos-development-collection/pull/2966#issuecomment-682469739

PR #2733, opened 2020-05-05
BUGFIX: Increase memory limit for node in jenkins build

Sebobo:
This should solve the out of memory errors with 5.2.0 production build https://jenkins.neos.io/job/neos-build-minified/18450/console


kdambekalns:
My guess: you need to `export` the variable…

PR #2735, opened 2020-05-27
BUGFIX: Improve usability for certain editors in creation dialog

gerhard-boden:
Certain editors like the reference editor need a certain amount of space to present results. We now add additional space to the bottom of the editor if it is placed at the end of the dialog.

before
after

### DIFF
--- a/packages/neos-ui/src/Containers/Modals/NodeCreationDialog/index.js
+++ b/packages/neos-ui/src/Containers/Modals/NodeCreationDialog/index.js
@@ -155,6 +155,16 @@ export default class NodeCreationDialog extends PureComponent {
         );
     }
 
+    editorNeedsAdditionalSpace(elementConfiguration) {
+        const editorPaths = [
+            'Neos.Neos/Inspector/Editors/ReferenceEditor',
+            'Neos.Neos/Inspector/Editors/ReferencesEditor',
+            'Neos.Neos/Inspector/Editors/SelectBoxEditor'
+        ];
+
+        return editorPaths.includes(elementConfiguration.ui.editor);
+    }
+
     render() {
         const {isOpen, configuration} = this.props;
 
@@ -163,6 +173,7 @@ export default class NodeCreationDialog extends PureComponent {
         }
 
         const {validationErrors, isDirty} = this.state;
+        const elements = Object.keys(configuration.elements).filter(elementName => Boolean(configuration.elements[elementName]));
 
         return (
             <Dialog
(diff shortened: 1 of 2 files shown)


packages/neos-ui/src/Containers/Modals/NodeCreationDialog/index.js
```diff
@@ -155,6 +155,16 @@ export default class NodeCreationDialog extends PureComponent {
         );
     }
 
+    editorNeedsAdditionalSpace(elementConfiguration) {
+        const editorPaths = [
```
Sebobo, line 159: Hum, not sure if this is a good idea with a static list of editors.
What if I have my own editor based on one of them or at least on the selectbox?

Isn't it possible to change the styling to check for our selectbox and if it's the last element add some margin?
Or improve the editors css to adapt to the situation?

Sebobo: Maybe #2735 can be fixed too, if we find a more general solution.

gerhard-boden: Hi, thanks for the feedback! It's been a while, but I think the problem was that the editor is not aware where it is rendered (or at what position). I'm not too happy with the solution either, but it's the most efficient way I got rid of this annoying editor problem. Or find a way how the select options can overflow the dialog, so nothing is cut off, but that's not a simple task either and brings its own set of issues with it 🙈

Sebobo:
I'm all for fixing this. But maybe with a different approach:

I saw that the dropdown doesn't set `aria-expanded` or any other hint that it's open.
If we had that then we could also give an open dropdown some additional space below more easily via css.
So adding that would improve accessibility and selecting the last editor + a child of it that is expanded allows us to add some margin at the bottom or a different way of making it nicer. 

Also saw that our selectboxes dropdown have two ul's wrapping each other. Maybe we should fix that at some point too ;)

bwaidelich:
@gerhard-boden This would be a great fix for the next version IMO. What do you think of @Sebobo's comments?

gerhard-boden:
It's definitely worth a try and a good idea! Sure beats my approach, because we could do without the hard coded editor paths!
Need to make time to try it out though, because I'm about to move into a freshly build apartment and the necessary coordination and moving hassle is eating up a lot of my spare time currently 🙈

Will work on this as soon as time allows it!

markusguenther:
BTW: We also had something like flyouts but it was reverted in the end https://github.com/neos/neos-ui/pull/2221

Sebobo:
Hi @gerhard-boden, wouldn't https://github.com/neos/neos-ui/pull/2799 fix this issue too?

markusguenther:
Thanks for all the effort you put into that topic @gerhard-boden 
I will close this PR in favour of https://github.com/neos/neos-ui/pull/2799

Hope that it is also in your  interest.

gerhard-boden:
Very good, @grebaldi delivered's approach is much more nuanced than my additional margin and therefore a better solution since it will work also work with custom editors

Issue #2710, opened 2020-04-21
Missing inline link editor options


ComiR:
In contrast to #2214 or #2535, I am now missing the normal link editor options in the inline link editor, especially `nodeTypes` to be able to link only to specific node types.

https://github.com/neos/neos-ui/commit/f8436887e147fdff7d9f5537e3bd529f379aa077 did move the the `linkingOptions` (i.e. `….inline.editorOptions.linking`) out of the other options, so now only this specific subset of options can be used here.

The LinkEditor (in the inspector) on the other hand doesn't make use of the `linkingOptions`. Right now I'm fine with that, but this also would probably have worked before.

PR #2736, opened 2020-05-27
BUGFIX: Allow additional options to inline link editor

gerhard-boden:
Make it possible to pass options to the inline link editor through configuration, so the inline link editor is fully configurable like the link editor within the inspector.

resolves: #2710

### DIFF
--- a/packages/neos-ui-editors/src/Library/LinkInput.js
+++ b/packages/neos-ui-editors/src/Library/LinkInput.js
@@ -81,16 +81,17 @@ export default class LinkInput extends PureComponent {
     };
 
     getDataLoaderOptions() {
-        const contextForNodeLinking = $get('options.startingPoint', this.props) ?
-            Object.assign({}, this.props.contextForNodeLinking, {
-                contextNode: this.props.options.startingPoint
-            }) :
+        const options = {...this.props.options, ...this.props.linkingOptions};
+
+        const contextForNodeLinking = $get('startingPoint', options) ?
+            {...this.props.contextForNodeLinking, contextNode: options.startingPoint} :
             this.props.contextForNodeLinking;
+
         return {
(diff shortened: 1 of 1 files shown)


gerhard-boden:
This is targeting 4.0 intentionally, since this was (unofficially?) possible before, but stopped working after https://github.com/neos/neos-ui/commit/f8436887e147fdff7d9f5537e3bd529f379aa077

Sebobo:
If this is a feature we cannot put it into 4 or 5.0 - 5.2. So either we agree its a bugfix or have it in 5.3

Sebobo:
I think we can simply have this as bugfix. I mean it apparently worked before and should work again.
The solution is also okay I think, but then we have to also extend the docs and the schema file.
Would it work exactly like it did before?

gerhard-boden:
I'm really not sure if it was ever documented before, but people obviously expected the inline link editor to be configurable like the link editor within the inspector. I don't think it would break anything. I mean if you have configuration in place that currently does nothing it would work after the update, but nobody would but configure something and expect the configuration to have no effect I guess 😂 

Where should I provide the documentation how to configure the inline link editor?

gerhard-boden:
I can also adjust my commit prefix after we made a final decision

Sebobo:
The docs are actually in Neos.Neos in the `PropertyReferenceEditor`. So would be a PR there as soon as we finished this.
Ah and I see the schema file doesn't contain the formatting and linking options yet.
Would maybe then be another PR to extend that from the Neos UI package then.

gerhard-boden:
Ah, that part of the documentation I did find, but I thought it was limited to the inspector editors, since I thought it was an "implementation secret" that we use the same React component within the CK Editor and the inspector. Since we have no other place I guess it's the most logical?

Issue #2450, opened 2019-04-09
Previewbutton href attribute is incorrect depending on the current state 


davidspiola:
### Description
The "preview" button does not work correclty right after creating a new page node. If you refresh the UI, the redirect in the href attribute is correct.

Related issue: #2173

#### After "create new page"
/neos/redirect?node=/sites/neosdemo/features/shortcuts/shortcut-to-child-node/node-v2q0rfysg5096@live

#### After UI reload
/neos/redirect?node=%2Fsites%2Fneosdemo%2Ffeatures%2Fshortcuts%2Fshortcut-to-child-node%2Fnode-v2q0rfysg5096%40live%3Blanguage%3Dde

##### Issues:
- url is not encoded
- dimension/language parameter is missing

### Steps to Reproduce

1. Create a new page node and check the href attribute of the preview button right after you hit the "publish" button
2. Refresh the the UI (F5) and check the link again 

#### Expected behavior

The link should be the same regardless of the current page state.

### Affected Versions

Neos: 4.2
UI: ^3.1.0

danielkestler:
This is probably my fault, I integrated this feature and thought I did it the same way as the default – noticed the missing url-encoding, but did not have any problems with it. Sorry. I think somebody with more expertise in this area should maybe have a look ...

DrillSergeant:
Not sure if I just experience a behaviour of the same cause:
When you translate a document, publish it and hit the preview-button, the document is not opend in the new dimension, but in the "source" dimension.
Shall I open another ticket for this or is this the same bug?

dlubitz:
@DrillSergeant Is this what you have experienced? 
https://github.com/neos/neos-ui/issues/2641

DrillSergeant:
Yes, sounds like the same behaviour

Sebobo:
I created a fix in #2744 but I didn't unterstand why the target dimension was modified in the old code.
From what I can see, we already have the correct url provided by the backend response.
I tested a while with the language dimension and my workspace and things seemed fine.
So it would be great if someone could test my fix.

PR #2744, opened 2020-07-03
BUGFIX: Use correct preview url after publishing

Sebobo:
The server response already supplies the correct
preview url for the backend incl. the dimensions
and target workspace.

Resolves: #2450


Sebobo:
Thx :)
But I will only merge it if someone else has tested it with some dimensions and workspaces :)

Issue #2742, opened 2020-06-18
Wrong dependencies to build-essentials and positional-array-sorter


kdambekalns:
### Description

Trying to build JS code for the UI inspector and I am getting options has an unknown property `localIdentName` when building. 

#### Actual behavior

Turns out that is coming from `@neos-project/build-essentials` in version 5.1.1 that I have installed. Even though… I have `@neos-project/neos-ui-extensibility` as version 5.2.1 installed.

Why does `@neos-project/neos-ui-extensibility` 5.2.1 depend on version 5.1.1 of both `@neos-project/build-essentials` and `@neos-project/positional-array-sorter`?

### Affected Versions

Neos: 5.2.1
UI: 5.2.1

Sebobo:
Seems the build scripts mixed this up.

kdambekalns:
My problem right now: If I require the correct version in my package, I simply get both versions, and the build still fails. Any way to work around that in userland code?

kdambekalns:
> Seems the build scripts mixed this up

Probably that, these lines have only been touched by Jenkins in the (recent) past.

Sebobo:
I started updating a plugin to 5.2 yesterday but haven't finished yet. I will probably run into the same problem and will look for a quick fix.

PR #2745, opened 2020-07-03
BUGFIX: update build-essentials to 5.2.1

andrehoffmann30:
Update @neos-project/build-essentials and @neos-project/positional-array-sorter to fix build problem with css-loader because the used css-loader version expects a different configuration syntax than in version 5.1.1 of @neos-project/build-essentials.

I upgrade an existing ui plugin and tried to use the newest version of the neos-ui-extensibility package. When i tried to build the plugin, the build failed with an error.
Bildschirmfoto 2020-07-03 um 13 21 40

After examining the version I saw, that the dev dependencies were incorrect.

### DIFF
--- a/packages/neos-ui-extensibility/package.json
+++ b/packages/neos-ui-extensibility/package.json
@@ -22,8 +22,8 @@
     "typescript": "3.2.2"
   },
   "dependencies": {
-    "@neos-project/build-essentials": "5.1.1",
-    "@neos-project/positional-array-sorter": "5.1.1",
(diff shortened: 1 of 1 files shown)


Sebobo:
The problem is that this is not something we should have to do but the Jenkins build is usually doing that. See #2742.
And you now set it to 5.2.1 but with the next bugfix release it should already be 5.2.2. 
Therefore I will check during the next release if it works properly.

andrehoffmann30:
Thank you for your response @Sebobo . Then I close this pull request, because it isn't need.

PR #2749, opened 2020-07-06
TASK: Upgrade dependencies for extensibility

markusguenther:
Upgraded the build-essentials and the positional-array-sorter to version 5.2.1


Sebobo:
Just denied the same in #2745 ;)
The dependency has to be changed by jenkins or it will not help us at all

markusguenther:
Ok your right will check that later.

PR #2751, opened 2020-07-07
BUGFIX: Manually adjust versions

markusguenther:
The versions of build-essentials and positional-array-sorter are wrong and does not fit the version of the current release. That leads to the problem that in the upcoming release version will be not replaced. The new release only replaces the current version number with the new released version. So the two packages will never get updated.

This is a result of a wrong up merge so that we had not the latest version.

Fixes: #2742

### DIFF
--- a/packages/neos-ui-extensibility/package.json
+++ b/packages/neos-ui-extensibility/package.json
@@ -22,8 +22,8 @@
     "typescript": "3.2.2"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


markusguenther:
We need to adjust it once manually and the next release will then again replace all packages of the extensibility. We don`t need to touch Jenkins here because the scripts working fine and just get stock because the mismatching version numbers.

Issue #2652, opened 2020-01-17
Document-Tree refresh crashes backend


eWert-Online:
### Description

When refreshing the document tree, the inspector starts loading for an infinite amount of time. When you afterwards try to click in an Inline-Editable field, the backend crashes.

### Steps to Reproduce

1. Refresh the Document Tree (second Button from the right)

#### Expected behavior

- The document tree refreshes
- The inspector doesn't start loading

#### Actual behavior

![document-tree-bug](https://user-images.githubusercontent.com/7459510/72614009-6fcb5a00-3931-11ea-9477-f334c91bff4c.gif)

1. The inspector starts loading
2. When clicking in an inline-editable field, the whole backend crashes

### Affected Versions

Neos: 5.1.3
UI: 5.1.2

Sebobo:
Can reproduce, will look into it.

Sebobo:
PR is ready to fix this.

PR #2756, opened 2020-07-14
BUGFIX: Prevent endless spinner in inspector after tree refresh

Sebobo:
The already loaded node data was overwritten
after the refresh and therefore lost all
stored data including the data required for
the inspector to show the node.

Resolves: #2652


Sebobo:
Yay teamwork <3

PR #2757, opened 2020-07-15
FEATURE: Configurable Asset Constraints

bwaidelich:
See https://github.com/neos/neos-development-collection/issues/2984

PR #2759, opened 2020-07-15
FEATURE: Configurable Asset Constraints

bwaidelich:
See neos/neos-development-collection#2984

### DIFF
--- a/packages/neos-ui-backend-connector/src/Endpoints/index.ts
+++ b/packages/neos-ui-backend-connector/src/Endpoints/index.ts
@@ -1,5 +1,4 @@
-import {getElementInnerText, getElementAttributeValue, getContextString} from './Helpers';
-import {urlWithParams, encodeAsQueryString} from '@neos-project/utils-helpers/src/urlWithParams';
+import {urlWithParams, searchParams, getElementInnerText, getElementAttributeValue, getContextString} from './Helpers';
 
 import fetchWithErrorHandling from '../FetchWithErrorHandling/index';
 import {Change, NodeContextPath, WorkspaceName, DimensionCombination, DimensionPresetCombination, DimensionName} from '@neos-project/neos-ts-interfaces';
@@ -60,7 +59,7 @@ export default (routes: Routes) => {
             changes
         })
     })).then(response => fetchWithErrorHandling.parseJson(response))
-    .catch(reason => fetchWithErrorHandling.generalErrorHandler(reason));
+        .catch(reason => fetchWithErrorHandling.generalErrorHandler(reason));
 
     const publish = (nodeContextPaths: NodeContextPath[], targetWorkspaceName: WorkspaceName) => fetchWithErrorHandling.withCsrfToken(csrfToken => ({
         url: routes.ui.service.publish,
@@ -75,7 +74,7 @@ export default (routes: Routes) => {
             targetWorkspaceName
         })
     })).then(response => fetchWithErrorHandling.parseJson(response))
-    .catch(reason => fetchWithErrorHandling.generalErrorHandler(reason));
+        .catch(reason => fetchWithErrorHandling.generalErrorHandler(reason));
 
     const discard = (nodeContextPaths: NodeContextPath[]) => fetchWithErrorHandling.withCsrfToken(csrfToken => ({
         url: routes.ui.service.discard,
@@ -90,7 +89,7 @@ export default (routes: Routes) => {
             nodeContextPaths
         })
     })).then(response => fetchWithErrorHandling.parseJson(response))
-    .catch(reason => fetchWithErrorHandling.generalErrorHandler(reason));
+        .catch(reason => fetchWithErrorHandling.generalErrorHandler(reason));
 
     const changeBaseWorkspace = (targetWorkspaceName: WorkspaceName, documentNode: NodeContextPath) => fetchWithErrorHandling.withCsrfToken(csrfToken => ({
         url: routes.ui.service.changeBaseWorkspace,
@@ -106,7 +105,7 @@ export default (routes: Routes) => {
             documentNode
         })
     })).then(response => fetchWithErrorHandling.parseJson(response))
-    .catch(reason => fetchWithErrorHandling.generalErrorHandler(reason));
+        .catch(reason => fetchWithErrorHandling.generalErrorHandler(reason));
 
     const copyNodes = (nodes: NodeContextPath[]) => fetchWithErrorHandling.withCsrfToken(csrfToken => ({
         url: routes.ui.service.copyNodes,
@@ -121,7 +120,7 @@ export default (routes: Routes) => {
             nodes
         })
     })).then(response => fetchWithErrorHandling.parseJson(response))
-    .catch(reason => fetchWithErrorHandling.generalErrorHandler(reason));
+        .catch(reason => fetchWithErrorHandling.generalErrorHandler(reason));
 
     const cutNodes = (nodes: NodeContextPath[]) => fetchWithErrorHandling.withCsrfToken(csrfToken => ({
         url: routes.ui.service.cutNodes,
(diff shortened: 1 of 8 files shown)


packages/neos-ui/src/manifest.dataloaders.js
```diff
@@ -221,7 +221,7 @@ manifest('main.dataloaders', {}, globalRegistry => {
             }).then(() => {
                 // Trigger query
                 const assetProxySearchApi = backend.get().endpoints.assetProxySearch;
-                const resultPromise = assetProxySearchApi(searchTerm, '', options);
+                const resultPromise = assetProxySearchApi(searchTerm, options);
```
bwaidelich, line 224: @Sebobo asked (in #2757) why this param can be removed.
=> This was referring to the `assetSourceIdentifier` param, that is now replaced by `options.constraints.assetSources` (See https://github.com/neos/neos-ui/pull/2759/files#diff-d34d33e9dc47b1d77ad5f29e5a091011R260)

Thinking about it, it probably makes sense to keep it though for b/c..

bwaidelich: OTOH, is that really part of the public API?

packages/neos-ui-editors/src/Editors/Image/index.js
```diff
@@ -210,10 +210,13 @@ export default class ImageEditor extends Component {
...
     handleChooseFromMedia = () => {
-        const {secondaryEditorsRegistry} = this.props;
+        const {secondaryEditorsRegistry, options} = this.props;
         const {component: MediaSelectionScreen} = secondaryEditorsRegistry.get('Neos.Neos/Inspector/Secondary/Editors/MediaSelectionScreen');
 
-        this.props.renderSecondaryInspector('IMAGE_SELECT_MEDIA', () => <MediaSelectionScreen type="images" onComplete={this.handleMediaSelected}/>);
+        // set media type constraint to "image/*" if it is not explicitly specified via options.constraints.mediaTypes
+        const constraints = {...options.constraints, mediaTypes: (options.constraints && options.constraints.mediaTypes) || ['image/*']};
+
+        this.props.renderSecondaryInspector('IMAGE_SELECT_MEDIA', () => <MediaSelectionScreen constraints={constraints || {}} onComplete={this.handleMediaSelected}/>);
```
Sebobo, line 219: Here we still should keep the `type` for now as mentioned in the other PR.

bwaidelich: Ah, keep it on the _calling_ side, too. Good point, and adjusted!

packages/neos-ui-backend-connector/src/Endpoints/index.ts
```diff
@@ -257,8 +257,9 @@ export default (routes: Routes) => {
             return getElementInnerText(assetProxy, '.local-asset-identifier');
         });
 
-    const assetProxySearch = (searchTerm = '', assetSourceIdentifier = '', options: {assetsToExclude: string[]} = {assetsToExclude: []}) => fetchWithErrorHandling.withCsrfToken(() => ({
-        url: urlWithParams(routes.core.service.assetProxies, {searchTerm, assetSourceIdentifier}),
+    const assetProxySearch = (searchTerm = '', options: {assetsToExclude: string[], constraints: any} = {assetsToExclude: [], constraints: {}}) => fetchWithErrorHandling.withCsrfToken(() => ({
```
bwaidelich, line 260: TODO: Re-add "assetSourceIdentifier" and add to "constraints" param if set

packages/neos-ui-editors/src/SecondaryEditors/MediaSelectionScreen/index.js
```diff
@@ -24,13 +24,33 @@ class MediaSelectionScreen extends PureComponent {
...
+        }
         const mediaBrowserUri = $get('routes.core.modules.mediaBrowser', neos);
-
         return (
-            <iframe name="neos-media-selection-screen" src={`${mediaBrowserUri}/${type}.html`} className={style.iframe}/>
+            <iframe name="neos-media-selection-screen" src={`${mediaBrowserUri}/assets/index.html?${this.encodeAsQueryString({constraints})}`} className={style.iframe}/>
         );
     }
+
+    encodeAsQueryString = (obj, prefix) => {
```
markusguenther, line 38: Maybe we put this into the `utils-helpers` package. Seems general and can be used also in other places. What do you think @bwaidelich ??

bwaidelich: Good call. There is already a similar helper in https://github.com/neos/neos-ui/blob/master/packages/neos-ui-backend-connector/src/Endpoints/Helpers.ts#L42

But I wasn't sure wether we want to introduce that cross-package dependency here. Also the existing helper doesn't (yet) work recursively yet IIRC and doesn't remove `null` params.

If that dependency is not an issue we could just extend the existing helper and use that

bwaidelich: @markusguenther @Sebobo @dimaip do you guys think its fine to have a dependency from the `neos-ui-editors` package to the `neos-ui-backend-connector` package (or is that the case already?) and – if so – should we extend the existing `urlWithParams` helper so that it can deal with nested objects and `null` params?

bwaidelich: *ping* :)

dimaip: Hey @bwaidelich!
I'd rather extract this helper from both places and put it into `utils-helpers`.
But I don't think it's the end of the world if you leave it here either, though extracting it is perhaps a better move.

bwaidelich: @markusguenther @dimaip Thanks for your feedback, I'll try to move it into `utils-helper` later today, seems like a good place for it! :)

packages/neos-ui-backend-connector/src/Endpoints/index.ts
bwaidelich: Done

packages/neos-ui/src/manifest.dataloaders.js
bwaidelich: I re-added it, you never know...

packages/neos-ui-editors/src/SecondaryEditors/MediaSelectionScreen/index.js
bwaidelich: OK, later _today_ didn't work out but almost :)

bwaidelich:
Mh. can someone tell whether the breaking e2e tests are related to this change?

markusguenther:
> Mh. can someone tell whether the breaking e2e tests are related to this change?

Will check it. Video tells me that it is hanging somehow. https://app.saucelabs.com/tests/dce33035f2e347fc9505bf1b44d7f767#3

bwaidelich:
> Video tells me..

Wow, I didn't know they exist. Whoever has set up all these CI checks deserves a medal!

I partly reverted my last commit to see if that fixes the tests

markusguenther:
The medal is for @dimaip and @johannessteu ;)

markusguenther:
Awesome feature :) ❤️

dimaip:
Great feature @bwaidelich!
And yes, I'm terribly happy every time our test setup catches something :)

bwaidelich:
> And yes, I'm terribly happy every time our test setup catches something :)

Absolutely. In this case it just caught a cold (or something else that led to a false negative) – but better this way then the other way around :)

dimaip:
Oh, I thought it caught something legit

PR #2760, opened 2020-07-22
TASK: Update @friendsofreactjs/react-css-themr

markusguenther:
Updates to the latest version


markusguenther:
Package is broken for our use case :(

PR #2764, opened 2020-08-11
!!! BUGFIX: Speed up move actions to prevent timeouts

Sebobo:
**What I did**

WIth this change instead of returning the information for the moved node and all its
descendants only the changes to the new and old parent and the information
about the old and new context path are submitted.

The already stored nodes are then updated with this information instead of being replaced
completely with the new information from the server.

This improves the response time by ~50% and memory consumption by 25% - 80%.
Large move operations still need a lot of memory and some time to compute
the updated workspace information. This issue will be fixed in a future PR.

This change also has the side effect, that after a move action the moved node can be focused again, 
as the information is now available where it went.
Before the old was removed and a redirect to the parent happened.

**Possible breaking change**

As the `Neos.Neos.Ui:RemoveNode/Main` feedback from the server was replaced with a `Neos.Neos.Ui:UpdateNodePath/Main` this change can be breaking for plugins that listened to it.
They will now have to listen to both.

**How I did it**

Replace the old `remove` + `update` feedbacks with a single `UpdatePath` feedback.
The Feedback contains the old and new path and the UI can update the known nodes with this information.

**How to verify it**

Move some nodes with child nodes.
The node tree should still be consistent after the operation.

Example with 4500 changed nodes (including the fix in https://github.com/neos/neos-development-collection/pull/3015):

Performance difference for 4500 changed nodes

### DIFF
--- a/Classes/Domain/Model/Changes/AbstractCopy.php
+++ b/Classes/Domain/Model/Changes/AbstractCopy.php
@@ -14,6 +14,7 @@ namespace Neos\Neos\Ui\Domain\Model\Changes;
 use Neos\ContentRepository\Domain\Model\NodeInterface;
 use Neos\ContentRepository\Domain\Service as ContentRepository;
 use Neos\Flow\Annotations as Flow;
+use Neos\Neos\Ui\Domain\Model\Feedback\Operations\UpdateNodeInfo;
 
 abstract class AbstractCopy extends AbstractStructuralChange
 {
@@ -22,6 +23,21 @@ abstract class AbstractCopy extends AbstractStructuralChange
      * @var ContentRepository\NodeServiceInterface
      */
     protected $contentRepositoryNodeService;
+
+    /**
+     * Perform finish tasks - needs to be called from inheriting class on `apply`
+     *
+     * @param NodeInterface $node
+     * @return void
+     */
+    protected function finish(NodeInterface $node): void
+    {
+        $updateNodeInfo = new UpdateNodeInfo();
+        $updateNodeInfo->setNode($node);
+        $updateNodeInfo->recursive();
+        $this->feedbackCollection->add($updateNodeInfo);
+        parent::finish($node);
+    }
 
     /**
      * Generate a unique node name for the copied node
--- a/Classes/Domain/Model/Changes/AbstractCreate.php
+++ b/Classes/Domain/Model/Changes/AbstractCreate.php
@@ -16,6 +16,7 @@ use Neos\ContentRepository\Domain\Model\NodeType;
 use Neos\ContentRepository\Domain\Service\NodeServiceInterface;
 use Neos\ContentRepository\Domain\Service\NodeTypeManager;
 use Neos\Flow\Annotations as Flow;
+use Neos\Neos\Ui\Domain\Model\Feedback\Operations\UpdateNodeInfo;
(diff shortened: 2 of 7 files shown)


Sebobo:
There seems to be a possible bug when moving the a page that is currently being viewed. Will investigate.

Sebobo:
OK, I had overlooked the complexity in the removal event that was part of the previous move event. Thanks to @markusguenther for finding the issue.

It's now mostly behaving the same but actually a tiny bit smarter as the focus is now kept correctly after the move.
Even if a parent of the currently selected node was moved (documents and content). The remove is not able to do that in all cases.

So the removal action could also be improved in the same way. But not sure if that should be in here...

Sebobo:
I fear this would be now breaking if someone listened to the new node removed feedback. The old one is still fired.

markusguenther:
Do you really think that someone is using this? I personally don't know a project that use actively the events of the UI.

kitsunet:
> Do you really think that someone is using this? I personally don't know a project that use actively the events of the UI.

Dangerous, I have made the same assessment over the years just to get to know after release someone was using it. If it's public API in any way you must assume it's used by _someone_. Even if it's not public API but then you can say I don't care because it's not public API :D

dimaip:
Great change!
Just curious, why is it labelled as a Bugfix? I'd call it a task and aim for master...

kdambekalns:
> Just curious, why is it labelled as a Bugfix?

Guess: To be able to *not* have to target master!? 🙊

Sebobo:
> Great change!
> 
> Just curious, why is it labelled as a Bugfix? I'd call it a task and aim for master...

@dimaip because its part of the improvements that Sebastian and me started with https://github.com/neos/neos-development-collection/pull/3015.
Of course its bigger, but a huge improvement for every Neos site.
Putting it into master would mean that many sites will get into sometime next year maybe.
If we consider the breaking topic then it would be in v6 and a long time until sites get it.

Sebobo:
And regarding whether its a real bug... for users it creates bugs when their requests time out because they move their pages.

kitsunet:
I still think this is all hairy in terms of a bugfix. Maybe we should think about an earlier major release instead 🙊

Sebobo:
Many projects are LTS only. So the question is: "now", "in some months when 5.3 is considered great", "2022 with 6.3".

Targeting 5.3 sounds like a compromise maybe.

Besides that. More feedback to the implementation is also welcome ;)

skurfuerst:
Hey, 

To me actually this is a nasty performance bug we should have caught earlier... So IMHO our users will thank us putting this into more releases.

I consider the "breakingness" really low actually - I'd still mark it as breaking under extremely rare circumstances; and have a neos.io blog post explaining the "breakingness". We'll need this for the other performance fix as well (the one with Doctrine EntityManager).

All the best,
Sebastian

kitsunet (approved):
After looking at how the feedback is documented and exposed and the possible cases, this makes sense to merge, even as a bugfix.

skurfuerst (approved):

Sebobo:
Thx everyone. 

I didn't fix the styleci stuff as it was not related to my change. We can do a separate PR to clean those up.

markusguenther:
Awesome improvement 🙏

PR #2769, opened 2020-08-28
BUGFIX: Avoid URLs to be rendered twice in link- and reference editors

kdambekalns:
This applies https://github.com/neos/neos-ui/pull/2711 to the 4.0
branch.


kdambekalns:
e2e test fails with a connection error related to Saucelabs. 🤷‍♂️

markusguenther (approved):
Tested that locally and the e2e tests also fail locally caused by react warnings.
So I filter the warnings. We need to tackle this soonish to get rid of such side effects in the pipeline.

Thanks to @bwaidelich and @kdambekalns for the fix. We need to be aware when the up merging happens that the change already exist in 5.1 and newer.

PR #2772, opened 2020-09-09
TASK: Limit dependencies to neos 5.0

markusguenther:
With the current configuration the test distribution always installs the latest 5.x neos and that is not compatible since the preview url change.

So we should limit that to neos 5.0 and adapt that for the upcoming versions.

**What I did**
Changed version constraint from caret version to tilde version.

**How to verify it**
Check end to end tests of the PR.

### DIFF
--- a/Tests/IntegrationTests/TestDistribution/DistributionPackages/Neos.TestSite/composer.json
+++ b/Tests/IntegrationTests/TestDistribution/DistributionPackages/Neos.TestSite/composer.json
@@ -3,6 +3,7 @@
     "description": "A dummy site package that will be replaced by fixtures during tests",
     "type": "neos-site",
     "require": {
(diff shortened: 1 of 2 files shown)


markusguenther:
Video of the end to end test: https://app.saucelabs.com/tests/21980edf8901428e8a39c8db2db5ca2f

Sebobo (approved):

kdambekalns (approved):

kdambekalns:
When upmerging this need to be adjusted to 5.1, 5.2, 5.3 respectively?

markusguenther:
> When upmerging this need to be adjusted to 5.1, 5.2, 5.3 respectively?

Correct...
The issue was actually that we got always the newest 5.x release and as 5.1 introduced the new preview url it was in 5.0 completely broken.

PR #2777, opened 2020-09-14
BUGFIX: Prevent preview button to link to the personal user workspace

bwaidelich:
With this change the `PreviewButton` of the UI never links to a preview
of the node in the personal user workspace.
Instead it:

* Is disbled if the node is hidden/missing in the base workspace
* Points to a `/neos/redirect?<node-in-base-workspace>` URL that redirects to the current node
  in its base workspace

For the "usual editing case" (personal workspace with target == `live`)
that means, that clicking the button will lead to the version in the live
workspace (i.e. `/the/final/url.html`).

For shared workspaces the redirect will point to a preview URL for that
workspace (i.e. `/neos/preview?node=<node-in-shared-workspace>`)

Fixes: neos/neos-development-collection#3019

### DIFF
--- a/packages/neos-ui/src/manifest.js
+++ b/packages/neos-ui/src/manifest.js
@@ -220,6 +220,13 @@ manifest('main', {}, globalRegistry => {
     });
 
     //
+    // When the server advices to update the nodes preview URL, dispatch the action to do so
+    //
+    serverFeedbackHandlers.set('Neos.Neos.Ui:UpdateNodePreviewUrl/Main', (feedbackPayload, {store}) => {
+        store.dispatch(actions.UI.ContentCanvas.setPreviewUrl(feedbackPayload.newPreviewUrl));
+    });
+
+    //
     // When the server advices to reload the children of a document node, dispatch the action to do so.
     //
     serverFeedbackHandlers.set('Neos.Neos.Ui:NodeCreated/Main', (feedbackPayload, {store}) => {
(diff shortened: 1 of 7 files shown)


jonnitto:
You're awesome! Thank you for this PR! ❤️

packages/neos-ui/src/manifest.js
```diff
@@ -219,6 +219,13 @@ manifest('main', {}, globalRegistry => {
         store.dispatch(actions.CR.Workspaces.update(feedbackPayload));
     });
 
+    //
+    // When the server advices to update the nodes preview URL, dispatch the action to do so
+    //
+    serverFeedbackHandlers.set('Neos.Neos.Ui:UpdateNodePreviewUrl/Main', (feedbackPayload, {store}) => {
+        store.dispatch(actions.UI.ContentCanvas.setPreviewUrl(feedbackPayload.newPreviewUrl));
```
skurfuerst, line 226: I think we still need to discuss thus part :)

skurfuerst: I think that is incorrect as soon as you publish multiple document nodes at once...

bwaidelich: very good catch indeed..

bwaidelich:
The preview route has been introduced with 5.1 (c255b2933be434531efc2cdba446f60ee6d2773b). I'll close this one and create it against the 5.1 branch accordingly!

PR #2780, opened 2020-09-15
Bugfix/3019 prevent preview button from linking to personal workspace 2

bwaidelich:
With this change the `PreviewButton` of the UI never links to a preview
of the node in the personal user workspace.
Instead it:

* Is disbled if the node is hidden/missing in the base workspace
* Points to a `/neos/redirect?<node-in-base-workspace>` URL that redirects to the current node
  in its base workspace

For the "usual editing case" (personal workspace with target == `live`)
that means, that clicking the button will lead to the version in the live
workspace (i.e. `/the/final/url.html`).

For shared workspaces the redirect will point to a preview URL for that
workspace (i.e. `/neos/preview?node=<node-in-shared-workspace>`)

Fixes: neos/neos-development-collection#3019


bwaidelich:
grml

PR #2782, opened 2020-09-16
BUGFIX: Fix spacing on dimension-switcher dropdown

c4ll-m3-j4ck:
fixes #2766 

Added some more consistent spacing to the Dimension Switcher in the backend-ui.

Verify by using the forked version and compiling the UI and trying it out in the backend.

Before:
before

After:
before

PR #2783, opened 2020-09-17
BUGFIX: Fix spacing on dimension-switcher dropdown

c4ll-m3-j4ck:
Followup to #2782, targeting 5.1


markusguenther (approved):
Thanks @c4ll-m3-j4ck for fixing that issue. Everything looks fine and I tested this change now with neos 5.1 as we moved the switcher in this version.

Special thanks for the first contribution in the neos-ui. Would be awesome to see more in the future.
🎉 🥇

PR #2785, opened 2020-09-22
BUGFIX: Reduce the number of NodeTypes to run policy check against #2784

### LINKED ISSUE #2784: Improve performance of /neos/ui-services/get-additional-node-metadata endpoint
### Description

The endpoint `/neos/ui-services/get-additional-node-metadata` is called by the backend multiple times and can take many seconds to finish. In our case we have around 600 NodeTypes and this request can take 1-2 minutes.

The input for the call is a set of nodes which are processed here https://github.com/neos/neos-ui/blob/master/Classes/Controller/BackendServiceController.php#L465.

The runtime of this method depends on
* the number of nodes given as the input
* the number of NodeTypes configured
* the time needed for policy checks

Most of the time is consumed by https://github.com/neos/neos-ui/blob/master/Classes/Service/NodePolicyService.php#L109.

I'll provide a PR to reduce the number of NodeTypes processed during the policy checking.

### Steps to Reproduce

Have a look at the debug console in the neos backend of your browser. In bigger projects this request often takes a few seconds.

### Affected Versions

Neos: 4.x and up

UI: 4.x and up

gjwnc:
fixes #2784

**What I did**

In `getDisallowedNodeTypes(NodeInterface $node)` the list of configured NodeTypes is filtered using the NodeType constraints of the given `$node`. Abstract NodeTypes are still included which might be relevant for the policy check.

**How I did it**

Before running the policy check, `array_filter` is called used on all NodeTypes (`$this->nodeTypeManager->getNodeTypes()`). The filter method then checks the constraints depending if `$node` is auto-created or if the NodeType to filter is a superType or the NodeType of the node itself.

**How to verify it**

One need to compare the runtime of of the /get-additional-node-metadata endpoint in the Neos Backend UI on a project with configured CreateNodePrivilege policy.

This is a screenshot for a Neos 4.3 project without the code changes: 
image

And this one has the code changes in place:
image

### DIFF
--- a/Classes/Service/NodePolicyService.php
+++ b/Classes/Service/NodePolicyService.php
@@ -119,7 +119,7 @@ class NodePolicyService
         $nodeNodeType = $node->getNodeType();
 
         $superTypes = [];
-        $generateSuperTypes = function(array $nodeTypes, &$superTypes) use (&$generateSuperTypes) {
+        $generateSuperTypes = function (array $nodeTypes, &$superTypes) use (&$generateSuperTypes) {
             foreach ($nodeTypes as $nodeType) {
                 $superTypes[$nodeType->getName()] = $nodeType;
                 $generateSuperTypes($nodeType->getDeclaredSuperTypes(), $superTypes);
@@ -131,7 +131,7 @@ class NodePolicyService
         try {
             $parentNode = $node->findParentNode();
             $parentNodeType = $parentNode->getNodeType();
-        }catch(NodeException $exception){
+        } catch (NodeException $exception) {
             // no parentNode found
         }
 
@@ -150,16 +150,15 @@ class NodePolicyService
                 if ($parentNodeType && $parentNodeType->allowsGrandchildNodeType($nodeName, $nodeType)) {
                     return true;
                 }
-            }
-            else if ($nodeNodeType->allowsChildNodeType($nodeType)) {
+            } elseif ($nodeNodeType->allowsChildNodeType($nodeType)) {
                 return true;
             }
             // check if the nodeType is a supertype
-            else if (isset($superTypes[$nodeType->getName()])) {
+            elseif (isset($superTypes[$nodeType->getName()])) {
                 return true;
             }
             // check if the nodeType is the same as that of the current $node's nodeType
-            else if ($nodeType->getName() == $nodeNodeType->getName()) {
+            elseif ($nodeType->getName() == $nodeNodeType->getName()) {
                 return true;
             }
 
--- a/Classes/Service/NodePolicyService.php
+++ b/Classes/Service/NodePolicyService.php
@@ -136,17 +136,17 @@ class NodePolicyService
         }
 
         // check if the node is auto-created
-        $isAutCreated   = false;
+        $isAutoCreated   = false;
         if ($parentNodeType && array_key_exists((string)$node->getNodeName(), $parentNodeType->getAutoCreatedChildNodes())) {
-            $isAutCreated = true;
+            $isAutoCreated = true;
         }
 
         // filter the set of configured nodeTypes
         $nodeName = (string)$node->getNodeName();
 
-        $constraintAndSuperTypeFilter = function ($nodeType) use ($nodeName, $nodeNodeType, $superTypes, $isAutCreated, $parentNodeType) {
+        $constraintAndSuperTypeFilter = function ($nodeType) use ($nodeName, $nodeNodeType, $superTypes, $isAutoCreated, $parentNodeType) {
             // check if the nodeType is mentioned in the constraints
-            if ($isAutCreated) {
+            if ($isAutoCreated) {
                 if ($parentNodeType && $parentNodeType->allowsGrandchildNodeType($nodeName, $nodeType)) {
                     return true;
                 }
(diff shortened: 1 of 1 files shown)


Classes/Service/NodePolicyService.php
```diff
@@ -114,14 +115,68 @@ public function getDisallowedNodeTypes(NodeInterface $node): array
...
+        $parentNodeType = null;
+        try {
+            $parentNode = $node->findParentNode();
+            $parentNodeType = $parentNode->getNodeType();
+        } catch (NodeException $exception) {
+            // no parentNode found
+        }
+
+        // check if the node is auto-created
+        $isAutCreated   = false;
```
dimaip, line 139: isAutoCreated?

Classes/Service/NodePolicyService.php
```diff
@@ -114,14 +115,68 @@ public function getDisallowedNodeTypes(NodeInterface $node): array
...
+            // check if the nodeType is the same as that of the current $node's nodeType
+            elseif ($nodeType->getName() == $nodeNodeType->getName()) {
+                return true;
+            }
+
+            // ignore other nodeType
+            return false;
+        };
+
+        $nodeTypes = array_filter($this->nodeTypeManager->getNodeTypes(), $constraintAndSuperTypeFilter);
```
dimaip, line 169: For my taste I'd call it smth like `filteredNodeTypes`

dimaip:
Can't really say something meaningful about this, I guess only @kitsunet can...

Classes/Service/NodePolicyService.php
gjwnc: Thanks! I've fixed it

Classes/Service/NodePolicyService.php
gjwnc: Yes, you are right. I've changed the variable name according to your suggestion.

markusguenther:
@gjwnc Hi,

I tried to update the PR to the latest 5.3, but seems that I as team member has no permission to change the PR :/
Are you willing to rebase to origin/5.3 or should I take your commits and open a new PR?

Thanks for the effort :)

Screenshot 2021-10-15 at 13 39 57

markusguenther:
Thanks for updating the branch 💚

Sebobo:
Thx, will try the patch in a larger project and test it.

Classes/Service/NodePolicyService.php
```diff
@@ -114,14 +115,68 @@ public function getDisallowedNodeTypes(NodeInterface $node): array
...
+        $generateSuperTypes = function (array $nodeTypes, &$superTypes) use (&$generateSuperTypes) {
+            foreach ($nodeTypes as $nodeType) {
+                $superTypes[$nodeType->getName()] = $nodeType;
+                $generateSuperTypes($nodeType->getDeclaredSuperTypes(), $superTypes);
+            }
+        };
+        $generateSuperTypes($nodeNodeType->getDeclaredSuperTypes(), $superTypes);
+
+        $parentNodeType = null;
+        try {
```
Sebobo, line 131: Couldn't this try/catch be replaced by a `getParent` and checking whether its not null?

gjwnc: Yes that would be possible, but `getParent()` is deprecated and in the comment for `getParent()` it is mentioned that you should use `findParentNode()`.

Sebobo: True, just wondered if the try/catch stuff might have a negative impact again. But not sure how many times the method is actually called.

Sebobo:
Did you project use a lot of node related policies? I could see a small difference of ~15% in my customers project which is quite large, but they don't use node policies.

gjwnc:
We use EditNodePropertyPrivilege and Create|RemoveNodePrivilege very often in the project because we limit access to the site and language dimension.

What do you mean by difference? The returned array of the `getDisallowedNodeTypes()`?  
It's quite some time, that I worked on this, but as I read it now again, I think in the final result, I miss the set of nodeTypes the policy check is not running against. Because this set is disallowed by definition for the given node's nodeType, right?

Sebobo:
I meant the performance improvement is ~15%, but I have to make more checks with pages that have more nodes.

And regarding the set of NodeTypes. I think you also include them in line 169.

mhsdesign:
should/can this new code be extracted into a new method? just a lot going on...

gjwnc:
@mhsdesign Do you mean split `getDisallowedNodeTypes()` into a protected methods or creating a completely separate new method along the original `getDisallowedNodeTypes()`?  
I would just move the anonymous functions for `$generateSuperTypes` and `$constraintAndSuperTypeFilter` to separate methods.
Since this is more than a year old I'm questioning if it will be merge sometime soon - if it's worth it currently. It's basically a nice to have I noticed quite some time ago.

Classes/Service/NodePolicyService.php
mhsdesign: https://php.watch/articles/php-exception-performance

just less performant, when an exception will be thrown.

mhsdesign: but an exception will only be thrown when `$this->isRoot() === true`:
https://github.com/neos/neos-development-collection/blob/c14cc4ab46e7d9be75157bb5aa80f96fcc2dff14/Neos.ContentRepository/Classes/Domain/Model/Node.php#L2026

so we could instead check for that first and never get an exception.

Sebobo (approved):

Sebobo:
I'm fine with merging this change, but it would be great if someone could test it in a project where the policies are extensively used.

kitsunet (approved):
Great idea, I needed a couple of reads, but seems sound to me and works. So I would agree to merge but have to bring up the question of version. I am not sure this can be declared a bug, strictly speaking. Performance is a feature, and such optimizations always have the risk of breaking something :D Would at least want to hear some other opinions, but would probably otherwise merge in ~24h as I feel it's a good change.

Sebobo:
I would feel better to merge into 8.0 honestly.

mhsdesign:
So better merge this into 8.0 as Feature or as (Performance) Bug in 5.3?

Issue #2093, opened 2018-08-23
Refresh NodeTree does not work when the NodeType filter is used


bwaidelich:
### Description

When the NodeType filter is active, hitting the "Refresh" button results in a broken Node Tree & Inspector

### Steps to Reproduce

1. Use the `NodeType` filter to filter the tree for a node type or preset
2. Hit the "Refresh" button above the tree

#### Expected behavior

Node tree should be reloaded, inspector reflect the currently selected node

#### Actual behavior

The Node Tree shows an invalid result and the inspector keeps loading:

![image](https://user-images.githubusercontent.com/307571/44513029-93e32700-a6bc-11e8-891f-7946184cdc93.png)

The latter even happens if the filter is reset before hitting the "Refresh" button!

### Affected Versions

Neos: dev-master

UI: dev-master  de87d33

sbruggmann:
For me not reproducable anymore.
@bwaidelich Needs this issue still work?

bwaidelich:
Thanks for the pointer – I assume this one was resolved with #2076

bwaidelich:
I have to reopen this one as the issue still persists for me unfortunately.
Not by filtering for node types, but for custom *presets*.

## Steps to reproduce

Add some preset via `Settings.yaml`:

```yaml
Neos:
  Neos:
    userInterface:
      navigateComponent:
        nodeTree:
          presets:
            'default':
              baseNodeType: 'Neos.Neos:Document,!Neos.Demo:Chapter'
            'legalPages':
              ui:
                label: 'Test Preset'
              baseNodeType: 'Neos.Demo:Chapter'
```

...follow the "steps to reproduce" from above

jirgn:
I just stumbled over this issue and can confirm it still exists. 
In Addition to the described Behaviour there is also a Problem when adding in (preset) filtered State.

Description:
When adding a Node in (preset) filtered State, the new Node-Entry is not shown in Node-Tree.

Steps to Reproduce:
* configure as described above
* filter by preset ('legalpages')
* add new chapter document

Expected Result:
* new chapter is shown in the (filtered) Node-Tree

Actual Behaviour:
* no change in the Node-Tree (no new Node/Document)

DrillSergeant:
We start using the custom presets for a current project and now I am affected of this issue.

I could sponsor a bugfix if anyone could do this in the next 2 weeks.

dimaip:
Reproduced, I'm at it

bwaidelich:
[...]

PR #2787, opened 2020-09-24
BUGFIX: fix reload when filtering or searching

dimaip:
Fixes: https://github.com/neos/neos-ui/issues/2093

**What I did**

Currently if you reload the document tree nodes during searching or filtering by preset, it leads to very odd results. This PR fixes it.

**How I did it**

I store the current query and filter preset inside redux state and do `commenceSearch` instead of reloading the default nodes when the refresh button is clicked.

**How to verify it**

Put this into Settings.yaml:

```yaml
Neos:
  Neos:
    userInterface:
      navigateComponent:
        nodeTree:
          presets:
            'default':
              baseNodeType: 'Neos.Neos:Document,!Neos.Demo:Chapter'
            'legalPages':
              ui:
                label: 'Test Preset'
              baseNodeType: 'Neos.Demo:Chapter'
```

Toggle the preset and hit reload. Also try searching for something and hit reload.

### DIFF
--- a/packages/neos-ui-redux-store/src/UI/PageTree/index.ts
+++ b/packages/neos-ui-redux-store/src/UI/PageTree/index.ts
@@ -13,6 +13,8 @@ export interface State extends Readonly<{
     intermediate: NodeContextPath[];
     loading: NodeContextPath[];
     errors: NodeContextPath[];
+    query: string;
+    filterNodeType: string;
 }> {}
 
 export const defaultState: State = {
@@ -21,7 +23,9 @@ export const defaultState: State = {
     hidden: [],
     intermediate: [],
     loading: [],
-    errors: []
+    errors: [],
+    query: '',
+    filterNodeType: ''
 };
 
 //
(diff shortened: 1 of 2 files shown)


dimaip:
Please check if it does the trick for you @DrillSergeant

bwaidelich:
I finally came around testing this, and it seems as though it doesn't solve the issue for me at least.
With the setup described in the issue I get the following behavior now:


2787_test_1

bwaidelich:
Also, not new with this fix but related:
2787_test_2

markusguenther (approved):

markusguenther:
Thanks @dimaip for taking care. And it is awesome that Internezzo is sponsoring the fix of this bug.
This PR really solves the issue with the refresh of the tree.

The solution is pretty clean and straight forward. But @bwaidelich is right that some issues still exists.
When you delete or create a node with the filter then the tree is broken.

When I combine this change with PR #2788 the issue for deleting a node in the tree with a filter is also solved.
But when you create a new node the tree miss these node and you need to press the refresh tree button.

I would merge this as it solves the issue with the refresh and the filter and search. And will better tackle the last missing piece in a new PR.

Thanks to all of you for solving this ❤️

dimaip:
Thanks for testing, guys! I'll try to fix the other issue soonish too.

Issue #1691, opened 2018-03-06
BUG: Navigating the Document Tree makes filtered nodes disappear


bwaidelich:
### Description

When a node type filter is active in the `navigateComponent`, when navigating the node tree previously filtered nodes disappear..

This is not easy to explain, here's a video of the bug in action:

![broken_node_filter](https://user-images.githubusercontent.com/307571/37031524-6b83862c-213f-11e8-988f-3309c6de491a.gif)

I don't understand what is happening exactly, but I can confirm that this works in the "old UI".

Using `Redux DevTools` to debug this I can see a lot of removed nodes in the state of the  `ADD` event but I'm not sure if that helps ;)

![image](https://user-images.githubusercontent.com/307571/37031624-c8ad32da-213f-11e8-8016-1adb5b7e63cd.png)

### Background

When there are a lot of (Document) Nodes on one level, this can significantly slow down the NodeTree. For this reason it can make sense to hide certain node types using the [Node tree base node type](http://neos.readthedocs.io/en/stable/References/Configuration/Configuration.html#node-tree-base-node-type).

sbruggmann:
Not reproducable in Neos 3.3 with UI 2.x and Neos 4.2 with UI 3.x

Sebobo:
@bwaidelich do you still get this error?

bwaidelich:
Yes, unfortunately the issue still exists for me – but not when filtering for node types, but for custom *presets*.

## Steps to reproduce

Add some preset via `Settings.yaml`:

```yaml
Neos:
  Neos:
    userInterface:
      navigateComponent:
        nodeTree:
          presets:
            'default':
              baseNodeType: 'Neos.Neos:Document,!Neos.Demo:Chapter'
            'legalPages':
              ui:
                label: 'Test Preset'
              baseNodeType: 'Neos.Demo:Chapter'
```

...follow the "steps to reproduce" from above

DrillSergeant:
We start using the custom presets for a current project and now I am affected of this issue.

I could sponsor a bugfix if anyone could do this in the next 2 weeks.

markusguenther:
As already mentioned in the PR the general issue seems to be resolved but we have some glitches left.
So when you move a node inside another node the tree does not show that until you refresh the tree.

Or when you create a new node and the filter is enabled, the new node is also not part of the tree until you refresh the tree.
I would say we open a new PR for that. But it works much better than before. Thanks to @dimaip and @DrillSergeant ❤️ 

[...]

PR #2788, opened 2020-09-24
BUGFIX: merge children instead of replacing them

dimaip:
Fixes: https://github.com/neos/neos-ui/issues/1691

**What I did**

When you toggle a custom tree preset and then navigate the tree, the children of the node missing in the default preset would disappear. I fixed it.

**How I did it**

When we are navigating to some page, the node information about this node is being read from the script tags rendered on that page and merged into the redux state. Now it so happens that those inline script tags are rendered according to the default preset, and so the nodes that exist only in the preset get removed from the redux state.

The only way I could think of fixing it is to actually merge children instead of overwriting them. I first merge them, then make sure there are now duplicates left comparing them by context paths.

**How to verify it**

See the issue description basically.

### DIFF
--- a/packages/neos-ui-redux-store/src/CR/Nodes/helpers.ts
+++ b/packages/neos-ui-redux-store/src/CR/Nodes/helpers.ts
@@ -54,3 +54,19 @@ export const getNodeOrThrow = (nodeMap: NodeMap, contextPath: NodeContextPath) =
     }
     return node;
 };
+
+// We have to merge children and not just override them, as there still may be children from different node tree preset, and we shouldn't overwrite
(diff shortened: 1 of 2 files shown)


dimaip:
@DrillSergeant could you look into this one as well please?

markusguenther (approved):
Thanks @dimaip for resolving this issue. As already mentioned in the PR regarding the search and refresh with presets this solves the issue but one little piece is missing.

When you create a new node the node is not listed in the filtered tree.
But will address this in a new PR.

Thanks again to @dimaip and Internezzo which sponsored the issue 🎉

PR #2789, opened 2020-09-25
TASK: Upgrade react-collapse to version 5

markusguenther:
Upgrading react-collapse to the latest version for security reasons.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,8 @@
   "resolutions": {
     "moment": "^2.20.1"
   },
-  "engines" : {
-    "node" : "~12"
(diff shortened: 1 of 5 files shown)


markusguenther:
This time I also tested the firefox and we do not have the issue like in https://github.com/neos/neos-ui/pull/1874 😄

dimaip (approved):
Oh wow, I wonder why it works this time!

markusguenther:
Don't know what's different now in version 5 but as I failed several times regarding this issue. I first tested this in Firefox ;)

PR #2791, opened 2020-10-04
TASK: Rename legacy fusion objects to new names

daniellienert:


Sebobo (approved):
Looks good by reading.
Thx!

PR #2794, opened 2020-10-08
BUGFIX: fix creating and moving nodes with tree preset activated

dimaip:
Fixes: https://github.com/neos/neos-ui/issues/1691

TODO:

- [ ] write E2E tests for this

**What I did**

Creating or moving nodes when tree preset was activated used to cause them to disappear from the tree.
That happened because the node information rendered with NodeInfo helper used to always use the default preset, and not the one that is currently active.

**How I did it**

I'm passing the currently active tree preset along with create and copy/move change requests, and then use it server-side to render the correct child nodes with UpdateNodeInfo feedback.

**How to verify it**

1. Set up some tree presets, e.g.:

```yaml
Neos:
  Neos:
    userInterface:
      navigateComponent:
        nodeTree:
          presets:
            'default':
              baseNodeType: 'Neos.Neos:Document,!Neos.Demo:Document.Chapter'
            'legalPages':
              ui:
                label: 'Test Preset'
              baseNodeType: 'Neos.Demo:Document.Chapter'
```

2. Toggle the preset. Try to create a node. Observe that it appears in the tree.
3. Move a node via DnD.
4. Move a node via toolbar buttons.
5. Copy a node via toolbar buttons.
6. Make sure nothing else breaks

### DIFF
--- a/Classes/Domain/Model/Changes/AbstractStructuralChange.php
+++ b/Classes/Domain/Model/Changes/AbstractStructuralChange.php
@@ -51,6 +51,33 @@ abstract class AbstractStructuralChange extends AbstractChange
     protected $cachedSiblingNode = null;
 
     /**
+     * Used when creating nodes within non-default tree preset
+     *
+     * @var string|null
+     */
+    protected $baseNodeType = null;
+
+    /**
+     * Set the baseNodeType
+     *
+     * @param string $baseNodeType
+     */
+    public function setBaseNodeType($baseNodeType)
+    {
+        $this->baseNodeType = $baseNodeType;
+    }
+
+    /**
+     * Get the baseNodeType
+     *
+     * @return string|null
+     */
+    public function getBaseNodeType()
+    {
+        return $this->baseNodeType;
+    }
+
+    /**
      * Get the insertion mode (before|after|into) that is represented by this change
      *
      * @return string
@@ -133,6 +160,7 @@ abstract class AbstractStructuralChange extends AbstractChange
 
         $updateParentNodeInfo = new UpdateNodeInfo();
         $updateParentNodeInfo->setNode($node->getParent());
+        $updateParentNodeInfo->setBaseNodeType($this->baseNodeType);
         $this->feedbackCollection->add($updateParentNodeInfo);
 
         $this->updateWorkspaceInfo();
--- a/Classes/Domain/Model/Feedback/Operations/UpdateNodeInfo.php
+++ b/Classes/Domain/Model/Feedback/Operations/UpdateNodeInfo.php
@@ -33,6 +33,28 @@ class UpdateNodeInfo extends AbstractFeedback
 
     protected $isRecursive = false;
 
+    protected $baseNodeType = null;
+
+    /**
+     * Set the baseNodeType
+     *
+     * @param string $baseNodeType
+     */
+    public function setBaseNodeType($baseNodeType)
+    {
+        $this->baseNodeType = $baseNodeType;
+    }
+
+    /**
+     * Get the baseNodeType
+     *
+     * @return string|null
+     */
+    public function getBaseNodeType()
+    {
+        return $this->baseNodeType;
+    }
+
     /**
      * Set the node
      *
@@ -122,7 +144,7 @@ class UpdateNodeInfo extends AbstractFeedback
     public function serializeNodeRecursively(NodeInterface $node, ControllerContext $controllerContext)
     {
         $result = [
-            $node->getContextPath() => $this->nodeInfoHelper->renderNodeWithPropertiesAndChildrenInformation($node, $controllerContext)
+            $node->getContextPath() => $this->nodeInfoHelper->renderNodeWithPropertiesAndChildrenInformation($node, $controllerContext, $this->baseNodeType)
         ];
 
         if ($this->isRecursive === true) {
--- a/Classes/Controller/BackendController.php
+++ b/Classes/Controller/BackendController.php
@@ -178,13 +178,14 @@ class BackendController extends ActionController
 
     /**
      * @param NodeInterface $node
+     * @param string $presetBaseNodeType
      * @throws StopActionException
      */
-    public function redirectToAction(NodeInterface $node)
+    public function redirectToAction(NodeInterface $node, string $presetBaseNodeType = null)
     {
         $this->response->getHeaders()->setCacheControlDirective('no-cache');
         $this->response->getHeaders()->setCacheControlDirective('no-store');
-        $this->redirect('show', 'Frontend\Node', 'Neos.Neos', ['node' => $node]);
+        $this->redirect('show', 'Frontend\Node', 'Neos.Neos', ['node' => $node, 'presetBaseNodeType' => $presetBaseNodeType]);
     }
 
     /**
--- a/Classes/Domain/Model/Changes/AbstractCreate.php
+++ b/Classes/Domain/Model/Changes/AbstractCreate.php
@@ -65,6 +65,7 @@ abstract class AbstractCreate extends AbstractStructuralChange
     {
         $updateNodeInfo = new UpdateNodeInfo();
         $updateNodeInfo->setNode($node);
+        $updateNodeInfo->setBaseNodeType($this->baseNodeType);
         $updateNodeInfo->recursive();
         $this->feedbackCollection->add($updateNodeInfo);
         parent::finish($node);
--- a/Classes/Domain/Model/Changes/MoveAfter.php
+++ b/Classes/Domain/Model/Changes/MoveAfter.php
@@ -52,6 +52,7 @@ class MoveAfter extends AbstractMove
 
             $updateParentNodeInfo = new UpdateNodeInfo();
             $updateParentNodeInfo->setNode($parent);
+            $updateParentNodeInfo->setBaseNodeType($this->baseNodeType);
 
             $this->feedbackCollection->add($updateParentNodeInfo);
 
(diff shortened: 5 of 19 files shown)


dimaip:
Please give it a try @DrillSergeant!

Classes/Domain/Model/Changes/AbstractStructuralChange.php
```diff
@@ -50,6 +50,33 @@ abstract class AbstractStructuralChange extends AbstractChange
...
+     * @var string|null
+     */
+    protected $baseNodeType = null;
+
+    /**
+     * Set the baseNodeType
+     *
+     * @param string $baseNodeType
+     */
+    public function setBaseNodeType($baseNodeType)
```
markusguenther, line 65: I don`t write that often PHP Code in the project but since all supported versions use modern PHP versions we should use typings.... don`t know if we have a rule for that but see @daniellienert is always adjusting that ;)

Classes/Domain/Model/Feedback/Operations/UpdateNodeInfo.php
```diff
@@ -33,6 +33,28 @@ class UpdateNodeInfo extends AbstractFeedback
...
     protected $isRecursive = false;
 
+    protected $baseNodeType = null;
+
+    /**
+     * Set the baseNodeType
+     *
+     * @param string $baseNodeType
+     */
+    public function setBaseNodeType($baseNodeType)
```
daniellienert, line 43:
```suggestion
    public function setBaseNodeType(string $baseNodeType): void
```

daniellienert: Absolutely ;)

Classes/Domain/Model/Changes/AbstractStructuralChange.php
daniellienert, line 65:
```suggestion
    public function setBaseNodeType(string $baseNodeType): void
```

Classes/Domain/Model/Changes/AbstractStructuralChange.php
```diff
@@ -50,6 +50,33 @@ abstract class AbstractStructuralChange extends AbstractChange
...
+    {
+        $this->baseNodeType = $baseNodeType;
+    }
+
+    /**
+     * Get the baseNodeType
+     *
+     * @return string|null
+     */
+    public function getBaseNodeType()
```
daniellienert, line 75:
```suggestion
    public function getBaseNodeType(): string
```

markusguenther:
I am really thankful that you tackle the issue and passthru the base nodetype. I now tested it in brave and firefox and for some reason move actions often work correct but sometimes not.

If it not works it is correct in the default preset, but the tree could not expand like in the end of my video. It is really strange. Maybe in issue caused by the tree merge we did before to fix the issue. So maybe we can check if this change can be removed with the current change. I will not find the time for that before next week. Would be awesome if you can check that out.

markusguenther:
Ah damn movie was to big :/

Classes/Domain/Model/Changes/AbstractStructuralChange.php
bwaidelich: Careful, It gotta be
```suggestion
    public function getBaseNodeType(): ?string
```

Classes/Domain/Model/Feedback/Operations/UpdateNodeInfo.php
bwaidelich: Did you just comment on your own suggestion? :)

Classes/Domain/Model/Feedback/Operations/UpdateNodeInfo.php
```diff
@@ -33,6 +33,28 @@ class UpdateNodeInfo extends AbstractFeedback
...
+    {
+        $this->baseNodeType = $baseNodeType;
+    }
+
+    /**
+     * Get the baseNodeType
+     *
+     * @return string|null
+     */
+    public function getBaseNodeType()
```
bwaidelich, line 53:
```suggestion
    public function getBaseNodeType(): ?string
```

bwaidelich:
> Ah damn movie was to big :/

The screenshot too, apparently

DrillSergeant:
I tested and the behaviour is way better than before: after adding a subchapter-node, the new node does not dissapear

But I found a faulty behaviour by executing this steps:
1. Ceate a filter preset: (I used the same as Dimitri did above)
```
Neos:
  Neos:
    userInterface:
      navigateComponent:
        nodeTree:
          presets:
            'default':
              baseNodeType: 'Neos.Neos:Document,!Neos.Demo:Document.Chapter'
            'legalPages':
              ui:
                label: 'Test Preset'
              baseNodeType: 'Neos.Demo:Document.Chapter'
```
2. Select the filter preset in the backend
3. Create a new sub-chapter "test" (I did it below "II. The Pool of Tears")
Result:
image
4. Move the new node "test" into anther chapter (I moved it into "I. Down the Rabbit-hole")
Result:
image
5. Move the new node "test" back into the first node ("II. The Pool of Tears" in my case)
**Broken Result:**
image

DrillSergeant:
Another bug found, happens when inserting node, no matter if the filter-preset is active or not:
1. Select a node
2. Add a node with insert-mode "insert after" the selected node
image

**Expected result:**
The node (I named it "Subpage 1.5") is inserted between "Subpage #1" and "Subpage #2"

**Actual result:**
The node is inserted after  "Subpage #3"
image

DrillSergeant:
I will stop testing for now beacuse fixes of the two just reported behaviours might be better first.

markusguenther:
> > Ah damn movie was to big :/
> 
> The screenshot too, apparently

Updated the post ... in the second one you can see that it is expanded with the preset but the page is missing.

markusguenther:
Puh ... seems that the presets are not that trivial stuff.

dimaip:
Thanks for thorough testing, guys!
I have ideas how to fix that. Will try tonight.

dimaip:
This is probably the most convoluted bugfix I ever had to do for the UI...

1. Merging of children was a really really bad idea here https://github.com/neos/neos-ui/pull/2788. I reverted it.
2. The only proper fix is to pass current preset with all content canvas requests. This is not exactly trivial, since we set the iframe src from multiple places, but it seems it generally enough to do it on tree navigation + inline navigation. Other places like reload of dimensions doesn't seem to be affected by this bug, so I left them intact.
3. There's one more bug that remains: if you toggle a preset, then click on a node that doesn't have Chapter nodes as direct children, all of its children would disappear. We need to adjust filtering in NodeInfo helper, thought technically it could go into a different PR.

Please test test test, this change is really tricky! /cc @DrillSergeant @markusguenther @bwaidelich

markusguenther:
Thank you for all the effort you put into that fix ❤️ 
Will check that out 🙇

DrillSergeant:
Thanks @dimaip for you work. I am in meetings all day, will check it out as soon as I can.

DrillSergeant:
Gooooooood job, I tested (about 30min) without any problem. Ok, one time I managed that a new node was not displayed anymore after some actions, but I cannot reproduce that now.

The current state seems pretty stable to me and even if there is hidden gremlin left it is a much better state than the current released version. So I would love to see this released.

Of course, if someone else could test as well, would not be a bad idea.

dimaip:
Thanks for testing!
Still I'd love to hear from @markusguenther and maybe @bwaidelich could help to sanity-check this?

markusguenther:
Screen-Recording-2020-10-12-at-1

Thanks again for working on that topic @dimaip and thanks to @DrillSergeant for the testing session.
I pulled the current state und rebuild my UI with `make clean && make setup` but still have some strange issue with loosing nodes when moving stuff around.

I can see that it is obviously a hard PR. I am happy that you tackle that Dmitri. But if I did nothing wrong in updating my test instance it is not working. I see that it is better than before @DrillSergeant but to be honest I am not sure how we proceed release wise.

dimaip:
Thanks for testing @markusguenther, I'll try to reproduce it tonight.

dimaip:
@markusguenther I tried to reproduce the issue by doing the exact same steps as you are doing, but I just couldn't... Are you absolutely sure have rebuild the latest code (`make build`) and you have `frontendDevelopmentMode: true`?

markusguenther:
I tested this again and I have still the issue. And I have frontendMode on and the branch is builded....
I have this also in multiple browsers and the cache has been cleared.

[Screen-Recording-2020-10-15-at-13-46-55-01.gif](https://postimg.cc/PP84kNrp)

As it is way better than before we should maybe merge it and can fix that later if it is not just me with the issue :)

dimaip:
Reproduced!
If I move Chapter 5 into 7 than 6 into 7 it works fine.
If I move first 6 into 7 then 5 into 7 I get your behaviour!
Will try to track it down.

markusguenther:
@dimaip So it is not just me and my computer that is mad to you :)

DrillSergeant:
@dimaip @markusguenther nice you can reproduce that now. This is what I mentioned:
> Ok, one time I managed that a new node was not displayed anymore after some actions, but I cannot reproduce that now.

dimaip:
So the steps to reproduce:

1. Toggle the tree preset
2. Move a Chapter node
3. Refresh the tree
4. Observe the node disappear. It won't re-appear even after reload.

It's a backend issue, a bug in `getNodeDataForParentAndNodeType` as far as I can tell: it doesn't find nodes that have been moved but not yet published, but it's really hard for me to dig through that CR code... Maybe some backend people could lend a hand? @bwaidelich, could you?

dimaip:
I'd say that this PR is safe to merge, the outstanding issue is totally unrelated to anything in the frontend code.

markusguenther:
I merged it now .. it is way better than before and I will create an issue for the remaining issue. So hope we can fix the last little step :)

Thanks again to Internezzo for the sponsoring and @dimaip for taking care ❤️

Issue #2776, opened 2020-09-14
Feature request: insertion-anchor with mode-preselection


DrillSergeant:
I often see that editors click on the +-icon to insert a content-node as child to a content-collection-node but they end up inserting not a child but a sibling to the clicked content node because they dont realize that the default insert-mode is "insert below".

We have quite a lot of content nodetypes that are also content-collections where the editor most of the time wants to insert a child  and not a sibling.

I wish there would be an option to set the default-mode (globally or even better per node type) so the insert dialog could look like this right after clicking the +-icon:
![image](https://user-images.githubusercontent.com/883552/93068512-dea5b200-f67c-11ea-8056-69211ac8ee33.png)

I could imagine to configure that via NodeTypes.yaml or as value for the data attribute `data-__neos-insertion-anchor`.

If anyone is up to implement this, I would be open to help sponsoring it.

dimaip:
I believe the default should be the insert-into mode if it's available. I'd just make it the default, instead of making it configurable (don't like having configs if it's avoidable).
Wdyt @bwaidelich @kitsunet?

bwaidelich:
> I believe the default should be the insert-into mode if it's available

+1

DrillSergeant:
I just tested with the demo-package. Here are the steps to reproduce the behaviour:

1. Create a content-nodetype that is also a content-collection:
```
'Neos.Demo:Content.Test':
  superTypes:
    'Neos.Neos:Content': true
    'Neos.Neos:ContentCollection': true
  ui:
    label: i18n
    icon: 'icon-file-text'
```
2. Fusion-Object for rendering (not needed to reproduce, just to make the example complete)
```
prototype(Neos.Demo:Content.Test) < prototype(Neos.Neos:ContentComponent) {

    content = Neos.Neos:ContentCollectionRenderer

    renderer = afx`
        <div>
            {props.content}
        </div>
    `
}
```
3. Insert the a content node of the new type somewhere
4. Click the +-icon of the inserted content-element

**Expected result**
You end up in the insert-dialog with mode = insert into

**Actual result**
You end up in the insert-dialog with mode = insert after

dimaip:
Makes sense, will adjust, but I doubt this can go in as a bugfix, it will target master.

DrillSergeant:
Would be fine with me

DrillSergeant:
[...]

PR #2797, opened 2020-10-12
FEATURE: set Into as the default insertion mode for the content tree

dimaip:
Resolves: https://github.com/neos/neos-ui/issues/2776

**What I did**

Set Into as the default insertion mode for the content tree.

**How to verify it**

Try to insert content and document nodes. Into should be the default position for content nodes, if it's allowed.

### DIFF
--- a/packages/neos-ui-redux-store/src/CR/Nodes/index.ts
+++ b/packages/neos-ui-redux-store/src/CR/Nodes/index.ts
@@ -143,7 +143,7 @@ const commenceRemovalMultiple = (contextPaths: NodeContextPath[]) => createActio
  * @param {String} preferredMode (optional) The default mode to use in the nodetype selection dialog. Currently not used withing the system but may be useful for extensibility.
  * @param {String} nodeType (optional) If set, then the select nodetype step would be skipped completely. Currently not used withing the system but may be useful for extensibility.
  */
-const commenceCreation = (referenceNodeContextPath: NodeContextPath, referenceNodeFusionPath: FusionPath, preferredMode: InsertPosition = InsertPosition.AFTER, nodeType: NodeTypeName | null = null) => createAction(actionTypes.COMMENCE_CREATION, {
+const commenceCreation = (referenceNodeContextPath: NodeContextPath, referenceNodeFusionPath: FusionPath, preferredMode: InsertPosition = InsertPosition.INTO, nodeType: NodeTypeName | null = null) => createAction(actionTypes.COMMENCE_CREATION, {
     referenceNodeContextPath,
     referenceNodeFusionPath,
     preferredMode,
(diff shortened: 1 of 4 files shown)


dimaip:
@DrillSergeant please give it a try!

Also @bwaidelich @markusguenther @Sebobo, do you think Into makes sense as the default mode for the page tree as well (the way I currently did it) or does it only make sense for the content tree?

markusguenther:
Sometimes it would be helpful to have that, but I dont edit that much content.
I knoe that you dislike to much configs but guess this would be a feature that would be good to have as configuration.

What do the others think. If there is no majority for a config I am also fine with changing the default to into.

dimaip:
@markusguenther haha yeah you are right, I try to resist configs as much as I can, they make the product bloated and hard to integrate...

Basically we have these choices:

a) both document and content trees have INTO as the default mode
b) document tree inserts AFTER by default, content tree INTO (it might be what people would actually want, but it would be inconsistent and confusing)
c) add a config

Wdyt people?

DrillSergeant:
**Content-Tree**
Tested, works like I expected it

**Page-Tree**
I like the consistent way of "insert into" as default. Feels good to me when I try to think like an editor.

**My preference (in order of which one I like most)**
1. a)
2. c)
3. b)

Each one is ok for me, as long as I can manage/configure to have "insert into" as default for the content-tree.

Sebobo:
@dimaip the fallback for "Insert to" is "Insert after" correct?
Because due to constraints that would often behave different.

Personally I prefer insert into in most cases. I wouldn't make it configurable.

dimaip:
@Sebobo yes!

dimaip:
Thanks for feedback and testing!
I'll wait till tomorrow evening and merge.

dimaip:
@bwaidelich I kinda feel the same, it's just that my command of English language is not so good to be able to put it into words so well as you :)

bwaidelich:
> it's just that my command of English language is not so good

I love your command of English! :)

BTW: Should we make this a feature and change the subject to something like:
```
!!! FEATURE: Change default insertion mode from AFTER to INSERT for collection nodes
```

dimaip:
Why `!!!`? It's not breaking any apis...

Do we all agree to do as @bwaidelich suggests (b)? /cc @DrillSergeant @markusguenther @Sebobo 
If yes, I'll get to coding.

Sebobo:
> Why `!!!`? It's not breaking any apis..

It breaks the editors minds 🤪

bwaidelich:
> It's not breaking any apis...

It's "breaking" behavior (in making it more reasonable).. But, you're right, it's an edge case

Sebobo:
I agree with @bwaidelich, gogogo

DrillSergeant:
Ok for me

markusguenther:
I am also fine with Bastians suggestion :) 
Thanks for bringing that up 👍

dimaip:
Done! Please review!

markusguenther (approved):
Thanks @dimaip for bringing that up.
Tested that now and it works nice.

Only downside is that you also get the default INSERT when you push the add button in the Content Tree and have the Page node selected.
But guess that is not that important and more an edge case.

dimaip:
Lovely! Thanks everyone for active and agile participation!

PR #2799, opened 2020-10-17
BUGFIX: Allow dropdown contents to break out of overflow-restricted areas

### LINKED ISSUE #1214: Options of the select box editor in the node creation dialog should visible

grebaldi:
Hi there!

This PR enhances the DropDown component in two ways:

1. It allows drop down contents to break out of overflow-restricted areas like the content creation dialog.
2. It positions dropdown contents above the dropdown header, if there's not enough space below it.

All of the above are only effective if the DropDown is marked as `scrollable`. Otherwise, it is assumed that the call-site of the DropDown component is aware of any screen-related size-restrictions and reserves enough space below wherever the DropDown is placed.

|**Example:** Creation Dialog, breakout|**Example:** Creation Dialog, positioning|**Example:** Inspector|
|-|-|-|
|Peek 2020-10-17 16-40|Peek 2020-10-17 16-41|Peek 2020-10-17 16-46|

## How is this done?

If scrollable, the `ShallowDropDownContents` component gets a fixed positioning and ist rendered directly to the Host frames `document.body` via `React.createPortal`. This way, it is no longer subject to any `overflow: ...`-Restriction within the ancestry of its parent DropDown.

To ensure the correct positioning of `ShallowDropDownContents`, its screen position and dimensions are calculated relative to its closest `DropDownWrapper` component. `DropDownWrapper` provides a DOM reference of itself via the React Context API. Layout recalculation happens on mount, `scroll` on `document.body` and `resize` on `window`. Recalculation is throttled via `requestAnimationFrame`.

Additionally, whenever `DropDownWrapper` is not visible whithin its closest scrollable element, `ShallowDropDownContents` is hidden to avoid interference with other interactive elements.

## Further notes

<s>Unfortunately, I couldn't figure out how to get e2e-Tests to run locally. And since this feature cannot be reliably tested in a `jsdom`-environment alone, I didn't add any tests at all.</s>

<s>If that's a deal-breaker for this PR, any help on the testing part would be greatly appreciated :)</s>

<s>**Edit:** Since e2e-Tests are red in CircleCI and I couldn't reproduce the error that is logged, I'll continue to figure out how to run them locally. Once that's done, I'll probably be able to add some tests for this feature as well.</s>

**Edit:** I figured it out and added some tests :) But I remain open to suggestions as to how this might be tested more thoroughly. As of right now, the tests cover only some rough positioning logic and more could be done, but Testcafe has its limits...

## TODOs

- [x] Figure out how to run e2e-Tests locally
- [x] Fix issue with DimensionSwitcher
- [x] Add e2e-Tests for new DropDown behavior

---

fixes #1214

### DIFF
--- a/packages/react-ui-components/src/DropDown/contents.tsx
+++ b/packages/react-ui-components/src/DropDown/contents.tsx
@@ -1,5 +1,48 @@
 import React, {PureComponent, ReactNode} from 'react';
+import ReactDOM from 'react-dom';
 import mergeClassNames from 'classnames';
+
+/**
+ * Helper function to find the closest ancestor element that currently
+ * has a scrollbar and could possibly cause the given element not to
+ * be visible on screen.
+ *
+ * @param HTMLElement el
+ * @return HTMLElement
+ */
+function getScrollContainer(el: HTMLElement): HTMLElement {
+    if (el.scrollHeight > el.clientHeight) {
+        const { overflowY } = getComputedStyle(el);
+
+        if (overflowY === 'auto' || overflowY === 'scroll') {
+            return el;
+        }
+    }
+
+    if (el.parentElement) {
+        return getScrollContainer(el.parentElement);
+    } else {
+        return document.body;
+    }
+}
+
+/**
+ * Helper function to determine whether the given element is visible
+ * within its ancestor scroll context.
+ *
+ * @param HTMLElement el
+ * @return boolean
+ */
+function isInScrollView(el: HTMLElement): boolean {
+    const scrollContainer = getScrollContainer(el);
+    const elementBoundingBox = el.getBoundingClientRect();
+    const scrollContainerBoundingBox = scrollContainer.getBoundingClientRect();
+
+    return (
+        elementBoundingBox.top >= scrollContainerBoundingBox.top &&
+        elementBoundingBox.bottom <= scrollContainerBoundingBox.bottom
+    );
+}
 
 interface ShallowDropDownContentsTheme {
     readonly 'dropDown__contents': string;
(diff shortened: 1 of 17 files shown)


Sebobo:
Hi @grebaldi,

awesome change thx!
As it's marked bugfix, did you want to target master / 6.0 or another branch?

markusguenther (approved):
Awesome @grebaldi ... thanks for providing such a nice PR. Welcome back after quite some time. While testing I spotted that a fusion for OpensAboveIn was missing so it could not be rendered in the backend. But added that.

Really nice solution.

patricekaufmann:
If it gets rendered to the host frame's document body, it will be affected by the site stylesheets, correct? E.g. any generic `ul, li, div, etc...` rules might apply and break the layout. A solution for this would be to use the Shadow DOM, although it would require self contained css on the dropdown.

packages/react-ui-components/src/DropDown/contents.tsx
```diff
@@ -1,6 +1,49 @@
 import React, {PureComponent, ReactNode} from 'react';
+import ReactDOM from 'react-dom';
 import mergeClassNames from 'classnames';
 
+/**
+ * Helper function to find the closest ancestor element that currently
+ * has a scrollbar and could possibly cause the given element not to
+ * be visible on screen.
+ *
+ * @param HTMLElement el
```
dimaip, line 10: I'm not sure there's any use in jsdoc in TS files

dimaip (approved):
Impressive!
So good to see a PR from you again, you definitely set high standards for the rest of us here! ❤️

dimaip:
> @patricekaufmann If it gets rendered to the host frame's document body, it will be affected by the site stylesheets, correct?

There's no such thing as host frame in our terminology. We a have a host document (i.e. the UI) and a guest iframe with the website rendered into it.
The SelectBox and most other UI (besides inline buttons) are rendered within the host document, so they are not in any way affected by the guest styles.

markusguenther:
> As it's marked bugfix, did you want to target master / 6.0 or another branch?

Maybe 5.3 as it is less hassle. For 4.0 the end to end stuff needs to be adjusted and maybe other things too.
But would be nice to have that also for the latest LTS release.

dimaip:
Oh yeah, I totally missed the fact that it was marked a bugfix. It looks more than a simple bugfix, so I assumed it's ok to aim master with it, but yeah, if we are labelling it as a bugfix I guess it should go to 4.0. If there's too much work rebasing it, than at least to 5.3 as you suggest.

markusguenther:
Closed in favour of https://github.com/neos/neos-ui/pull/2854
Thanks for that awesome fix @grebaldi ❤️

PR #2801, opened 2020-10-19
TASK: Upgrade slugify to version 1.1.0

markusguenther:
**What I did**
Upgraded slugify to the latest version

**How to verify it**

Change the page title to a for instance Cyrillic name and update the uri path segment.
We also have a test for that in out end to end tests ;)

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "@fortawesome/react-fontawesome": "^0.1.4",
     "@friendsofreactjs/react-css-themr": "^3.3.2",
     "@neos-project/brand": "^1.1.0",
-    "@sindresorhus/slugify": "^0.9.1",
+    "@sindresorhus/slugify": "^1.1.0",
     "amator": "^1.0.1",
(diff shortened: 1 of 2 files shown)


markusguenther:
Tests are fine ... the styleCI stuff is unrelated.

dimaip:
Could you please apply the suggested patch? We need to get the tests back to green.

markusguenther:
> Could you please apply the suggested patch? We need to get the tests back to green.

Hi @dimaip what do you mean with suggested patch? The StyleCI stuff? To be honest don't get it why they always change their base and suggest changes you did not touch in your PR.

dimaip:
@markusguenther yeah, the styleci one.
I don't know either, but it's the way it works, it just spontaneously complains about untouched parts of code for whatever reason...
So we either remove styleci altogether or keep on applying their patches when they popup. I'm against just ignoring them.

markusguenther:
With the updated and cleaned up CI pipeline this is now green again ;)

Issue #2798, opened 2020-10-14
Remove classes marked as deprecated and can be removed by 6.0


markusguenther:
### Description

We have a upcomming major release in some weeks and therefore we need to check which deprecations we need to remove.
Sadly in the ui we did not deprecate that much stuff and also did not tag it to versions.

So we have to check history when the marked deprecations has been added.

### TODOs

- [x] Remove the legacy aloha configuration handling (#2802)
- [x] Remove deprecated isDisabled Button prop (#2803)
- [x] Make the content UI a backend module with its own module privilege (#2793)
- [x] Remove legacy fusion code of the old emberjs content module (#2849)

markusguenther:
@bwaidelich Guess it is too early to just support the  constraints.mediaTypes right. So it is more a thing for the 7.0 right?

markusguenther:
The PR from @daniellienert also removes some deprecations :)
https://github.com/neos/neos-ui/pull/2793

bwaidelich:
@markusguenther since the todos seem to be done, can this be closed?

PR #2802, opened 2020-10-19
!!! TASK:  Remove the legacy aloha configuration handling

markusguenther:
This is a breaking task. So you need to transform your configuration to the current editorOptions. We also offer a core migration for that (Neos.Neos.Ui-20190319094900).

Relates: #2798


dimaip (approved):
Oh wow, I guess we have to do it, but it's going to explode most of legacy installations I bet!

markusguenther:
@dimaip But as we have core migrations the adjustment should not be so hard :)
We need to remove it at any point. And the next is version 7.0 ;)

Issue #2805, opened 2020-10-20
JS error when using NodeType filter


bwaidelich:
### Description

When using the NodeType filter with the latest Neos UI version the following JS error occurs when trying to select any node:
```
[...] urlAppendParams is not a function
```

### Background

This probably happened while upmerging #2794 
This change introduced a new Helper method [urlAppendParams](https://github.com/neos/neos-ui/blob/ed9e04cdcaf5774914a60d866a8589709b4704f8/packages/neos-ui-backend-connector/src/Endpoints/Helpers.ts#L55-L63) that no longer exists in 5.3

And, since URL related helpers have been moved to a dedicated file with 6e4b39c7309c4b2eef351b79ea8a03899571ecb9 we can't easily re-add it unfortunately.

### Affected Versions

UI: 5.3.2

Related: #1691

markusguenther:
Will be fixed in 5.3.3 🥇

PR #2806, opened 2020-10-20
BUGFIX: Fix NodeType filter regression

bwaidelich:
Fixes a regression introduced with upmerging #2794 since
the new helper `urlWithParams` wasn't added.

Note:

Instead of adding the helper to the endpoint helpers, this
extends the already existing `urlWithParams` helper such that
it keeps existing query params.

Fixes: #2805

### DIFF
--- a/packages/utils-helpers/src/urlWithParams.ts
+++ b/packages/utils-helpers/src/urlWithParams.ts
@@ -3,7 +3,11 @@
 //
 export function urlWithParams(urlString: string, params: any = {}): string {
     const queryString = encodeAsQueryString(params);
-    return urlString + (queryString === '' ? '' : '?' + queryString);
+    if (queryString === '') {
+        return urlString;
+    }
+    const url = new URL(urlString);
+    return urlString + (url.search === '' ? '?' : '&') + queryString;
 }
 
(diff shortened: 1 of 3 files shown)


packages/utils-helpers/src/urlWithParams.ts
```diff
@@ -3,7 +3,11 @@
 //
 export function urlWithParams(urlString: string, params: any = {}): string {
     const queryString = encodeAsQueryString(params);
-    return urlString + (queryString === '' ? '' : '?' + queryString);
+    if (queryString === '') {
+        return urlString;
+    }
+    const url = new URL(urlString);
+    return urlString + (url.search === '' ? '?' : '&') + queryString;
```
bwaidelich, line 10: Note: This is just a very basic extension. It won't merge all params.
i.e. `urlWithParams('http://neos.io?foo=bar', {foo: 'baz'})` would return `http://neos.io?foo=bar&foo=baz`.

If anyone feels like a proper implementation that supports PHP-compatible query strings and merges the params recursively.. feel free :)

markusguenther (approved):
Thanks for finxing and taking care @bwaidelich 🙇  ❤️

dimaip (approved):
Thanks for the quick fix!

Issue #2807, opened 2020-10-21
Exception when clicking the preview button without a preset


markusguenther:
### Description

When you click the preview button and have no node type filter enabled the rendered page will show an exception that the presetBaseNodeType is missing.

### Steps to Reproduce

1. Go to content Module in the backend
2. Select a page in the PageTree (without a filter!)
3. Click the preview Button

#### Expected behavior

The page will be rendered.

#### Actual behavior

Exception

![Screenshot 2020-10-21 at 09 07 35](https://user-images.githubusercontent.com/1014126/96685383-50100900-137d-11eb-88ce-aed58b6c156b.png)

### Affected Versions

Neos: 5.1

UI:5.1.11

DrillSergeant:
Should this be gone with UI 5.3.3? I still have it using v5.3.3.

markusguenther:
It is gone with 5.3.4 just released ;)

PR #2808, opened 2020-10-21
BUGFIX: Prevent exception with missing presetBaseNodeType

markusguenther:
Make the argument presetBaseNodeType optional. In higher branches that parameter was also missing in the method arguments of redirectToAction.

Thanks to Paavo Limbo for reporting the issue, we appreciate it!

Fixes: #2807

### DIFF
--- a/Classes/Controller/BackendController.php
+++ b/Classes/Controller/BackendController.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace Neos\Neos\Ui\Controller;
 
 /*
@@ -189,13 +190,14 @@ class BackendController extends ActionController
      * @param string $presetBaseNodeType
      * @throws StopActionException
      */
-    public function redirectToAction(NodeInterface $node, string $presetBaseNodeType = null)
+    public function redirectToAction(NodeInterface $node, string $presetBaseNodeType = '')
     {
         $this->response->setComponentParameter(SetHeaderComponent::class, 'Cache-Control', [
             'no-cache',
             'no-store'
         ]);
-        $this->redirectToUri($this->linkingService->createNodeUri($this->controllerContext, $node, null, null, false, ['presetBaseNodeType' => $presetBaseNodeType]));
+        $arguments = !empty($presetBaseNodeType) ? ['presetBaseNodeType' => $presetBaseNodeType] : [];
+        $this->redirectToUri($this->linkingService->createNodeUri($this->controllerContext, $node, null, null, false, $arguments));
     }
 
     /**


markusguenther:
For neos-ui 5.2 it will look more like that.

```
/**
     * @param NodeInterface $node
     * @param string $presetBaseNodeType
     * @throws MissingActionNameException
     * @throws StopActionException
     * @throws UnsupportedRequestTypeException
     * @throws \Neos\Flow\Http\Exception
     * @throws \Neos\Flow\Persistence\Exception\IllegalObjectTypeException
     * @throws \Neos\Flow\Property\Exception
     * @throws \Neos\Flow\Security\Exception
     * @throws \Neos\Neos\Exception
     */
    public function redirectToAction(NodeInterface $node, string $presetBaseNodeType = ''): void
    {
        $this->response->setComponentParameter(SetHeaderComponent::class, 'Cache-Control', [
            'no-cache',
            'no-store'
        ]);
        $arguments = !empty($presetBaseNodeType) ? ['presetBaseNodeType' => $presetBaseNodeType] : [];
        $this->redirectToUri($this->linkingService->createNodeUri($this->controllerContext, $node, null, null, false, $arguments));
    }
```

The upmerge mess this up and since 5.2 the parameter was missing and that lead to the exception.
But to be honest ... I don`t know how the parameter should get a value?

Did that work in 4.3 @dimaip ???

markusguenther:
If not we could maybe just remove that part of the code... I don`t see that it has an impact for the preview link. Maybe somewhere else.

bwaidelich (dismissed):

bwaidelich:
@markusguenther is this bug really present in `5.1`
https://github.com/neos/neos-ui/blob/5.2/Classes/Controller/BackendController.php#L202-L208 is obviously broken, but IMO we just need to fix the method signature according to 5.1

bwaidelich:
As discussed with @markusguenther : I'll create a new PR against `5.2` with the minimal fix for now.
As a follow-up we should check whether this argument really used at all

bwaidelich:
The minimal invasive version as discussed: #2809 

I would suggest to merge #2809, close this one and investigate if we can get rid of the action parameter alltoghether for 4.x+

dimaip:
@markusguenther I don't think I tested the redirect, too bad we didn't have a test for it :(
Should this be closed now?

markusguenther:
@dimaip Guess we can close this PR but can discuss if the parameter in the redirect is really needed because I did not find the place where it is used. So the question is more why you added that ;)

dimaip:
@markusguenther I don't get the question. The `presetBaseNodeType` has to be passed with the redirect. It's used from Fusion to render correct child node metadata.

markusguenther:
@dimaip Thanks for the offer ... maybe later. I just did not find the part where the parameter is used. And when I debugged it, it was always empty. So I asked myself yesterday if we can remove the parameter, But if this is used somewhere it is fine :)

dimaip:
@markusguenther it's used here https://github.com/neos/neos-ui/blob/04ccc1fb176b88df504e1ef1a6b9ee1da6da2601/Classes/Fusion/Helper/NodeInfoHelper.php#L141

dimaip:
Closing this one, but I just noticed that 5.0 is broken: https://github.com/neos/neos-ui/pull/2810

PR #2810, opened 2020-10-22
BUGFIX: pass presetBaseNodeType when redirecting

dimaip:
I think the upmerge went wrong and we need this.


bwaidelich (approved):
I don't really understand what went wrong here, but this makes sense obviously

dimaip:
@bwaidelich upmerge gone wrong obviously, but I'm not sure why, I wasn't the one who was doing it https://github.com/neos/neos-ui/commit/9e8db31d3ad5fe0039b50634ceee88450cec0a03

PR #2812, opened 2020-10-30
TASK: Adapt circle ci config to use nvm node version

markusguenther:
Circle CI now uses node version 14 instead of our version 12 and fails on switchingh the version.


markusguenther:
For the latest version I now rised the node version to the current LTS of node as it makes sense.
For the older branches I will dive deeper to solve that :)

https://github.com/neos/neos-ui/pull/2816

PR #2817, opened 2020-11-02
TASK: Make testing distribution work with more neos installations

markusguenther:
It can happen that you have a neos distribution that uses the neos nodetypes and want to execute the neos-ui e2e tests. Then the own node types are not used because they have the same name as the basic neos nodetypes.

**What I did**
Add the _Test to the name of the nodetypes.

**How to verify it**
Run the tests


markusguenther:
Will rebase that when https://github.com/neos/neos-ui/pull/2816 is merged... then the checkout of the circle CI is working again.

markusguenther:
I missused this PR a bit to change the Circle CI config for master. I base on a new docker image with php 7.4 and need to bring it back to life ;)

markusguenther:
@dimaip @Sebobo What do you think about that? It does not hurt and helps when you use the end to end tests with a real neos distribution.

dimaip (approved):
Awesome change!

markusguenther:
@dimaip great that you like it :)

Issue #2595, opened 2019-11-28
Neos-UI does not support language label groups for plurals


markusguenther:
### Description

Sometimes you are using language labels with arguments and it can happen that you need a singular and plural of a word depending on the argument. For instance for the amout of errors.

XLIFF supports that and the Flow translator as well. But when you want to use that labels in the JS part of the neos-ui you have no possibility to get the plural. You always get the singular if this is the first value in the group.

```xml
<group id="validationErrorTooltip" restype="x-gettext-plurals">
    <trans-unit id="validationErrorTooltip[0]" xml:space="preserve">
        <source>{tabName} – {amountOfErrors} validation issue</source>
    </trans-unit>
    <trans-unit id="validationErrorTooltip[1]" xml:space="preserve">
        <source>{tabName} – {amountOfErrors} validation issues</source>
    </trans-unit>
</group>
```

```javascript
const notificationTooltipLabelPieces = i18nRegistry.translate(
    'validationErrorTooltip',
    '',
    {
        tabName: tabLabel,
        amountOfErrors: number
    },
    'Neos.Neos.Ui'
);
```

So in this case the label will be in the UI always `Our Tab – 2 validation issue` instead of `Our Tab – 2 validation issues`.

Would like to change that and have the plural also in the UI. 

### Steps to Reproduce

1. Create group labels like in the description
2. use the label with i18nRegistry.translate
3. Label will be always the first label in the group

#### Expected behavior

Label should be `Our Tab – 2 validation issues` when using error number higher than one.

#### Actual behavior

Label is singular `Our Tab – 2 validation issue`

### Affected Versions

Neos: all

UI: all since ever as the code comes from the old Emberjs UI

markusguenther:
It turns out that it is not just an neos-ui issue. The UI consumes the language labels from the xliff endpoint via json. The JSON has been build with the XliffService of teh Neos.Neos package an this service also does not respect the plural parameter.

```php
foreach ($file->getTranslationUnits() as $key => $value) {
    $valueToStore = !empty($value[0]['target']) ? $value[0]['target'] : $value[0]['source'];
    if ($this->scrambleTranslatedLabels) {
        $valueToStore = str_repeat('#', UnicodeFunctions::strlen($valueToStore));
    }
    $this->setArrayDataValue($labels, str_replace('.', '_', $packageKey) . '.' . str_replace('/', '_', $sourceName) . '.' . str_replace('.', '_', $key), $valueToStore);
}```

[...]

PR #2827, opened 2020-11-09
FEATURE: Use language plurals in I18n registry 

markusguenther:
Sometimes you are using language labels with arguments and it can happen that you need a singular and plural of a word depending on the argument. For instance for the amout of errors.

XLIFF supports that and the flow translator as well. But when you want to use that labels in the JS part of the neos-ui you have no possibility to get the plural. You always get the singular if this is the first value in the group.

The following PR extended the `XliffService` to create a response that also respects plurals, so that the UI is able to consume the plural labels as well. https://github.com/neos/neos-development-collection/issues/2786

This change is not breaking as it uses singular labels when no plural exists. Take a look at the unit tests for instance.

**What I did**

Added a quantity parameter to the translate function of the I18n registry. With that quantity parameter you are able to fetch the plural of a label. If the label has no plural available we use the existing singular.

**How to verify it**

For instance try the tabs validation in the inspector.
In the old version the inspector tooltip was always singular and we just replaced the amount of issues.

Screenshot 2020-11-09 at 12 28 53

As we are now able to consume the plurals we can render a better matching label :)

Screenshot 2020-11-09 at 12 25 36

Resolves: #2595

### DIFF
--- a/packages/neos-ui-i18n/src/registry/I18nRegistry.js
+++ b/packages/neos-ui-i18n/src/registry/I18nRegistry.js
@@ -59,6 +59,21 @@ const substitutePlaceholders = function (textWithPlaceholders, parameters) {
     return result.join('');
 };
 
+const getPluralForm = (translation, quantity = 0) => {
+    const translationHasPlurals = translation instanceof Object;
+
+    // no defined quantity or less than one returns singular
+    if (translationHasPlurals && (!quantity || quantity <= 1)) {
+        return translation[0];
+    }
+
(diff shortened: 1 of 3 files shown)


markusguenther:
The change also works without the PR https://github.com/neos/neos-development-collection/issues/2786 but then you will just get the singular ;)

kdambekalns (approved):
Looks good by reading (if you trust my JS skills)

markusguenther:
@kdambekalns I trust you :)

PR #2831, opened 2020-11-16
TASK: Hotfix to replace node[__identity] in preview uri

markusguenther:
This is just a temporary fix. We should remove this when the routing handling is fixed.

**What I did**
The backend was currently broken because the preview uri contained `[__identity]` .. this is a kind of a dirty quickfix and needs to be removed afterwards.

**How I did it**
relaced node[__identity] with node in the created preview URI

**How to verify it**
Just open the backend ;)

Relates: https://github.com/neos/neos-development-collection/pull/3183

### DIFF
--- a/Classes/Fusion/Helper/NodeInfoHelper.php
+++ b/Classes/Fusion/Helper/NodeInfoHelper.php
@@ -413,7 +413,10 @@ class NodeInfoHelper implements ProtectedContextAwareInterface
         }
 
         // Create an absolute URI without resolving shortcuts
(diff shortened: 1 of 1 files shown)


kdambekalns (changes requested):
Fix should be like in https://github.com/neos/neos-development-collection/pull/3183

bwaidelich:
Thanks for this.
I can confirm that Karstens [PR](https://github.com/neos/neos-development-collection/pull/3183) will fix this (once the dependencies are resolved)

markusguenther:
Awesome that we found the right place to fix that :)

Issue #2828, opened 2020-11-12
Validation is no longer enforced for inline editable properties


bwaidelich:
### Description

Since #2538 validation could be activated for inline editable properties.
Due to some regression this no longer works with Neos 5.0+

### Steps to Reproduce

Adjust the headline node type definition in the demo site:

```yaml
'Neos.Demo:Content.Headline':
  properties:
    'title':
      # ...
      validation:
        'Neos.Neos/Validation/StringLengthValidator':
          minimum: 1
          maximum: 5
```

Open up the homepage (or any other page that contains a headline node) and enter a headline that exceeds 5 characters

#### Expected behavior

The inline editable should get a red border & display the validation error (see #2538).
It should be impossible to persist / publish the invalid state.

#### Actual behavior

The validation is completely ignored.

### Affected Versions

Neos: 5.0+
UI: 5.1+ (probably also 5.0)

markusguenther:
The neos.ui has a automated test for that and has the nodetype headline.

```
'Neos.TestNodeTypes:Content.Headline':
  superTypes:
    'Neos.Neos:Content': true
  ui:
    label: Headline_Test
    icon: icon-header
    position: 200
    inlineEditable: true
    help:
      message: 'Markdown **test**.'
  properties:
    title:
      type: string
      defaultValue: '<h1>Enter headline here</h1>'
      ui:
        inlineEditable: true
        inline:
          editorOptions:
            placeholder: '<h2>Enter headline here...</h2>'
            formatting:
              h1: true
              h2: true
              h3: true
              h4: true
              h5: true
              a: true
      validation:
        'Neos.Neos/Validation/NotEmptyValidator': []
        'Neos.Neos/Validation/StringLengthValidator':
          minimum: 1
          maximum: 255
```

This leads to the following result.

![Uploading Screenshot 2020-11-16 at 21.18.05.png…]()

markusguenther:
The issue was bit tricky and the error can have multiple reasons.


1. The validators are stored for each node so when the Validators has been changed the existing nodes still use the old config.
2. When you have a validator with just stringLength and expect a minimum of 1 it does not work and you need additionally the NotEmpty validator.
3. The StringLength validator with a pretty short maximum will often fail because the inline elements have values with markup. So the HTML tags blow that up.

The PR now strips the tags off and that resolves the maximum validation, but for the minimum of 1 we need a NotEmpty validator as well.

dimaip:
Thanks so much for digging into this Markus!

[...]

PR #2832, opened 2020-11-16
BUGFIX: Remove tags from to validated content 

markusguenther:
The StringLength validator will fail when the minimum value is pretty small.
The reason is that the value contains the HTML tags as well. So we now strip the tags from the to be validated value.

Fixes: #2828

### DIFF
--- a/packages/neos-ui-guest-frame/src/initializePropertyDomNode.js
+++ b/packages/neos-ui-guest-frame/src/initializePropertyDomNode.js
@@ -2,6 +2,7 @@ import {$get, $contains} from 'plow-js';
 
 import {actions} from '@neos-project/neos-ui-redux-store';
 import {validateElement} from '@neos-project/neos-ui-validators';
+import {stripTags} from '@neos-project/utils-helpers';
 
 import {getGuestFrameWindow, closestContextPathInGuestFrame} from './dom';
 
@@ -67,7 +68,8 @@ export default ({store, globalRegistry, nodeTypesRegistry, inlineEditorRegistry,
                         actions.Changes.persistChanges([change])
                     ),
                     onChange: value => {
-                        const validationResult = validateElement(value, $get(['properties', propertyName], nodeType), globalRegistry.get('validators'));
+                        const purgedValue = stripTags(value);
+                        const validationResult = validateElement(purgedValue, $get(['properties', propertyName], nodeType), globalRegistry.get('validators'));
                         // Update inline validation errors
                         store.dispatch(
                             actions.CR.Nodes.setInlineValidationErrors(contextPath, propertyName, validationResult)


Classes/Fusion/Helper/NodeInfoHelper.php
```diff
@@ -413,7 +413,10 @@ public function uri(NodeInterface $node = null, ControllerContext $controllerCon
         }
 
         // Create an absolute URI without resolving shortcuts
-        return $this->linkingService->createNodeUri($controllerContext, $node, null, null, true, [], '', false, [], false);
+        $uri = urldecode($this->linkingService->createNodeUri($controllerContext, $node, null, null, true));
+        // @todo remove this hack when the uri is again without the [__identity] part.
+        $uri = str_replace('node[__identity]=', 'node=', $uri);
```
dimaip, line 418: I don't understand this code

markusguenther: The current master was broken and the uri contains node[__indentity]= instead of node= ... this is just a QuickFix to make the UI work again. Will discuss the better solution with @bwaidelich or maybe @kdambekalns today.

kdambekalns: I am confused, too… this fix has nothing to do with validation and I'd never ever look for it in this PR. Which is bad, since researching the past is about a quarter of software development… This must go in a separate PR, IMHO.

kdambekalns (changes requested):

Classes/Fusion/Helper/NodeInfoHelper.php
markusguenther: Ah now I understand you... this is just because my PR was based on the branch for the quickfix. Otherwise all test will fail because master is broken. Hope that we find a substainable solution for master soonish and then I just need to rebase the PR.

markusguenther: So this comes from https://github.com/neos/neos-ui/pull/2831

markusguenther: Saw that the issue was for neos 5.1. So I now changed the base branch and the discussion is resolved anyways ;)

kdambekalns (approved):
Looks good by reading, consistent with the findings in the issue.

bwaidelich (changes requested):
Just to prevent that this is merged. I'll comment #2828

markusguenther:
validation

markusguenther:
It turns out that there is no bug. For further information check the related issue ;)

bwaidelich:
As mentioned in #2828 we already allow HTML to be stripped via #2579 
usage:
```yaml
      validation:
        'Neos.Neos/Validation/StringLengthValidator':
          maximum: 10
          stripHtml: true
```

markusguenther:
> As mentioned in #2828 we already allow HTML to be stripped via #2579 
> 
> usage:
> 
> ```yaml
> 
>       validation:
> 
>         'Neos.Neos/Validation/StringLengthValidator':
> 
>           maximum: 10
> 
>           stripHtml: true
> 
> ```

ignoreHtml ;)

dimaip:
Nevertheless thanks for your efforts here to get to the bottom of this @markusguenther ❤️

bwaidelich:
> ignoreHtml ;)

Ah, good catch, fixed the example

> Nevertheless thanks for your efforts here to get to the bottom of this @markusguenther 

+100

Issue #2824, opened 2020-11-06
UriPathSegmentEditor ignores disabled & readonly flag


bwaidelich:
### Description

The `UriPathSegmentEditor` that was introduced with #2574 ignores the `disabled` and `readonly` [editorOptions](https://neos.readthedocs.io/en/stable/References/PropertyEditorReference.html#property-type-string-textfieldeditor-single-line-text-editor-default)

### Steps to Reproduce

1. Disable the `uriPathSegment` property via `NodeTypes.yaml`:

```yaml
'Neos.Neos:Document':
  properties:
    'uriPathSegment':
      ui:
        inspector:
          editorOptions:
            disabled: true
            # or:
            #readonly: true
```

2. Enter the Neos backend and open up the property inspector for the `uriPathSegment` property of a corresponding document node

#### Expected behavior

Text field for the uriPathSegment cannot be altered via direct input nor via the new synchronize button

#### Actual behavior

The textfield input is protected but it can be changed via the button:

![image](https://user-images.githubusercontent.com/307571/98369060-a4cba900-2038-11eb-9bb6-b25df22c0ee1.png)

![image](https://user-images.githubusercontent.com/307571/98369067-a85f3000-2038-11eb-9b5d-8bca72b0db0a.png)

### Affected Versions

UI: 5.1.0+

PR #2833, opened 2020-11-16
BUGFIX: Add disabled property to sync button

markusguenther:
The sync button did not respect the editorOption disabled and therefore the uriPathSegment could be updated even when the property was disabled.

Fixes: #2824

### DIFF
--- a/packages/neos-ui-editors/src/Editors/UriPathSegment/index.js
+++ b/packages/neos-ui-editors/src/Editors/UriPathSegment/index.js
@@ -75,7 +75,7 @@ export default class UriPathSegment extends PureComponent {
                         id="neos-UriPathSegmentEditor-sync"
                         size="regular"
                         icon="sync"
-                        disabled={finalOptions.disabled}
+                        disabled={finalOptions.disabled || finalOptions.readonly}
                         onClick={() => commit(slug)}
                         className={style.syncButton}
                         style="neutral"
--- a/packages/neos-ui-editors/src/Editors/UriPathSegment/index.js
+++ b/packages/neos-ui-editors/src/Editors/UriPathSegment/index.js
@@ -12,6 +12,24 @@ const defaultOptions = {
     maxlength: null,
     readonly: false
 };
+
+const SynchronizeButton = (title, slug, commit) => {
+    return (
+        <div style={{flexGrow: 0}}>
+            <IconButton
+                id="neos-UriPathSegmentEditor-sync"
+                size="regular"
+                icon="sync"
+                onClick={() => commit(slug)}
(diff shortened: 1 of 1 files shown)


markusguenther:
**Readonly:**
Screenshot 2020-11-16 at 22 40 19

**Disabled:**
Screenshot 2020-11-16 at 22 40 37

markusguenther:
The vulnerability check will be resolved in the PR https://github.com/neos/neos-ui/pull/2834

packages/neos-ui-editors/src/Editors/UriPathSegment/index.js
```diff
@@ -75,6 +75,7 @@ export default class UriPathSegment extends PureComponent {
                         id="neos-UriPathSegmentEditor-sync"
                         size="regular"
                         icon="sync"
+                        disabled={finalOptions.disabled || finalOptions.readonly}
```
bwaidelich, line 78: Ha, that was easy. And we love easy, don't we! ;)

I didn't test this yet, but I'm wondering: Would it make sense to hide the icon completely if the editor is read only?
Not sure about it and this fix would definitely improve the situation already of course!

Sebobo (approved):

packages/neos-ui-editors/src/Editors/UriPathSegment/index.js
markusguenther: We can also hide it.

Sebobo: I'm fine with both.

bwaidelich: Me too. That was just a brainwave, both make sense as long as it's easy to tell that the icon/button is disabled. Is that the case? (can't test it right now)

markusguenther: Removing the button is also easy ... 
Question is remove button as well for disabled as it is more consitent?

bwaidelich: > remove button as well for disabled as it is more consitent?

Yes, with "read only" above I meant all the cases where the button doesn't have an effect

markusguenther:
**Read only:** 
Screenshot 2020-11-17 at 20 10 03

**Disabled:**
Screenshot 2020-11-17 at 20 10 31

**Regular:**
Screenshot 2020-11-17 at 20 12 07

markusguenther:
After talking with @Sebobo I moved the synchronize button back to the class itself instead of an external component.

PR #2848, opened 2020-11-19
TASK: Rename neos backend container

markusguenther:
Rename the container from `neos-new-backend-container` to `neos-backend-container`. The emberjs content module has been removed for 5.0 and now we rename the backend container as we have no old and new one anymore.

This can cause issues when the third party code rely on the neos-new-backend-container id.

Related: #2828
Fixes: #2849


bwaidelich (approved):
+1 by reading

jonnitto (approved):
Fine, by reading

PR #2850, opened 2020-11-20
BUGFIX: Synchronize language labels

markusguenther:
The used language IDs were not in sync and therefore it could happen that the ui could not find a language label. Especially when you have used not english or german as backend language. That lead to a JS error for some users.

**What I did**
Streamline the indention of the language files and synchronize the existing language labels with the english base language.

Fixes: #2819


markusguenther:
Closed in favour of #2851

PR #2851, opened 2020-11-20
BUGFIX: Synchronize language labels

dogawaf:
The used language IDs were not in sync and therefore it could happen that the ui could not find a language label. Especially when you have used not english or german as backend language. That lead to a JS error for some users.

What we did
Streamline the indentation of the language files and synchronize the existing language labels with the english base language.
Also update the french translation

Fixes: #2819

Supersedes #2850

### DIFF
--- a/Resources/Private/Translations/af/Main.xlf
+++ b/Resources/Private/Translations/af/Main.xlf
@@ -1,267 +1,345 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" version="1.2">
-  <file original="" product-name="Neos.Neos.Ui" source-language="en" datatype="plaintext" target-language="af">
-    <body>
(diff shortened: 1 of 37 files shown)


markusguenther:
Thank you so much for translating all the labels. Need to check some of them because they are not used from the Neos.Neos.Ui package and obviously use same labels from Neos.Neos.

Will verify that and maybe also discuss with the team because it make no sense to have labels in both packages.

Great that you created your first neos-ui PR ❤️

markusguenther:
Thank you so much for translating all the labels. Need to check some of them because they are not used from the Neos.Neos.Ui package and obviously use same labels from Neos.Neos.

Will verify that and maybe also discuss with the team because it make no sense to have labels in both packages.

Great that you created your first neos-ui PR ❤️

markusguenther:
Ok turns out we already have an issue to discuss this problem of language label location :)

https://github.com/neos/neos-ui/issues/2421

markusguenther (approved):

PR #2853, opened 2020-11-24
TASK: Upgrade react-textarea-autosize to version 8.3

markusguenther:
As we get warnings in the console regarding the react lifecycle methods we needed to upgrade the library.

Screenshot 2020-11-24 at 09 47 40

This upgrade resolves the issues with the react lifecylce and brings also some bugfixes.


markusguenther:
Hmm as we already use the same version in neos-ui 4.0 I will better change the target to 4.0.

markusguenther:
Ok needed to go back to neos 5.2 because for the 4.0 branch we also need to update react :/

PR #2360, opened 2019-02-16
TASK: Remove legacy UI switch

danielkestler:
Remove legacy UI switch, closes #2321 as it is no longer needed.

PR #2857, opened 2020-12-02
TASK: Remove legacy "neos!" route

bwaidelich:
Removes the deprecated "Backend Login New UI Route" that was added
in order to explicitly switch between "classic" and "new" UI.

Related: #2360


Sebobo (approved):

markusguenther (approved):
Thanks for cleaning that up ❤️

PR #2858, opened 2020-12-02
BUGFIX: Fix position of guestFrameApplication in Page prototype

bwaidelich:
The `head.javascriptDocumentInformation` path was renamed to  `...javascriptBackendInformation`
a long time ago (#98) but the `@position` of `head.guestFrameApplication` part was never updated accordingly.

With the current Flow version this led to an exception:
```
The positional string "after javascriptDocumentInformation" (defined for key "guestFrameApplication") references a non-existing key.
```


markusguenther (approved):
Great that you changed that already.
Tested locally but lets wait for the UI tests.

PR #2859, opened 2020-12-02
BUGFIX: Fix position of guestFrameApplication in Page prototype

bwaidelich:
The `head.javascriptDocumentInformation` path was renamed to  `...javascriptBackendInformation`
a long time ago (#98) but the `@position` of `head.guestFrameApplication` part was never updated accordingly.

Related: #2858

### DIFF
--- a/Resources/Private/Fusion/Prototypes/Page.fusion
+++ b/Resources/Private/Fusion/Prototypes/Page.fusion
@@ -60,7 +60,7 @@ prototype(Neos.Neos:Page) {
         }
 
         guestFrameApplication = Neos.Fusion:Template {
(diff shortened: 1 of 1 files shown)


markusguenther:
@bwaidelich Hmm why 2 PRs? Normally we would upmerge the changes ... but that will take time of course :/

bwaidelich:
> but that will take time of course :/

Yes, that was the main reason. Also the PR description is slightly different

markusguenther:
Ah ok. Will apply styleCi stuff and upmerge then 😆

Thanks for finding that piece of outdated code.

PR #2868, opened 2020-12-17
TASK: Change auto formatting to https

daniellienert:
Nowadays most of the URLs are or at least should be accessible through https, so this should be the better default.


dimaip (approved):
Makes sense!

PR #2876, opened 2021-02-11
BUGFIX: SearchOperation must use the NodeSearchServiceInterface

dfeyer:
The current implementation did not allow to use custom NodeSearchService.

### DIFF
--- a/Classes/FlowQueryOperations/SearchOperation.php
+++ b/Classes/FlowQueryOperations/SearchOperation.php
@@ -18,7 +18,7 @@ use Neos\ContentRepository\Domain\Service\NodeTypeManager;
 use Neos\Eel\FlowQuery\FlowQuery;
 use Neos\Eel\FlowQuery\Operations\AbstractOperation;
 use Neos\Flow\Annotations as Flow;
-use Neos\Neos\Domain\Service\NodeSearchService;
+use Neos\Neos\Domain\Service\NodeSearchServiceInterface;
 
 /**
  */
@@ -26,7 +26,7 @@ class SearchOperation extends AbstractOperation
 {
     /**
      * @Flow\Inject
-     * @var NodeSearchService
+     * @var NodeSearchServiceInterface
      */
     protected $nodeSearchService;
 


kdambekalns (approved):
Makes sense! And… welcome back, @dfeyer!

Sebobo (approved):

Sebobo:
Thx!

But the PR should then go into some older branch right?

nlx-lars:
I like 👍 

But the NodeSearchServiceInterface does not support the fourth argument `$contextNode` which is passed in the SearchOperation:

https://github.com/neos/neos-development-collection/blob/28ae85b32d7192d40478885b48f2e6d48c6080fd/Neos.Neos/Classes/Domain/Service/NodeSearchServiceInterface.php#L27

Only the NodeSearchService has that.

kdambekalns:
Valid points, but at least PHP silently ignores "overflowing" method parameters, so the `$contextNode` would do no harm.

Then again, would this be an issue we need to tackle? Making that part of the interface? Or get rid of the parameter? 🤔

nlx-lars:
@kdambekalns  I don't think this would be a problem for now. I'd like to have the `$contextNode` in the interface anyway, I stumbled upon that problem a few times already. 

I guess we could change the interface later (probably Neos 8.0 since this is breaking?)

dfeyer:
> I guess we could change the interface later (probably Neos 8.0 since this is breaking?)

If it's an optional parameter it's not breaking ;)

nlx-lars:
@dfeyer but existing implementations wouldn't be compatible with the method signature, or am I missing something?

dfeyer:
> @dfeyer but existing implementations wouldn't be compatible with the new method signature, or am I missing something?

Yes modern PHP is a bit more picky about this ;) I'm too old, some bad practice remain in my brain since too long

kdambekalns:
> If it's an optional parameter it's not breaking ;)

Well, you could still the method, but when _implementing_ the interface, you must match the signature. As such, breaking. And you would be very old to not remember, at least since 5.0 this would have been breaking: https://3v4l.org/EaJQN

markusguenther:
Could not change this branch to the lowest maintained 4.0. So I created a new PR, hope that is ok.

markusguenther:
Is it ok for the others to backport that to version 4.0? @kdambekalns @Sebobo

dfeyer:
Closed in favor of #2881

PR #2877, opened 2021-02-23
TASK: Remove custom view and use views configuration

daniellienert:
Previously a custom view was used to configure fusion paths
which is not necessary and not best practice.

Also properties were configured that do not exist anymore


dimaip (approved):
Looks good, thanks!

mficzel (approved):
Looks fine, i would have configured the view in "initlializeView" but this is much more flexible as it allows other packages to add pathes. I always thought we still have to invent something like this but it is already there.

Issue #2282, opened 2019-01-10
Nodes can be hidden, even though they should not be hideable


kdambekalns:
### Description

Having a node with an auto-created childNode of a custom type here.

When selecting that childNode in the structure tree, it cannot be hidden using the icons in the tree or the content area.

But the inspector has a checkbox, and that is still active, so the node can be hidden.

#### Expected behavior

The node cannot be hidden at all.

#### Actual behavior

The node can be hidden using the inspector.

### Affected Versions

Neos: 4.0.9

UI: reproduced with 1.3.4 & 1.4.1

kdambekalns:
To cross-check, I added the `HidableMixin` to `Neos.Neos:ContentCollection` and the behaviour is the same: not hideable via the icons, but using the inspector. So the logic to disallow hiding seems to be inconsistent.

kitsunet:
technically you shouldn't be able to hide a childnode because the parent relies on it being there.
if it's optional you might as well let the user create it on demand, so the checkbox in the inspector should be disabled

kdambekalns:
I agree. The problem here is exactly that - it can be hidden, even though parts of the system already think it should not be possible…

bwaidelich:
Ooh, this change broke many pages for us.
In our scenario we want to let Neos create pre-defined child nodes hidden by default in order to allow the editors to activate those nodes at some point in the future (for example to announce the winner of a competition etc).

We don't want to allow the editor to _add_ or _delete_ the child nodes so using auto-created child nodes seemed like the right tool for us..

> technically you shouldn't be able to hide a childnode because the parent relies on it being there.

I beg to differ. It allows me to specify child nodes with a predefined node name, but it is not said that the parent has to rely on those child nodes not to be hidden (unless I manually disable/hide the `_hidden` property from the inspector)

grebaldi:
Hi  @bwaidelich,

I'm sorry this is causing trouble for you :(

Unfortunately, I don't see a proper way out of here other than to revert this change. And actually, if hidden auto-created child nodes are a scenario that should be supported, the StructureTree behavior should be changed as well.

I guess, we'll need a discussion for the latter. (I also could imagine something like adding an `optional` or `required` flag to the `childNodes` configuration of the NodeType schema)

As for reverting this change: Are you aware of any projects that would face serious problems by the ability to hide auto-created child nodes, @kdambekalns &  @kitsunet? If not, I'd say we should just revert.

[...]

PR #2890, opened 2021-03-22
BUGFIX: Ensure that auto-created child nodes cannot be hidden via the inspector

grebaldi:
Hello there,

this is a small fix to prevent auto-created child nodes from being hidden via the inspector and thus make that prevention consistent with the node tree toolbars.

**What I did**

I added an explicit check to the `InspectorEditorEnvelope` to see, whether we're currently looking at an auto-created child node and whether we're currently rendering the editor for the `_hidden` property. If so, the editor will be initialized with the option `disabled` set invariably to `true`.

That condition is also considered in the `isPropertyEnabled` method of the `Inspector` component (though I'm not sure about whether that makes sense).

**How to verify it**

Add this NodeType configuration to your setup:

```yaml
'Neos.Neos:ContentCollection':
  superTypes:
    'Neos.Neos:Hidable': true
```
And then check, whether the default page main content collection can still be hidden via the inspector.

solves: #2282

### DIFF
--- a/packages/neos-ui/src/Containers/RightSideBar/Inspector/InspectorEditorEnvelope/index.js
+++ b/packages/neos-ui/src/Containers/RightSideBar/Inspector/InspectorEditorEnvelope/index.js
@@ -38,6 +38,16 @@ export default class InspectorEditorEnvelope extends PureComponent {
         commit: PropTypes.func.isRequired
     };
 
+    get options() {
+        // This makes sure that auto-created child nodes cannot be hidden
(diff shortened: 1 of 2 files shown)


markusguenther (approved):
Thanks again @grebaldi for this PR. 
It works and the code looks fine 🙂

bwaidelich:
@grebaldi do you have an idea/opinion on how I could best work around this limitation? (see https://neos-project.slack.com/archives/C0U0KEGDQ/p1631092388025900)

Issue #2767, opened 2020-08-20
when changing node type, re-rendering in UI might fail


skurfuerst:
### Steps to Reproduce

1. Use the demo site
2. use the two column on the homepage
3. change the node type to a four-column
  -  EXPECTED: the backend is re-rendered with a 4 column
  - ACTUAL: the backend is re-showing the two-column.
4. reload the backend
  - now, the BE shows the 4-column.
5. change back to two-column
  - EXPECTED: the backend is re-rendered with a 2 column
  - ACTUAL: the backend is re-rendered with a 4-column; with the two right columns being an error.
6. reload the backend
  - the error disappears

### Problem description

- in the `nodeDomAddress` for the change, the node type is encoded in the fusion path.
- when we **change** the node type, this is wrong....

### Solution Idea

- we want to send TWO nodeDomAddresses: one for the node itself, and one for the closest rendered parent.
- then the server can decide which one to use for the refresh; and in case the node type changes, use the outer one.

### Affected Versions

probably all versions

skurfuerst:
reminder to self: once fixed (e.g. in Property.php in Neos UI), we also need to adjust the Event Sourced CR Property implementation

Sebobo:
Yes that's something a bit similar to what I wanted to fix in general. There are often NodeTypes in which the collection children need to know about each other. Like for setting the correct css class. 
In this case the integrators always set something like "reloadPageIfChanged". But instead reloading the closest parent content element would be the desired option.

grebaldi:
This should have been closed by  #2892, but wasn't because I misspelled the magic word :facepalm:

PR #2892, opened 2021-03-25
BUGFIX: Always try to re-render content by re-evaluating its closest ContentCase

grebaldi:
Hi there,

out-of-band rendering in the UI works via so-called DomAddresses that contain the information what fusion path needs to be used to re-render a certain content element. That fusion path usually leads through a `Neos.Neos:ContentCase` ending up at a NodeType-specific fusion prototype.

Out-of-band rendering happens when a node properties change and a rendered content element needs to be updated. If the property that was changes was the NodeType of that content element, the behavior described above runs into a problem. The prototype used for rendering is no longer the prototype that `Neos.Neos:ContentCase` would lead to. This problem is described in issue  #2767.

This PR is an attempt at solving that problem by manipulating the above mentioned fusion path, so that it always tries to render the appropriate ContentCase instead of the final prototype.

**Before:**

https://user-images.githubusercontent.com/2522299/112519158-63932600-8d9a-11eb-8fae-bbd9eb8982b7.mp4

**After:**

https://user-images.githubusercontent.com/2522299/112519178-67bf4380-8d9a-11eb-8908-ec91b5c36833.mp4

**How to verify it**

Follow the reproduction steps in #2767.

solves: #2767

### DIFF
--- a/Classes/Domain/Model/RenderedNodeDomAddress.php
+++ b/Classes/Domain/Model/RenderedNodeDomAddress.php
@@ -80,7 +80,7 @@ class RenderedNodeDomAddress implements \JsonSerializable
      *
      * @return string
      */
-    public function getFusionPathForContentRendering()
+    public function getFusionPathForContentRendering(): string
     {
         $fusionPathForContentRendering = $this->getFusionPath();
         $fusionPathForContentRendering = preg_replace(
--- a/Classes/Domain/Model/RenderedNodeDomAddress.php
+++ b/Classes/Domain/Model/RenderedNodeDomAddress.php
@@ -84,7 +84,7 @@ class RenderedNodeDomAddress implements \JsonSerializable
     {
         $fusionPathForContentRendering = $this->getFusionPath();
         $fusionPathForContentRendering = preg_replace(
-            '/(\/itemRenderer\<Neos\.\Neos\:ContentCase\>)\/default\<Neos\.Fusion\:Matcher\>\/element(\<.*\>)$/',
+            '/(\/itemRenderer<Neos\.Neos:ContentCase>)\/default<Neos\.Fusion:Matcher>\/element(<[^>]+>)$/',
             '$1',
             $fusionPathForContentRendering
         );
--- a/Classes/Domain/Model/RenderedNodeDomAddress.php
+++ b/Classes/Domain/Model/RenderedNodeDomAddress.php
@@ -73,6 +73,26 @@ class RenderedNodeDomAddress implements \JsonSerializable
     }
 
     /**
+     * Get the fusion path that should be used for rendering the addressed
+     * content. For most contents that would be the closest `Neos.Neos:ContentCase`
+     * within the path rather than the actual prototype that was used to
+     * render the content.
+     *
+     * @return string
+     */
+    public function getFusionPathForContentRendering()
+    {
+        $fusionPathForContentRendering = $this->getFusionPath();
+        $fusionPathForContentRendering = preg_replace(
+            '/(\/itemRenderer\<Neos\.\Neos\:ContentCase\>)\/default\<Neos\.Fusion\:Matcher\>\/element(\<.*\>)$/',
+            '$1',
+            $fusionPathForContentRendering
+        );
+
+        return $fusionPathForContentRendering;
+    }
+
+    /**
      * Serialize to json
      *
      * @return array
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -135,6 +135,26 @@ jobs:
       - store_artifacts:
           path: /home/circleci/app/Data/Logs
 
+  php-unittests:
+    environment:
+      FLOW_CONTEXT: Production
+    docker:
+      - image: quay.io/yeebase/ci-build:7.2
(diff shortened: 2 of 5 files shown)


markusguenther (approved):
I like the solution and it works fine :). 
Thanks @grebaldi

markusguenther:
Maybe also @bwaidelich or @kitsunet want to take a look :)

Classes/Domain/Model/RenderedNodeDomAddress.php
```diff
@@ -72,6 +72,26 @@ public function getFusionPath()
...
 
+    /**
+     * Get the fusion path that should be used for rendering the addressed
+     * content. For most contents that would be the closest `Neos.Neos:ContentCase`
+     * within the path rather than the actual prototype that was used to
+     * render the content.
+     *
+     * @return string
+     */
+    public function getFusionPathForContentRendering()
```
bwaidelich, line 83:
```suggestion
    public function getFusionPathForContentRendering(): string
```

Classes/Domain/Model/RenderedNodeDomAddress.php
```diff
@@ -72,6 +72,26 @@ public function getFusionPath()
...
+    {
+        $fusionPathForContentRendering = $this->getFusionPath();
+        $fusionPathForContentRendering = preg_replace(
+            '/(\/itemRenderer\<Neos\.\Neos\:ContentCase\>)\/default\<Neos\.Fusion\:Matcher\>\/element(\<.*\>)$/',
```
bwaidelich, line 87: The `\N` seems incorrect, but a few more escapes aren't required IMO
```suggestion
            '/(\/itemRenderer<Neos\.Neos:ContentCase>)\/default<Neos\.Fusion:Matcher>\/element(<.*>)$/',
```

bwaidelich (commented):
Thanks a lot for taking care! <3
I remember that I stumbled upon this issue as well and had no clue how to solve it best.

I'm just wondering..
Is that hard coded `/itemRenderer<Neos.Neos:ContentCase>/default<Neos.Fusion:Matcher>/element` prefix lookup something that should creep into the Neos UI knowledge?

I _think_ it is fine in this case since the bug is rather annoying and the solution seems pretty pragmatic.
But, especially since this targets already published branches, we should have some test cases for this IMO covering for example the scenario of two nested Content Cases.
Not sure how hard it is to test this behavior but I'm happy to help.

Apart from that two comments regarding cosmetics mostly

grebaldi:
Thank you for the review @bwaidelich!

> Is that hard coded `/itemRenderer<Neos.Neos:ContentCase>/default<Neos.Fusion:Matcher>/element` prefix lookup something that should creep into the Neos UI knowledge?

My guess would be that the architecture of the Neos UI API is more likely to change than this specific path slice. Imho, depending on this specific structure should be safe for now. But I do agree that this a leaked implementation detail that belongs behind a contract somewhere. 

> [...] we should have some test cases for this IMO covering for example the scenario of two nested Content Cases.
> Not sure how hard it is to test this behavior but I'm happy to help.

Agreed. I could extract some data samples and write up a unit test case. But I'm pretty clueless about how to achieve anything beyond that. The biggest problem would be to come up with some cases in which rendering the closest `Neos.Neos:ContentCase` isn't sufficient. I tried, but I can't think of any :sweat_smile:

bwaidelich:
> My guess would be that the architecture of the Neos UI API is more likely to change than this specific path slice

That's a fair point!

> The biggest problem would be to come up with some cases in which rendering the closest
> Neos.Neos:ContentCase isn't sufficient

I also can't think of other cases. I'm thinking more about tests that verify that we are really matching the _closest_ ContentCase.

Which makes me think:
For a Fusion path of
`/itemRenderer<Neos.Neos:ContentCase>/default<Neos.Fusion:Matcher>/element<First>/itemRenderer<Neos.Neos:ContentCase>/default<Neos.Fusion:Matcher>/element<Second>`

The replaced path would be `/itemRenderer<Neos.Neos:ContentCase>` right now. Is that correct?

grebaldi:
> The replaced path would be `/itemRenderer<Neos.Neos:ContentCase>` right now. Is that correct?

The regular expression ends on `$` - it should only match this pattern at the very end of the string. So, your example should be converted to:

```
/itemRenderer<Neos.Neos:ContentCase>/default<Neos.Fusion:Matcher>/element<First>/itemRenderer<Neos.Neos:ContentCase>
```

But I agree that this should be tested.

bwaidelich:
> it should only match this pattern at the very end of the string

The problem is, that the `<.*>` is pretty greedy and matches the nested part completely, see: https://regex101.com/r/x9KmyF/1

What about:
`/(\/itemRenderer<Neos\.Neos:ContentCase>)\/default<Neos\.Fusion:Matcher>\/element(<[^>]+>)$/`
?

grebaldi:
I added a Unit Test for `RenderedNodeDomAddress` and adjusted the circle-ci config to include php unit tests.

I also adjusted the regular expression according to your suggestions, @bwaidelich.

bwaidelich (approved):
Awesome, thank you!
+1 (by reading)

Issue #2878, opened 2021-02-23
removeFormat does not work


MarcoPNS:
In the Neos Docs it says that you can use the formatting option "removeFormat: true" for the CKEditor. Apparently, however, it was never actually added.

I have tested it with Neos 7, but it is probably a general problem since the CKEditor 5 is used. The required package @ckeditor/ckeditor5-remove-format is not in the dependencies.

ref: https://ckeditor.com/docs/ckeditor5/latest/features/remove-format.html

MarcoPNS:
a older issue also had `wait for removeformat to be implemented` as todo: #2002

markusguenther:
Has been released with version 5.3.6

Issue #2002, opened 2018-07-12
META: CKE5 integration


dimaip:
### How to test?

The default text editor is now configurable in Settings.yaml:

```
Neos:
  Neos:
    Ui:
      frontendConfiguration:
        defaultInlineEditor: 'ckeditor5'
```

### TODO:

- [x] wait for  removeformat to be implemented in https://github.com/ckeditor/ckeditor5/issues/908
- [x] test, make sure it's rock solid, then release a major version making CKE5 the default editor
- [ ] ctrl+k should open link editor

DrillSergeant:
@dimaip , could you change the handling of link-schemes in CKE5 integration? In CKE4 the possible schemes are mentioned in Packages/Application/Neos.Neos.Ui/packages/neos-ui-ckeditor-bindings/src/EditorToolbar/LinkIconButton.js in this construct:
```
const isUri = str =>
    str && Boolean(str.match('^(https?://|mailto:|tel:|#)'));
```
Sometimes (especially for intranet-projects) there are more url-schemes that should work (like "notes://" for lotus-notes documents). In CKE4 I cannot extend that because it's hardcoded and so it takes a lot of work to make linking to those schemed-links possible in RTE.
I would love if you could make this configurable via settings to add additional schemes so the CKE5-link editor accepts those custom schemes.
I just had a slack call with @kitsunet where he mentioned even some more ideas. Maybe you guys can talk/discuss. Remember, a simple solution for my usecase would help for many cases I think. I guess (hope) it's not so hard to check settings for some extra schemes.

DrillSergeant:
@dimaip ah nice, thank you.

danielkestler:
We can close this, can we?

PR #2899, opened 2021-04-09
BUGFIX: Add removeFormat button to CKEditor5 toolbar

grebaldi:
Hi there,

this PR adds `@ckeditor/ckeditor5-remove-format` to the dependencies of `@neos-project/neos-ui-ckeditor5-bindings` and uses it to realize the formerly non-functional `removeFormat` option for CK Editor.

solves #2878

### DIFF
--- a/packages/neos-ui-ckeditor5-bindings/package.json
+++ b/packages/neos-ui-ckeditor5-bindings/package.json
@@ -19,6 +19,7 @@
     "@ckeditor/ckeditor5-link": "^11.0.2",
     "@ckeditor/ckeditor5-list": "^12.0.2",
     "@ckeditor/ckeditor5-paragraph": "^11.0.2",
+    "@ckeditor/ckeditor5-remove-format": "^10.0.0",
     "@ckeditor/ckeditor5-table": "^13.0.0",
     "@ckeditor/ckeditor5-widget": "^11.0.2",
     "@neos-project/neos-ui-backend-connector": "5.3.6",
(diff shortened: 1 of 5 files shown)


markusguenther (approved):
It works great. Thanks for that @grebaldi , but isn't it more a feature than a task?

https://user-images.githubusercontent.com/1014126/114239966-12348a80-9987-11eb-9b51-42c61f10a22d.mp4

grebaldi:
I marked it as "BUGFIX", because I took it that `removeFormat` was supposed to work but didn't. I have to admit though that I'm clueless on the guarantees we make for CKEditor 5 when it comes to LTS releases. 

The `Ctrl+K` addition may be more of a feature - I just added it, because it was mentioned as a requirement in #2002. It's probably worth separating the commits into dedicated PRs, because the second one changes quite a bit in several packages. I'm optimistically confident however, that both commits would survive upmerges quite well. 

Wdyt?

markusguenther:
Did not mention that because I think that it will break something at the up merge. Was just a mention because there was the rule no Features in released branches. TBH I would not block this from getting into 5.3 at all, but don't know what the others think.

Checked also the old emberjs UI but there we did not have a button like that. We had a button `keep/discard formating`
But I like the PR :)

So if @dimaip @Sebobo does not have something against that, I am open to merge it as it is. It does not break anything :)

Sebobo:
I love the improvements, but tbh. the title is IMO misleading or rather inconclusive and two barely related changes combined.

So I would prefer having this separated.
The remove format I could accept as bugfix, but the other should be a feature for master.

grebaldi:
Yep, I was unsure about that as well. I'll go ahead and open a separate PR to address #2002.

Issue #2741, opened 2020-06-16
Drag & Drop between trees not validated


rolandschuetz:
### Description

It's possible to drag content nodes from the content tree onto document nodes from the document tree. The UI doesn't do anything but also gives no success nor error feedback.

### Steps to Reproduce

1. Drag a content element from the content tree onto the document tree
2. Drop it anywhere.

#### Expected behavior

Feedback that the position is invalid during drag.

#### Actual behavior

Blue valid drop locations

### Affected Versions

UI: 5.2.1

grebaldi:
Should have been closed by #2905 - not sure, why it wasn't :/

PR #2905, opened 2021-04-22
BUGFIX: Prevent drag and drop across trees

grebaldi:
Hi there,

this PR contains a small bugfix to prevent drag and drop across the Document and Content trees. This never actually worked in the first place, but the UI gave visual indication that this would be a valid operation.

**Before:**

https://user-images.githubusercontent.com/2522299/115692548-62382780-a35f-11eb-8936-f41082e5b067.mp4 

**After:**

https://user-images.githubusercontent.com/2522299/115692594-6bc18f80-a35f-11eb-8be9-50bd42ce4c86.mp4

fixes: #2741


markusguenther (approved):
Thanks again for that fix @grebaldi 
Works like charm ❤️

PR #2920, opened 2021-07-05
TASK: Refactor controller to not use static fusion path

daniellienert:
The fusion path is now dynamically determined by controller / action
This enables additional actions with separate templates for that controller.

The path `Neos.Neos.Ui.BackendController.index` is used instead and the `initializeView` hook with `$view->setFusionPath('backend');` was removed.

### DIFF
--- a/Classes/Controller/BackendController.php
+++ b/Classes/Controller/BackendController.php
@@ -120,12 +120,6 @@ class BackendController extends ActionController
      */
     protected $splashScreenPartial;
 
-    public function initializeView(ViewInterface $view)
-    {
-        /** @var FusionView $view */
-        $view->setFusionPath('backend');
-    }
-
     /**
      * Displays the backend interface
      *
--- a/Resources/Private/Fusion/Backend/Root.fusion
+++ b/Resources/Private/Fusion/Backend/Root.fusion
@@ -13,6 +13,6 @@ include: ./Application.fusion
 #
 # Neos UI (Content Module) entry point
 #
-backend = Neos.Neos.Ui:Application {
+Neos.Neos.Ui.BackendController.index = Neos.Neos.Ui:Application {
     title = 'Neos CMS'
 }


kdambekalns (approved):
Looks good by reading…

grebaldi (changes requested):
Hi @daniellienert,

`Resources/Private/Fusion/Backend/Prototypes/RenderConfiguration.fusion` has been moved by this change. That leads to several exceptions like this one:

```
Could not include Fusion file "resource://Neos.Neos.Ui/Private/Fusion/Prototypes/RenderConfiguration.fusion"
resource://Neos.Neos.Ui/Private/Fusion/Backend/Root.fusion:4

Exception Code | 1347977017
-- | --
Neos\Fusion\Exception
2021072513041021ee63
/tmp/neos/Development/SubContextddev/Cache/Code/Flow_Object_Classes/Neos_Fusion_Core_Parser.php
641
Packages/Application/Neos.Fusion/Classes/Core/Parser.php
```

This is because it is referenced here:
https://github.com/neos/neos-ui/blob/050e58b92e8248acd48051eaa9ec0e8abe0e2768/Resources/Private/Fusion/Backend/Root.fusion#L3

Yet, updating the include won't cut it, because the prototype is also used here:
https://github.com/neos/neos-ui/blob/050e58b92e8248acd48051eaa9ec0e8abe0e2768/Resources/Private/Fusion/Prototypes/Page.fusion#L11

So, I'd suggest just to move the file back to its original path.

Other than that, the PR looks good :) 

After moving back `Resources/Private/Fusion/Backend/Prototypes/RenderConfiguration.fusion` to `Resources/Private/Fusion/Prototypes/RenderConfiguration.fusion`, I also recommend rebasing the PR against `master`. After that, E2E Tests are likely to go green again.

Sebobo:
Probably also a rebase is needed here to make e2e tests work again.

grebaldi (approved):

markusguenther:
Also rebased this change and make it work with the latest state.

markusguenther (approved):
Looks good to me, but I also adjusted something to make it work ;)

mhsdesign:
Im not super sure that the fusion contents are 1 to 1 the same ... aaand we have to remember when upmerging to use the new 9.0 configuration

mhsdesign (approved):
no yes its the same

mhsdesign (changes requested):
But now i fear with the refactoring for 9.0 in https://github.com/neos/neos-ui/pull/3682 (see commit https://github.com/neos/neos-ui/pull/3682/commits/b5b918db273cde421859dc4c9fadcc524ca18f2e) we have waited a bit too long ^^

i wouldnt want to have different backend controller fusion behaviour for 8.3 and 8.4 and 9.0

so instead i would propose to incorporate this cleanup into the mentioned pr or wait for it to be merged and target 9.0

/ cc @grebaldi

grebaldi:
@mhsdesign:
> so instead i would propose to incorporate this cleanup into the mentioned pr or wait for it to be merged and target 9.0

I'd be fine with targetting `9.0`, but I would propose to merge this PR first. Then I can rebase https://github.com/neos/neos-ui/pull/3682 on `9.0` and incorporate the change accordingly.

mhsdesign:
> I would propose to merge this PR first

i only agree with this part ;) So i rebased it onto your pr and this pr targets now `feature/multi-app/01-foundation` its wayyy easier this way imo ^^

If you agree @grebaldi please just hit the merge button - as https://github.com/neos/neos-ui/pull/3682 is your pr ^^

mhsdesign (approved):

PR #2922, opened 2021-07-09
BUGFIX: Use setHttpHeader instead of deprecated setComponentParameter

daniellienert:


albe (approved):

grebaldi:
E2E tests are failing, because of a deprecated Guzzle function used here:
https://github.com/neos/flow-development-collection/blob/7.0/Neos.Flow/Classes/Mvc/ActionResponse.php#L6

And a broken version Constraint in Neos.Http.Factories 7.0.7:
https://github.com/neos/flow-development-collection/blob/7.0.7/Neos.Http.Factories/composer.json#L12

The latter was fixed in 7.0.8:
https://github.com/neos/flow-development-collection/blob/7.0.8/Neos.Http.Factories/composer.json#L12

But for some reason, composer in Circle CI is pulling Neos.Http.Factories 7.0.7 even though it is pulling the correct version of Neos.Flow (7.0.8).

I'll investigate further...

daniellienert:
Backported the fix for the deprecated guzzle function

Sebobo:
Probably a rebase is now needed to make the e2e tests work?

Classes/Fusion/ExceptionHandler/PageExceptionHandler.php
```diff
@@ -11,6 +11,7 @@
  * source code.
  */
 
+use GuzzleHttp\Psr7\Message;
 use function GuzzleHttp\Psr7\str;
```
grebaldi, line 15: This can be removed, right?

grebaldi:

They're still failing due to:
```
Exception in line 44 of /home/circleci/app/Data/Temporary/Production/Cache/Code/Flow_Object_Classes/Neos_Flow_Mvc_DispatchMiddleware.php: Call to undefined function GuzzleHttp\Psr7\stream_for()
```

As can be seen here: https://app.circleci.com/pipelines/github/neos/neos-ui/1657/workflows/89db7e3e-f5bf-4a60-9678-04ae8e601851/jobs/7301/artifacts

After the bugfix release of Neos.Fusion.Form, the E2E tests of Neos.Neos.Ui just need to be re-run with an updated cache key prefix (as described here: https://github.com/neos/neos-ui#just-the-end-to-end-tests-fail). There's most likely no need for a rebase.

kdambekalns:
Neos starts fine now, but e2e still fails…

grebaldi:
I had another look into it. We have two issues:

1. **Another package messes with the version constraints**

If I ask composer about the installed version of `guzzlehttp/psr7`, I get this:
```sh
$ composer why guzzlehttp/psr7 2.0.0
guzzlehttp/guzzle    7.3.0  requires  guzzlehttp/psr7 (^1.7 || ^2.0)           
neos/fusion-form     2.0.3  requires  guzzlehttp/psr7 (*)                      
neos/http-factories  7.0.7  requires  guzzlehttp/psr7 (^1.4, !=1.8.0 || ~2.0)
```

Okay, `guzzlehttp/guzzle` allows for `guzzlehttp/psr7 2.0.0` to be installed, so I guess composer prefers that and therefore tosses `neos/http-factories 7.0.8` in favor of `7.0.7`.

But who needs `guzzlehttp/guzzle` at `7.3.0`?:
```sh
$ composer why guzzlehttp/guzzle
neos/buildessentials  9999999-dev  requires  guzzlehttp/guzzle (^7.0) 
```
:disappointed: 

Well, we do have a `require-dev` section in the test distribution that looks like this:
```json
  "require-dev": {
      "neos/buildessentials": "@dev",
      "mikey179/vfsstream": "^1.6",
      "phpunit/phpunit": "^8.1"
    },
```

Before meddling any further with those constraints, I suggest, we remove the `require-dev` section altogether. I'll create a separate PR for that in the coming days, so we can check whether this works.

2. **There's a problem with the Circle CI cache keys for the composer cache.**

This is a bit more complicated.

I started a new build (with newly generated cache key prefix) for this PR to verify @kdambekalns observation. Indeed that build (No. 7384) failed.

Then I restarted the build for another PR (#2937) to verify, whether the problem affects `5.3` as well. The build failed, but it's likely false-negative, because - as you can see in the following log - the composer cache was erroneously pulled from build no. 7384:

https://app.circleci.com/pipelines/github/neos/neos-ui/1673/workflows/8c4b3c03-d5f8-4af5-888c-f78124905200/jobs/7390/parallel-runs/0/steps/0-102

I'm not an expert regarding Circle CI, but apparently this configuration:

```yaml
  - &restore_composer_cache
    keys:
      - composer-cache-v{{ .Environment.CACHE_VERSION }}-{{ checksum "Tests/IntegrationTests/TestDistribution/composer.json" }}
      - composer-cache-v{{ .Environment.CACHE_VERSION }}-
```
expresses a fallback between the first and the second cache key. I suspect, both are being tested and both are being written. As a result, it's possible that all E2E tests are running with whatever dependencies have been first persisted into `composer-cache-v{{ .Environment.CACHE_VERSION }}-`.

In the coming days, I will also create a separate PR to address this issue. This one would have to be merged into `5.3` though. 

**Bottom line:**
So, the E2E tests for *this* PR will have to wait for upmerges of the fixes described above and require a rebase against `7.0`.

albe:
> If I ask composer about the installed version of `guzzlehttp/psr7`, I get this:
> 
> ```shell
> $ composer why guzzlehttp/psr7 2.0.0
> guzzlehttp/guzzle    7.3.0  requires  guzzlehttp/psr7 (^1.7 || ^2.0)           
> neos/fusion-form     2.0.3  requires  guzzlehttp/psr7 (*)                      
> neos/http-factories  7.0.7  requires  guzzlehttp/psr7 (^1.4, !=1.8.0 || ~2.0)
> ```
> 
> Okay, `guzzlehttp/guzzle` allows for `guzzlehttp/psr7 2.0.0` to be installed, so I guess composer prefers that and therefore tosses `neos/http-factories 7.0.8` in favor of `7.0.7`.
> 
> But who needs `guzzlehttp/guzzle` at `7.3.0`?:
> 
> ```shell
> $ composer why guzzlehttp/guzzle
> neos/buildessentials  9999999-dev  requires  guzzlehttp/guzzle (^7.0) 
> ```
> 
> 😞
> 
> Well, we do have a `require-dev` section in the test distribution that looks like this:
> 
> ```json
>   "require-dev": {
>       "neos/buildessentials": "@dev",
>       "mikey179/vfsstream": "^1.6",
>       "phpunit/phpunit": "^8.1"
>     },
> ```
> 
> Before meddling any further with those constraints, I suggest, we remove the `require-dev` section altogether. I'll create a separate PR for that in the coming days, so we can check whether this works.

Yeah, that guzzle dependency in the build-essentials was added through my build script for the release notes I added (needed to do http requests to GH API) - see https://github.com/neos/BuildEssentials/pull/51/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R9

What still strikes me as odd is that composer thinks it should downgrade one second-level dependency in order to install a higher version of a third-level dependency 🤷‍♂️

We discussed two options today in the unicorn meeting - one is to add a "conflicts" with guzzle/psr7 2.0 to flow, which should hopefully prevent composer from prefering that 2.0 version unless it thinks it should downgrade Flow then. The other is to get rid of all the guzzle function calls that were breaking and hence be compatible with 2.0 again.

grebaldi:
Brief Update on my findings:

1. Removing the fallback cache-key did nothing

This log shows, that `guzzlehttp/psr7 2.0.0` is still being installed: https://app.circleci.com/pipelines/github/neos/neos-ui/1679/workflows/4ccbe018-d3a3-410b-a20c-4f98ec83fc58/jobs/7409/parallel-runs/0/steps/0-105

2. Removing `require-dev` from the TestDistribution's composer.json also did nothing:

This log shows that `guzzlehttp/psr7 2.0.0` is still being installed: https://app.circleci.com/pipelines/github/neos/neos-ui/1681/workflows/874c4403-89c9-4c39-9cbb-c0682bb274bc/jobs/7422/parallel-runs/0/steps/0-105

This however is the first occurrence of this issue that I can't reproduce locally. The only difference I can spot is that Circle CI appears to be running composer 1, while I am running composer 2. I will have to repeat reproduction using composer 1 to verify.

@albe:
 > We discussed two options today in the unicorn meeting - one is to add a "conflicts" with guzzle/psr7 2.0 to flow, which should hopefully prevent composer from prefering that 2.0 version unless it thinks it should downgrade Flow then.

Is it possible to have a canary release to test that theory?

> The other is to get rid of all the guzzle function calls that were breaking and hence be compatible with 2.0 again.

Probably the better option tbh. As far as I understand the issue however, we would have to do that as a bugfix release starting at 5.3, since the issue is present there as well. 

I could prepare a PR for the UI package, but I guess other packages are affected as well?

Wdyt?

kdambekalns:
TBH I don't understand what is missing here. Is it just the broken test? Is it only broken for this PR? If not, I'd say this one is fone, and the tests need to be fixed independently. If yes, what is the cause and how is it related to this change?

Issue #2926, opened 2021-07-22
Don't load nodedata for structure tree from guest frame


Sebobo:
### Description

During the performance investigations in the last weeks we got aware, that the content tree loads a lot of information from the rendered content.
Explains very well, that the tree breaks when the rendering breaks. Also a lot of information is loaded from the CR to make sure that everything is in the guest frame.
As the content tree already has an endpoint to get fresh and complete information from the backend with its little reload button, why don’t we just ignore the data from the guest frame and always load the information from the endpoint?
I think that would help quite a bit with performance (ui & rendering) and solve those issues with the tree not getting the information it needs.

Also found out, that the `loadingDepth` of the Content Tree also has no real performance benefit (contrary to the documentation) because of this as the tree loads all metadata of all rendered or unrendered nodes…

### Steps to Reproduce

Set the following setting: 

```
Neos:
  Neos:
    userInterface:
      navigateComponent:
        structureTree:
          loadingDepth: 1
```          

Then check which nodedata is loaded from the Neos backend endpoint `get-additional-node-metadata`.
Also see in the html of the guest frame that tags like `<script>(function(){(this['@Neos.Neos.Ui:Nodes']..` are 
rendered for all invisible nodes and even other document nodes.

#### Expected behavior

* Content tree loads without errors even when rendering in guest frame fails
* Content tree only loads nodes defined by `loadingDepth`
* No nodedata for unrendered nodes needs to be appended to page rendering (See `AugmentationAspect`)

#### Actual behavior

* Content tree doesn't load properly when rendering fails and needs manual reload.
* Content contains script tags for other document nodes and unrendered nodes.
* Content tree loads all content nodes on page, even when `loadingDepth` is set to a low value.

### Affected Versions

Neos: all until 7.1 so far

UI: all until 7.1 so far

Sebobo:
Some progress already:

1. Disabled the rendering of `neosUiNonRenderedNodeMetadata`
2. Content Tree only shows visibly rendered nodes, db queries down from 119 -> 78
3. Adapted toggle and children loading sagas from PageTree to ContentTree
4. Content nodes are now loaded from backend when toggling nodes 🎉
5. ❌ Inspector broken when selecting a loaded childnode
6. ❌ Invisible nodes like empty content collections still missing in content tree

PR #2927, opened 2021-07-23
FEATURE: Only load content tree nodes on demand

Sebobo:
Resolves: #2926

**What I did**

The content tree now behaves the same as the document tree. Only loading the nodes defined by `loadingDepth` initially and not relying purely on the nodedata rendered in the guestframe.
Toggling nodes with children also mimimcs the behaviour of the document tree and reloads the children when needed.

That reduces the number of initial db requests during page rendering in the backend from 119 to 78 in my test.

For large number of content nodes on a page this should improve performance considerably, especially if they are not all rendered.

**How I did it**

- [x] Removed the `nonRenderedNodes` rendering in the `Neos.Neos:Page` prototype
- [x] Adapt toggle/loading/updating mechanics from document tree to content tree
- [x] Load all necessary nodes defined by `loadingDepth` when reloading or switching documents and merge them with data from content (might make loading on toggle in some cases unnecessary later) 

**How to verify it**

Work with content tree like before

Example video shows page with content that is not rendered, uncollapsing the parent node then loads the nodes (see loading spinner).
The other page renders all content, no further nodes need to be loaded.

https://user-images.githubusercontent.com/596967/126767022-9c7414d1-37e4-4692-875b-2f7500fe4f8c.mov


Sebobo:
Failing tests in master seems unrelated as it also fails for the other open PRs.

grebaldi:
Tests are green again :)

grebaldi (changes requested):
Hi @Sebobo,

I've encountered an issue, though I have not identified a root cause yet.

So, I have a Content NodeType that has two auto-created content collections:
```yaml
'Vendor.Site:Content.Section':
  superTypes:
    'Vendor.Site:Content': true
// ...
  childNodes:
    contents:
      type: Neos.Neos:ContentCollection
      constraints:
        nodeTypes:
          'Vendor.Site:Content': true
    hidden-contents:
      type: Neos.Neos:ContentCollection
      constraints:
        nodeTypes:
          'Vendor.Site:Content': true
```

The collection `contents` is being rendered with content element wrapping, while `hidden-contents` is not being rendered at all.

This leads to this behavior:

| Before | After |
|-|-|
| Screenshot_2021-07-25_15-00-52 | https://user-images.githubusercontent.com/2522299/126901010-2d21160f-4c60-4113-842f-12c68083126a.mp4 |

I'll continue to investigate, but maybe you have an immediate idea what might be causing this :)

Sebobo:
Hi @grebaldi thanks for testing.

Your example was one of the first issues I encountered after turning off the unrendered nodes rendering and had fixed. And with your nodetype definition and the following Fusion I'm not able to reproduce:

```
prototype(Vendor.Site:Content.Section) < prototype(Neos.Neos:ContentComponent) {
    renderer = afx`
        <div>
            <Neos.Neos:ContentCollection nodePath="contents" />
        </div>
    `
}
```

See in my video how it has part of the tree initially retrieved from the content and then it automatically loads the missing second collection:

https://user-images.githubusercontent.com/596967/126942731-1a5968a9-788d-4dc6-80e9-d9aec8656287.mov

So maybe anything special in your Fusion, that I did differently?

And can you check if the initial event for `watchCurrentDocument` is triggered for the content tree and executes a flowquery?

Sebobo:
Fixed two bugs related to selecting content nodes that are not on the current page.

@grebaldi did you have another chance to verify your issue?
Maybe we should have a call and look at it together.

markusguenther (approved):
Tested the PR and did not see any issues like @grebaldi 
So can not reproduce it. But maybe Wilhelm can check it again :)

The PR looks pretty clean. The change works like a charm on my machine ;)
Thanks to @Sebobo for this improvement 💚

packages/neos-ui-sagas/src/UI/ContentTree/index.js
```diff
@@ -63,3 +64,102 @@ export function * watchNodeFocus({configuration}) {
...
+
+        if (!children) {
+            return;
+        }
+
+        const childrenAreFullyLoaded = children
+        .filter(childEnvelope => nodeTypesRegistry.hasRole(childEnvelope.nodeType, 'content') || nodeTypesRegistry.hasRole(childEnvelope.nodeType, 'contentCollection'))
+        .every(
+            childEnvelope => Boolean($get(['cr', 'nodes', 'byContextPath', $get('contextPath', childEnvelope)], state))
+        );
```
grebaldi, line 84:
```suggestion
        const checkIfChildrenAreFullyLoadedRecursively = contextPath => {
            return $get(['cr', 'nodes', 'byContextPath', contextPath, 'children'], state)
            .filter(childEnvelope => nodeTypesRegistry.hasRole(childEnvelope.nodeType, 'content') || nodeTypesRegistry.hasRole(childEnvelope.nodeType, 'contentCollection'))
            .every(
                childEnvelope =>
                    $get(['cr', 'nodes', 'byContextPath', $get('contextPath', childEnvelope)], state)
                    && checkIfChildrenAreFullyLoadedRecursively($get('contextPath', childEnvelope))
            );
        }
        const childrenAreFullyLoaded = checkIfChildrenAreFullyLoadedRecursively(contextPath);
```

grebaldi (commented):
I still saw the issue and debugged a bit. 

I could assert that the initial check for `childrenAreFullyLoaded` in the `watchCurrentDocument` saga wasn't performed recursively - that was the cause for the issue I've seen.

I added a suggestion that should fix it :)

Sebobo:
@grebaldi thx!

Can't see a difference, but makes sense :)
Commited the change.

grebaldi (approved):

packages/neos-ui-sagas/src/UI/ContentTree/index.js
```diff
@@ -63,3 +64,107 @@ export function * watchNodeFocus({configuration}) {
...
+
+        const checkIfChildrenAreFullyLoadedRecursively = contextPath => {
+            return $get(['cr', 'nodes', 'byContextPath', contextPath, 'children'], state)
+            .filter(childEnvelope => nodeTypesRegistry.hasRole(childEnvelope.nodeType, 'content') || nodeTypesRegistry.hasRole(childEnvelope.nodeType, 'contentCollection'))
+            .every(
+                childEnvelope =>
+                    $get(['cr', 'nodes', 'byContextPath', $get('contextPath', childEnvelope)], state)
+                    && checkIfChildrenAreFullyLoadedRecursively($get('contextPath', childEnvelope))
+            );
+        }
```
grebaldi, line 88:
```suggestion
                    $get(['cr', 'nodes', 'byContextPath', $get('contextPath', childEnvelope)], state) &&
                    checkIfChildrenAreFullyLoadedRecursively($get('contextPath', childEnvelope))
            );
        };
```

grebaldi (commented):
Hi @Sebobo,

I added another suggestion that should fix those eslint complaints that were introduced by my other suggestion. Sorry for that :sweat_smile:

PR #2928, opened 2021-07-24
TASK: Fix e2e tests

grebaldi:
TODO

Failures were related to a Guzzle deprecation fixed in https://github.com/neos/neos-development-collection/pull/3368
An unfixed version of Neos.Fusion got stuck in the CircleCI cache. Resetting cache keys as described in https://github.com/neos/neos-ui#just-the-end-to-end-tests-fail fixed that.

### DIFF
--- a/Tests/IntegrationTests/Fixtures/1Dimension/createNewNodes.e2e.js
+++ b/Tests/IntegrationTests/Fixtures/1Dimension/createNewNodes.e2e.js
@@ -158,7 +158,7 @@ test('Supports secondary inspector view for element editors', async t => {
     await t.switchToMainWindow();
 
     await t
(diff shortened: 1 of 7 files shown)


markusguenther:
@grebaldi As https://github.com/neos/neos-development-collection/pull/3368 is merged we can remove the Draft mode or not? 

Thanks for working the whole sunday 🙏

markusguenther (approved):
Thanks @grebaldi for digging into that.
Looks good to me 💚

grebaldi:
Hi @markusguenther,

the changes in here were supposed to go to some of the lower branches as well. But since none of them were actually the cause for the E2E tests failing, I think we can also leave it as-is. Wdyt?

Issue #2914, opened 2021-05-25
Allow custom previews in selectboxes


Sebobo:
### Description

Currently it's only possible to use FontAwesome Icons supplied with Neos as "icons" for the `SelectBoxEditor`.

It's often requested to also allow custom previews for each item like the `AssetEditor` or `NodeEditor` allow.

I implemented this already for a customer with a custom `SelectBoxEditor` that uses the `SelectBox_Option_MultiLineWithThumbnail` component for rendering and setting the `preview` option on the datasource instead of the `icon`.

Therefore I would integrate a new option that allows to switch to the other option component. This should be a rather small code change.

Sebobo:
Put the plugin into a package for older Neos versions https://github.com/Sebobo/Shel.Neos.PreviewSelectBox.

jonnitto:
This is freaking awesome! :heart:

Issue #2092, opened 2018-08-23
Move resourcepath logic onto serverside


mstruebing:
As mentioned in this PR: https://github.com/neos/neos-ui/pull/2087#issuecomment-415170992

The logic to transform resource uris from `yaml` files should be moved onto the server side as the current one can lead to errors.

grebaldi:
I've been looking into this and there's a couple of problems here:

1. To resolve this on the server-side, we can either use a dedicated `NodeTypePostprocessor` or an Aspect. Both aren't good options performance-wise, because we would iterate through all configured node properties and resolve resource URIs (potentially costly). This operation would be performed every time the NodeType schema is resolved, even though the resolution of those resource URIs is only ever relevant to the UI and should not add cost to any userland code.
2. The resolution of resource URIs (at least in the case of help thumbnails) is implemented incorrectly. A resource URI `resource://Vendor.Site/Thumbnails/my-editor.png` will be resolved to `/_Resources/Static/Packages/Vendor.Site/Thumbnails/my-editor.png`. This is wrong. What should resolve to `/_Resources/Static/Packages/Vendor.Site/Thumbnails/my-editor.png` is `resource://Vendor.Site/Public/Thumbnails/my-editor.png` (notice the `/Public/` segment). This means that everybody who has configured help thumbnails via resource URI has done so with a wrong resource URI. We won't get out of that one without a breaking change for those people :(

We also need to consider other places in which resource URIs are manually resolved or `/_Resources/Static/...` paths are hard-coded. The places I found are:

1. (no surprise) Property help thumbnails
2. `/_Resources/Static/Packages/Neos.Neos/Images/dummy-image.svg` is hard coded in some places
3. The `<ResourceIcon/>` component (introduced via https://github.com/neos/neos-ui/pull/2979)

For (1.) - considering the bigger picture - I'd suggest the following plan:

1. Move `Neos\Neos\Controller\Backend\SchemaController` into `Neos.Neos.Ui` (while leaving the original intact - this is only to be able to apply UI-specifics to that endpoint)
2. Resolve resource URIs in `SchemaController`

Case (2.) can be easily solved by either passing the dummy image URI as configuration or add the `dummy-image.svg` as a component to the build. I'd be fine with either one.

[...]

PR #2930, opened 2021-07-27
FEATURE: Support preview uris in datasource selectboxeditor

Sebobo:
**What I did**

Options returned from a datasource can now contain a `preview` with an absolute image uri which will then be shown in the dropdown.

**How I did it**

Use new option component supporting previews for data source based selectboxeditor.
Works almost identically to the AssetOption of the AssetEditor.

**How to verify it**

Bildschirmfoto 2021-07-27 um 08 42 42

Also added a test.

Resolves: #2914

### DIFF
--- a/packages/neos-ui-editors/src/Editors/SelectBox/DataSourceBasedSelectBoxEditor.js
+++ b/packages/neos-ui-editors/src/Editors/SelectBox/DataSourceBasedSelectBoxEditor.js
@@ -7,6 +7,7 @@ import MultiSelectBox from '@neos-project/react-ui-components/src/MultiSelectBox
 import {selectors} from '@neos-project/neos-ui-redux-store';
 import {neos} from '@neos-project/neos-ui-decorators';
 import {shouldDisplaySearchBox, searchOptions, processSelectBoxOptions} from './SelectBoxHelpers';
+import PreviewOption from '../../Library/PreviewOption';
 
 const getDataLoaderOptionsForProps = props => ({
     contextNodePath: props.focusedNodePath,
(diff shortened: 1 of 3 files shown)


markusguenther (approved):
Works pretty fine and is a great improvement for the UI.

jonnitto (approved):
Really nice addition, like always ❤️

JPLudin:
In contrast to your plugin Shel.Neos.PreviewSelectBox this seems to works only for data sources, not for manually defined values, is that on purpose?

```
            'iconName':
              type: string
              ui:
                label: Icon
                inspector:
                  group: general
                  editorOptions:
                    values:
                      activity:
                        preview: /Resources/Private/Fusion/Presentation/Atom/Icon/Assets/activity.svg
                        label: activity
                      ..
```

mhsdesign:
I guess it should be a in the format of a flow resource stream wrapper: `resource://My.Package/Public/lol.svg` but you might need to (wrongly) leave out the public part see https://github.com/neos/neos-ui/issues/2092#issuecomment-1606055787

rolandschuetz:
No, neither version works

Sebobo:
Hi @JPLudin, both my plugin and this change do the same, and both mention that you should absolute URIs, as relative URIs are not reliable.
If one works for you I think this must be a coincidence depending on the environment.

You can check in the dev tools which uri it tries to open and then adjust your uri. Or use a datasource and properly resolve it to absolute Uris to have no problem with environments/hosting, etc.

rolandschuetz:
@Sebobo As described above, exactly this code works with your plugin, but not with the code.

Issue #2867, opened 2020-12-17
AssetEditor / AssetUpload does not send the correct propertyName to the server


Benjamin-K:
### Description

I tried to add an AssetManipulator to automatically tag assets with predefined tags. The problem is: The propertyName of the asset field is not send to the server / is always an empty string as the propertyName is not handed over from the AssetEditor to AssetUpload component. That's why the default value (`''`)  is used.

### Steps to Reproduce

1. Create a node type with an asset property (or use an existing one).
2. Go to the Neos backend, then select or create a node type with an asset.
3. Open the developer console (Network tab).
4. Upload a new asset on that property.
5. Check which data has been sent and see, that the `propertyName` always contains an empty string.

#### Expected behavior

The `propertyName` contains the correct value as defined in the nodetype definition.

#### Actual behavior

The `propertyName` is always an empty string.

### Affected Versions

Neos: 6.3.1
UI: 5.3.1

(I did not yet checked the lowest affected version)

This issue replaces neos/neos-development-collection#3253

Benjamin-K:
Affected files:

- https://github.com/neos/neos-ui/blob/master/packages/neos-ui-editors/src/Editors/AssetEditor/index.js#L212
- https://github.com/neos/neos-ui/blob/master/packages/neos-ui-editors/src/Library/AssetUpload.js

PR #2932, opened 2021-07-29
BUGFIX: Provide propertyname on asset upload

Sebobo:
**What I did**

The AssetEditor receives the `propertyName` as in its `identifier` prop. This prop had to be forwarded to the `AssetUpload` component.

Resolves: #2867 

**How to verify it**

See #2867


grebaldi (approved):
Verified manually:
image

Works like a charm :+1:

PR #2933, opened 2021-07-29
TASK: Upgrade postcss and related dependencies

markusguenther:
Updates the postcss related dependencies to newer versions.


mhsdesign:
i think we can close this. Since now having yarn 3 its not easy to bump dependencies in a bugfix...
And either way we are about to get rid of some post css dependencies or is this somehow security relevant?

Sebobo:
Old, if required let's make a new one. And I still don't know why we need to think about security in our frontend context. Security is done in PHP IMHO.

Issue #1717, opened 2018-03-23
Properties in inspector should be hidden on restricted document nodes


ComiR:
#### Steps to Reproduce
1. Create a `Neos\Neos\Security\Authorization\Privilege\NodeTreePrivilege` for a document node
2. As a user without said privilege, go to that document node

#### Expected behavior
The inspector stays empty.

#### Actual behavior
The inspector presents to you every property to edit (which fails obviously/"fortunately")

### Affected Versions
Neos: 3.3.9
UI: 1.0.9

daniellienert:
This issue seems still open. I encounter the same behavior on current master (2018-12-10) when using an edit node privilege.

grebaldi:
I'm currently trying to reproduce the issue in 5.3, where the situation seems to have worsened. Instead of the faulty inspector behavior described above, I get a blank page and a null pointer:

![image](https://user-images.githubusercontent.com/2522299/127741443-21619a14-6fc1-4740-b14b-0292f721217a.png)

I'll try my best to come up with a fix for both :)

Issue #2913, opened 2021-05-17
Backend crashes if no nodes are available


PRGfx:
### Description

If a user has no access to a site i.e. through a `NodeTreePrivilege`, the backend can't load any nodes and the UI crashes.

### Steps to Reproduce

1. setup a neos-base-distribution
2. setup "content"
    ```
    ./flow doctrine:migrate
    
    ./flow site:create --node-name site1 --node-type Neos.Demo:Document.Homepage --package-key Neos.Demo --name Site1
    ./flow domain:add --site-node-name site1 --hostname site1.local
    
    ./flow site:create --node-name site2 --node-type Neos.Demo:Document.Homepage --package-key Neos.Demo --name Site2
    ./flow domain:add --site-node-name site2 --hostname site2.local
    
    ./flow user:create admin password Ad Min --roles Neos.Neos:Administrator,Neos.Demo:MyTestEditor2
    ```
3. configure policies
    ```yaml
    privilegeTargets:
      'Neos\Neos\Security\Authorization\Privilege\NodeTreePrivilege':
        'Neos.Demo:MyTestSite1':
          matcher: 'isDescendantNodeOf("/sites/site1")'
        'Neos.Demo:MyTestSite2':
          matcher: 'isDescendantNodeOf("/sites/site2")'
    roles:
      'Neos.Demo:MyTestEditor1':
        parentRoles:
          - 'Neos.Neos:RestrictedEditor'
        privileges:
          -
            privilegeTarget: 'Neos.Demo:MyTestSite1'
            permission: GRANT
      'Neos.Demo:MyTestEditor2':
        parentRoles:
          - 'Neos.Neos:RestrictedEditor'
        privileges:
          -
            privilegeTarget: 'Neos.Demo:MyTestSite2'
            permission: GRANT
    ```
4. visit the backend for Site1 and Site2

#### Expected behavior

Site2 should work fine, but for Site1 the content-module should render some kind of error message that the current site is not available/the user is not allowed to access this backend-module. Maybe if the user has access to other sites, there could be a selection to switch the site.

#### Actual behavior

Site2 works fine, but for Site2 the react app unmounts, resulting in a white screen.

<details>
<summary>stacktrace</summary>

```
TypeError: can't convert null to object
    preprocessViewConfiguration index.js:134
    preprocessViewConfigurationDebounced index.js:158
    Lodash 3
    componentDidUpdate index.js:111
    React 2
    unstable_runWithPriority scheduler.production.min.js:18
[...]

Issue #2172, opened 2018-10-12
Opening a page without NodeTreePrivilege causes errors


ComiR:
### Description
If you navigate to a document node you are allowed to read, but don't have the `NodeTreePrivilege`, the frame loads and suddenly the whole UI changes to complete blank due to js errors below:
<details>
<summary>console.error messages</summary>

```
utils.js:225 uncaught at initializeGuestFrame TypeError: Cannot read property 'children' of null
    at Neos.Neos.Ui/JavaScript/Host.js?ba57087a:73803:37
    at Array.filter (<anonymous>)
    at Neos.Neos.Ui/JavaScript/Host.js?ba57087a:73802:56
    at Neos.Neos.Ui/JavaScript/Host.js?ba57087a:159323:20
    at Neos.Neos.Ui/JavaScript/Host.js?ba57087a:159316:24
    at Array.reduce (<anonymous>)
    at Neos.Neos.Ui/JavaScript/Host.js?ba57087a:159315:29
    at Neos.Neos.Ui/JavaScript/Host.js?ba57087a:159316:24
    at Array.reduce (<anonymous>)
    at Neos.Neos.Ui/JavaScript/Host.js?ba57087a:159315:29
```
```
checkPropTypes.js:19 Warning: Failed prop type: The prop `documentNode` is marked as required in `NodeVariantCreationDialog`, but its value is `undefined`.
    in NodeVariantCreationDialog (created by Neos(NodeVariantCreationDialog))
    in Neos(NodeVariantCreationDialog) (created by Connect(Neos(NodeVariantCreationDialog)))
    in Connect(Neos(NodeVariantCreationDialog)) (created by Modals)
    in div (created by Modals)
    in Modals (created by Neos(Modals))
    in Neos(Modals) (created by App)
    in div (created by App)
    in App (created by Root)
    in Neos (created by Root)
    in Provider (created by Root)
    in div (created by Root)
    in Root (created by DragDropContext(Root))
    in DragDropContext(Root)
```
```
react.development.js:217 Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: null.

Check the render method of `SecondaryToolbar`.
    in SecondaryToolbar (created by Connect(SecondaryToolbar))
    in Connect(SecondaryToolbar) (created by Neos(Connect(SecondaryToolbar)))
    in Neos(Connect(SecondaryToolbar)) (created by App)
    in div (created by App)
    in App (created by Root)
[...]

PR #2935, opened 2021-08-01
BUGFIX: Make Guest Frame and Inspector more resilient when there's no focusable document

grebaldi:
**What I did**

I prevented the UI from crashing when an account without respective privileges navigates to a document that is matched by a `NodeTreePrivilege`.

**How I did it**

I added several null-checks in places that caused runtime errors when a `NodeTreePrivilege` prevented access to augmented document metadata.

**How to verify it**

1. You'll need a configured NodeTreePrivilege that hides at least one node within the document tree. An example would be:

```yaml
privilegeTargets:
  'Neos\Neos\Security\Authorization\Privilege\NodeTreePrivilege':
    'Vendor.Site:MyNodeTreePrivilege':
      matcher: 'nodeIsOfType("Vendor.Site:Document.Privileged")'

roles:
  'Neos.Neos:Administrator':
    privileges:
      -
        privilegeTarget: 'Vendor.Site:MyNodeTreePrivilege'
        permission: GRANT
```
(This would hide all nodes of type `Vendor.Site:Document.Privileged`)

2. Try to reach the restricted document with an account that doesn't have the above privilege by:
  - Linking to the document, or
  - Entering the backend-URI of the document directly

**Before the fix** you would have seen a blank screen with a console error:
127741443-21619a14-6fc1-4740-b14b-0292f721217a

**After the fix** you should see the document without editable content and with an empty inspector.

fixes: #1717,  #2172


grebaldi:

After the bugfix release of Neos.Fusion.Form, the E2E tests of Neos.Neos.Ui just need to be re-run with an updated cache key prefix (as described here: https://github.com/neos/neos-ui#just-the-end-to-end-tests-fail).

Sebobo:
Would probably also fix #2913 ?

grebaldi:
Sounds like it. I'll try to reproduce that locally and verify.

Sebobo (approved):
Works as expected 👍  
Though I still get quite a few proptypes warning in the console. But they don't cause a crash, so that's fine for now.

Sebobo:
Should we still merge this one now, or do you want to wait for the e2e tests to work again?

grebaldi:
I'd say wait, so we can verify E2E-Tests without having to worry about too many additional changes that were introduced in the meantime.

PR #2936, opened 2021-08-01
TASK: Add docker-compose setup to run E2E tests locally

grebaldi:
**What I did**

Since I had continuous trouble running the E2E locally on my machine in an efficient way, I came up with a docker-compose setup to ease that task. So, I thought I share it :)

Currently, running the tests includes a lot of (automated) setup work, so each test run takes quite a bit of time. The setup is designed to leave the repository in a clean state, which means that the local composer dependencies have to be installed with each run. I couldn't figure out how to cache that. The global composer cache of the php-image however lives in a docker volume, so at least, subsequent runs are significantly faster than the very first.

**How to verify it**

When the PR branch is checked out, simply run:

```sh
make clean
make setup
make test-e2e-docker
```

By default, the setup will try to run `chromium`, but you can also use a different browser by passing the `browser` env variable to the `test-e2e-docker` make target:

```sh
make test-e2e-docker browser=firefox
```


grebaldi:

After the bugfix release of Neos.Fusion.Form, the E2E tests of Neos.Neos.Ui just need to be re-run with an updated cache key prefix (as described here: https://github.com/neos/neos-ui#just-the-end-to-end-tests-fail).

Issue #2916, opened 2021-05-27
White page after discard all


gerdemann:
### Description

I get a white page when I click on "discard all" and then want to edit something inline again.

### Steps to Reproduce

1. Edit some properties
2. Click on "discard all"
3. Click in an inline editable field

#### Expected behavior

I can edit the property.

#### Actual behavior

I get a white page with the following error in console:
<details>
<summary>stacktrace</summary>
<br>
<pre>
index.js:43 Uncaught TypeError: Cannot read property 'contextPath' of null
    at InlineUI.render (index.js:43)
    at gi (react-dom.production.min.js:182)
    at fi (react-dom.production.min.js:181)
    at xo (react-dom.production.min.js:263)
    at Qj (react-dom.production.min.js:246)
    at Kj (react-dom.production.min.js:246)
    at yj (react-dom.production.min.js:239)
    at react-dom.production.min.js:123
    at w.unstable_runWithPriority (scheduler.production.min.js:18)
    at cg (react-dom.production.min.js:122)
render @ index.js:43
gi @ react-dom.production.min.js:182
fi @ react-dom.production.min.js:181
xo @ react-dom.production.min.js:263
Qj @ react-dom.production.min.js:246
Kj @ react-dom.production.min.js:246
yj @ react-dom.production.min.js:239
(anonymous) @ react-dom.production.min.js:123
w.unstable_runWithPriority @ scheduler.production.min.js:18
cg @ react-dom.production.min.js:122
fg @ react-dom.production.min.js:123
gg @ react-dom.production.min.js:122
Mj @ react-dom.production.min.js:240
notify @ Subscription.js:23
notifyNestedSubs @ Subscription.js:65
handleChangeWrapper @ Subscription.js:70
dispatch @ redux.js:221
(anonymous) @ clipboardMiddleware.js:22
(anonymous) @ localStorageMiddleware.js:21
(anonymous) @ middleware.js:66
focusSelectedNode @ initializeGuestFrame.js:110
(anonymous) @ initializeGuestFrame.js:125
</pre>
</details>

### Affected Versions

Neos: 5.3
UI: 5.3.8

dlubitz:
Same here! I've noticed it multiple times but couldn't reproduce it. Thanks @gerdemann .

I could narrow it down a bit more. The error happens if you click an inline editable element AFTER `/neos/preview?node=....` request has been started and BEFORE you get the response. 

After getting the response page is fully reloaded and selecting elements works as expected.

[...]

PR #2942, opened 2021-08-11
BUGFIX: Prevent error when trying to edit a node during discard

Sebobo:
Resolves: #2916


grebaldi (commented):
Code looks fine, but unfortunately doesn't solve #2916 for me. I left a comment with more detail at the issue.

grebaldi:
Maybe we can put a transparent overlay over the guest frame while it is busy to prevent the UI from crashing. Wdyt, @Sebobo?

Sebobo:
@grebaldi ok thx for checking. 

The overlay sounds like a good idea as we probably want to prevent any interaction. 
Maybe making it very slightly opaque (darken?) and with a different cursor?
So it's clear that the user has to wait a tiny bit. But it should not flicker.

grebaldi:
Sounds good. I would say, it shouldn't be darkened by much (maybe just 10%). Appropriate cursor would be `wait`, I think. To prevent flickering, I'd suggest to let the overlay enter and exit with a very slight transition (maybe only in the 50-100ms range). Wdyt?

Issue #2909, opened 2021-05-05
ImageCropper provides unnecessary controls


PRGfx:
### Description

![image](https://user-images.githubusercontent.com/3868790/117126186-b51bd100-ad9a-11eb-94a9-309674424534.png)

- The AspectRatioDropDown provides an option to clear the selection even if `allowCustom` is set to `false`. I expect it to only let me choose one of the aspect ratio `options`
- I get the `AspectRatioItem` controls for changing the aspect ratio of the selected preset. I would only expect these controls if `allowCustom` is `true` and no preset is selected.

It somewhat undermines the concept of cropping presets

### Steps to Reproduce

1. Fresh 5.3 Neos project, Neos.Demo imported
2. disable custom aspect ratio
 ```yaml
'Neos.Demo:Content.Image':
  properties:
    image:
      ui:
        inspector:
          editorOptions:
            crop:
              aspectRatio:
                allowCustom: false
```
3. crop image in inspector

### Affected Versions

tested with the following versions but as there weren't changes to the editor in a while, maybe longer

Neos: 5.3.2
UI: 5.3.7

PR #2943, opened 2021-08-11
BUGFIX: Prevent custom image ratios when disabled

Sebobo:
The option `allowCustom` hides the dropdown selection
for a custom ratio, but without this patch the controlls still
allowed setting a custom ratio by modifying the numbers.

Now the ratio controls are disabled if the custom ratio is disabled.

Resolves: #2909

Bildschirmfoto 2021-08-11 um 13 23 48


grebaldi (approved):
Thanks for taking care of this, @Sebobo!

Code looks good and I could verify manually that everything works as expected :+1:

PR #2944, opened 2021-08-13
BUGFIX: Scroll focused content tree node into view

Sebobo:
Fixes: ##2872

**What I did**

Scroll to the focused node in the content tree if necessary.

**How I did it**

Replace the animation library call with native browser smooth scroll feature.
Smooth scroll doesn't work in Safari yet, where it just jumps which IMO is fine until they support it.

**How to verify it**

Click on nodes in the guest frame on a page with a lot of nodes

Example:

https://user-images.githubusercontent.com/596967/129373547-8b02b803-5070-4812-be9e-b96109edba0d.mov


Sebobo:
I actually don't know why the old code didn't work anymore, but as the amator animation library is only used in two places, I think we should just get rid of it soonish.

jonnitto (approved):

PR #2945, opened 2021-08-28
FEATURE: Consumer Api for @neos-project/neos-ui-guest-frame

mhsdesign:
This makes it possible to reuse the guest iframe api for getting the guest iframe and fx. dispatching events.

you can then create a custom extension and import from '@neos-project/neos-ui-guest-frame' as if you have this file '@neos-project/neos-ui-guest-frame/src/dom' https://github.com/neos/neos-ui/blob/master/packages/neos-ui-guest-frame/src/dom.js installed.

```js
import {getGuestFrameDocument, dispatchCustomEvent} from '@neos-project/neos-ui-guest-frame'
```

slack discussion over here: https://neos-project.slack.com/archives/C0U0KEGDQ/p1629923508013600

> **Note:**
> 
> Internally it seems that to access the ./dom file of the @neos-project/neos-ui-guest-frame one need to import directly from it like:
> 
> ```js
> import {getGuestFrameDocument, dispatchCustomEvent} from '@neos-project/neos-ui-guest-frame/src/dom'
> ```
> (i hope that its not to confusing that its now aliased by '@neos-project/neos-ui-guest-frame'.
> I took an examle from the '@neos-project/neos-ui-ckeditor5-bindings/src/ckEditorApi' wich is accessible like '@neos-project/neos-ui-ckeditor5-bindings' through the consumer api)
> 

**What I did**
The already compliled neos ui sets on the window object imported code from certain modules: https://github.com/neos/neos-ui/blob/master/packages/neos-ui/src/apiExposureMap.js
I added here an Object which contains everything from the '@neos-project/neos-ui-guest-frame/src/dom'

The "@neos-project/neos-ui-extensibility" package allows to use 'normal' imports, which will be resolved by getting the object set on the window.

### DIFF
--- a/packages/neos-ui-extensibility/scripts/helpers/webpack.config.js
+++ b/packages/neos-ui-extensibility/scripts/helpers/webpack.config.js
@@ -70,6 +70,7 @@ module.exports = function (neosPackageJson) {
                 '@neos-project/neos-ui-i18n': '@neos-project/neos-ui-extensibility/src/shims/neosProjectPackages/neos-ui-i18n/index',
                 '@neos-project/neos-ui-redux-store': '@neos-project/neos-ui-extensibility/src/shims/neosProjectPackages/neos-ui-redux-store/index',
                 '@neos-project/neos-ui-views': '@neos-project/neos-ui-extensibility/src/shims/neosProjectPackages/neos-ui-views/index',
+                '@neos-project/neos-ui-guest-frame': '@neos-project/neos-ui-extensibility/src/shims/neosProjectPackages/neos-ui-guest-frame/index',
                 '@neos-project/utils-redux': '@neos-project/neos-ui-extensibility/src/shims/neosProjectPackages/utils-redux/index'
             }
         }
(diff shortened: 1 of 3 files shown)


grebaldi (approved):
Hi @mhsdesign,

:+1: by reading. I'm not able to test this locally right now, but I think this change makes perfect sense.

mhsdesign:
btw i tested it locally by using the local package link feature of yarn for the neos-ui-extensibility package. And created so an working extension which is importing from the '@neos-project/neos-ui-guest-frame' package.

markusguenther (approved):
Thanks for this PR. Great that you fought your way through there and solved your problem in Slack. 

Looks pretty solid 👍

mhsdesign:
Btw, since this is only available since Neos.Ui 7.2 https://github.com/neos/neos-ui/releases/tag/7.2.0
i made a polyfill: https://github.com/mhsdesign/MhsDesign.PolyfillNeosUiGuestFrameApi

PR #2975, opened 2021-10-24
BUGFIX: EditNodePrivilege combined with EditNodePropertyPrivilege

ComiR:
This is a rebased version of #2165:

> This PR contains multiple fixes to get to the point where the behavior of the UI corresponds to the behavior of the ContentRepository.
>
> When you get an abstaining EditNodePrivilege (resulting in a "soft" DENY) in combination with an EditNodePropertyPrivilege GRANT, the UI ignores the latter and denies editing. The ContentRepository on the other hand (or rather Flows SecurityInterceptor) sees both privileges matching, one abstaining and the other granting, so it will allow access to the editProperty method.

Please cautiously check if the rebased version still works as intended as I only rebased this "blindly" (i.e. no running Neos instance) since I currently don't have a running dev environment and unfortunately not enough time to set it up atm.

### DIFF
--- a/Classes/Aspects/AugmentationAspect.php
+++ b/Classes/Aspects/AugmentationAspect.php
@@ -12,7 +12,6 @@ namespace Neos\Neos\Ui\Aspects;
  */
 
 use Neos\ContentRepository\Domain\Model\NodeInterface;
-use Neos\ContentRepository\Service\AuthorizationService;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Aop\JoinPointInterface;
 use Neos\Flow\Persistence\Exception\IllegalObjectTypeException;
@@ -22,6 +21,7 @@ use Neos\Fusion\Service\HtmlAugmenter;
 use Neos\Neos\Domain\Service\ContentContext;
 use Neos\Neos\Ui\Domain\Service\UserLocaleService;
 use Neos\Neos\Ui\Fusion\Helper\NodeInfoHelper;
+use Neos\Neos\Ui\Service\NodePolicyService;
 
 /**
  * - Serialize all nodes related to the currently rendered document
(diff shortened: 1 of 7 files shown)


markusguenther:
Would be nice if @bwaidelich and/or @kitsunet could also take a look at this.

jonnitto:
Perhaps we get @ComiR one time back to the Neos Universe 😉

Classes/Aspects/AugmentationAspect.php
```diff
@@ -193,16 +193,42 @@ public function editableElementAugmentation(JoinPointInterface $joinPoint)
...
     /**
      * @param NodeInterface $node
-     * @param boolean $renderCurrentDocumentMetadata
-     * @return boolean
+     * @param bool $renderCurrentDocumentMetadata
+     * @return bool
      * @throws IllegalObjectTypeException
      */
-    protected function needsMetadata(NodeInterface $node, $renderCurrentDocumentMetadata)
+    protected function needsMetadata(NodeInterface $node, bool $renderCurrentDocumentMetadata): bool
```
kitsunet, line 200: Currently trying to get more insights here, I think we shouldn't only look at inline editable properties, as we probably need the metadata also if only inspector properties are allowed. And at that point I am really not sure anymore if the performance gains are still there or if we maybe just skip that check altogether and just check for isInBackend() but that is TBD.

kitsunet (commented):
Only commenting for now, i will do a second pass, but I guess the change itself will work and fix the issue just fine, but as I comment I am not sure we should do this in the UI. We have no more generalised lower level code for this at the moment and would need to create it.

Note to self:
How to evaluate if $something is allowed. especially methods, and additionally: is what some NodePrivileges do a good idea (do some split code path between method and node privilege subjects)

markusguenther:
Thanks for the notes @kitsunet ... 
I would not be able to check that :)

mhsdesign:
I would like to close this as this has not received input in a long time - original from 2018, here the rebased untested version for 5.3 - and is also not applicable for Neos 9.x

PR #2979, opened 2021-11-09
FEATURE: Adds resource icon component 

markusguenther:
With these change, the icon component from a font-awesome only version to a more flexible component.
The idea is to have the possibility to use all icons the integrator wants.

In the first step, this change extracts the current icon behavior to an own FontAwesomeIcon component and adds a new ResourceIcon component. The icon component acts then as a de facto factory that checks if the icon property contains a resource URI or just a regular icon string.

The font-awesome icons are the default, and Neos will come now with an additional resource based SVG Icon.
In future, it would be nice to have something like Icon Providers so that people are able to easily register another IconProvider. For instance, a MaterialUIIconProivder.

The regular icons will work like before, but we are able to use now also resources like that:

```
'Neos.Demo:Document.Chapter':
  ui:
    icon: 'resource://Neos.Demo/Images/screen-heart.svg'
```

The integrators should now just be aware of colors and viewports of the SVG files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -175,6 +175,7 @@
     "react-motion": "^0.5.0",
     "react-portal": "^4.2.0",
     "react-redux": "^7.1.3",
+    "react-svg": "^11.1.2",
     "react-test-renderer": "^16.12.0",
     "react-textarea-autosize": "^8.3.0",
     "recharts": "^1.0.0-alpha.6",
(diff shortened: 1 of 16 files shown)


markusguenther:
Current state is that we get the custom icons.
Need to make it work with the current node version.

Screenshot 2021-11-09 at 23 00 43
Screenshot 2021-11-09 at 23 00 33

jonnitto (approved):
Fine by reading

Kleisli:
@markusguenther Thank you, a small but great improvement! 
I was just going down that rabbit hole, but was lost in the react code..

markusguenther:
@Kleisli It is ok when you don't know all about the React part. But great that you like it :)

markusguenther:
We need to document this feature as well :)

PR #2980, opened 2021-11-10
TASK: Increase node version 16

markusguenther:
**What I did**
Changed the node version from version 14 to 16.
Node 16 is the current active LTS.

**How I did it**
Change the version for nvm and in the package.json


markusguenther:
Some of our unit tests actually does not work with node 16.
For instance:
 * SelectBox tests from react-ui-editors
 * saveHooks from neos-ui-sagas
 * FlowQuery from neos-ui-backend-connector

PR #2985, opened 2021-11-22
BUGFIX: Set color of links in help messages

jonnitto:
Links in help messages are very hard to read, so I set the color.

### Before (creation dialog):
CleanShot 2021-11-22 at 11 06 49@2x

### After (creation dialog):
CleanShot 2021-11-22 at 11 07 03@2x

### Before (inspector):
CleanShot 2021-11-22 at 11 07 19@2x

### After (inspector):
CleanShot 2021-11-22 at 11 07 30@2x


markusguenther (approved):
Thanks for this tiny but nice improvement 💙

PR #2986, opened 2021-11-22
FEATURE: Add range editor

jonnitto:
Include the range editor from https://github.com/Sebobo/Shel.Neos.RangeEditor


markusguenther (approved):
Thanks to @Sebobo for the foundational work and @jonnitto for bringing that to Neos.UI.

markusguenther:
Nice new feature.

**_Suggestion_**

- In a next iteration, usage of Neos constants for the colors (Neos blue for instance)
- When the editor is disabled, also change the blue slider to greyish color
- more space between unit and the input field
- make default value work :)

But that are little tweaks we can achieve in the next step :)

markusguenther:
We need to document this editor as well :)

jonnitto:
Issues are fixed: https://github.com/neos/neos-ui/pull/2998

PR #2988, opened 2021-11-30
BUGFIX: Allow more domains in the email address validator

### LINKED ISSUE #2987: Email Validator doesn't allow certain domains
Some domains aren't allowed by the email address validator, e.g. `.swiss`.

There are some TLD defined here: 
https://github.com/neos/neos-ui/blob/fbda2de621f82a5a33169265fd643e16588dd737/packages/neos-ui-validators/src/EmailAddress/index.tsx#L9

Does it make sense to have a hardcoded list like that? I mean, there are popping more and more Top-Level-Domains up all the time.

CleanShot 2021-11-30 at 15 46 58@2x

jonnitto:
Fix #2987

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,6 +131,7 @@
     "he": "^1.1.1",
     "hoist-non-react-statics": "^3.3.1",
     "immer": "^2.1.4",
+    "isemail": "^3.2.0",
     "lodash.assign": "^4.2.0",
     "lodash.debounce": "^4.0.8",
     "lodash.defaultsdeep": "^4.6.1",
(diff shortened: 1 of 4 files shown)


jonnitto:
I think the test will fail because of `email@domain.web`…

markusguenther:
I now removed the regex and use the isemail library as a replacement.
To be more compatible with the PHP part, we now also use the test cases of the flow EmailAddressValidatorTest.

The reason is that we had four addresses in the UI unit tests that should be invalid, but they are valid and the PHP Validator also returns valid in these cases.

* あいうえお@domain.com
* email@domain.web
* email@domain
* email@111.222.333.44444

jonnitto:
Approve from my side

PR #2989, opened 2021-12-03
FEATURE: Expose parentNode for ClientEval in right side bar

markusguenther:
In some circumstances, we need to use ClientEval to hide a property, for instance. Actually it is just possible to use the current node data or some child information. But we have not much information about the parent.

As the neos-ui knows which node is the parent, it is quite helpful to use also parent node information in ClientEval. This PR adds the parent node to the context so that we have `node` and `parentNode` available in ClientEval.

**What I did**
Added the parentNode of the current focused node to the props, and therefore it is possible to use the parentNode also in the context of ClientEval.

**How to verify it**
If you have a node type property, you are able to use node and parentNode in the ClientEval.

```
'chapterImage':
      ui:
        inspector:
          hidden: 'ClientEval:console.log(parentNode)'
```

Resolves: #2990

### DIFF
--- a/packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
+++ b/packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js
@@ -35,11 +35,14 @@ import style from './style.css';
         const shouldPromptToHandleUnappliedChanges = selectors.UI.Inspector.shouldPromptToHandleUnappliedChanges(state);
         const shouldShowUnappliedChangesOverlay = isDirty && !shouldPromptToHandleUnappliedChanges;
         const shouldShowSecondaryInspector = selectors.UI.Inspector.shouldShowSecondaryInspector(state);
+        const focusedNode = selectors.CR.Nodes.focusedSelector(state);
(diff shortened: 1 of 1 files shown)


jonnitto (approved):
By reading

Sebobo:
Awesome!

Did you check if this works with referenced nodes where the parent might not has been loaded?

markusguenther:
> Awesome!
> 
> Did you check if this works with referenced nodes where the parent might not has been loaded?

Did not test it and TBH don't know how to reproduce that with the Demo package 🙈

Sebobo:
> > Awesome!
> > Did you check if this works with referenced nodes where the parent might not has been loaded?
> 
> Did not test it and TBH don't know how to reproduce that with the Demo package 🙈

It should be possible to check that if you create a content reference to a node which has some kind of special behaviour based on its parent. The referenced node should behave the save even though the parent might not be rendered on the same page.

I fear that we have cases in which the parent was not requested from the backend and the inspector would then behave different for the same node.

markusguenther:
Thanks for the Feedback @Sebobo 

I would say we take it as it is now and when we have a solid test case, we fix eventual bugs.
It adds just a new option and does not break the behavior of the old CientEvals with `node`.

PR #2991, opened 2021-12-06
BUGFIX: Open links in new window

jonnitto:
When links are defined in the help message, they open currently always in the same window. With this fix, the will open in a new window


markusguenther (approved):
Tested and works like charm.

Simply add  a markdown link to the help message.

```
ui:
    help:
      message: 'The Chapter node will work with all other chapter nodes on the same level to form a browsable book. [google](https://google.com)'
```

PR #2994, opened 2021-12-12
BUGFIX: publishing readonly workspace

### LINKED ISSUE #2796: A restricted editor is able to click on publish to live and gets raw exceptions back
### Description

restrictedEditor

### Steps to Reproduce

1. Log with an user with the RestrictedEditor role
2. Change something
3. Hit the publish to live button

#### Expected behavior

Publish to live is not enabled even when changes are made

Better: The editor cannot change content when own workspace is targeted to live

#### Actual behavior
See above

### Affected Versions

Neos: 5.3

UI: 5.3

markusguenther:
**What I did**
The readOnly property has always been overwritten when the content has been updated.
Therefore, the publishing assumed to be allowed, even when the workspace was readOnly.

We now take care that the state keeps the readOnly property of the personal workspace.

**How to verify it**
Create a restricted editor and try to edit content on live.

closes #2796


markusguenther:
Example Video after the fix.
https://user-images.githubusercontent.com/1014126/145717140-5fe560f0-bb70-4e64-97fc-bc23faf9188e.mp4

markusguenther:
Question: Should we also prevent editing and creating new content on read only, or is this to much breaking so that we take it as a feature for 8.0?

markusguenther:
Found another little issue on changing the workspace. The server feedback never returned the current value for the workspace permission. So how should the UI decide if a user is able to write or not.

lorenzulrich:
> Question: Should we also prevent editing and creating new content on read only, or is this to much breaking so that we take it as a feature for 8.0?

In my opinion, it is highly illogical to be able to edit in a workspace you don't have write access to. Editing content is already disabled if you are in preview central (at least inline editing), so I assume the UI has some mechanism to work that way.

jonnitto (approved):

markusguenther:
@jonnitto As you already approved the PR. Would you like to see the prevention of editing also as part of this bug fix PR? Or do you see that more as a feature? I am really unsure as it is fixing a UX bug but is also a bit breaking the behavior.

Technically, the user edit the content in its own personal workspace and can just not publish that to the base workspace.
From an UX perspective, it does not make sense to be able to edit that.

jonnitto:
For me it is a bugfix

markusguenther:
OK Karsten gave a thumbs up in Slack, which I take as approval. So will prevent editing and creation of content later :)

markusguenther:
Adjusted now the PR so that we are not able to edit content, create new content or move content, when the workspace is in read only mode.

We have the following changes:
-  disable the buttons in the document tree toolbar and the content tree toolbar
- disable dragging of nodes in the trees
- disable inline editing
- hide the inline editable toolbar as the scroll to node should be possible but when we remove the toolbar completely we need to reimplement the focus stuff when you click on the node in the content tree.
- disable editors in the ride sidebar
- disable apply button in the right sidebar and the dialog

Hope that I have catched everything.

jonnitto (approved):

Issue #2978, opened 2021-11-09
The appearance of some checkboxes in Neos might be confusing for some users 


MarcoSchiffmann:
### Description
In one of our customer projects I discovered that some of the checkboxes seem to confuse editors. In this particular case the checkboxes for the search lead to some confusion.

For the editor the checked or unchecked state is not clearly visible. They would expect that the checkbox is empty when unchecked instead in Neos the checkbox is just not highlighted anymore.

<img width="286" alt="Bildschirmfoto 2021-11-09 um 14 49 24" src="https://user-images.githubusercontent.com/93990638/140935759-8951c77c-3d11-4f63-9e80-e0f81e0f37f0.png">

Hence when checking and unchecking checkboxes in the backend the appearance of the checkbox should change to avoid confusion. Simply highlighting is not sufficient in my eyes.

### Steps to Reproduce

1. Select any checkbox in the backend 
2. Try to check and uncheck the checkbox 

#### Expected behavior

Checkbox should be empty when unchecked and filled out or checked when activated.

Something like this, state unchecked: 

<img width="126" alt="Bildschirmfoto 2021-11-09 um 14 59 24" src="https://user-images.githubusercontent.com/93990638/140937446-c674edae-99c6-44b7-998d-63696c122396.png">

#### Actual behavior

Checkbox is not highlighted when unchecked or highlighted when checked

markusguenther:
Hello Marco!

As you describe this now, it makes totally sense that it could be confusion to editors.
I am a bit unsure if we declare this as a bug, but if no one complains against that, I would handle it as a bug.
As it is more or less a UI glitch.

I hope that others also would categorize it like that.

WDYT @dimaip @Sebobo @jonnitto ??

MarcoSchiffmann:
Hi Markus, thanks for picking this up. If you need any input from the client/customer perspective just let me know.

dimaip:
I agree, the standard UX pattern is just to have an empty square, e.g. 
![image](https://user-images.githubusercontent.com/837032/145808987-86c3c34e-f778-459b-b6ab-0d708026e12f.png)

markusguenther:
Ok, thanks for your input. So let's wait for a review, and then we have this issue less :)
BTW great to see you @dimaip :)

markusguenther:
This fix will be deployed with Version 5.3.13.
Thanks for reporting that issue, and for the feedback.

PR #2995, opened 2021-12-12
BUGFIX: Don't render check mark for unchecked checkbox

markusguenther:
For editors, it can be confusing that the checkbox renders a checkmark even when the checkbox is unchecked. The editor need to know that a gray out checkmark means that the checkbox is unchecked.

**What I did**
Conditionally render the checkmark icon.

**How to verify it**
Login to the backend and look for a checkbox in the right inspector.

resolves: #2978

### DIFF
--- a/packages/react-ui-components/src/CheckBox/__snapshots__/checkBox.spec.tsx.snap
+++ b/packages/react-ui-components/src/CheckBox/__snapshots__/checkBox.spec.tsx.snap
@@ -11,14 +11,6 @@ exports[`<CheckBox/> should render correctly. 1`] = `
     onChange={[Function]}
     type="checkbox"
   />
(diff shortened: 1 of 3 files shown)


markusguenther:
**Unchecked:**
Screenshot 2021-12-12 at 21 57 43

**Checked:**
Screenshot 2021-12-12 at 21 58 04

mhsdesign:
i was wondering if a border radius might fit better but the style does match the one of a text box so i guess a user will understand it...
image

markusguenther:
@mhsdesign Guess that will become important if we also support a light theme anytime :)

jonnitto (approved):

markusguenther:
Thanks for the review :)

PR #2996, opened 2021-12-15
FEATURE: Add custom validation message for input validation

### LINKED ISSUE #2997: Feature: Make Regex-Validation user-readable
When using the RegularExpressionValidator for input-fields, currently the validation shows an error message: **The given subject did not match the pattern (/^[a-z0-9\-]+$/i)** 

Regular users/editors mostly do not understand Regex. To help them understand what they are allowed to enter into the field a custom message would be helpful.

pKallert:
**What I did**
I added an option that allows people who understand regex to write custom validation error messages for users who do not understand regex. 

**How I did it**
I added an option called "validationErrorMessage" to the valodator options that is shown in the neos-ui if it is set. 

**How to verify it**
Add the option "customValidationError" in any yaml file where the Neos.Neos/Validation/RegularExpressionValidator is used. This is for example the uriPathSegment in the NodeTypes.Document.yaml in Neos.Neos. 

```
'Neos.Neos/Validation/RegularExpressionValidator':
          regularExpression: '/^[a-z0-9\-]+$/i'
          validationErrorMessage: 'Please enter only lowercase letters, numbers or minus'
```

Bildschirmfoto 2021-12-15 um 12 26 56

closes #2997


pKallert:
closes #2997

markusguenther:
Hello again @pKallert,

Thanks for that improvement. I will check that soon.
For instance, if the multilanguage handling works out of the box or if we need adjustments here.

pKallert:
I tried it out locally and it does not work with i18n, and I am not sure how to make it work.

jonnitto:
Awesome, thank you @pKallert! Already can't wait to have it ;-)

markusguenther:
To be able to use the localization, we need to extend the response of the node type schema endpoint.
Therefore, I have created another PR in the Neos Dev collection.

I guess it is better to offer the replacement of the validation message for every validator. So, it would be nice to use the option in every validator, to be more consistent. @pKallert WDYT and would you like to adjust the other validators as well?

https://github.com/neos/neos-development-collection/pull/3538

pKallert:
I will try 😄

pKallert:
@markusguenther  I added the validationErrorMessage to all Validations in the Neos UI I could find. I hope everything works 😅

jonnitto (approved):
Fine by reading

jonnitto (changes requested):
I think you need to adjust the tests, too

markusguenther:
Need to check that, but sometimes it is enough to update the snapshots.

I have no that PHP part ready and need a review here.
But guess I now caught all cases for the validation.
https://github.com/neos/neos-development-collection/pull/3538#issuecomment-995974196

markusguenther:
For the most cases of our validators, it works pretty well.
Some validators like the `StringLengthValidator` or `NumberRangeValidator` have lot of other returns and also use some hard-coded responses.

So guess for the future we need to extend that a bit, so that the special validators can have multiple custom labels.

Also found an issue with the `LabelValidator` in the docs and will change the docs here, as the option regularExpression is not used in the validator itself. But that just as side note.

jonnitto (approved):

jonnitto:
And here we have the first **awesome** feature for 8.0! Thank you @pKallert for pushing this forward!

markusguenther:
Guess when it is ok for the others. It is free to merge - the worst thing that can happen without the Neos/Neos adjustment is that the label is still an i18n String.

Screenshot 2021-12-17 at 11 00 49


As the NodeType schema, endpoint does not replace the labels.

mhsdesign (approved):
👍 by reading

markusguenther:
If you need help with testing this. Feel free to ask @mhsdesign

PR #2998, opened 2021-12-16
BUGFIX: Range editor styling issues

jonnitto:
* Pass default options to the editor
* Remove transition of thumb (opacity gets never changed)
* Add more space between the unit and the input field
* Add border-radius (Similar to a text field)
* Remove border from thumb
* Use Neos constants
* Add hover and focus state to thumb
* Style disabled state

#### Enabled state
Enabled state of range editor

#### Disabled state
Disabled state of range editor

### DIFF
--- a/packages/neos-ui-editors/src/Editors/Range/index.js
+++ b/packages/neos-ui-editors/src/Editors/Range/index.js
@@ -68,7 +68,7 @@ class RangeEditor extends PureComponent {
     };
 
     render() {
-        const options = Object.assign({}, this.constructor.defaultProps.options, this.props.options);
+        const options = {...this.constructor.defaultProps.options, ...this.props.options};
         const {value} = this.state;
 
         return (
--- a/packages/neos-ui-editors/src/Editors/Range/index.js
+++ b/packages/neos-ui-editors/src/Editors/Range/index.js
@@ -72,7 +72,7 @@ class RangeEditor extends PureComponent {
         const {value} = this.state;
 
         return (
-            <div className={style.rangeEditor + (options.disabled ? ' ' + style.rangeEditorDisabled : '' )}>
+            <div className={style.rangeEditor + (options.disabled ? ' ' + style.rangeEditorDisabled : '')}>
                 <input
                     type="range"
                     min={options.min}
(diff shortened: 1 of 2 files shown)


packages/neos-ui-editors/src/Editors/Range/index.js
```diff
@@ -68,11 +68,11 @@ class RangeEditor extends PureComponent {
     };
 
     render() {
-        const {options} = this.props;
+        const options = Object.assign({}, this.constructor.defaultOptions, this.props.options);
```
mhsdesign, line 71: i think i dont understand `this.constructor.defaultOptions` but it seems having to do with: `static defaultProps options` right?

jonnitto: Fixed it. Strange that it worked also like that

mhsdesign: i looks like the 'static defaultProps' is a react thing and works out of box: https://stackoverflow.com/questions/38786973/how-to-set-component-default-props-on-react-component

so maybe your good to go using only this.props.options and the rest is handeled backstage...

packages/neos-ui-editors/src/Editors/Range/index.js
```diff
@@ -68,11 +68,11 @@ class RangeEditor extends PureComponent {
     };
 
     render() {
-        const {options} = this.props;
+        const options = Object.assign({}, this.constructor.defaultProps.options, this.props.options);
```
mhsdesign, line 71:
```suggestion
        const {options} = this.props;
```

mhsdesign: i looks like the 'static defaultProps' is a react thing and works out of box: https://stackoverflow.com/questions/38786973/how-to-set-component-default-props-on-react-component

so maybe your good to go using only this.props.options and the rest is handeled backstage...

jonnitto: Nope, this didn't work out, but maybe @markusguenther has a better opinion as my approach

markusguenther: No that would not work. But we can use the spread operator here.
```const options = {...this.constructor.defaultProps.options, ...this.props.options};```

markusguenther (approved):
This looks good to me :)
Thanks, @jonnitto ❤️

PR #3000, opened 2021-12-16
BUGFIX: Do not allow null in CountValidation

### LINKED ISSUE #2999: Bugfix: Passing wrong value to CountValidator makes everything go white
### Description
Passing null to the CountValidator makes the page go white

### Steps to Reproduce

1. Add a new property to a yaml file of type integer 
2. Add the 'Neos.Neos/Validation/CountValidator' as validation
3. Set minimum and maximum 

`
Testing:
      type: integer
      ui:
        label: 'blub'
        reloadPageIfChanged: true
        inspector:
          group: 'document'
      validation:
        'Neos.Neos/Validation/CountValidator':
          minimum: 1
          maximum: 255
`

#### Expected behavior
An Error message should be shown that the count validator does not work here. 

#### Actual behavior
The whole Backend is white and there is an JS error Uncaught TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at Count (index.tsx:32) [...]

### Affected Versions

Neos: 7.3

UI: 7.3

pKallert:
closes #2999 

**What I did**
Since passing null to the CountValidator makes the page go white, null should throw an error message that it cannot be counted 

**How I did it**
If the value passed to the CountValidator is not an object or is null, an error message is shown. 

**How to verify it**
Validating a property of type integer does not make the page go white anymore, but shows an error message below the input field 

Bildschirmfoto 2021-12-16 um 13 34 24

Bildschirmfoto 2021-12-16 um 13 30 38

### DIFF
--- a/packages/neos-ui-validators/src/Count/index.tsx
+++ b/packages/neos-ui-validators/src/Count/index.tsx
@@ -26,7 +26,7 @@ const Count = (value: any, validatorOptions: CountOptions) => {
         return 'The maximum is less than the minimum.';
     }
 
-    if (typeof value !== 'object') {
+    if (typeof value !== 'object' || value === null) {
         return <I18n id="content.inspector.validators.countValidator.notCountable"/>;
     }
 


markusguenther (approved):
Thanks for that catch @pKallert 
In the end, the usage with an integer property is not intended for the CountValidator.
As it checks if the given value is an object or array
 * and its amount is between minimum and maximum
 * specified in the validation options.
 
 But this miss usage should not break the whole UI.

bwaidelich:
I missed this one, but usually all validators allow `null` in order to allow for _optional_ values.
If null should be forbidden, the `NotEmptyValidator` has to be used additionally.

Not 100% sure whether this applies here though

markusguenther:
But is this really intended? I mean, it feels wrong that you can break the UI by just misconfigure a validator.
When this is not the right way, then it is more or less my fault :/

bwaidelich:
> it feels wrong that you can break the UI by just misconfigure a validator.

Absolutely. The validator should ignore `null` values.
At least that's how all the Flow validators work by default (see https://github.com/neos/flow-development-collection/blob/master/Neos.Flow/Classes/Validation/Validator/AbstractValidator.php#L153)

Issue #3006, opened 2021-12-22
BUG: Grave Accent/Backticks in input fields / Noto Sans


pKallert:
### Description
When entering grave accents in an input field, they do not behave as single letters. This makes deletion difficult 

### Steps to Reproduce

1. Enter any number of grave accents in a text-input field with a text: ````´´´´´Testing 
2. double click the field to select everything 
3. delete the selected text 

#### Expected behavior
The grave accents should be shown as single letters and be deleted after everything is selected 

![Bildschirmfoto 2021-12-22 um 11 42 03](https://user-images.githubusercontent.com/91674611/147081156-bb35af8b-a4b1-4942-a476-484398f206c8.png)

#### Actual behavior
The grave accents are all in one place and cannot be deleted by selecting with a double click. 

![Bildschirmfoto 2021-12-22 um 11 41 43](https://user-images.githubusercontent.com/91674611/147080904-de25be1d-09b8-4592-844d-b474ae2ef6b7.png)

### Affected Versions

Neos: 7.3

UI: 7.3

mhsdesign:
Yes same issue here.
It is caused by the used font '**Noto Sans**'.

I tried this font out online with backticks but it will also not work: https://fonts.google.com/noto/specimen/Noto+Sans

The problem occurs not only in input fields, but wherever this font is used and where individual backticks/grave accents may appear.

I lately had a corrupted message in the Neos ui somewhere (i cant recall where excactly but i recreated it ^^)
![image](https://user-images.githubusercontent.com/85400359/147285048-4a8e64db-6a4b-4faa-92c2-abb3e5e15693.png)
expected:
This is just `` a demo with ```` some é funny ´´ backticks ```'

mhsdesign:
The font that the current Neos UI uses: https://github.com/neos/neos-ui-compiled/blob/master/Resources/Public/Styles/Font-61c3431e4e0ac62b136bc26b427ee05f.ttf is outdated. (Not updated in 3 Years) the command `fc-query Font-61c3431e4e0ac62b136bc26b427ee05f.ttf` says fontversion: 68157. (funny enough, it seems to be the same version as what the google servers are currently hosting (fonts.google....): https://github.com/googlefonts/noto-fonts/issues/736#issuecomment-768889258)

[...]

PR #3007, opened 2021-12-28
BUGFIX: Update the NotoSans font

markusguenther:
Some accents signs does not work in the Neos backend and the reason is our used NotoSans font.
The issue with the font has been fixed, and this PR updates the font to version v20201206-phase3 as of 2020.12.06

Resolves: #3006

**What I did**
Exchanged the font file with the current version.

**How to verify it**
Use characters like the accent sign.

Screenshot 2021-12-28 at 18 24 16


bwaidelich (commented):
Great, thanks for taking care.
Makes sense to me – I just wonder if that license change might be an issue (or whether we have to mention it at least)

markusguenther:
Good point @bwaidelich,
we should mention it, but as we don't modify the font, it seems that we are more or less on the same level. But as I am not  a Lawyer, it is maybe a good thing to mention it explicitly.

jonnitto (approved):

markusguenther:
Merged it now and will mention that in the release notes.
It is BTW interesting that we have no license mentioning file in neos/neos, and we also use noto sans here.

Issue #3005, opened 2021-12-22
BUG: PlainInputMode in Link Editor still shows dropdown line


pKallert:
### Description

When editing links with the Link Editor, there sometimes is a small line of 4px height below the input field. 

### Steps to Reproduce

1. Add a new Node of Type 'Shortcut'
2. Change the 'Target mode' to 'Selected Target'
3. Add a link starting with https://

#### Expected behavior
 The input field should switch from a select-field to a simple input field if a uri is detected and the dropdown area below the input-field should disappear.

#### Actual behavior
The dropdown is still showing as a very small line. 

![Bildschirmfoto 2021-12-22 um 09 37 15](https://user-images.githubusercontent.com/91674611/147063118-fd54a05c-879b-4f28-b4bf-8161b579c775.png)

### Affected Versions

Neos: 7.3

UI: 7.3

pKallert:
As far as I found out, the switch to PlainInputMode in the Selectbox only hides the dropdown content and not the dropdown itself. Because the dropdown itself still has a padding, it shows as a small line. 

This could also be solved with css by moving the padding, but where is the fun in that 🤷

markusguenther:
Thanks for reporting that @pKallert ... I have added a small PR. Would be nice if you can verify that it works as expected.

markusguenther:
Will be released with next UI version (5.3.13 and newer)

PR #3008, opened 2021-12-28
BUGFIX: Reset padding for empty dropdown contents

markusguenther:
**What I did**
Adds an extra class of the select box contents section when the content is empty to reset the padding.

**How to verify it**

Screenshot 2021-12-28 at 19 50 20

Resolves: #3005


mhsdesign (commented):
is padding: 0 !important; not a hack? wouldnt it make more sense to extract the padding 2 px from theme.selectBox__contents and apply it conditionally seperately?

markusguenther:
I now inverted the behavior... so we have the padding just when we have items.

PR #3014, opened 2022-01-27
 BUGFIX: Handle asset proxy for inline links

markusguenther:
The inline link handing uses the LinkLookup data loader to fetch node and asset data.
The LinkLookup is a combination of the NodeLookup and NeosAssetLookup data loaders.

As the whole backend can work with asset proxies, the usage of NeosAssetLookup needs to be replaced by the AssetLookup data loader.

**How I did it**
Replaced the NeosAssetLookUp with the AssetLookup in the LinkLookup data loader.
So that we not search only for local assets. Additionally, we need some adjustments to handle asset proxies for the link input component.

**How to verify it**
Activate an asset source that is not the local Neos source, Unsplash for instance.
When the asset source is working, you are able to find assets in the asset editor from Unsplash.

When you now create an inline link and search for an image,  you should also see Unsplash results.
That was not possible before.

https://user-images.githubusercontent.com/1014126/151379459-fbba3c49-423b-4f39-9a0b-d874b5d4dfb9.mp4

### DIFF
--- a/packages/neos-ui-redux-store/src/UI/Remote/index.ts
+++ b/packages/neos-ui-redux-store/src/UI/Remote/index.ts
@@ -22,6 +22,8 @@ export const defaultState: State = {
 export enum actionTypes {
     START_SAVING = '@neos/neos-ui/UI/Remote/START_SAVING',
     FINISH_SAVING = '@neos/neos-ui/UI/Remote/FINISH_SAVING',
+    LOCK_PUBLISHING = '@neos/neos-ui/UI/Remote/LOCK_PUBLISHING',
+    UNLOCK_PUBLISHING = '@neos/neos-ui/UI/Remote/UNLOCK_PUBLISHING',
     START_PUBLISHING = '@neos/neos-ui/UI/Remote/START_PUBLISHING',
     FINISH_PUBLISHING = '@neos/neos-ui/UI/Remote/FINISH_PUBLISHING',
     START_DISCARDING = '@neos/neos-ui/UI/Remote/START_DISCARDING',
@@ -60,6 +62,16 @@ const startDiscarding = () => createAction(actionTypes.START_DISCARDING);
 const finishDiscarding = () => createAction(actionTypes.FINISH_DISCARDING);
 
 /**
+ * Marks that an publishing process has been locked.
+ */
+const lockPublishing = () => createAction(actionTypes.LOCK_PUBLISHING);
+
+/**
+ * Marks that an publishing process has been unlocked.
+ */
+const unlockPublishing = () => createAction(actionTypes.UNLOCK_PUBLISHING);
+
+/**
  * Should be called once the server informs the client that a node has been created.
  */
 const documentNodeCreated = (contextPath: NodeContextPath) => createAction(actionTypes.DOCUMENT_NODE_CREATED, {contextPath});
@@ -70,6 +82,8 @@ const documentNodeCreated = (contextPath: NodeContextPath) => createAction(actio
 export const actions = {
     startSaving,
     finishSaving,
+    lockPublishing,
+    unlockPublishing,
     startPublishing,
     finishPublishing,
     startDiscarding,
@@ -89,6 +103,14 @@ export const reducer = (state: State = defaultState, action: InitAction | Action
             break;
         }
         case actionTypes.FINISH_SAVING: {
+            draft.isSaving = false;
+            break;
+        }
+        case actionTypes.LOCK_PUBLISHING: {
+            draft.isSaving = true;
+            break;
+        }
+        case actionTypes.UNLOCK_PUBLISHING: {
             draft.isSaving = false;
             break;
         }
--- a/packages/neos-ui-backend-connector/src/Endpoints/index.ts
+++ b/packages/neos-ui-backend-connector/src/Endpoints/index.ts
@@ -291,7 +291,7 @@ export default (routes: Routes) => {
                     assetProxyIdentifier
                 };
             });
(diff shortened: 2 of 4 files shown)


lorenzulrich:
One thing that came to my mind: For the `AssetEditor`, the asset sources available are configurable:
https://neos.readthedocs.io/en/stable/References/PropertyEditorReference.html#property-type-image-neos-media-domain-model-imageinterface-imageeditor-image-selection-upload-editor
I think we should have something similar here (as part of the `inlineEditor` options).

markusguenther:
@lorenzulrich Can take a look, but that would become more a feature when we add this.

markusguenther:
It is just my personal point of view, maybe the others have a different opinion.

WDYT @dimaip @skurfuerst @Sebobo @bwaidelich ???

Sebobo:
From the video I wonder how to differentiate between local and remote assets. When entering some search word, there are probably many more external results as local, even though the local one might be the desired one.

markusguenther:
@Sebobo Yes already wrote with @lorenzulrich about that. The asset editor groups the output. But the old code did not do that for the nodes and assets at all. So guess that grouping is more a feature than a bug fix ;)

Sebobo:
Yes I would just see it as a new bug if due to the change it becomes harder so selected the assets I could simply find before. But I haven't tried the change yet to see if it's really a problem or not.

jonnitto:
I would merge this one as a Bugfix and then tackle the other issues as a feature

markusguenther:
Yeah the grouping and the asset-source filter would be a follow-up feature, I guess.

lorenzulrich:
If this one is merged as a bugfix without taking into account my comment before, people will suddenly have more assets available than before which is clearly a noticeable change. Even worse, since the configuration options change would be a feature, people could not do anything about it. I think the full thing (selection, grouping, configuration) should go into 7 or nothing at all.

markusguenther:
The grouping is also included now. Take a look into the options for filtering the asset-sources.

Screenshot 2022-02-14 at 16 40 04

markusguenther:
The grouping is now also available for the external links for shortcuts. As this part uses the same linkInput implementation.

markusguenther:
**Update:**

The version now works with the regular editors and inline editors. The only difference is, where we define the asset-source constraints, for instance.

For inline editors we need to add the constraints to the linking section. This is an example for the text property:
```
'Neos.Demo:Content.Text':
  properties:
    text:
      ui:
        inline:
          editorOptions:
            linking:
              constraints:
                assetSources: ['unsplash']
```

For editors in the right sidebar, like the shortcut target, we need to add the constraints like mentioned in the top comment of Lorenz. So this works like the existing AssetEditor.
```
# A shortcut.
# "Shortcuts" can be linked to but instead of displaying content themselves they redirect to another "Document" node
'Neos.Neos:Shortcut':
  properties:
    target:
      ui:
        inspector:
          editorOptions:
            constraints:
              assetSources: ['unsplash']
```

markusguenther:
Is this commit ok for everybody?
If not, I revert that and create a new PR for 8.0 and introduce that we can lock the publish button.

https://github.com/neos/neos-ui/pull/3014/commits/161a1f04d4fbb6c044a85d74f3a84c9ce07805ad

jonnitto (approved):
Works fine. After reading @lorenzulrich comment, I've to say that he is right. It needs these features to fix the bug. 🤷🏽

Sebobo:
If I understand correctly only users will see a difference in a Neos instance with multiple Asset Sources correct?
For everyone else everything will stay the same?

We should already create a connected issue for the JSON schema so it can be updated after this is released.

lorenzulrich:
> If I understand correctly only users will see a difference in a Neos instance with multiple Asset Sources correct?
> For everyone else everything will stay the same?

There is another difference: If you have only the Neos Asset Source, it will still group Pages ("Documents") and Neos Assets. Before this change, there was no grouping at all; results from Pages and Assets were in one list.

lorenzulrich:
@markusguenther I can confirm that I tested this code (applied as a patch to Neos 4.3) and it works flawlessly.

packages/neos-ui-redux-store/src/UI/Remote/index.ts
```diff
@@ -59,6 +61,16 @@ const startDiscarding = () => createAction(actionTypes.START_DISCARDING);
  */
 const finishDiscarding = () => createAction(actionTypes.FINISH_DISCARDING);
 
+/**
```
kitsunet, line 64: this seems unrelated? (everything in this file)

kitsunet (commented):
See question about unrelated changes. rest looks good!

packages/neos-ui-redux-store/src/UI/Remote/index.ts
markusguenther: Nah, it was a little improvement. Because when you click publishing really fast, it can happen that you save the link with the asset proxy and not the asset, as it was not imported yet. So I came up with this... so that the publishing is blocked until the import succeeds or fails.

kitsunet (approved):

packages/neos-ui-redux-store/src/UI/Remote/index.ts
markusguenther: Also plan to add an API for that in future versions, as this is more a quick fix that misuse the saving state.
Would be nice to have an API to lock the publishing.

lorenzulrich:
Thanks everyone involved for bringing this into 5.3+!

Issue #3035, opened 2022-02-21
Bug: Line break in header row/column option of table dropdown


esdete2:
### Description

The label of header row/column option is below the corresponding checkbox.

### Steps to Reproduce

1. Place a table inside of an inline text element
2. Click on the row or column dropdown in ckeditor options

#### Expected behavior

The label should be next to the checkbox.

#### Actual behavior

![Bildschirmfoto 2022-02-21 um 10 01 35](https://user-images.githubusercontent.com/34043608/154922149-be8c79b1-23b0-4cfa-abaa-0e127acbcf5e.png)

In this commit the checkbox is changed from `display: inline-block` to `display: block`, which causes it to take up the full width:  https://github.com/neos/neos-ui/commit/5647b84f24492762f019273ac001a34850f2fac7#diff-715d745c65e8adde10e1120c36b7f62567b0d5bb2b9d9bf3cd1429542d61d0e6R4

### Affected Versions

Neos: 7.3.1

UI: 7.3.1

markusguenther:
Thanks for reporting the issue. Tested that and can confirm the issue :)

markusguenther:
Should also affect all Neos versions since 5.3 BTW.

markusguenther:
This is a regression of #2976 but will be fixed in an upcoming version of the UI up from version 5.3.

PR #3038, opened 2022-02-22
BUGFIX: Prevent label from wrapping

markusguenther:
The dropdowns for the column and row settings of the CKEditor tables wrap the label after the checkbox. To prevent that, we now inline the checkbox. So that the text is in the same level.

fixes: #3035

### DIFF
--- a/packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/TableDropDown.css
+++ b/packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/TableDropDown.css
@@ -7,10 +7,10 @@
     text-align: left;
 }
 .checkBox {
+    display: flex;
     height: var(--spacing-GoldenUnit);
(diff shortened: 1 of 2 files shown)


jonnitto (approved):

Sebobo:
Why does the issue only appear here if it was a general component?

markusguenther:
> Why does the issue only appear here if it was a general component?

Good question, did not investigate this. But tested it in the sidebar and there was it no issue.

markusguenther:
Ok now investigated the issue and the surrounding component is different.
The dropdown in the CKEditor uses `display: block:` and in the sidebar we have `display: flex`

markusguenther:
Reverted the change and adjusted now the TableDropDown component which wraps the Checkbox in this case.

Sebobo (approved):

Issue #3041, opened 2022-02-25
Introduce second icon property for larger or more detailed icons


markusguenther:
### Description

If you use custom icons, you want to use detailed icons, which describes the use case better, in the node type model. But then the detailed icon is not the best choice for the document or content tree.

So, this feature will bring a new configuration to have a different icon for the node type dialog tiles. We can now configure an icon and a `largeIcon` setting for the node type.

#### Expected behavior

Being able to configure a second icon for a node type for the select node dialog.

#### Actual behavior

We have just one icon property.

PR #3042, opened 2022-02-25
FEATURE: Introduce icon configuration for nodetypes

markusguenther:
**What I did**
Added a new option `largeIcon` and the node type item component checks if we configure a larger icon and render this instead if the regular icon. The larger icon also use the doubled size.

**How to verify it**

Configure a node type with the property `largeIcon` and check the select node type dialog after that.

```
'Neos.Demo:Content.Text':
  ui:
    icon: 'icon-file-text'
    largeIcon: 'resource://Neos.Demo/Images/customIcon.svg'
```

Screenshot 2022-02-25 at 18 48 22

resolves: #3041


packages/neos-ui/src/Containers/Modals/SelectNodeType/nodeTypeItem.js
```diff
@@ -31,8 +31,10 @@ class NodeTypeItem extends PureComponent {
 
     render() {
         const {ui, name} = this.props.nodeType;
-        const icon = $get('icon', ui);
         const label = $get('label', ui);
+        const useLargeIcon = ('largeIcon' in ui);
```
jonnitto, line 35: what's the difference between `('largeIcon' in ui);` and $get('icon', ui). Why not use the same method?

markusguenther: This checks if we have the `largeIcon` property in the options. 
And then we use the `icon` or the `largeIcon`

jonnitto:
What I'm missing is the documentation part…

markusguenther:
@jonnitto We have no documentation in Code... so when we merge this we need to adjust the docs in the neos/neos package.

Sebobo:
Then we should call the parameter not `largeIcon` but directly one of `XS, sm, etc...` right?
And the component decides which to use, with a fallback mechanism. Wdyt?

markusguenther:
> Then we should call the parameter not `largeIcon`

TBH, I had also issued to name the property. My goal was to have an extra icon for the select node dialog and choose `largeIcon` caused by the bigger representation in the view. But prefixing the icon with the size sounds not so nice.

As we have an 8.0 we can maybe change the whole naming. Make something like `primaryIcon` and `secondaryIcon` and for these settings I can imagine something like:

```
primaryIcon:
    size: '2x'
    icon: 'resource://Neos.Demo/Images/customIcon.svg'
```

We, of course need to keep the old icon for compatibility reasons, as we did not deprecate the icon before.
Perhaps we can also provide a migration. Question is if we want to handle the sizing as well for the `secondaryIcon` as it is used for the tree and sidebar.

Sebobo:
then how about introducing `icon-lg` instead of `largeIcon` and leave the rest as you did. Then we could extend it when needed

markusguenther:
So, just `icon-lg` and no extra size? Or keep the idea of splitting the icon and size? So we have `icon-2x`, `icon-3x` and so on?

mhsdesign:
> As we have an 8.0 we can maybe change the whole naming. Make something like `primaryIcon` and `secondaryIcon` and for these settings I can imagine something like:

i like that the 'normal' config for icon is flat like `ui.icon: 'icon-file-text'` as one usually doesnt need to configure much there.

the name `largeIcon` doesnt really fit into this maybe `icon-lg` really is better.

I at first thought the `largeIcon` would in any case overrule the normal icon - maybe name the property so that this is more intuitive, but i cant think of anything fitting ...

Sebobo:
> How about naming the config key "preview" and use this for the creation dialog with a fallback to "icon" if no preview exists.

That would have the advantage that the integrator would know what they are defining in yaml. 
Only setting various sizes and not really understanding where they are used doesn't help so much.

markusguenther:
Ok so we use the `preview` key and drop the ability to adjust the scaling?
Because I also like that the integrators are able to upscale the icon in the select dialog a bit.

mficzel:
Maybe display previews larger and allow icon names in there aswell as svg resources. That way we get away with a single configuration option.

jonnitto:
So the name would be `previewIconSize`, right?

markusguenther:
But having a bigger size for the preview is maybe irritating to the users. But if this is not the most wanted feature I will skip that and start with the extra preview icon 🙃

jonnitto:
I would like to have the option for the size. I would have a current use case for this…

mficzel:
Fine if there is an actual useCase ... i was just against adding things that might be useful one day.

jonnitto:
Naming things…
Just a question (Sorry for that):

Why it is called `preview` when the size setting is `previewIconSize`?  Shouldn't it be named `previewIcon`?

markusguenther:
Can also name it like this. Named it preview as it was a suggestion and had two likes ;)

markusguenther:
So now we have something like this :)

```
'Neos.Demo:Content.Text':
  ui:
    icon: 'icon-file-text'
    previewIcon: 'resource://Neos.Demo/Images/customIcon.svg'
    previewIconSize: '3x'
```

jonnitto (approved):

mhsdesign (approved):

markusguenther:
Thanks for the feedback and reviews 💙

Issue #2855, opened 2020-11-25
Remove Internet Explorer 11 Support


markusguenther:
### Description

The IE 11 is dead and even some of the libraries we use (ckeditor 5) are not compatible anymore.
So we should remove the IE support and get rid of some polyfills and so one.

As it is to short before the 6.0 release we should tackle that for the next 6.1 version :) 

- [x] Remove ckeditor version 4
- [x] Check if we still need polifills
- [ ] Cleanup styles and remove potential IE hacks

markusguenther:
It has been a while ;)
Actually, we are shipping CKEditor 4 and 5. The version 4 is for compatibility reasons because this one worked in the older IE browsers. With version 8.0 I plan to drop this CKEditor version 4.

markusguenther:
- [ ] Remove ckeditor version 4
- [ ] Check if we still need polifills
- [ ] Cleanup styles and remove potential IE hacks

mhsdesign:
are the `'./polyfills.js'` still needed? https://github.com/neos/neos-ui/blob/master/polyfills.js

albe:
Anything left to do on this?

mhsdesign:
with #3200 we are now targeting es2020, removed (some if not all) polyfills and im working via #3266 on removing the autoprefixer.

i think thats the last task - but hard to get started
> Cleanup styles and remove potential IE hacks

maybe it will up itself eventually over time ...

mhsdesign:
I could find no comments referencing internet explorer, ie, microsoft, or ms (vendor prefixes and stuff)
so i think we can close this ;)

PR #3045, opened 2022-02-28
FEATURE: Remove neos-ui-ckeditor-bindings and ckeditor 4

markusguenther:
**What I did**
This PR removes the compatibility layer of the CKEditor to be able to run Neos in the old Internet-Explorer 11.
The whole CKEditor 4 bindings will be removed and the `defaultInlineEditor` configuration is removed.

We now use only the CKEditor in version 5.

Relates: #2855


jonnitto (approved):
Nice. Less code == faster

markusguenther:
Updated to the latest state of master

markusguenther:
> wtf 416 files 😂 i guess i can only see if it works by testing it
> 
> one can find out what actually changed by unticking Deleted files:
> 
> image

Yeah, as this removes the whole CKEditor version 4 stuff it is a lot.
The packages path from the CKEditor binding can be excluded as this has been removed and the assets from the resource folder. So the more important part is where the switch of the versions take place.

mhsdesign:
so this basically removes this whole `setDefaultInlineEditor` api - that one can change the editor right?
then the node yaml schemas should also be updated - if they exist ^^

markusguenther:
> `setDefaultInlineEditor`

Yes, when this is merged, we also need to check docs for that config.

markusguenther:
@mhsdesign did you find something while the review? Or can you also approve this change?

mhsdesign (approved):
i still wanted to test this out locally - sorry for keeping you waiting ^^ - conceptionally i a can approve already

markusguenther:
I resolved the conflicts again. And if you want to try it locally feel free to test this first.
Don't want to push you to an approval.

PR #3069, opened 2022-03-11
BUGFIX: Allow zero as value in selectbox

pKallert:
**What I did**
When using the 'Neos.Neos/Inspector/Editors/SelectBoxEditor' with a value 0, the label is not displayed in the field when loading a page

**How I did it**
The value is currently checked for javascript 'falsy' values, which includes 0. This leads to the SelectBox_HeaderWithSearchInput always being displayed if the value is 0. 

**How to verify it**
1. Use the following property 
`autoplay:
      type: integer
      defaultValue: ''
      ui:
        label: 'Autoplay'
        reloadPageIfChanged: true
        inspector:
          group: 'image'
          editorOptions:
            values:
              '0':
                label: 'No Autoplay'
              '1000':
                label: '1 Second'`

2. Chose 'No Autoplay' as a option in the dropdown
3. Reload Page
4. 'No Autoplay' is shown as the active option

Old behavior:
https://user-images.githubusercontent.com/91674611/157832243-e729df73-435d-4006-af4f-e85ed0f858c3.mov

### DIFF
--- a/packages/react-ui-components/src/SelectBox/selectBox.js
+++ b/packages/react-ui-components/src/SelectBox/selectBox.js
@@ -257,7 +257,7 @@ export default class SelectBox extends PureComponent {
 
         if (
             displaySearchBox && (
-                !value ||
+                (!value && value !== 0) ||
                 this.state.isExpanded ||
                 plainInputMode
             )


jonnitto (approved):

packages/react-ui-components/src/SelectBox/selectBox.js
```diff
@@ -257,7 +257,7 @@ export default class SelectBox extends PureComponent {
 
         if (
             displaySearchBox && (
-                !value ||
+                (!value && value !== 0) ||
```
mhsdesign, line 260: i cant tell what `value` is excactly supposed to be when its not selected since it just says `value: PropTypes.any` but i assume its `null`. maybe we should do no type conversion at all and just check if its not null?

as far as i can see `value` can only be a string or an int when it was set (since it was a php array key in the yaml config) 
so it must be null or undefined when nothing is selected right?

markusguenther:
@pKallert As the type can be much more than integer, I used now a more general approach to test if the value is empty.
The Fix itself just works as you proposed the change :)

packages/react-ui-components/src/SelectBox/selectBox.js
markusguenther: Adjusted the PR a bit, maybe it is more general for you :)

markusguenther:
Technically, I can approve that the change is working ;)

packages/utils-helpers/src/isEmpty.ts
```diff
@@ -0,0 +1,24 @@
...
+// Checks if the given item is empty
+//
+export default function isEmpty(item: any): boolean {
+    if (item == null) {
+        return true;
+    }
+
+    if (typeof item === 'number' || typeof item === 'boolean') {
+        return false;
+    }
```
mhsdesign, line 11: not to be nitpicking: and it doesnt matter in this case but `isEmpty(false)` will be `false` ... is this right?
otherwise 👍👍👍

```suggestion
    if (typeof item === 'number') {
        return false;
    }

    if (typeof item === 'boolean') {
        return item;
    }
```

markusguenther: Correct `isEmpty(false)` and `isEmpty(true)` will be not empty, and every number is also not empty.
Maybe we can discuss if `isEmpty(NaN)` should be empty. Because `NaN` is a number but not really. So, this would be false.

mhsdesign: and hu there are 'functions' and so on - i must say i regret now that i said anything 😂

mhsdesign: i think as long as we dont try to reuse this rather specific `isEmpty` were all good to go and this can be merged ...

with specific i mean, in which language is `0` truthy and `''` falsy ? not in php or js at least ... so that makes this unpredictable ^^

mhsdesign (approved):

packages/utils-helpers/src/isEmpty.ts
Sebobo: In my use cases I usually desire a differentiation if something is "set" or not. That means is it null/undefined or any other value.

But on the other hand the truthiness and falsiness of PHP also usually fulfils my demands except a check on an actual `false` value.

Sebobo: Or at least we should rename it to its use case. Like `isOptionSet`

mhsdesign: i agree with @Sebobo,
when we figure out what the value can be and is when its not set - probably `null` or `undefined` we dont need to check for array lenght or object stuff - because the value will never be an object fx.

PR #3072, opened 2022-03-14
TASK: Remove lodash.set

markusguenther:
We use `lodash.set` only for one case and as it has some security issues we should get rid of that.

**What I did**
- Replaced the usage with `plow-js`.
- removed the `lodash.set` package

**How to verify it**
-  should work like before


markusguenther:
With that PR we resolve the following issue:
https://security.snyk.io/vuln/SNYK-JS-LODASHSET-1320032

jonnitto (approved):

PR #3075, opened 2022-03-14
TASK: Increase default node version to 16

markusguenther:
**What I did**
Increased the default node version to 16 as it is the Active LTS version


kdambekalns:
The failure is a timeout in the unitests step. 🤷‍♂️

markusguenther:
https://support.circleci.com/hc/en-us/articles/360045268074-Build-Fails-with-Too-long-with-no-output-exceeded-10m0s-context-deadline-exceeded-

Need to find an equal for node :D

markusguenther:
The test is now running longer, but for some reason it does not come to an end. But the last state comes after approximately 4 minutes, and then nothing happens. So there is an issue with the test somewhere.

kdambekalns:
The duplicate use of `--forceExit` presumably is something to really force it? Like `-vvvv` or the like?

markusguenther:
Ok, even with forceExit the tests does not go further. But also don't show an error :/
Locally, it stops at the same place. Need to investigate the issue :/

Screenshot 2022-08-31 at 16 02 41

markusguenther:
> The duplicate use of `--forceExit` presumably is something to really force it? Like `-vvvv` or the like?

Force Jest to exit after all tests have completed running. This is useful when resources set up by test code cannot be adequately cleaned up.

markusguenther:
Found the issue and now need to find a solution ;)
used now ```--stream --concurrency=1 ``` and so the output was better.

Screenshot 2022-08-31 at 16 12 22

markusguenther:
Ok it turns out that the failing UI tests is not caused by us. In test cafe the hammerhead request fails, and therefore we have a not expected error in the log.

https://github.com/DevExpress/testcafe/issues/7097

kdambekalns (approved):

Issue #3070, opened 2022-03-11
BUG: Option with value 0 in selectbox 


pKallert:
### Description
After adding a SelectBoxEditor with one of the value 0, the label is not shown when the option is active. Instead 'Choose'  is shown 

### Steps to Reproduce

1. Add a new property: 

```yaml
autoplay:
      type: integer
      defaultValue: ''
      ui:
        label: 'Autoplay'
        reloadPageIfChanged: true
        inspector:
          group: 'image'
          editor: 'Neos.Neos/Inspector/Editors/SelectBoxEditor'
          editorOptions:
            values:
              '0':
                label: 'No Autoplay'
              '1000':
                label: '1 Second'
```
2. Chose 'No Autoplay' as the active option in the dropdown 
3. Click 'Apply'
#### Expected behavior
The label for the value 0 ('No Autoplay') is shown as active. 

#### Actual behavior
The label is not shown and instead the 'Chose' placeholder is shown: 

https://user-images.githubusercontent.com/91674611/157833078-48949e8a-717d-4424-85dc-8eb3ac45ef3f.mov

### Affected Versions

UI: 5.3

PR #3085, opened 2022-03-18
TASK: Renames isEmpty helper to isNil

markusguenther:
This is a change related to the issue #3070
In the PR #3069 we discussed after merge that the naming could be different.

Therefore, we rename the helper now afterwards.


Sebobo:
Name is definitely better. But I wonder why an empty string is considered as "not set".
Also why does it iterate over the keys and return false if a key exists?

jonnitto:
Should the naming not be something like `isOptionNotSet`? `isOptionSet(['Jon Doe'])` return `false` Otherwise, `true` and `false` need to be switched

mhsdesign:
> Also why does it iterate over the keys and return false if a key exists?

as far as i know its not too easy to find out if a js object is empty `{} === {}` wont work for example ...

mhsdesign:
btw i feel like we are all nitpicking over a total overengineer solution (https://github.com/neos/neos-ui/pull/3069)

**-** and by saying this im probably nitpicking the solution 😁

Sebobo:
This is sadly a very important but dangerous method that needs to be thoroughly defined, documented and changing it later will be super breaky. So its important to have several people nitpick :)

markusguenther:
Ok, as already asked in the PR before, maybe `isNil` is enough. It solves the initial error and does not that much magic as mentioned before. Hope that can help with all the concerns :)

mhsdesign (approved):
Thanks for your effort 👍

Sebobo (approved):

markusguenther:
That merge was too quick. The tests were red because when the value is an empty string, it does not behave the same now.
So will have another PR as bug fix ;)

mhsdesign:
we can also 'revert'

PR #3086, opened 2022-03-18
BUGFIX: Render search also for empty string

markusguenther:
With the introduction of `isNil()` the check for empty strings fail. So, we need to check also for empty strings and render the select box with search header.


mhsdesign (approved):
sorry for the extra round 😇

PR #3091, opened 2022-03-24
FEATURE: Adds restore button for user impersonation

markusguenther:
With Neos 8.0 we introduce the impersonation feature. So, it is possible to impersonate as a different user from the user management module.

This change will add related endpoints to the neos-ui to be able to communicate with the Neos.Neos API. The impersonation state will be saved in the Redux store, and a saga handles the restore of the origin user.

Besides the data handling stuff, we also add a new button to the user menu.

Related Neos PR:
https://github.com/neos/neos-development-collection/pull/3648

https://user-images.githubusercontent.com/1014126/159938292-884a17d0-e138-40e0-83ea-65366000396e.mp4

### DIFF
--- a/Resources/Private/Fusion/Backend/Root.fusion
+++ b/Resources/Private/Fusion/Backend/Root.fusion
@@ -195,6 +195,18 @@ backend = Neos.Fusion:Template {
                     controller = 'Service\\ContentDimensions'
                     action = 'index'
                 }
+                impersonateStatus = Neos.Fusion:UriBuilder {
+                    package = 'Neos.Neos'
+                    controller = 'Backend\\Impersonate'
+                    action = 'status'
+                    format = 'json'
+                }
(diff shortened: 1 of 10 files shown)


jonnitto (approved):

markusguenther:
It works locally, as the e2e tests make issues since the 8.0 Neos changes. I need to rework the CircleCI stuff.
Currently I am not so confident in the CircleCI tests.

markusguenther:
Screenshot 2022-03-24 at 16 21 36

Test failing because the backend is not able to start, and the URI is not available.

markusguenther:
OK now just the test after the second site import fails. And that happens since the PHP 8.0 change.
Screenshot 2022-03-24 at 16 59 22

kdambekalns (approved):

PR #3096, opened 2022-04-04
DOCS: Update Component Wiring

mhsdesign:
This seems to be out of date.

see EditorEnvelope/index.js https://github.com/neos/neos-ui/blob/6ce277470258599c1d29f9ec2f5983e7e44d4338/packages/neos-ui-editors/src/EditorEnvelope/index.js#L31

### DIFF
--- a/packages/neos-ui/src/manifest.js
+++ b/packages/neos-ui/src/manifest.js
@@ -108,11 +108,10 @@ manifest('main', {}, globalRegistry => {
 
         Every component gets the following properties (see EditorEnvelope/index.js)
 
-        - identifier: an identifier which can be used for HTML ID generation
+        - id: an identifier which can be used for HTML ID generation
         - label: the label
-        - node: the current node
         - value: The value to display
-        - propertyName: Name of the property to edit (of the node)
+        - identifier: Name of the property to edit (of the node)
         - options: additional editor options
         - commit: a callback function when the content changes.
             - 1st argument: the new value
@@ -124,6 +123,8 @@ manifest('main', {}, globalRegistry => {
             - 2nd argument: a callback function which can be used to render the secondary inspector. The callback function should return the secondary inspector content itself; or "undefined/null" to close the secondary inspector.
 
             Example usage: props.renderSecondaryInspector('IMAGE_CROPPING', () => <MySecondaryInspectorContent />)
+        
+        the current node is not passed and must be acquired via @connect($transform({ node: selectors.CR.Nodes.focusedSelector }))
 
     `));
 


mhsdesign:
nope im not sure if my proposed fix is 100% correct - when looking deeper into the code - editors are created at two places, but the props dont match exactly.

i just know that logging `this.props` will not result in a `node` as property (from the inspector). What are the desired editor props? the from above in the (outdated) docs?

we should make a Ts interface with jsdoc for that ...

mhsdesign:
Thats what i have currently ^^, 
note that it conflicts with https://docs.neos.io/cms/manual/extending-the-user-interface/react-extensibility-api#component-wiring 
where it seems that id and identifier have been mixed up.

```ts
/**
 * Component Wiring
 * 
 * {@link https://docs.neos.io/cms/manual/extending-the-user-interface/react-extensibility-api#component-wiring}
 */
export type EditorProps<Options = {}, Value = any> = {
    /**
     * an identifier which can be used for HTML ID generation
     * @example
     *  "__neos__editor__property---{propertyName}"
     */
    id?: string;
    /** name of the node property to edit */
    identifier: string;
    /**
     * label of the node property
     * to be possibly translated via i18n.translate(label)
     * 
     * recommended, when {@link RegisteredEditor.hasOwnLabel} = true is used
     *
     *  @example
     * ```jsx
     *  <Label htmlFor={this.props.id}>
     *      <I18n id={this.props.label} />
     *      {this.props.renderHelpIcon()}
     *  </Label>
     * ```
     */
    label: string;
    /** additional editor options of the `editorOptions` */
    options: Options;
    /**
     * currently edited property value of the node
     * is reactive - eg updates on discard and commit
     */
    value?: Value;
    /**
     * @param secondaryInspectorName identifier, used to implement toggling of the inspector when calling this method twice.
     * @param secondaryInspectorComponentFactory factory for the secondary inspector content or undefined|null to close the secondary inspector
     * @example
     *  props.renderSecondaryInspector('IMAGE_CROPPING', () => <MySecondaryInspectorContent />)
     */
    renderSecondaryInspector(
        secondaryInspectorName: string,
        secondaryInspectorComponentFactory: () => JSX.Element | undefined | null
    ): void;
    /**
     * register value change of the node property, which can be applied or discarded
     * 
     * @param value the new value. 
     * @param hooks an object whose keys are saveHooks to be triggered, the values are hook-specific options: Example: `{'Neos.UI:Hook.BeforeSave.CreateImageVariant': nextImage}`
     */
    commit(value: Value, hooks?: Record<string, unknown>): void;
    /**
     * unsets the property
     * 
     * @param value an empty string to unset the property
     */
    commit(value: "", hooks?: Record<string, unknown>): void;

    // unofficial api
    /** name of the editor `Foo.Bar/EditorName` */
    editor: string;
    /**
     * renders toggleable button
     * 
     * see for an examle:
     * {@link EditorProps.label}
     */
    renderHelpIcon(): JSX.Element | "";
    hooks?: object;
    editorRegistry: EditorRegistry;
    i18nRegistry: I18nRegistry;
    /** Saves any pending changes (As if you clicked the Apply button) */
    onEnterKey(): void;
    helpMessage?: string;
    helpThumbnail?: string;
    highlight?: boolean;
};

interface RegisteredEditor {
    /** Editor component to use */
    component: React.ComponentType<any>;
    /**
     * No label of the property and helpmessage will be rendered above the component.
     * 
     * You need to render the label internally in the component
     * @example
     * ```jsx
     *  <Label htmlFor={this.props.id}>
     *      <I18n id={this.props.label} />
     *      {this.props.renderHelpIcon()}
     *  </Label>
     * ```
     */
    hasOwnLabel?: boolean;
}
```

mhsdesign:
a certain inconsistency happends du to the fact that the node creation dialog and the inspector render the [`<EditorEnvelope />`](https://github.com/neos/neos-ui/blob/744d50cd2503f64af1c1baf78c756269d34f7116/packages/neos-ui-editors/src/EditorEnvelope/index.js
) (which will render your final `<EditorComponent />`) differently (in two places with different props)

## The truth so far

### [`<NodeCreationDialog />`](https://github.com/neos/neos-ui/blob/744d50cd2503f64af1c1baf78c756269d34f7116/packages/neos-ui/src/Containers/Modals/NodeCreationDialog/index.js#L237)

my nice ts interface *does not* match the props https://github.com/neos/neos-ui/pull/3096#issuecomment-1249164371

Props received in the  `<EditorComponent />`
passed via NodeCreationDialog:
- identifier (this time `${elementName}--creation-dialog`)
- label
- editor
- helpMessage
- helpThumbnail
- options
- commit
- value
- hooks
- onKeyPress (when enter apply: event.key === 'Enter'  && onEnterKey())
- onEnterKey (applies pending changes)
- renderSecondaryInspector

passed via EditorEnvelope
- id (`__neos__editor__property---{propertyName}`)
- renderHelpIcon
- className

passed via EditorEnvelope @neos
- editorRegistry
- i18nRegistry

### [`<InspectorEditorEnvelope />`](https://github.com/neos/neos-ui/blob/744d50cd2503f64af1c1baf78c756269d34f7116/packages/neos-ui/src/Containers/RightSideBar/Inspector/InspectorEditorEnvelope/index.js#L76)

my nice ts interface matches the props https://github.com/neos/neos-ui/pull/3096#issuecomment-1249164371

Props received in the  `<EditorComponent />`
passed via InspectorEditorEnvelope:
- options
- highlight
- identifier (this time node property name)
- value
- hooks
- commit
- label
- editor
- options
- renderSecondaryInspector
- commit
- onEnterKey (applies pending changes)
- helpMessage
- helpThumbnail

passed via EditorEnvelope
- id (`__neos__editor__property---{propertyName}`)
- renderHelpIcon 
- className

passed via EditorEnvelope @neos
- editorRegistry
- i18nRegistry

also the use of passing all props via `{...this.props}` in EditorEnvelope as in InspectorEditorEnvelope is scary as things might become api that clearly arent (this way `i18nRegistry`, and `editorRegistry` show up in my tests and composition of the typescript interface)

i think we should explicitly pass the props [`<EditorEnvelope />:L85`](https://github.com/neos/neos-ui/blob/744d50cd2503f64af1c1baf78c756269d34f7116/packages/neos-ui-editors/src/EditorEnvelope/index.js#L85
) to the `<EditorComponent />` for a 100% clear api.


## Misleading documentation:

### [docs.neos.io](https://docs.neos.io/cms/manual/extending-the-user-interface/react-extensibility-api#component-wiring)
```diff
- identifier (an identifier which can be used for HTML ID generation)
label
value
+ propertyName (name of the node property to edit)
options
commit
renderSecondaryInspector:
```
(+green whats new -red where the value is inconsistent)

### [the inline editors registry docs](https://github.com/neos/neos-ui/blob/744d50cd2503f64af1c1baf78c756269d34f7116/packages/neos-ui/src/manifest.js#L109)
```diff
- identifier (an identifier which can be used for HTML ID generation)
label
+ node (the current node)
value
+ propertyName (Name of the property to edit (of the node))
options
commit
renderSecondaryInspector
```
(+green whats new -red where the value is inconsistent)

mhsdesign:
we need to fix the editor component props api - in a way that is non breaking.

the problematic things are
 - `id` and `identifier` as they are inconsistent an their naming is not the finest ^^
 - `node` is not available but would seem logical - sure one can grep it via `node: selectors.CR.Nodes.focusedSelector(state)`
 - props that dont belong to the api `editorRegistry` `i18nRegistry` `hooks` 
    - `helpMessage`, `helpThumbnail` serve no use (at least not on first thought see `renderHelpIcon`)
 - commit behaves odd as one need to pass `""` to unset
 - the naming of `onEnterKey`
 - `onKeyPress` is useless (essentially: event.key === 'Enter' && onEnterKey())
 - `className` yes sure its a className but what does it do? it will render an orange or red boxshadow / borderline around the element the class is applied to indicate editing or failure.
   - wouldnt that make `highlight` useless?

proposed fixes
- introduce the prop `propertyName` (currently edited property name) (new in NodeCreationDialog)
- introduce the prop `htmlIdentifier` as alias for `id`
- introduce a better way to unset properties instead of calling `commit("")` by passing a prop `reset` which calls commit
- introduce the prop `applyPendingChanges` as alias for `onEnterKey`

proposed cleanups
remove
- `editorRegistry`
- `i18nRegistry`
- `onKeyPress`
- `identifier`
- `id`
- `onKeyPress`


the question is how we propose to cleanup and fix the api.
I would like to avoid fixing things only for new Ui versions - so that we get into a state where old plugins stop working and new ones must be adjusted for old neos version.
I really like that not much changes in the ui for plugin writers so things generally keep on working over long times.

all of these fixes and be implemented in older neos versions, so one could use the improved api today and we would wait for breaking cleanups for a major version.

but the important thing is in my opinion, that one plugin would still work for a range of neos ui versions.

PRGfx:
Some collected thoughts on that:

I would expect that at least there aren't many editors specifically for the node creation dialog, so I think adjusting these changes to the inspector use case shouldn't be too bad.

Regarding prop renaming I would suggest something more descriptive for the HTML ID. While I was thinking about htmlId (like `htmlFor`), maybe it makes sense to include a hint about where it will be applied. As with refs on more complex components where I like an `inputRef` vs a `containerRef` etc. Will the ID be applied to the envelope or a wrapper around the editor?

`propertyName` sounds reasonable to me however the creation dialog may show elements now related to properties e.g. elements that will then be used some other way by the node templates package. I'm not sure what to propose instead. Technically it could be mapped from the element name.

Similar things go for `node`: in the creation dialog there is no node yet, I think. Best it could be is `node?` I guess...

i18nRegistry seems reasonable to have imo.

I don't understand your remark on the help properties. They are used in another render method.

+1 for explicit props passing.

PR #3110, opened 2022-04-20
BUGFIX: Check canEdit policy instead of hidden property

Sebobo:
**What I did**

Fix the copy & paste error by checking if a node is allowed to be edited instead of checking for the hidden property which was actually meant for the `visibilityToggleAllowed` check.

**How to verify it**

Create a `Policy.yaml` with the following content:

```yaml
privilegeTargets:
  'Neos\ContentRepository\Security\Authorization\Privilege\Node\EditNodePropertyPrivilege':
    'My.Site:DisableHidden':
      matcher: nodeIsOfType("Neos.Neos:Document") && nodePropertyIsIn(["_hidden", "hidden"])
```

Without the fix, no actions are possible on the selected document node. With the fix, only the visibility is disabled.


markusguenther (approved):
Tested that and it works fine, thanks @Sebobo

PR #3120, opened 2022-05-23
TASK: Update plow-js to v3.0.0

grebaldi:
This PR contains an update for `plow-js` v3.0.0 (as tracked here: https://github.com/grebaldi/plow-js/pull/21).

`plow-js` v3.0.0 isn't released yet, that's why currently all packages point to `plow-js` v3.0.0-canary2.

I'd like to use this PR to gather some feedback. I did a smoke test in a limited Neos UI setup, but it would certainly be nice to get some feedback from 1 or 2 real-world projects.

If feedback's positive, I'll do a final release of `plow-js` v3.0.0 and update this PR accordingly.


mhsdesign:
@markusguenther this got now closed too...

Sebobo:
Yes it now needs to target master, but I cannot change it. 
@grebaldi can you make the change?

mhsdesign:
we need a new pr i guess (i cant even reoopen as the branch is deleted)

Sebobo:
Only the target branch is deleted

mhsdesign:
does need a rebase either way - and probably yarn install with the yarn cache stuff

Issue #3129, opened 2022-07-01
BUG: backend crashes if user is in a non existing edit mode


haase-fabian:
### Description

After removing a custom edit mode. Users who have still selected that mode can no longer login to the backend. 

### Steps to Reproduce

1. add a custom edit mode
2. change the edit mode to that custom edit mode with a user
3. remove that custom edit mode
4. relog with that user

#### Expected behavior

seeing the backend of the neos installation

#### Actual behavior

Basically the backend crashes right after logging in.
blank space

1. TypeError: Cannot read properties of undefined (reading 'backgroundColor')
2. TypeError: Cannot read properties of undefined (reading 'isEditingMode')
3. uncaught at application  TypeError: Cannot read properties of undefined (reading 'backgroundColor')

### Affected Versions

Neos: 8.0.3
UI: 8.0.1

haase-fabian:
A fallback to the inline edit mode in case the selected mode is not defined should be sufficient to solve this issue.

PR #3130, opened 2022-07-01
BUGFIX: fallback to a defined edit mode if edit mode is undefined

haase-fabian:
**What I did**

Whenever an edit mode is removed and a user has still selected set mode the backend would crash.
Now it can keep loading allowing the user to select another edit mode.  

see #3129

**How I did it**

I added a fallback whenever I found `editPreviewMode[currentPreviewMode] `being requested.
I chose to use a non hard-coded fallback in case the edit modes get renamed or removed.

I tested it on the master branch. But committed to 5.3.
I did not find other occurences of  `editPreviewMode[` but I can't garantee the issue is completly fixed on older versions. 

**How to verify it**

change an entry in the table `neos_neos_domain_model_userpreferences` for example to `a:1:{s:30:"contentEditing.editPreviewMode";s:7:"inPlXXX";}`

Previous behaviour was recieving a blank page. Now the backend is at least shown:

Screenshot 2022-07-01 at 21-42-04 Neos Error


Sebobo:
Hi @haase-fabian, thx!

I had to rebase the PR as we don't create fixes for 5.3 anymore and I improved the behaviour further to actually switch the mode and reload: #3193
So I will close this PR but your change and mention is also contained in the new one :)

Issue #3031, opened 2022-02-19
[UX-Feature] Filtering node types selects the first result by default [and inserts it by hitting `enter`]


putzwasser:
### Description

When you filter for node types the creation/insert dialog it would be nice if the dialog would automatically highlights the first filter result.

![image](https://user-images.githubusercontent.com/26040044/154813641-bc828424-4e26-4e46-8a4e-a953efdc03db.png)

In this case »Text/Bild Kombination«.

If you’d hit enter it would "instantly" insert the currently selected node type (in this case) below node type, which you clicked on to spawn the dialog.

#### "Bonus"
Selecting the other filtered ones via the arrow keys would be nice. But that would mean:
1. arrow keys are just for navigation between the nodes, you couldn’t move the cursors in the input field anymore
2. you could only move the cursor by clicking
3. you could only add characters or remove all of them until (using `backspace`) until you get to the point you wanted to edit

I would be fine with that limitation.

mhsdesign:
i would expect this ux behavior:

- type `bil`
- nodes are filtered, first item in list is semi highlight selected so enter works
- arrow keys and focus still is in the input field.
- press tab to move out of the input field and then use arrow keys for selecting. (enter would save and close)
- shift tab to move back into the search bar.

escape at anytime will close this window.

danielkestler:
We could also open this dialog with `Cmd` + `K` like all the cool apps do.

Sebobo:
The search field should also be autofocused when the dialog opens.

PR #3132, opened 2022-07-06
FEATURE: Improvements for the Node Creation Dialog

danielkestler:
Add improvements to the Node Creation Dialog like autofocusing the filter input field, keyboard navigation and insertion, see #3031

**What I did**
- autofocus node type filtering input element
- auto select first matching node type
- add keyboard navigation for up/down navigating the node type selection
- add enter keypress for inserting node

**How I did it**
- extended the state to hold a list of all filtered node types and the user selection that can then be toggled via the keyboard up/down arrows 

**How to verify it**
- open node type insertion dialog
- enter some filter text into the filter input
- use up/down arrows to select node type
- press enter to insert node

Bildschirmfoto 2022-07-06 um 15 50 57


danielkestler:
We should make the input field more prominent to better fit our changes.

Bildschirmfoto 2022-07-07 um 08 26 55

Bildschirmfoto 2022-07-07 um 08 29 11

Sebobo:
Cool. Like it much better 👍

jonnitto:
Do we really want the autofocus feature? If you edit the page on a device without a keyboard (e. g. iPad), the software keyboards pop up

Sebobo:
I love having autofocus in general. 
Is it possible to have a feature/device check whether we enable that?

jonnitto:
Unfortunately, currently, there is no direct way to detect if a virtual keyboard appears on the screen using JavaScript.

markusguenther:
It is such a nice improvement. Do we want to discuss the autofocus this week? Or do we take it as it is?
We have still time until 17th of November (feature freeze)

@jonnitto

jonnitto (approved):

packages/neos-ui/src/Containers/Modals/SelectNodeType/nodeTypeFilter.js
```diff
@@ -15,19 +15,31 @@ const NodeTypeFilter = ({onChange, filterSearchTerm, i18nRegistry}) => {
...
 
     return (
         <div className={style.nodeTypeDialogHeader__filter}>
-            {filterSearchTerm ? (
-                <IconButton icon="times" onClick={handleResetFilter}/>
-            ) : (
-                <Icon icon="filter" padded="right"/>
-            ) }
+            <div className={style.nodeTypeDialogHeader__filterIconSearch}>
+                <Icon icon="search" size="md"/>
```
markusguenther: The size md is invalid. We have just `"lg","xs","sm","1x","2x","3x","4x","5x","6x","7x","8x","9x","10x"`

packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
```diff
@@ -182,10 +188,97 @@ export default class SelectNodeType extends PureComponent {
...
+
+            if (filteredNodeTypes.length > 0) {
+                filteredNodeTypesFlat.push(...filteredNodeTypes);
+                filteredNodeTypesGrouped.push(value);
+            }
+
+            return true;
+        });
+
+        const focusedNodeType = (filterSearchTerm !== '' && filteredNodeTypesFlat.length > 0) ? filteredNodeTypesFlat[0].name : null;
```
markusguenther: It happens that the property name is not defined. that leads to errors in the console.
```Uncaught TypeError: can't access property "name", filteredNodeTypesFlat[(index + 1)] is undefined```

File: packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js [hunk 3]
```diff
@@ -182,10 +188,97 @@ export default class SelectNodeType extends PureComponent {
...
+    }
+
+    handleKeyDown = event => {
+        if (event.key === 'ArrowUp' || event.key === 'ArrowDown') {
+            event.preventDefault();
+
+            const {filteredNodeTypesFlat, focusedNodeType} = this.state;
+            const index = filteredNodeTypesFlat.findIndex(element => element.name === focusedNodeType);
+
+            if (event.key === 'ArrowUp') {
```
markusguenther, line 271: Here we need to check if the index is valid ;)

packages/neos-ui/src/Containers/Modals/SelectNodeType/nodeTypeFilter.js
markusguenther: Replaced it now with `1x` if it is wrong, we can change it before the 8.2 release ;)

packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
markusguenther: that was not the root case but added a check to be save that it is working

packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
markusguenther: fixed that issue

markusguenther (approved):
Thanks, @danielkestler, for the improvements. Looking forward to more UX changes ;)

PR #3134, opened 2022-07-07
TASK: Improve rendering of outlines for focused or hovered nodes

danielkestler:
Sometimes, the outlines of selected or hovered nodes in inline editing were not correctly rendered – due to CSS 
specificity issues. That lead to nodes being selected and then the toolbar floating around not really visually connected to the node.

Also, different gaps for hovering (dashed) and selected (solid) made these lines "seem jumping around" sometimes. This PR fixes it.

**What I did**

Fix strange outline rendering for selected nodes.

**How I did it**

Adjusting CSS for better specificity and equalize outline offset.

**How to verify it**

Selecting nodes in inline editing should now always render the correct outlines.

Bildschirmfoto 2022-07-07 um 13 17 49
Bildschirmfoto 2022-07-07 um 13 17 44

### DIFF
--- a/packages/neos-ui-guest-frame/src/InlineUI/NodeToolbar/Buttons/AddNode/index.js
+++ b/packages/neos-ui-guest-frame/src/InlineUI/NodeToolbar/Buttons/AddNode/index.js
@@ -68,12 +68,11 @@ export default class AddNode extends PureComponent {
         if (preferredMode === 'before' || preferredMode === 'after') {
             return isAllowedToAddSiblingNodes;
         }
-        else if (preferredMode === 'into') {
+        if (preferredMode === 'into') {
             return isAllowedToAddChildNodes;
         }
-        else {
-            return isAllowedToAddChildOrSiblingNodes;
-        }
+
+        return isAllowedToAddChildOrSiblingNodes;
     }
 
     render() {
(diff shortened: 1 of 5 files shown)


Sebobo:
> Okay, well, this is not perfect yet. But to be honest, it was not the real intent of this PR – which comes here:
> 
> Bildschirmfoto 2022-07-07 um 14 25 40
> 
> I moved the "Insert above" / "Insert below" / "Insert into" decision to the "AddButton" of the toolbar. This is still WIP, I just want to know, what do you think of such a change? Maybe it is visually to obtrusive?

Thx for your work to improve the current situation and I'm so happy to see new contributions from you!

I think the 3 button areas area a bit confusing. Maybe having the other buttons also on the left, but with a separator would be better?
Will the insertion mode in the dialog then match depending on which button is pressed?

danielkestler:
Thank you, will have another look at UX. The insertion mode is of course respected in the Node Creation Dialog.

mhsdesign:
@danielkestler should we just merge the css-outline changes separately since they are faster and easier to review? ^^

About the insertion situation. Great that you took the initiative!! I dont know if im alone with this opinon but i think there can be a lot improved - this might even be a topic for a small hangout if your up for it.

(maybe have those buttons show only on hover? idk - lets see how wordpress gutenberg handles it xD)

danielkestler:
@mhsdesign Thank you, yes we could, but they are not perfect yet – still on improving it, will let you know!

I will also have to take a look at Gutenberg and maybe Webflow.

Sebobo:
@danielkestler so what do you actually want to fix now in this PR? ;)
IMO we should split the button related changes into a separate PR.

mhsdesign:
The specifity issues should hopefully be fixed by now - otherwise we need a new issue;)

But more importantly the plus add buttons are now part of 9.1 so this can be closed 🥳 

https://github.com/neos/neos-ui/pull/4016

Issue #3165, opened 2022-07-29
AssetProxies wrongly reimport assets


kitsunet:
The link editor handles assets differently from the asset editor and doesn't correctly check if an asset is already available locally, this leads to unnecessary calls to import proxies and can lead to duplicates if a given proxy is implemented similarly to the Neos asset source.

PR #3166, opened 2022-07-29
!!!BUGFIX: AssetProxies wrongly reimport assets

kitsunet:
The link editor handles assets differently from the asset editor and doesn't correctly check if an asset is already available locally, this leads to unnecessary calls to import proxies and can lead to duplicates if a given proxy is implemented similarly to the Neos asset source. This change addresses the issue, and prevents duplicate imports.

A minor breaking change had to be made though to preserve the rest of the functionality. IF you use the asset proxy search or detail endpoints of our React UI the `loaderUri` (which is an internal implementation detail specifically for the link editor) has changed format and will now include only `assetProxy://<UUID>` for any asset that is already local (regardless if it comes from a proxy or not).

Further refactoring in this area is recommended in future versions.

Fixes: #3165


kitsunet:
Note, I had to make this a breaking change and fix the loaderUri as the linkEditor NEEDS the loaderUri to work because it operates on assets and nodes and needs the loaderUri to differentiate, so I saw no other way. I would assume the breakage is rather small though.

kdambekalns (approved):

jonnitto (approved):
Fine by reading

PR #3078, opened 2022-03-14
TASK: Increase PHP version for CircleCI

markusguenther:
We now use PHP version 8.0 as minimal version. So we need to adjust the CircleCI images.

PR #3175, opened 2022-08-31
TASK: Temporary disable of 2 dimension tests

markusguenther:
We have some issue on CircleCI since the PHP8 image. The switch of the dimension fixture is not working and then the 2dimension test is always running on the data of the 1 dimension fixture. As we have just one test case for the 2 dimensions, and it works locally, we disable that for a while and investigate the issue.


markusguenther:
@kdambekalns made the PR https://github.com/neos/neos-ui/pull/3078 run. So no need for that.

Issue #3057, opened 2022-03-02
AssetEditor: Cannot replace the Asset on clicking the Asset's title


lorenzulrich:
### Description

The Neos UI "suggests" that you could replace an Asset in the AssetEditor by clicking its title and searching another file.

### Steps to Reproduce

1. In an AssetEditor property, select an Asset and apply.
2. Click the Asset (not the X button to the right!): The Asset is "removed" and the field is turned into a search field.
3. Start typing a search term.

#### Expected behavior

Matching Assets would be listed and are clickable.

#### Actual behavior

No Asset but the currently used one can be found or selected.

### Demo

https://user-images.githubusercontent.com/1816023/156387922-e4adf609-c21e-460d-b029-35f0442c73a4.mp4

### Comment

As mentioned above (and can be seen in the video), if the Asset is removed first using the X button, replacing the Asset works.
So it would be perfectly fine to just do nothing on clicking the current Asset.

### Affected Versions

I reproduced this in 4.3 and 7.3

PR #3187, opened 2022-09-07
FEATURE: Linking referenced documents in the reference and references editor

FahimNsr:
**What I did**  
I added the feature, by clicking on referenced Document, the URL will change to the referenced Document.

**How I did it**
I got the Action "actions.UI.ContentCanvas.setSrc" from redux in reference and references components and set the new URL based on referenced Document by clicking on them.

**How to verify it**
Select a Document in reference and references editors and click on them.


packages/react-ui-components/src/MultiSelectBox/multiSelectBox.js
```diff
@@ -148,7 +148,8 @@ class MultiSelectBox extends PureComponent {
         SelectBox: PropTypes.any.isRequired,
         IconComponent: PropTypes.any.isRequired,
         IconButtonComponent: PropTypes.any.isRequired,
-        MultiSelectBox_ListPreviewSortable: PropTypes.any.isRequired
+        MultiSelectBox_ListPreviewSortable: PropTypes.any.isRequired,
+        setActiveContentCanvasSrc: PropTypes.func
```
Sebobo: Setting the content canvas src shouldn't be a concern of the SelectBox as its super specific and not reusable in this generic component which is also used in different contexts.

I would prefer when the SelectBox has a callback prop for clicking the header. This should be easily adjustable from what I see in your other code.
And then the reference editor which knows about Neos can have the handler which will then set the content canvas.

Sebobo (changes requested):

packages/react-ui-components/src/MultiSelectBox_ListPreviewSortable_DraggableListPreviewElement/multiSelectBox_ListPreviewSortable_DraggableListPreviewElement.js
```diff
@@ -127,20 +134,22 @@ export default class MultiSelectBox_ListPreviewSortable_DraggableListPreviewElem
                             icon={'grip-lines-vertical'}
                             className={theme.selectedOption__moveButton}
                             hoverStyle={'clean'}
-                            />
+                        />
                     )}
-                    <div className={theme.selectedOptions__innerPreview}>
+                    <div className={theme.selectedOptions__innerPreview} role="button" onClick={() => {
```
Sebobo: Not every option has an uri (not every reference is a document) and this would create an invalid action. So you need to check if an uri exists. This is another reason why the ReferenceEditor should supply the handling as it knows more about the options themselves.

Sebobo:
Thx for your feature. I think its great as I also missed it several times when working in the backend.
But we need to move the main logic into the ReferenceEditor instead of the SelectBox. Also you need to make sure that this feature is only active when valid urls exist.

FahimNsr:
Hi @Sebobo 

I fixed the mentioned problems and build the functionality, as you demanded, please review it again.

kind regards
Fahim

packages/react-ui-components/src/SelectBox/selectBox.js
```diff
@@ -187,9 +192,10 @@ export default class SelectBox extends PureComponent {
             disabled,
             className,
             DropDown,
-            SelectBox_ListPreview
+            SelectBox_ListPreview,
+            onReferenceClick
         } = this.props;
-
+        // console.log(this.props)
```
daniellienert: That should be removed

FahimNsr: done!

crydotsnake (changes requested):
When I click on a referenced document in the backend, which is no longer available (and therefore the URI no longer exists) an error message is thrown.

SCR-20221110-thl

So as @Sebobo already mentioned, your feature should only show up, if the referenced URI exists.

### Edit
I have found that the problem I described has nothing to do with this PR. The problem probably lies somewhere else.

markusguenther:
Hi @crydotsnake or @FahimNsr I tried to rebase the branch to the latest 8.2 but I am not allowed to push. So would be nice if someone of you can do that :)

daniellienert:
@markusguenther granted you access to the fork. Would you try to push again?

markusguenther:
Thanks @daniellienert ... will take a look

packages/react-ui-components/src/DropDown/header.tsx
```diff
@@ -56,6 +56,15 @@ export interface ShallowDropDownHeaderProps {
 
     readonly toggleDropDown: () => void;
 
+    /**
+     * This prop gets access to the onClick function for reference editor. It changes the url.
+     */
+    readonly onReferenceClick?: (args: string) => void;
```
Sebobo: This shouldn't be called `reference` as mentioned before. This is a presentational component and shouldn't have deeper knowledge about its usage. So `onItemClick` or something similar should be better.

packages/react-ui-components/src/DropDown/header.tsx
```diff
@@ -56,6 +56,15 @@ export interface ShallowDropDownHeaderProps {
...
+
+    /**
+     * This prop gets uri of refrenced document.
+     */
+    readonly referenceURL?: string;
```
Sebobo: Please nothing related to references or urls here.

packages/react-ui-components/src/MultiSelectBox_ListPreviewSortable_DraggableListPreviewElement/multiSelectBox_ListPreviewSortable_DraggableListPreviewElement.js
```diff
@@ -129,7 +137,11 @@ export default class MultiSelectBox_ListPreviewSortable_DraggableListPreviewElem
                             hoverStyle={'clean'}
                             />
                     )}
-                    <div className={theme.selectedOptions__innerPreview}>
+                    <div
+                        className={theme.selectedOptions__innerPreview}
+                        onClick={option.uri ? () => handleClick(option.uri) : undefined }
```
Sebobo: How about forwarding the whole `option` here and let the outside handle do whatever it wants with it?
Then we don't the to know about an `uri`.

packages/react-ui-components/src/SelectBox/selectBox.js
```diff
@@ -59,6 +59,11 @@ export default class SelectBox extends PureComponent {
          */
         onValueChange: PropTypes.func.isRequired,
 
+        /**
+         * This prop gets access to the onClick function for reference editor. It changes the url.
+         */
+        onReferenceClick: PropTypes.func,
```
Sebobo: Also rename the handler here

packages/react-ui-components/src/SelectBox/selectBox.js
```diff
@@ -213,7 +219,12 @@ export default class SelectBox extends PureComponent {
 
         return (
             <DropDown.Stateless className={theme.selectBox} isOpen={isExpanded} onToggle={this.handleToggleExpanded} onClose={this.handleClose}>
-                <DropDown.Header className={headerClassName} shouldKeepFocusState={false} showDropDownToggle={showDropDownToggle && Boolean(options.length)}>
+                <DropDown.Header
+                    className={headerClassName}
+                    shouldKeepFocusState={false}
+                    showDropDownToggle={showDropDownToggle && Boolean(options.length)}
+                    onReferenceClick={onReferenceClick} referenceURL={options[0] && options[0].uri ? options[0].uri : undefined }
```
Sebobo: Don't do anything related to reference here but pass the whole option.

Sebobo (changes requested):
For me there is still too much `reference` related code in the presentational component.
I think we can easily clean this up by not calling it reference and instead of handling an `uri` we can pass the whole option to the outside and let the handler in the actual `referenceEditor` do it's uri magic.

packages/react-ui-components/src/MultiSelectBox_ListPreviewSortable_DraggableListPreviewElement/multiSelectBox_ListPreviewSortable_DraggableListPreviewElement.js
Sebobo: I would directly call `handleClick` and as I said before, pass the whole option to the callback without checking for an uri. Let the callback do whatever (f.e. showing a modal if we need one later on.

Sebobo (changes requested):

FahimNsr:
Hi @Sebobo,

is that now correct?

packages/neos-ui-editors/src/Editors/References/index.js
```diff
@@ -26,13 +32,22 @@ export default class ReferencesEditor extends PureComponent {
...
-        disabled: PropTypes.bool
+        disabled: PropTypes.bool,
+        setActiveContentCanvasSrc: PropTypes.func.isRequired
     };
 
     handleValueChange = value => {
         this.props.commit(value);
     }
 
+    handleClick = src => {
```
Sebobo:
```suggestion
    handleClick = option => {
```

packages/neos-ui-editors/src/Editors/References/index.js
```diff
@@ -26,13 +32,22 @@ export default class ReferencesEditor extends PureComponent {
...
+        const {setActiveContentCanvasSrc} = this.props;
+
+        if (setActiveContentCanvasSrc) {
```
Sebobo:
```suggestion
        if (setActiveContentCanvasSrc && option && option.uri) {
```

packages/neos-ui-editors/src/Editors/References/index.js
```diff
@@ -26,13 +32,22 @@ export default class ReferencesEditor extends PureComponent {
...
+            setActiveContentCanvasSrc(src);
```
Sebobo:
```suggestion
            setActiveContentCanvasSrc(option.uri);
```

packages/react-ui-components/src/MultiSelectBox_ListPreviewSortable_DraggableListPreviewElement/multiSelectBox_ListPreviewSortable_DraggableListPreviewElement.js
```diff
@@ -129,7 +137,11 @@ export default class MultiSelectBox_ListPreviewSortable_DraggableListPreviewElem
                             hoverStyle={'clean'}
                             />
                     )}
-                    <div className={theme.selectedOptions__innerPreview}>
+                    <div
+                        className={theme.selectedOptions__innerPreview}
+                        onClick={handleClick(option.uri)}
```
Sebobo:
```suggestion
                        onClick={handleClick(option)}
```

packages/react-ui-components/src/MultiSelectBox_ListPreviewSortable_DraggableListPreviewElement/multiSelectBox_ListPreviewSortable_DraggableListPreviewElement.js
```diff
@@ -119,6 +119,14 @@ export default class MultiSelectBox_ListPreviewSortable_DraggableListPreviewElem
 
         const handleRemoveItem = () => onRemoveItem(index);
 
+        const handleClick = src => {
```
Sebobo:
```suggestion
        const handleClick = option => {
```

packages/react-ui-components/src/MultiSelectBox_ListPreviewSortable_DraggableListPreviewElement/multiSelectBox_ListPreviewSortable_DraggableListPreviewElement.js
```diff
@@ -119,6 +119,14 @@ export default class MultiSelectBox_ListPreviewSortable_DraggableListPreviewElem
...
+            const {onItemClick} = this.props;
+
+            if (onItemClick) {
+                onItemClick(src);
```
Sebobo:
```suggestion
                onItemClick(option);
```

Sebobo (changes requested):
Almost, see my suggestions

packages/react-ui-components/src/MultiSelectBox_ListPreviewSortable_DraggableListPreviewElement/multiSelectBox_ListPreviewSortable_DraggableListPreviewElement.js
FahimNsr: option or option.uri?

Sebobo: You should always pass the option itself out of the select box, as it cannot know whether an option has an uri and the ReferenceEditor knows more about its options.

packages/neos-ui-editors/src/Editors/References/index.js
```diff
@@ -26,13 +32,22 @@ export default class ReferencesEditor extends PureComponent {
...
 
     handleValueChange = value => {
         this.props.commit(value);
     }
 
+    handleClick = option => {
+        const {setActiveContentCanvasSrc} = this.props;
+
+        if (setActiveContentCanvasSrc && option && option.uri) {
+            setActiveContentCanvasSrc(option.uir);
```
Sebobo: A typo

```suggestion
            setActiveContentCanvasSrc(option.uri);
```

Sebobo (changes requested):

Sebobo (approved):
It's fine for me now. What do @mhsdesign and @markusguenther think?

mhsdesign:
I will try this out - otherwise a little video would help too :P

crydotsnake (changes requested):
I don't know if it's just me or if I'm misunderstanding something, but for me the functionality is a bit broken. When you reference something, you are redirected directly to the reference. But that should not be so. And as soon as you switch back to the page with the reference NodeType and click on the NodeType, you are automatically redirected to the reference again. Just to be sure, i also cleared the cache and build the JS files before.

https://user-images.githubusercontent.com/39345336/208378939-bb0041f6-ab2f-4d19-9b19-7fecbfc6442a.mov

It would be great if someone could try to reproduce my behavior

Sebobo:
@crydotsnake thx for testing, I can reproduce the issue

Sebobo:
I found the error and created a fix, but I need permission to push to this branch

Sebobo:
@crydotsnake can you test again

crydotsnake (approved):
From the functionality its working great now!

packages/react-ui-components/src/MultiSelectBox_ListPreviewSortable_DraggableListPreviewElement/multiSelectBox_ListPreviewSortable_DraggableListPreviewElement.js
```diff
@@ -89,6 +89,14 @@ export default class MultiSelectBox_ListPreviewSortable_DraggableListPreviewElem
         IconButton: PropTypes.any.isRequired
     }
 
+    handleClick = () => {
+        const {onItemClick, option} = this.props;
+
+        if (onItemClick) {
+            onItemClick(option);
+        }
```
grebaldi, line 98: `onItemClick` is missing in the `propTypes` declarations:

```javascript
onItemClick: PropTypes.func,
``` 

`<MultiSelectBox_ListPreviewSortable/>` and `<MultiSelectBox/>` should also declare it.

packages/react-ui-components/src/MultiSelectBox_ListPreviewSortable_DraggableListPreviewElement/multiSelectBox_ListPreviewSortable_DraggableListPreviewElement.js
```diff
@@ -129,7 +137,11 @@ export default class MultiSelectBox_ListPreviewSortable_DraggableListPreviewElem
                             hoverStyle={'clean'}
                             />
                     )}
-                    <div className={theme.selectedOptions__innerPreview}>
+                    <div
+                        className={theme.selectedOptions__innerPreview}
+                        onClick={this.handleClick}
+                        role="button"
```
grebaldi:
```suggestion
                        role={this.props.onItemClick ? "button" : null}
```

Likewise, `role="button"` only makes sense, if the element is actually supposed to trigger an action on click.

packages/react-ui-components/src/MultiSelectBox_ListPreviewSortable_DraggableListPreviewElement/multiSelectBox_ListPreviewSortable_DraggableListPreviewElement.js
```diff
@@ -129,7 +137,11 @@ export default class MultiSelectBox_ListPreviewSortable_DraggableListPreviewElem
                             hoverStyle={'clean'}
                             />
                     )}
-                    <div className={theme.selectedOptions__innerPreview}>
+                    <div
+                        className={theme.selectedOptions__innerPreview}
+                        onClick={this.handleClick}
```
grebaldi:
```suggestion
                        onClick={this.props.onItemClick ? this.handleClick : null}
```

packages/neos-ui-editors/src/Editors/Reference/index.js
```diff
@@ -48,6 +56,7 @@ export default class ReferenceEditor extends PureComponent {
             options={sanitizeOptions(options)}
             value={value}
             onValueChange={this.handleValueChange}
+            onReferenceClick={() => value && setActiveContentCanvasSrc(options[0].uri)}
```
grebaldi: As far as I can tell, `onReferenceClick` is never triggered anywhere inside the `<SelectBox/>` implementation.

Looks to me like a leftover from an earlier stage of this PR. So, the prop is probably supposed to be called `onItemClick`, but the `<SelectBox/>` is missing an implementation for that as well.

packages/react-ui-components/src/MultiSelectBox_ListPreviewSortable_DraggableListPreviewElement/multiSelectBox_ListPreviewSortable_DraggableListPreviewElement.js
grebaldi: Outside the `ReferencesEditor` use case, most instances of the `<MultiSelectBox/>` have no way of informing this click handler. It should therefore only be set, if it can actually perform an action.

grebaldi (changes requested):
Hi @FahimNsr,

thanks for your PR :slightly_smiling_face:  I think, it'll make a great feature for the Reference(s)Editor.

I was able to verify the functionality for the `ReferencesEditor` case. But I noticed some problems still:

- Imho, the `setActiveContentCanvasSrc` action should not be triggered, when the Inspector is in `dirty` state (especially not, when there are changes in the respective `references` property). At least not without triggering the infamous "Throw away?"-Dialog first, because that is what would happen in any other scenario in which the user tries to navigate away while there are active changes in the Inspector
- Likewise, the `setActiveContentCanvasSrc` action should not be triggered, when the `ReferencesEditor` is used from inside the `NodeCreationDialog`. Otherwise, the user can accidentally change the page they're looking at while they're trying to create a node - and this is quite hard to notice even, because the dialog covers the page almost entirely.

I was unable to get this functionality to work for the `ReferenceEditor` (singular). A closer Code Review leaves me under the impression that the implementation for it is (unintentionally) missing. (See my code comments below)

Besides that, I added some smaller nitpick comments.

grebaldi:
On an additional note: Linking from the `ReferenceEditor` (singular) may be problematic because of the behavior described in #3057. Clicking on the selected item of a single select removes the item for some reason (I've noticed this too while testing this PR).

packages/neos-ui-editors/src/Editors/Reference/index.js
FahimNsr: I think we don't need to implement it in `<SelectBox/>`, instead, we name it onClick.

it works but https://github.com/neos/neos-ui/issues/3057https://github.com/neos/neos-ui/issues/3057 is still there.

packages/react-ui-components/src/MultiSelectBox_ListPreviewSortable_DraggableListPreviewElement/multiSelectBox_ListPreviewSortable_DraggableListPreviewElement.js
FahimNsr: I agree with you about `<MultiSelectBox_ListPreviewSortable/>` but why we need it in `<MultiSelectBox/>`

FahimNsr:
Hi @grebaldi,

I fixed the problems that you commented now it works in both editors if NodeCreationDialog isn't open and there are no changes in Inspector,

but as I said in the Comment section, https://github.com/neos/neos-ui/issues/3057 is still there, maybe we should wait for the other PR.

packages/react-ui-components/src/MultiSelectBox_ListPreviewSortable_DraggableListPreviewElement/multiSelectBox_ListPreviewSortable_DraggableListPreviewElement.js
grebaldi: Because it passes `onItemClick` to the other components and thus accepts it as a prop like here in the ReferencesEditor:
https://github.com/neos/neos-ui/blob/3cc02a49bc64a376388792785599debfe14a8f91/packages/neos-ui-editors/src/Editors/References/index.js#L74

packages/neos-ui-editors/src/Editors/Reference/index.js
```diff
@@ -48,6 +69,7 @@ export default class ReferenceEditor extends PureComponent {
             options={sanitizeOptions(options)}
             value={value}
             onValueChange={this.handleValueChange}
+            onClick={() => this.handleClick(options[0].uri)}
```
grebaldi: This works :) 

When I tested this in combination with #3341, I drilled through the `<SelecBox/>` call chain and noticed, that `onClick` is missing in the `propTypes` declarations of the following components:

- `SelectBox`
- `SelectBox_Header`

I'd like to suggest however, that `SelectBox` should accept `onHeaderClick` and forward it as `onClick` to the `SelectBox_Header`. This way, it's a clearer API.

grebaldi (changes requested):
Hi @FahimNsr,

thanks a lot for your adjustments. Works nicely :ok_hand: 

I tested this in combination with #3341 and noticed that I need to make adjustments over there. We need to coordinate, so #3341 won't break your new feature in the future.

I'd like to suggest the following tasks on your end:

- [ ] Rename `<SelectBox {/*...*/} onClick={/*...*/} {/*...*/}/>` -> `<SelectBox {/*...*/} onHeaderClick={/*...*/} {/*...*/}/>` in `ReferenceEditor`
- [ ] Add `onHeaderClick` to `<SelectBox/>` propTypes
- [ ] Forward `onHeaderClick` prop to `<SelectBox_Header onClick={onHeaderClick}/>` in `renderHeader` method of `<SelectBox/>`
- [ ] Add `onClick` to `<SelectBox_Header/>` propTypes
- [ ] Add `onItemClick` to `<MultiSelectBox/>` propTypes
- [ ] Add `onItemClick` to `<MultiSelectBox_ListPreviewSortable/>` propTypes

If you agree with these, I will make the following adjustments in #3341:

- [ ] Add `onHeaderClick` to `<SelectBox/>` propTypes
- [ ] Forward `onHeaderClick` prop to `<SelectBox_Header onClick={onHeaderClick}/>` in `renderHeader` method of `<SelectBox/>`
- [ ] Add `onClick` to `<SelectBox_Header/>` propTypes
- [ ] Call `onClick` (if set) in `handleListPreviewElementClick` method of `<SelectBox_Header/>`

What do you say?

**UPDATE:** Turns out, I actually needed this kind of API to fix the E2E-Test issue over at #3341 :)
This means, that #3341 now already implements the aforementioned TODOs, but we still need to harmonize of course.

FahimNsr:
Hi @grebaldi,

I pushed the requested changes.
I have to write `{/* eslint-disable-next-line jsx-a11y/no-static-element-interactions */}` for `<MultiSelectBox_ListPreviewSortable_DraggableListPreviewElement>` in line 142 because of the conditional role for div. I take a look at another role in other codes but we never used a conditional role for any Element in the whole Neos UI app.

About E2E-Test honestly I didn't get, how to fix it. I would be grateful if you help me to fix it.

Best regards,

Fahim

packages/react-ui-components/src/SelectBox/selectBox.js
```diff
@@ -59,6 +59,14 @@ export default class SelectBox extends PureComponent {
          */
         onValueChange: PropTypes.func.isRequired,
 
+        /**
+         * This prop gets called when an option was selected,
+         * the inspector is not in dirty status and
+         * the NodeCreationDialog is not open.
+         * It handles the linking of the reference editor.
+         */
```
grebaldi:
```suggestion
        /**
         * This prop gets called when the select box header element gets clicked.
         */
```
This is to avoid merge conflicts with: https://github.com/neos/neos-ui/pull/3341

packages/react-ui-components/src/MultiSelectBox/multiSelectBox.js
```diff
@@ -46,6 +46,14 @@ class MultiSelectBox extends PureComponent {
          */
         onValuesChange: PropTypes.func.isRequired,
 
+        /**
+         * This prop gets called when an option was selected,
+         * the inspector is not in dirty status and
+         * the NodeCreationDialog is not open.
+         * It handles the linking of the references editor.
+         */
```
grebaldi:
```suggestion
        /**
         * This prop gets called when one of the selected items gets clicked.
         */
```

This is for consistency.

grebaldi (changes requested):
Hi @FahimNsr,

thanks for your quick adjustments :)

I've adjusted #3341, but would like to kindly ask you for another 2 small adjustments, to ease the later upmerges: (see comments below).

To address your comments:

> I have to write {/* eslint-disable-next-line jsx-a11y/no-static-element-interactions */} for <MultiSelectBox_ListPreviewSortable_DraggableListPreviewElement> in line 142 because of the conditional role for div

Yep, I find it pretty hard to decide, how to implement the `role` assignment correctly here. I've taken a closer look at that `jsx-a11y` rule (over here: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-static-element-interactions.md), and apparently, `jsx-a11y` would prefer both arms of the ternary condition to provide a string literal.

If I understand the WAI-ARIA definition correctly, `null` could potentially be replaced with `"none"` in this case:
https://www.w3.org/TR/wai-aria-1.1/#none

But, there's a catch. According to WAI-ARIA, `"none"` is synonymous with `"presentation"`:
> In ARIA 1.1, the working group introduced none as a synonym to the presentation role, [...]
*(see: https://www.w3.org/TR/wai-aria-1.1/#none)*

And `jsx-a11y` has an opinion about that:
> Do not use the role presentation on the element: it removes the element's semantics, and may also remove its children's semantics, creating big issues with assistive technology.
*(see: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-static-element-interactions.md)*

Exactly for that reason, I find it weird that `jsx-a11y` decides to complain, if a conditional `role` assignment contains a `null` branch. Imho, it is perfectly fine to leave out the role, when there's no associated interaction (In fact, that's just the default behavior). And it would be otherwise highly problematic to suggest interaction where there is none...

So, I guess the unfortunate `eslint-disable-next-line` comment is the price that needs to be payed for semantic correctness.

> About E2E-Test honestly I didn't get, how to fix it. I would be grateful if you help me to fix it.

I took those results as random hiccups at first, but they seem to persist :sweat_smile: 

I'll take care of this and get back to you, if I find anything :)

grebaldi (approved):
Hi @FahimNsr,

thanks a lot for your adjustments :) Sorry for the ping-pong game :sweat_smile: 

I was unable to reproduce the E2E-test failures locally and found out that your PR is affected by a broader issue with the E2E-tests right now (that affects other PRs as well). Long story short: nothing needs to be done here, everything's :+1:

FahimNsr:
Hi @grebaldi,

I'm so happy about my first contribution to Neos UI. I hope I can contribute more to improve the Neos UI together.

best regards

Fahim

crydotsnake:
🎉 🎉

PR #3193, opened 2022-10-04
BUGFIX: Fallback to a defined edit mode if the selected edit mode is undefined

Sebobo:
**What I did**

Whenever an edit mode is removed and a user has still selected set mode the backend would crash.
Now it makes sure that no errors occur and falls back to the first selectable mode.

see #3129
replaces #3130

**How I did it**

All occurrences of the edit mode have been enhanced with additional checks to prevent errors.
In addition the selector will check that the selected one exists and will fallback if not.
The new mode will then be stored and the guest frame reloaded.

**How to verify it**

change an entry in the table `neos_neos_domain_model_userpreferences` for example to `a:1:{s:30:"contentEditing.editPreviewMode";s:7:"inPlXXX";}`

Previous behaviour was recieving a blank page. Now the backend will reload with a viable mode.


Sebobo:
This change was initially started by @haase-fabian and I rebased and added the update & reload of the guest frame  👍

haase-fabian:
thank you @Sebobo

JamesAlias (approved):
We need to update the build system quickly to make use of optional chaining and nullish coalescing asap 😄

mhsdesign:
ASAP!!!!!

PR #3194, opened 2022-10-04
BUGFIX: Auto-publish menu alignment 

### LINKED ISSUE #2993: Auto-Publish checkbox is just not properly aligned
### Description

The Auto-Publish checkbox  and label is just not properly aligned actually.

### Steps to Reproduce

1. Login to the backend 
2. Open the publish drop-down

#### Expected behavior

Checkbox and label are in the same line and the checkbox looks nice with aligned check mark

#### Actual behavior

Label is not in the same line and the check mark is not centered.
Screenshot 2021-12-11 at 14 00 49

### Affected Versions

Neos: 5.3

UI: 5.3.x

pKallert:
closes #2993 

**What I did**
The checkbox for the Auto-publish option in the primary menu was not correctly aligned. 
Should work now.

**How I did it**
I used flexbox to align the items next to each other, just like the other checkboxes in the backend do. 

To align the check svg in the checkbox I changed the alignment of the blue box background. I tested this on other instances of the checkbox in the backend and the changes did not break any checkbox as far as I could see. 

**How to verify it**

Previous: 
Bildschirmfoto 2022-10-04 um 16 08 13

Now: 
Bildschirmfoto 2022-10-04 um 16 07 28

### DIFF
--- a/packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/style.css
+++ b/packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/style.css
@@ -59,10 +59,6 @@
         }
     }
 
-    > label {
-        margin-bottom: 0;
-    }
-
     > a,
     > label,
     > button {
@@ -80,6 +76,12 @@
         > input {
             margin-right: .5em;
         }
+    }
+
+    > label {
+        margin-bottom: 0;
+        display: flex;
+        align-items: center;
     }
 
     .badge {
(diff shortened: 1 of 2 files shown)


Sebobo:
Thx! Works as expected but the problem is already in Neos 7.3 so it should be targeted on that branch.
Can you adjust it :)

Sebobo (changes requested):
Needs rebase

PR #3195, opened 2022-10-04
!!! BUGFIX: CKEditor now uses em for italic

### LINKED ISSUE #2906: Italics rendered as <i>, not <em>
### Description

I am seeing `<i>…</i>` in the output instead of `<em>…</em>` as I expect.

### Steps to Reproduce

1. Select some text in the inline editor and click the "Italics" button
2. An `<i>…</i>` element is created

### Expected Behaviour

That used to create an `<em>…</em>` element.

### Affected Versions

UI: 5.3.0 and above, probably earlier as well (any version using CKE5!?)

JamesAlias:
**What I did**
Fixed #2906 
**How I did it**
Created a new CKEditor Plugin replaces `<i>` tags with `<em>` tags.
**How to verify it**
- Open Neos. 
- Open a Page where you can use CKEditor to edit Text.
- Mark text and make it Italic.
- Inspect this text in Devtools.
- The text should be wrapped in an *em* tag

Screenshot 2022-10-04 at 6 29 37 PM

### DIFF
--- a/packages/neos-ui-ckeditor5-bindings/src/manifest.config.js
+++ b/packages/neos-ui-ckeditor5-bindings/src/manifest.config.js
@@ -12,6 +12,7 @@ import Essentials from '@ckeditor/ckeditor5-essentials/src/essentials';
 import Paragraph from '@ckeditor/ckeditor5-paragraph/src/paragraph';
 import Bold from '@ckeditor/ckeditor5-basic-styles/src/bold';
 import Italic from '@ckeditor/ckeditor5-basic-styles/src/italic';
+import ItalicWithEm from './plugins/italicWithEm';
 import Underline from '@ckeditor/ckeditor5-basic-styles/src/underline';
 import Strikethrough from '@ckeditor/ckeditor5-basic-styles/src/strikethrough';
 import Heading from '@ckeditor/ckeditor5-heading/src/heading';
@@ -128,6 +129,10 @@ export default ckEditorRegistry => {
         $get('formatting.h5'),
         $get('formatting.h6')
     )));
+
+    // Custom Plugin that automatically converts <i> to <em> for italics
+    // @fixes https://github.com/neos/neos-ui/issues/2906
+    config.set('italicWithEm', addPlugin(ItalicWithEm, $get('formatting.em')));
 
     //
     // @see https://docs.ckeditor.com/ckeditor5/latest/features/headings.html#configuring-heading-levels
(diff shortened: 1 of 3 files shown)


Sebobo:
Thx, works perfect.

But do we have to consider this a breaking change, as people might have relied on creating `i` tags?

Sebobo (approved):

Sebobo (changes requested):
IMO we should target this for 8.2 or 9.0

skurfuerst:
@Sebobo Isn't this configurable if I see the config right?

Sebobo:
@skurfuerst previously we had:
```
config.set('italic', addPlugin(Italic, $get('formatting.em')));
```

And now additionally (using the same option):
```
config.set('italicWithEm', addPlugin(ItalicWithEm, $get('formatting.em')));
```

So don't we even have to remove the first line?

mhsdesign:
I agree on the breakiness - and no this behaviour (is yet) not configurable - and i would not like it to be configurable -> as this would feel really hacky ^^

mhsdesign:
about the code: 

> previously we had:
> 
> ```
> config.set('italic', addPlugin(Italic, $get('formatting.em')));
> ```
> 
> And now additionally (using the same option):
> 
> ```
> config.set('italicWithEm', addPlugin(ItalicWithEm, $get('formatting.em')));
> ```
> 
> So don't we even have to remove the first line?

when looking at `ItalicWithEm` its clear why we still need `Italic` as its only an extension on top -> but it doesnt read well imo.

JamesAlias:
Proposal:

We can introduce an option that activates the conversion of `<i>` to `<em>`.

This way it's not breaking, the feature is available and it also reads better. win, win, win 😉

mhsdesign:
I asked in slack and the best solution would be


> [...] just document it and let people fix and adjust their CSS. I did add `i` to the CSS for `em` at some point,

> It's such a simple change. Documenting it should suffice.


https://neos-project.slack.com/archives/C050C8FEK/p1665386934728039?thread_ts=1665298045.139539&cid=C050C8FEK

so i think we dont need the config option ;)

Sebobo:
Then let's target 8.2 and add it to the release notes.

mhsdesign:
do we need a node migration do adjust everything to `<em>`? 

otherwise t would be mixed `em` and `i`... with a light direction to `em` (on every save)

Sebobo:
No wouldn't do that. We never migrate the users actual text content. As we also don't know whether they wrote that with CK or it's a custom property with an inspector editor or something

Sebobo (approved):

Sebobo (approved):

PR #3192, opened 2022-10-04
FEATURE: Use Yarn pnp mode

Sebobo:
**What I did**

Enable pnp mode for yarn to speed up initial install and build steps

**How I did it**

Adjusted yarn and webpack configurations

**How to verify it**

Run `make setup`, `make install`, `make build`

PR #3200, opened 2022-10-06
FEATURE: esbuild

### LINKED ISSUE #3113: Evaluate replacing webpack & babel with esbuild.
I will update this issue with further Infos later...

Lately I have been learning to love esbuild. And it's speed advantage is undeniable.

I tried to build the Neos Ui with esbuild but there are of course some stones in the way:

- Codesplitting
  - currently all the vendor packages are grouped in one separate bundle via webpack. I havent checked so far If tree shaking was applied. But what concerns esbuild there is yet no support for codesplitting with treeshaking for non esm builds. (imo it would make sense that treeshaking is not applied for the vendor build since we expose react fx as consumer api) 
- post css modules (find the right tooling) 
- clean up code that violates js/ts guidelines and is checked stricter by esbuild
- get ckeditor to work with esbuild

JamesAlias:
## TODO

- [x] make watch
- [x] remove added todos in code ...
- [x] revert adding fonts to git
- [x] fix esbuild build path
- [x] run typescript check (not done by esbuild)
- [x] drag and drop not working https://github.com/neos/neos-ui/pull/3200#issuecomment-1273199231
- [x] markdown help messages not working https://github.com/neos/neos-ui/pull/3200#issuecomment-1284011041
- [x] EmailAddressValidator not working https://github.com/neos/neos-ui/pull/3200#issuecomment-1284419262
- [x] !!! more node dependencies not working? (i fear only testing helps ... i dont have a way to check for that ... we can only encounter them at runtime ...) https://github.com/neos/neos-ui/pull/3200#issuecomment-1284454559
- [x] some list styles are still broken (missing normalize.css?)
- [x] .vanilla-css is missing (test with html nodetype -> codemirror)
- [x] include ck-editor css? -> not necessary!
- [x] Check extensibility (do custom plugins still work?)
- [ ] [*MOVED*](https://github.com/neos/neos-ui/issues/3262) Check `@neos-project/react-ui-components build-standalone-esm`
- [x] Remove old build setup stuff (webpack, babel, etc)
- [x] fix e2e tests -> via `keepNames: true`
- [x] fix jest tests -> https://github.com/neos/neos-ui/pull/3254
- [x] storybook
  - relies on own webpack config -> currently broken
  - dows anyone use storybook right now? https://github.com/neos/neos-ui/pull/3200#issuecomment-1271629882 (no one - so far)
- [x] finalize polyfill removal https://github.com/neos/neos-ui/issues/2855 - anything left? `./polyfills.js` was removed
- [x] side by side screenshot comparison -> currently the icons are a few pixel smaller? -> would break screenshot based testing if done by ourselfs or others. (maybe use postcss calc again instead of css `calc`)
- [x] What is our browser target ([esbuild target](https://esbuild.github.io/api/#target)) and do we still need postcss autoprefixer - remove root package.jsons browserslist
- [ ] [*MOVED*](https://github.com/neos/neos-ui/issues/3262) re add `main` fields and build subpackages command for npm subsplits
- [ ] [*MOVED*](https://github.com/neos/neos-ui/issues/3262) check if neos-ui-extensiblity still works (bable stuff) did we update babel?
- [ ] [*MOVED*](https://github.com/neos/neos-ui/issues/3262) fix scss files in lib esm build of react-ui-components
- [x] build size comparison (css gets bigger with `@include` vs `composes` ?)
  - OLD: `Host.css`: 135 KB (23,8 KB gzipped)
  - OLD: `Host.js`: 5,6 MB (1,5 MB gzipped) + `Vendor.js (with polyfills)`: 284 KB (90,1 KB gzipped)
  - NEW: `Host.css`: 118 KB (20,7 KB gzipped ❤️ )
  - NEW: `Host.js`: 3,7 MB (1,2 MB gzipped ❤️ )
- [x] build speed (M1 Max Macbook Pro)
  - OLD: 
Screenshot 2022-10-07 at 4 54 56 PM
  - NEW:
Screenshot 2022-10-07 at 4 51 56 PM

### optional todos:
- [ ] [*MOVED*](https://github.com/neos/neos-ui/issues/3263) replace style.css to style.module.pcss when used as module https://github.com/neos/neos-ui/pull/3200#issuecomment-1272485910
- [ ] [*MOVED*](https://github.com/neos/neos-ui/issues/3263) use pnp https://github.com/neos/neos-ui/pull/3192
- [x] split 'Neos.Neos.UI:Vendor' ? no rather not ...
- [x] dont use unsafe `eval` but `new Function`
- [x] find out why Vendor.js was used in iframe -> probably because it contained the `./polyfills.js`
- [x] move build files in Javascript and Styles folder :: no https://github.com/neos/neos-ui/pull/3200#issuecomment-1321088062
- [x] compare build files count - fonts etc with webpack
- [ ] [*MOVED*](https://github.com/neos/neos-ui/issues/3263) remove esbuild from yarn cache? -> as there would be linux-x64, darwin ... and many other builds for each system
- [x] what about make build-subpackes? -> we dont need it anymore, expect for distributing
- [x] todo do we still use tslib -> i think we can remove that
- [x] yarn dedupe -> doesnt do much in terms of build size -> probably only dev dependencies affected
- [x] find out why we have a dependency that uses nodes path?!
- [ ] [*MOVED*](https://github.com/neos/neos-ui/issues/3263) use postcss `composes` again instead of scss `@extends` https://github.com/neos/neos-ui/pull/3200#issuecomment-1281229469
- [ ] [*MOVED*](https://github.com/neos/neos-ui/issues/3263) Remove PostCSS css var generation from build step (on ice for this pr https://github.com/neos/neos-ui/pull/3200#issuecomment-1284529862)
- [x] rollback css var rename (and not using postcss var replace?) only for this pr so it becomes easier to review https://github.com/neos/neos-ui/pull/3200#issuecomment-1273261973
- [x] dont use buildsubpackages atrtefacts (dist) in the main bundle
- [ ] [*MOVED*](https://github.com/neos/neos-ui/issues/3262) -> cleanup `../dist/` in neos-ui-extensibility needed for https://github.com/neos/neos-ui/issues/3215
- [ ] [*MOVED*](https://github.com/neos/neos-ui/issues/3263) no need for `postcss-import` as esbuild does it itself
- [ ] [*MOVED*](https://github.com/neos/neos-ui/issues/3263) do we still need postcss-hexrgba -> esbuild minifies a lot -> so it may not even be preserved. What is the usecase?
- [x] what about the Host.js.LICENSE splitting in Resources/Public/JavaScript (now inside js ...)
- [x] what about the fonts in Resources/Public/Fonts -> shouldnt they be build artefacts in the "ui build" package?
- [ ] [*MOVED*](https://github.com/neos/neos-ui/issues/3263) analyze bundle with https://bundle-buddy.com/
- [ ] [*MOVED*](https://github.com/neos/neos-ui/issues/3263) replace `./Build/createVersionFile.sh` with esbuild define - but test would fails if not defined in jest
- [x] revert e2e changes https://github.com/neos/neos-ui/pull/3200#issuecomment-1321818375
- [ ] [*MOVED*](https://github.com/neos/neos-ui/issues/3263) extract webpack neos-ui-extensiblity adapter outside of workspaces (so we dont need babel installed anymore :P)
- [x] only use `keepNames: true` for e2e build `make build-testing`
- [x] remove bable.rc in every folder (probably because of jest and babel)
- [ ] [*MOVED*](https://github.com/neos/neos-ui/issues/3263) improve unit tests https://github.com/folke/esbuild-runner

WIP!
**What WE did**
Replace web pack with esbuild.

THANK YOU SO MUCH @NicoleNitschke and @mhsdesign ❤️ 

**How WE did it**
blood. sweat. tears.

**How to verify it**
tba


markusguenther:
Thanks for all the effort @JamesAlias 💙

The build_flow_app step can be solved with a rebase to get the latest changes in 8.2 :)

JamesAlias:
## TODO
- some list styles are still broken (missing normalize.css?)
- Check extensibility (do custom plugins still work?)

markusguenther:
Not sure, but the storybook is outdated, and I bet that no one uses that anymore. So maybe we can save some time and disable or remove the storybook. I did not use it for years now, can remember the sprint 2018 in Vienna... at some point it broke.

JamesAlias:
@markusguenther 
> Not sure, but the storybook is outdated, and I bet that no one uses that anymore. So maybe we can save some time and disable or remove the storybook. I did not use it for years now, can remember the sprint 2018 in Vienna... at some point it broke.

This is my hunch as well.
Does storybook run in CI? If not, we just disable it in makefile for the time beeing.

JamesAlias:
Unit Tests seem broken.. the CI says it does not find unit tests 🙈 👀 

This is probably due to missing `babelrc`

markusguenther:
> @markusguenther
> 
> > Not sure, but the storybook is outdated, and I bet that no one uses that anymore. So maybe we can save some time and disable or remove the storybook. I did not use it for years now, can remember the sprint 2018 in Vienna... at some point it broke.
> 
> This is my hunch as well. Does storybook run in CI? If not, we just disable it in makefile for the time beeing.

No, the storybook is not used in CI. So, it can be disabled.

markusguenther:
> Unit Tests seem broken.. the CI says it does not find unit tests 🙈 👀
> 
> This is probably due to missing `babelrc`

You're right, the old jest version need the whole babel stuff, and otherwise it cannot import modules.
Had the same issue with the babel upgrade PR, and that time the unit tests were the only missing piece.

packages/neos-ui/src/Containers/Modals/NodeCreationDialog/index.js
```diff
@@ -50,14 +58,35 @@ export default class NodeCreationDialog extends PureComponent {
...
     };
 
-    state = NodeCreationDialog.getDerivedStateFromProps(
-        this.props,
-        NodeCreationDialog.defaultState
-    );
+    constructor(props) {
+        super(props);
+
+        this.state = ((props, state) => {
```
mhsdesign: Todo: cleanup this code - we inlined getDerivedStateFromProps for testing - as this wouldnt work ...

mhsdesign: easiest fix would be extract it in a function outside of the class and use it here and also in  `static getDerivedStateFromProps`

packages/neos-ui/src/apiExposureMap.js
```diff
@@ -50,26 +50,26 @@ import ModelWriter from '@ckeditor/ckeditor5-engine/src/model/writer';
...
+import ViewDocumentFragment from '@ckeditor/ckeditor5-engine/src/view/documentfragment';
+import ViewDocumentSelection from '@ckeditor/ckeditor5-engine/src/view/documentselection';
+import ViewDOMConverter from '@ckeditor/ckeditor5-engine/src/view/domconverter';
+import ViewEditableElement from '@ckeditor/ckeditor5-engine/src/view/editableelement';
+import ViewElement from '@ckeditor/ckeditor5-engine/src/view/element';
+import ViewEmptyElement from '@ckeditor/ckeditor5-engine/src/view/emptyelement';
+import * as ViewFiller from '@ckeditor/ckeditor5-engine/src/view/filler';
+import ViewMatcher from '@ckeditor/ckeditor5-engine/src/view/matcher';
+import ViewNode from '@ckeditor/ckeditor5-engine/src/view/node';
+import * as ViewPlaceholder from '@ckeditor/ckeditor5-engine/src/view/placeholder';
```
mhsdesign, line 63: Has this `import * as` been tested - console.logeed ? ^^

mhsdesign: "looks" fine on first sight

image

packages/neos-ui/src/Containers/Modals/NodeCreationDialog/index.js
mhsdesign: done via https://github.com/neos/neos-ui/pull/3200/commits/09fc4c41fb56a8756056b34010b8ba24e43e152f

packages/neos-ui/src/Containers/Modals/NodeCreationDialog/index.js
```diff
@@ -50,51 +99,13 @@ export default class NodeCreationDialog extends PureComponent {
         secondaryInspectorComponent: null
     };
 
-    state = NodeCreationDialog.getDerivedStateFromProps(
-        this.props,
-        NodeCreationDialog.defaultState
-    );
```
mhsdesign, line 56: FOR REVIEWERS: This doesnt work anymore with esbuild ^^ - so we extracted the helpers to the module scope ;)

mhsdesign:
im thinking about hacking a php script together that renames all .css file to .module.pcss

via this regex: `import [a-zA-Z]+ from '.*\.s?css';` one can find all imports - lookup the file, rename it and adjust the import itself ... looks doable.

done in https://github.com/neos/neos-ui/pull/3217


```
Renamed "108" files
Replaced "123" imports
```

mhsdesign:
I also started trying out pnp - with say 50% success ^^ - building works, but not the build in the browser ;)

https://github.com/neos/neos-ui/pull/3218 

edit: update 90% success!!!

mhsdesign:
i think we should not be concerned about the https://www.npmjs.com/package/@neos-project/react-ui-components package, as there is currently a working version outside (8.2) - and if we cant provide a non breaking 8.3 -> we could release a 9.0 ;)

mhsdesign:
i noticed these issues sadly (https://github.com/neos/neos-ui/pull/3218#issuecomment-1272913932)

### console.error

something about `className` while clicking through the ui - i could not reproduce it yet ...

### ~Inspector buggy~ edit, this bug existed beforehand
see https://github.com/neos/neos-ui/issues/3219

<details>

the ui is sometimes buggy -> "looses" state and shows a blank inspector
This happens when using the selected element box to select a parent node, which is a shortcut

image

</details>

### drag and drop
drag and drop doesnt work anymore (node tree)

[Bildschirmaufzeichnung vom 10.10.2022, 10:12:25.webm](https://user-images.githubusercontent.com/85400359/194823489-fdd95b8d-6690-463d-a338-a04c7c46d3b2.webm)


## update 1
i tried updating react-dnd from v10.0.2 to v11 (v12 / v13) and v14
-> i could build this fine with updating react a bit https://github.com/neos/neos-ui/issues/3226
-> but it did not improve the situation

with v15 they removed the decorators, which we use and started only supporting hooks.
So we cannot go beyond v14 (latest is v16)

Transforming our dnd code to hooks is not something im looking forward too and also this would break our extensibility api

next i will try to check if the old examples from 10.0.2 will work with esbuild and try to debug this problem in a small enviroment
https://github.com/react-dnd/react-dnd/tree/v10.0.2/packages/documentation/examples-decorators

i also checked if react-dnd is included twice accidentally
and if the correct esm or source files are used instead of cjs or umd

also maybe try disabling minify and treeshaking to make sure no code is lost.

## update 2
i extracted the decorater react-dnd examples from the monorepo v10 and used esbuild to bundle it:
https://github.com/mhsdesign/react-dnd-example-decorators-esbuild

well it works :O 

minifying or treeshaking has no impact on the result - (neos ui still doesnt work, and the examples still work) 

## update 3
what i tried so far:

neos ui repo -> goal: fix that it works -> status: unsuccessful
- [x] use yarn 1
- [x] reinstall yarn 3
- [x] rebuild yarn cache, and node_modules links
- [ ] use pnpm
- [x] disable minifying and treeshaking

example react dnd -> goal: reproduce it that it doesnt work -> status: unsuccessful
- [x] use yarn 1 + esbuild
- [x] use yarn 3 (+node_modules) + esbuild 
- [x] use yarn 3 (+pnp) + esbuild
- [x] use pnpm + esbuild
- [x] disable minifying and treeshaking

## update 4
i extracted the react ui tree component and build it standalone.
this seems to work: (esbuild + yarn3)

https://github.com/mhsdesign/react-dnd-example-decorators-esbuild/tree/neos-tree

[Bildschirmaufzeichnung vom 19.10.2022, 11:15:49.webm](https://user-images.githubusercontent.com/85400359/196650530-287dd688-ee83-4003-9a4e-e014c3a78a16.webm)

for a second though, it had the same disfunctionality as in the ui (eg no blue insert lines etc) but then i reloaded and it was gone?!
... i even have the guess this could be OS related. Im on linux(ubuntu 22) and tried so far Firefox and Edge

maybe i need to extract more from the ui - or place some console logs on the events fired (on drag start and drop etc)

## update 5
yess omg are you kidding me

after rendering the full node tree in the ui alone (which still didnt work) i rendered only the demo tree from https://github.com/mhsdesign/react-dnd-example-decorators-esbuild/tree/neos-tree which suddenly stopped working when used in the ui build.

after commenting basically everything out and starting to compare the slightly different build stack i commented out everything i dont need including this line:
```
    define: {
        // put 'global' as empty object into glonal scope as some packages need it (nodeJS)
        // 'global': '{}'
    },
```
we added this so a node dependency doesnt crash - but it seems that react-dnd checks for `global` (which is not window) and  drops dead - ooooh boi

```
function getGlobalContext() {
  return typeof global !== 'undefined' ? global : window;
}
```

## Conclusion

we somehow have a dependency which uses global and because we where lazy we just defined in esbuild to insert for `global` an empty object `{}` like: we dont care.

i forgot about that line in the build script and dindt gave it much mlre attention - and considered it harmless

after playing debugging randomly and lately playing divide and conquer i found out that react-dnd uses either `global` (if defined) or `window` to store things - that broke it (as global was now everytime replaced with an empy object -> see esbuild 'define') :verschwitztes_lachen:

the hotfix is `global = window`

but ill try to narrow down the problem to fix it more accurately

i could also fix it by replacing `"global['PREVENT_CODEMIRROR_RENDER'] === true"` with `"false"` in the source code of the lib `react-codemirror2`

see https://github.com/neos/neos-ui/pull/3200/commits/ebc1e7078dd6bee57c8b4e61870d7fd97517cf78

but i looked at what webpack 4 did and they too took they lazy approach to just alias global to be window so we can do that too

mhsdesign:
i think we should rollback the `var(--neos-` change https://github.com/neos/neos-ui/issues/3201 it lets you loose the overview ^^

package.json
```diff
@@ -7,7 +7,6 @@
   "private": true,
   "resolutions": {
     "moment": "^2.20.1",
-    "d3-scale/d3-array": "2.3.3",
```
mhsdesign, line 10: why is this only hitting us now?


```
No matching export in "node_modules/d3-array/src/index.js" for import "InternMap"

    node_modules/d3-scale/src/ordinal.js:1:8:
      1 │ import {InternMap} from "d3-array";
        ╵         ~~~~~~~~~

```

how does webpack not fail? xD

mhsdesign:
about scss `@extends` vs postcss modules `composes`:

it seems the original people where very proud of it ^^
https://github.com/neos/neos-ui/wiki/Styling#css-modules-compose-feature


> ### CSS Modules' `compose` feature
> CSS Modules allow another way of code reuse: to extend other CSS classes, similar to SASS' `@extend` keyword, but without its drawbacks. I think that could be used to DRY some common traits between components, but should go through the same consideration as using variables, or else we risk to couple components too much together.


maybe with the proper css extensions: we can use another esbuild cssmodules plugin ... https://github.com/neos/neos-ui/pull/3217

edit about "another cssmodule esbuild plugin" we tried many things but currently no one supports this feature and other plugins wont even work properly in generall https://github.com/neos/neos-ui/pull/3200#issuecomment-1286912223

the currently used plugin "esbuild-style-plugin" is our best bet, but it doesnt work with `@composes` https://github.com/g45t345rt/esbuild-style-plugin/issues/16

mhsdesign:
## markdown issue - explaining why we need to upgrade `react-markdown` from 5 to 8

| before | after |
|------| -------|
|  image     |   image  |

we uses `react-markdown` to style help messages (in two places: the properties and node creation dialog)

in version 5 `react-markdown` uses the dependencies `vfile` (which uses `replace-ext`) and those mentioned libraries depend on node js path `require('path')`

with webpack4 (not 5 ^^) this is no propblem, as path will be autmatically mocked and made ready for the browser.

we did something similar for now - assuming that path is just imported accidentally and told esbuild to replace `require('path')` with `export default = {}`

this keeps the build working - but there are more issues lurking around such as `process.cwd()` not beeing defined and other ugly node stuff. 

so while we could use wierd plugins to archive the same forgiving webpack4 behaviour - or write plugins ourselves to mock the correct node api - i found out that `react-markdown` in v8 doesnt have this problem anymore since they changed/ or upgraded their dependencies.

we dont expose `react-markdown` in the consumer api - so this is not breaking.


### for reviewers: check that the markdown works as expected (for more complex usecases)

mhsdesign:
## `"Neos.Neos/Validation/EmailAddressValidator"` issue - used library is made for node-js

out of the box - the `EmailAddressValidator` will not work but complains that `Buffer` is not defined.

the with https://github.com/neos/neos-ui/pull/2988 introduced lib `isemail` replaced the previously used regex.

But the lib uses node js `Buffer.byteLength(content, 'utf8');` this needs to be polyfilled.
Also it uses `require("utils").types` which is not needed for browsers but needs to be removed for a sucessfull and clean bundle.

i patched these things now so they are working

```diff
diff --git a/lib/index.js b/lib/index.js
index 4f3d4cc33d13e68476dbe9ff0021081458af3291..087e62527d80e4e0f76773cc35a3c10b39b0a236 100644
--- a/lib/index.js
+++ b/lib/index.js
@@ -3,7 +3,8 @@
 // Load modules
 
 const Punycode = require('punycode');
-const Util = require('util');
+
+const Buffer = require('buffer/').Buffer
 
 // Declare internals
 
@@ -213,16 +214,9 @@ internals.isIterable = Array.isArray;
 if (typeof Symbol !== 'undefined') {
     internals.isIterable = (value) => Array.isArray(value) || (!!value && typeof value === 'object' && typeof value[Symbol.iterator] === 'function');
 }
-/* $lab:coverage:on$ */
-
 
-// Node 10 introduced isSet and isMap, which are useful for cross-context type
-// checking.
-// $lab:coverage:off$
-internals._isSet = (value) => value instanceof Set;
-internals._isMap = (value) => value instanceof Map;
-internals.isSet = Util.types && Util.types.isSet || internals._isSet;
-internals.isMap = Util.types && Util.types.isMap || internals._isMap;
+internals.isSet = (value) => value instanceof Set;
+internals.isMap = (value) => value instanceof Map;
 // $lab:coverage:on$
 
 

```

.yarnrc.yml
```yaml
packageExtensions:
  "isemail@3.2.0":
    dependencies:
      "buffer": "^5.0.0"
```

this issue was also mentioned here:
https://github.com/skeggse/isemail/pull/200#issuecomment-825537631

but it was recommended to use the lib https://www.npmjs.com/package/email-addresses

so maybe we should consider this library and check if it works with our email tests fine ;)

mhsdesign:
## How do we find out what else needs to be browserfied?

currently i found the above bugs only at runtime - which is really bad, we dont know what else to expect
- https://github.com/neos/neos-ui/pull/3200#issuecomment-1284011041
- https://github.com/neos/neos-ui/pull/3200#issuecomment-1284419262

somehow we need to check if the bundle has more hidden node references that come only to shine at runtime.

eg.
- process.env.NODE_ENV
- global
  - friends of global -> global.process
- Buffer

etc 

im against the previous webpack4 approach: just polyfill everything automatically - that is not well optimized ^^

packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/TableStyles.vanilla-css
```diff
@@ -51,36 +51,36 @@
 .ck.ck-icon *{
     fill:currentColor
 }
-.ck .ck-widget.ck-widget_with-selection-handler{
+.ck .ck-widget.ck-widget_with-selection-handle{
```
mhsdesign, line 54: this was very odd - but basically we fix the table handle (which can only be used in ck v16 to select a whole table)

since neos version 8.2 something changed about ckeditor (probably due to the new yarn 3 usage)

in webpack we disabled the icon: https://github.com/neos/neos-ui/pull/3212
but in this esbuild bundle we found the real cause and fixed it

image

packages/neos-ui-guest-frame/src/style.css
```diff
@@ -47,14 +47,10 @@
     border: 1px solid gray;
 }
 
-:global(.ck-widget__selection-handle .ck-icon) {
```
mhsdesign, line 50: here we undo it https://github.com/neos/neos-ui/pull/3200/files#r999875223

mhsdesign:
#3201 is on ice for this PR

so we do not have to concern in extensibility to
- [ ] Have old CSS Variables fall back to new `--neos-`-prefixed CSS Variables

packages/react-ui-components/src/SelectBox_Option_MultiLineWithThumbnail/style.css
```diff
@@ -34,9 +34,9 @@
     margin-right: .75em;
     display: inline-block;
     vertical-align: middle;
-    margin-left: -var(--spacing-Full);
-    margin-top: -var(--spacing-Half);
-    margin-bottom: -var(--spacing-Half);
+    margin-left: calc(var(--spacing-Full) * -1);
+    margin-top: calc(var(--spacing-Half) * -1);
+    margin-bottom: calc(var(--spacing-Half) * -1);
```
mhsdesign, line 39: we had css calc in place temporarely when we used real css variables - i forgot to undo it - but it will help in the future when we implement https://github.com/neos/neos-ui/issues/3201

Configuration/Settings.yaml
```diff
@@ -38,20 +38,20 @@ Neos:
       resources:
 
         javascript:
-          'Neos.Neos.UI:Vendor':
-            resource: '${"resource://" + Neos.Ui.StaticResources.compiledResourcePackage() + "/Public/JavaScript/Vendor.js"}'
-            position: 'start 1000'
+          # 'Neos.Neos.UI:Vendor':
+          #   resource: '${"resource://" + Neos.Ui.StaticResources.compiledResourcePackage() + "/Public/JavaScript/Vendor.js"}'
+          #   position: 'start 1000'
```
mhsdesign: todo cleanup

Makefile
```diff
@@ -93,23 +91,22 @@ build-react-ui-components-standalone:
...
 ## Watches (and polls) the source files on a file share.
 build-watch-poll:
-	NEOS_BUILD_ROOT=$(shell pwd) $(webpack) \
-		--progress --color --watch-poll --watch
+	echo "not implemented in esbuild, yet! PR Welcome!"
 
 # clean anything before building for production just to be sure
 ## Runs the production build.
 build-production:
 	make build-subpackages
```
mhsdesign: make build-subpackages

is out of order for now

esbuild.js
```diff
@@ -0,0 +1,104 @@
...
+    plugins: [
+        {
+            name: 'neos-ui-build',
+            setup: ({onResolve, onLoad, resolve}) => {
+                // exclude CKEditor styles
+                // the filter must match the import statement - and as one usually uses relative paths we cannot look for `@ckeditor` here
+                // the most correct way would be to look for all `/\.css/` - but this draws performance as we would intercept each css file
+                // so we use the the longest/most expressive regex that tries to match all ckeditor includes
+                // and luckely all includes look like `../theme/link.css` so we use the theme prefix `/theme\/[^.]+\.css$/`
+                // @todo i cant measure any time differences ... so using .css
```
mhsdesign, line 44: todo inline docs

mhsdesign:
# to be decided: is this breaking or non breaking?

pro
- we only change the internal build stack
- plugins keep working

con
- due to modernizing we kicked out the polyfills and will likely also stop using the autoprefixer -> this can be considered breaking - but i dont think that neos 8 runns sucessfully on internet explorer anymore.
- the npm packages react ui components and neos-ui extensibility will change so either they are a version ahead or we dont release them sonetime or we wait with this whole pr for 9

Sebobo:
@mhsdesign we don't bother about old browsers anymore. We removed compatibility with IE in most places quite a while ago.

What changes about those two packages?

mhsdesign:
## Css module plugins for esbuild

just for overview

|   npm package       |   supports postcss  | built in scss support |
| -------| --- | --- |
|   esbuild-style-plugin (currently in use)  | yes  (`composes` [bug](https://github.com/g45t345rt/esbuild-style-plugin/issues/16)) | yes | 
|   esbuild-plugin-postcss2  (unmaintained)  | yes (`composes` [bug](https://github.com/g45t345rt/esbuild-style-plugin/issues/16)) | no | 
|  esbuild-plugin-css-modules (unmaintained) | no (plain ast traversal) | no |
|  esbuild-plugin-simple-css-modules | no (plain ast traversal) | no |
| @deanc/esbuild-plugin-postcss | yes but no `composes` | no  | 
| esbuild-plugin-postcss  | yes but no `composes` | no  | 

or maybe try: https://github.com/igoradamenko/esbuild-plugin-webpack-bridge (webpack-bridge doesnt work with the webpack style-loader and css-loader plugins - as it doesnt support yet loader urls `!loader`...)

markusguenther:
Need to update https://github.com/neos/eslint-config-neos to make the linter run again...
The e2e tests fail because the backend is white... but need to test it locally and my macOS Ventura doesn't want to run testcafe :/

markusguenther:
As this is the most significant change code wise... I will try to work on this finally this week.
Hopefully we can make it for 8.2 🤞

markusguenther:
Code style stuff is also fixed and uses new dependencies now.

To-dos are still jest with babel 7 and check what's wrong with the e2e tests.

markusguenther:
Conflicts are also resolved and we have the latest state of 8.2 included.

Sebobo:
Then e2e tests fail as some `ReactSelectors` of test cafe don't work properly anymore. Especially in drop-downs and the tree. I adjusted to selectors but the one for the link editor dropdown still is a problem I haven't solved yet.

@markusguenther it seems the HostOnlyStyles are now in `Public` instead of `Public/Styles` and you added them to git.

markusguenther:
Thanks @Sebobo ❤️

Sebobo:
Damn, these tests drive me crazy and I give up for now.

I fixed several, but some stuff just feels random. Most problems seem to be related to the usage of portals, but this shouldn't be a problem as Dmitri fixed it years ago together with the TestCafe people and neither the old version nor the new make a difference here. Also some other portal stuff works fine.

What could be an issue is that the portals are not rendered inside the app, but the change here was also done 2 years ago and back then the tests didn't stop working.

 Other tests fail on my machine, because of my screen resolution, as they check for dropdown visibility depending on the scroll position, but if I change the fixture to try to fix this I break some other test with is then again unhappy with some position of some dropdown 🤦‍♂️
 
 I say: let's test some 3rd party plugins and then merge the PR in the next days. If a few tests keep behaving stupid I would even disable them for now and make some manual test checklist before a release.

Sebobo:
@mhsdesign as I said in a previous comment, the built Assets are now directly in `Resources/Public` instead of `Resources/Public/JavaScript|Styles` do we want to keep that or adjust it again? Sadly I don't know exactly where.

mhsdesign:
> @mhsdesign as I said in a previous comment, the built Assets are now directly in `Resources/Public` instead of `Resources/Public/JavaScript|Styles` do we want to keep that or adjust it again? Sadly I don't know exactly where.

We could simply use a bash script to move the files in place after build, as this is not supported yet: https://github.com/evanw/esbuild/issues/553#issuecomment-908504533

but by doing that it would mess with relative import parts (like fonts and etc)

i would just keep the output all in one folder ... any objections?

packages/react-ui-components/src/SelectBox_Option_MultiLineWithThumbnail/style.css
mhsdesign: it might be cause of some pixel shifting as noted by @Sebobo - ill try to remove it again for now ...

mhsdesign:
partyyyyyyy 🎉 lets gooo @Sebobo @markusguenther e2e ✅

markusguenther:
So just the unit tests are missing. Thanks for investigating the issue with the e2e tests.

mhsdesign:
and as discussed with @Sebobo we want to revert his changes and keep working with the react magic selector

mhsdesign:
hot! 
image

would be even hotter with pnp ;)  https://github.com/neos/neos-ui/pull/3192

packages/react-ui-components/package.json
```diff
@@ -9,7 +9,7 @@
   "main": "./src/index",
   "scripts": {
     "build-standalone-esm": "rm -Rf .postcssrc lib-esm && tsc --build tsconfig.esm.json && tsc --build tsconfig.esmtypes.json && rsync --prune-empty-dirs  -r --include='*/' --include='*.css' --include='*.svg' --exclude='*' src/ lib-esm -v && node scripts/generate-standalone-postcssrc.js",
```
mhsdesign: hmm we might need this now due to  https://github.com/neos/neos-ui/pull/3200#issuecomment-1281229469

--include='*.scss'

mhsdesign:
turns out we still need babel for eslint 🙄🤦

markusguenther:
> turns out we still need babel for eslint 🙄🤦

Therefore I released new versions of https://github.com/neos/eslint-config-neos ;)

markusguenther:
But guess you can also run `eslint` without `babel`... question is how much we need to change ;)

mhsdesign:
I dare you to revert my babel stuff - uups already done xD

I think we can just use the Typenscript eslint parser for all files and forget about babel (see the overrides config in eslint)

markusguenther:
Tests are now green... if you think that you can make it without much effort, go ahead. ;) @mhsdesign

markusguenther:
As we release the `@neos-project/react-ui-components` always as extra package ... we need to check what is wrong. But the most people I know seldom update the package version in the UI extension.

markusguenther:
Wow, what a ride. I am so thankful for all the time that has been spent on this PR and the PRs before that were abandoned. Thanks to @mhsdesign @JamesAlias @Sebobo and @grebaldi for all the discussions and the endless time 🥇🥇🥇🥇

markusguenther:
@mhsdesign already created all the issues for the follow-ups 💙

JamesAlias:
OMG you guys are crazy awesome! ❤️😍

markusguenther:
Thanks for all the work... we are brave enough and will resolve all upcoming issues with the extensibility.
But the most people don't use the latest packages of the extensibility.

mhsdesign:
wow this is great!!! what a run. Thank you so much again @JamesAlias for the initial effort ;) And great that we pulled it of.

PR #3203, opened 2022-10-06
TASK: Adjust neos-ui dependency

markusguenther:


Sebobo (approved):

markusguenther:
Ok seems to work... but we need to address the lowest maintained branch then and up merge the stuff

Issue #3013, opened 2022-01-19
Optimize width of DimensionSelector


cvette:
### Description

Right now there is no min-width for the dimension selector. If you have a short dimension label selected and open the drop-down, some of the longer labels are restricted way too much.

### Steps to Reproduce

1. Configure some dimension presets with both long and short labels
2. Select a preset in the Neos UI with a short label
3. Open Dropdown

#### Expected behavior

There should be a min width and probable a bit higher max width.

#### Actual behavior
![2022-01-19_11-37](https://user-images.githubusercontent.com/6884391/150114952-858e658b-2e65-43d0-9132-679152f9441c.png)

### Affected Versions

Neos: all versions with the new UI ?

UI: all versions of the new UI ?

NicoleNitschke:
I'll take this issue today. :)

skurfuerst:
merged :)

PR #3206, opened 2022-10-06
BUGFIX: add larger min-width to dimension switcher dropdown to prevent long options from clipping to early

NicoleNitschke:
**What I did**
* fixed issue #3013 

**How I did it**
* added a new min-width for the dimension switcher dropdown
* increased the max-width of the closed dropdown button

**How to verify it**
1. Configure some dimension presets with both long and short labels
2. Select a preset in the Neos UI with a short label
3. Open Dropdown

### DIFF
--- a/packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/index.js
+++ b/packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/index.js
@@ -97,6 +97,7 @@ class DimensionSelector extends PureComponent {
                     searchBoxLeftToTypeLabel={i18nRegistry.translate('Neos.Neos:Main:searchBoxLeftToType')}
                     threshold={0}
                     ListPreviewElement={DimensionSelectorOption}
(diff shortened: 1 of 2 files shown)


markusguenther:
Thanks for the fix @NicoleNitschke,

as it is a bug fix we should target the lowest maintained branch here.
In this case, it is 7.3. Do you want to give it a try and change that, or should one of the team take care?

The failing pipeline is maybe solved when you rebase the change to the latest version of 8.2 or with the new target 7.3 :)

NicoleNitschke:
> Thanks for the fix @NicoleNitschke,
> 
> as it is a bug fix we should target the lowest maintained branch here. In this case, it is 7.3. Do you want to give it a try and change that, or should one of the team take care?
> 
> The failing pipeline is maybe solved when you rebase the change to the latest version of 8.2 or with the new target 7.3 :)

@markusguenther I try to change it by myself and open another PR - thanks for your feedback. :)

NicoleNitschke:
@markusguenther the new PR is https://github.com/neos/neos-ui/pull/3210 :)

PR #3207, opened 2022-10-06
TASK: Remove JS decorators

JamesAlias:
**What I did**
Remove JS decorators because the rfc was rejected.

**How I did it**
Use function calls instead.

**How to verify it**

... this is WIP - there are many more occurrences

### DIFF
--- a/packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/LinkButton.js
+++ b/packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/LinkButton.js
@@ -10,15 +10,19 @@ import {selectors, actions} from '@neos-project/neos-ui-redux-store';
 
 import style from './LinkButton.css';
 
-@connect($transform({
-    isOpen: selectors.UI.ContentCanvas.isLinkEditorOpen
-}), {
-    toggle: actions.UI.ContentCanvas.toggleLinkEditor
-})
-@neos(globalRegistry => ({
+const connector = connect(
+    $transform({
+        isOpen: selectors.UI.ContentCanvas.isLinkEditorOpen
+    }), {
+        toggle: actions.UI.ContentCanvas.toggleLinkEditor
+    }
+);
+
+const applyNeosContext = neos(globalRegistry => ({
     i18nRegistry: globalRegistry.get('i18n')
-}))
-export default class LinkButton extends PureComponent {
+}));
+
+class LinkButton extends PureComponent {
     static propTypes = {
         formattingUnderCursor: PropTypes.objectOf(PropTypes.oneOfType([
             PropTypes.number,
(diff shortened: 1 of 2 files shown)


mhsdesign:
As discussed - this is a right step into the right direction - but even better would be functional components with hooks (and ts) ... which is quite impossible as this would equal an ui rewrite 2.0 ^^

since esbuild seems to work for now with these decorators we may leave them as is - it would only make the code a little better.

a little more profit would be in cases where typescript is used for components as the decorators cannot be typed but the hoc function syntax works with typescript like:

<details>

<summary>@neos-project/neos-ui-decorators.d.ts</summary>

this is not the same as in `@neos-project/neos-ui-decorators` this is written from hand by myself and works differently - more like `connect`

```ts
declare module "@neos-project/neos-ui-decorators" {
    import {
        InferableComponentEnhancerWithProps,
        ConnectedProps,
    } from "react-redux";
    import { GlobalRegistry } from "@neos-project/neos-ts-interfaces";

    export interface NeosContextInterface {
        globalRegistry: GlobalRegistry;
        configuration: {};
        routes: {};
    }

    /**
     * Infers the type of props that a neosifier will inject into a component.
     * we reuse this behavior from {@link ConnectedProps}
     */
    export type NeosifiedProps<TNeosifier> = ConnectedProps<TNeosifier>;

    export const NeosContext: React.Context<NeosContextInterface | null>;

    type MapRegistryToPropsParam<TStateProps> = (
        globalRegistry: GlobalRegistry
    ) => TStateProps;

    interface Neos {
        <TStateProps = {}, TOwnProps = {}>(
            mapRegistryToProps: MapRegistryToPropsParam<TStateProps>
        ): InferableComponentEnhancerWithProps<
            TStateProps & { neos: NeosContextInterface },
            TOwnProps
        >;
    }

    export const neos: Neos;
}
```

</details>



```ts
import { neos, NeosifiedProps } from "@neos-project/neos-ui-decorators"; // NeosifiedProps is basically ConnectedProps
import { connect, ConnectedProps } from "react-redux";

const neosifier = neos((globalRegistry) => ({
    secondaryEditorsRegistry: globalRegistry
        .get("inspector")
        .get("secondaryEditors")
}));

const connector = connect(
    (state) => ({
        node: selectors.CR.Nodes.focusedSelector(state),
    })
);


type NeosProps = NeosifiedProps<typeof neosifier>;

type StateProps = ConnectedProps<typeof connector>;

type OwnProps = {}

type Props = NeosProps & StateProps & OwnProps;

class YourComponent extends React.PureComponent<Props>
{
    public render()
    {
    }
}

export default neosifier(connector(YourComponent));
```

mhsdesign:
this should become a stale issue instead of pr for now ...

PR #3210, opened 2022-10-06
BUGFIX: add larger min-width to dimension switcher dropdown to prevent long options from clipping to early

NicoleNitschke:
**What I did**
* fixed issue #3013 

**How I did it**
* added a new min-width for the dimension switcher dropdown
* increased the max-width of the closed dropdown button

**How to verify it**
1. Configure some dimension presets with both long and short labels
2. Select a preset in the Neos UI with a short label
3. Open Dropdown


skurfuerst:
makes total sense ❤️

PR #3211, opened 2022-10-06
BUGFIX: label of checkbox "auto-publishing" is now clickable and looks nice

### LINKED ISSUE #3140: BUG: auto-publishing hitbox is to small
### Description

If you click here to enable it:
image

the dialog will close, but the checkbox auto-publishing will not be enabled.

you need to actually click directly on the checkbox:

image

### Affected Versions
UI version: v7.0.3

NicoleNitschke:
**What I did**
* fix issue #3140 
* made label of checkbox "auto-publishing" clickable and fixed its styling

**How I did it**
* used a pure css solution with absolute positioned label because the react-ui-component checkbox inside the react-ui-component select is buggy

**How to verify it**
* open the publish dropdown and see how it looks now
* click the "auto publish label" - the checkbox will be toggled and the dropdown closed

### Before:
grafik

### After:
grafik

### DIFF
--- a/packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/index.js
+++ b/packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/index.js
@@ -214,6 +214,12 @@ export default class PublishDropDown extends PureComponent {
                             </a>
                         </li>)}
                         <li className={autoPublishWrapperClassNames}>
+                            {
+                                /**
+                                 PLEASE NOTE: this additional styleClass is a fix, because react checkboxes inside a react select component are buggy,
+                                 for further information see https://github.com/neos/neos-ui/pull/3211
+                                 */
+                            }
                             <Label htmlFor="neos-PublishDropDown-AutoPublish" className={style.dropdownOptionCheckbox}>
                                 <
 // @ts-ignore
--- a/packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/index.js
+++ b/packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/index.js
@@ -214,15 +214,18 @@ export default class PublishDropDown extends PureComponent {
                             </a>
                         </li>)}
                         <li className={autoPublishWrapperClassNames}>
-                            <Label htmlFor="neos-PublishDropDown-AutoPublish">
+                            <Label htmlFor="neos-PublishDropDown-AutoPublish" className={style.dropdownOptionCheckbox}>
                                 <
 // @ts-ignore
                                 CheckBox
                                     id="neos-PublishDropDown-AutoPublish"
                                     onChange={toggleAutoPublishing}
                                     isChecked={isAutoPublishingEnabled}
+                                    className={style.dropdownOptionCheckbox__input}
                                     />
-                                <I18n id="Neos.Neos:Main:autoPublish" fallback="Auto-Publish"/>
+                                <span className={style.dropdownOptionCheckbox__label}>
+                                    <I18n id="Neos.Neos:Main:autoPublish" fallback="Auto-Publish"/>
+                                </span>
                             </Label>
                         </li>
                         <li className={style.dropDown__item}>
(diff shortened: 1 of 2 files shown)


crydotsnake (approved):
Hello @NicoleNitschke 

Tested it, and its working fine!

Thanks a lot for the fix!

packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/index.js
```diff
@@ -214,15 +214,18 @@ export default class PublishDropDown extends PureComponent {
                             </a>
                         </li>)}
                         <li className={autoPublishWrapperClassNames}>
-                            <Label htmlFor="neos-PublishDropDown-AutoPublish">
+                            <Label htmlFor="neos-PublishDropDown-AutoPublish" className={style.dropdownOptionCheckbox}>
```
skurfuerst, line 223: Great change! I'd love to have a comment here `{/* A JSX comment */}` explainig why this is needed; so that we understand it when looking at it in the future (also with link to the issue) : )

packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/style.css
```diff
@@ -46,6 +46,31 @@
     margin-right: 5px;
 }
 
+.dropdownOptionCheckbox {
```
skurfuerst, line 59: I'd love to have a comment here {/* A JSX comment */} explainig why this is needed; so that we understand it when looking at it in the future (also with link to the issue) : )

-> answering the question why a *new* checkbox is needed here :)

skurfuerst (changes requested):
really nice change, just a nitpick :)

packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/index.js
NicoleNitschke: Hi @skurfuerst I added the comment, thx for this feedback! :)

packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/style.css
NicoleNitschke: Hi @skurfuerst I added the comment, thx for this feedback! :)

skurfuerst:
@NicoleNitschke sorry, circle CI failed noch ;) (nitpicks)

Sebobo:
Does this PR interfere with https://github.com/neos/neos-ui/pull/3194 somehow? We merged that some days ago and it also fixes the checkbox alignment.

NicoleNitschke:
> @NicoleNitschke sorry, circle CI failed noch ;) (nitpicks)

@skurfuerst  CI works now :)

markusguenther:
> Does this PR interfere with #3194 somehow? We merged that some days ago and it also fixes the checkbox alignment.

The alignment was fixed in Paula's PR, but the click handling does not work for the label. That is fixed here as well.

markusguenther:
One thing that triggers me is that the dropdown closes after clicking the checkbox or label, but checked Neos 4.3 with the old UI, and it was always the behavior. So for me, it looks fine.

NicoleNitschke:
> One thing that triggers me is that the dropdown closes after clicking the checkbox or label, but checked Neos 4.3 with the old UI, and it was always the behavior. So for me, it looks fine.

@markusguenther I thought the same as you while fixing this issue, but unfortunately there is no simple solution to fix this. I see 2 ways:
1. Let the Dropdown open with very hacky JS. -> not a good idea
2. Use completely custom components instead of React-Ui-Components Select and Checkbox -> really big change for this feature

Furthermore, this is the only checkbox in the dropdown. For every other dropdown option, the select closes as well. So it could also be confusing for the user, why the dropdown stays open only in this use case. It's a tricky UX case. 
Just wanted to let you know I had thought about it a long time, hope this makes it a little bit less awful for you. :D

markusguenther:
Yeah, think for this change it is ok and if we want to change the behavior, we should open another PR for that. But it seems that it is there like always ;)

markusguenther (approved):

PR #3212, opened 2022-10-06
BUGFIX: hide big ckeditor icon when table is inserted

NicoleNitschke:
**What I did**
* CkEditor 5 inserts an icon with 100% width, when the user inserts a table
* this error appears only in version 8.2
* I hide this icon, because it's useless

**How I did it**
* CSS only

**How to verify it**
* simply insert a table with ckeditor and see how no icon appears 🤡

### Before
grafik

### After
grafik

### DIFF
--- a/packages/neos-ui-guest-frame/src/style.css
+++ b/packages/neos-ui-guest-frame/src/style.css
@@ -47,6 +47,10 @@
     border: 1px solid gray;
 }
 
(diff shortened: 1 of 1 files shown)


Sebobo (approved):

markusguenther (approved):
Works find, thanks for the fix @NicoleNitschke 💙

PR #3220, opened 2022-10-10
Bugfix: Add scroll to errors

### LINKED ISSUE #2572: Make the content in Error message scrollable
### Description

It can happen that the flash message has pretty long content. Like in the configuration module.
It would be nice if you can scroll the content of the flash message

### Steps to Reproduce

1. Log in to backend
2. Open configuration module
3. Hope that you have some more errors in the nodetypes or so

#### Expected behavior

Be able to read all the content of the flash message

#### Actual behavior

Bildschirmfoto 2019-10-21 um 13 54 06

### Affected Versions

Neos: 5.0

UI: master

pKallert:
similar neos/neos-development-collection#3917 

**What I did**
Error Messages in the UI were not scrollable, since they have a fixed position. Now they are.  

**How I did it**
I added a max-height to the UI error messages and made them scroll if there is too much content. 

**How to verify it**
Since I did not have an error message, I forced an error message on saving page changes. The long content can also be created with the devtools. 

New behavior for one long message: 

big_message

New behavior for multiple messages: 
multiple_messages

New behavior for one single small message:
Bildschirmfoto 2022-10-11 um 09 26 56

### DIFF
--- a/packages/neos-ui/src/Containers/FlashMessages/FlashMessage/style.css
+++ b/packages/neos-ui/src/Containers/FlashMessages/FlashMessage/style.css
@@ -8,6 +8,10 @@
     overflow: hidden;
     box-sizing: border-box;
     box-shadow: 0 0 14px rgba(#000, .15);
+
+    &:last-child {
+        margin-bottom: 0px;
+    }
 }
 
 .flashMessage--success {
--- a/packages/neos-ui/src/Containers/FlashMessages/style.css
+++ b/packages/neos-ui/src/Containers/FlashMessages/style.css
@@ -3,9 +3,11 @@
     z-index: var(--zIndex-FlashMessageContainer);
     top: 0;
     left: 50%;
-    width: 512px;
-    padding-top: 8px;
+    width: 516px;
+    margin-top: 8px;
     transform: translate(-50%, 0);
+    max-height: calc(100% - 16px);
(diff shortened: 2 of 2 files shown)


mhsdesign:
Hi ;) correct me when im wrong, but this is not the flash message rendered in the backend module "configuration" as shown in the linked issue https://github.com/neos/neos-development-collection/issues/3917

so this is only for the ui/content module ^^ which is fine.

also a screenshot would be really nice and helping ;)

pKallert:
I added a few screenshots of what it looks like on my local machine. 

The original Bug mentions the Flash messages and the configuration module, which is kind of confusing. 

There seems to be another Message-component in Neos.Neos: https://github.com/neos/neos-development-collection/tree/8.2/Neos.Neos/Resources/Public/JavaScript/Components/Notification  
and several custom message container in other packages. 

Since this is the only occurence of messages I could find in this project, I solved it here :)

mhsdesign (approved):
Thank you for the screenshots and videos and also your investigation into Neos.Neos ;)

I think we should move https://github.com/neos/neos-ui/issues/2572 to Neos.Neos.

This looks fine by reading!

Sebobo (approved):
Looks fine. Hope most users will never experience this 🤪

crydotsnake (approved):

PR #3221, opened 2022-10-11
Bugfix: Hover dimensionSwitcher and position of userDropdown

### LINKED ISSUE #3216: BUG: Upper toolbar - minor css ux/ui differences
when testing the esbuild build you have to look very carefully and become almost paranoid about the smallest anomalies - so I just found out these two tiny things (which also appear in the webpack build)

#### 1

image

image
(English US is not blue -> but should be)

#### 2

image

image
(The user when selected, moves the other elements to the left -> but the dialog should be position: absolute and should not affect other items)

UI: ... - 7.1 - 8.1 - ...

pKallert:
closes #3216 

**What I did**
There are two bugs in the upper right toolbar. Those should be fixed now. 

**How I did it**
I added a hover-state to the dimension Menu and set the User dropdown to absolute.

**How to verify it**
Opening the Usermenu should not move the dimensions menu: 
dropdown_desktop

Opening the usermenu on mobile: 
dropdown_mobile

Hover on the dimension Menu should color the text blue: 
Bildschirmfoto 2022-10-11 um 11 10 17

### DIFF
--- a/packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/style.css
+++ b/packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/style.css
@@ -4,6 +4,10 @@
 }
 .dropDown__header {
     background-color: transparent;
+
+    &:hover {
+        color: var(--colors-PrimaryBlue);
+    }
 }
 .dropDown__contents {
     width: fit-content;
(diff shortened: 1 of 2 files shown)


Sebobo (approved):
Works perfectly, thx!

crydotsnake (approved):

PR #3225, opened 2022-10-12
BUGFIX: Clicking on top menu SVG now triggers menu close

### LINKED ISSUE #3222: Bug: Usermenu and Dimensionmenu open at the same time
### Description
When the dimensions Menu is open, clicking on the usermenu down-button leads to a state where both usermenu and dimensions menu are open. 

### Steps to Reproduce

1. Open dimensions Menu / language menu
2. Open user menu by clicking on the "down" button, not on the username 

Other way around: 
1.  Click on the "down" button on the user menu to open the usermenu
2. Click on the "down" button to open the language menu

#### Expected behavior
I would expect the language menu to be closed and the user menu to be open

#### Actual behavior
The language menu is open and the usermenu is open and overlaying the language menu.
 
dropdown_error

pKallert:
closes #3222 

**What I did**
When clicking on any of the svgs in the right topmenu bar, other dropdowns were not closed. For example, after opening the  Language menu and clicking on the "down" or "user" icons to open the usermenu, the language menu would stay open. 
This should not happen anymore. 

**How I did it**
As far as I understood, SVGs have a pointer-event called: visiblePainted.  Apparently, this stops the click element from traveling. Setting the pointer-event to "none" allows the event to travel to the parents and children. I tried out different selectors, but this was the only one that worked. 
Since in the dimensions menu the svg is wrapped inside a span, I had to use a very broad selector to catch all SVGs inside the dropdown header. 
Also, I was not sure if this is the right place for adding the new CSS-declaration. If this should be placed anywhere else, suggestions are welcome 😸 

Link to documentation: https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events

**How to verify it**
Clicking on any SVG in the right topmenu now closes all other open menus. 
dropdown_right_mouse

### DIFF
--- a/packages/react-ui-components/src/DropDown/style.css
+++ b/packages/react-ui-components/src/DropDown/style.css
@@ -25,6 +25,10 @@
     &:focus{
         outline: 1px solid var(--colors-PrimaryBlue);
     }
+
+    svg{
+        pointer-events: none;
+    }
 }
 .dropDown__btn--withChevron {
     padding-right: var(--spacing-GoldenUnit);


mhsdesign:
Hi thanks ;)

Do i see that correctly that the mouse courser now doesnt change anymore? ^^

Sebobo:
Yes this seems to got lost. 

Love the change, thx. At some point we also have to fix the jumping of the other items when the user menu is opened.

mhsdesign:
the jumping has been adjusted already by @pKallert in https://github.com/neos/neos-ui/pull/3221 ;)

pKallert:
@mhsdesign I tried it out locally and the mouse cursor is still the normal cursor:pointer. There is apparently a setting in the video program that I missed for showing the right cursor in the GIF.  
I updated the comment with a GIF with the right cursor.

mhsdesign (approved):
thanks for looking into that ;)

markusguenther:
Thanks for the fix and taking care @pKallert 💙

PR #3228, opened 2022-10-14
TASK: translate rte toolbar style-select options

PRGfx:
**What I did**
The StyleSelect dropdown in the RTE toolbar now tries to translate the option labels.

**How I did it**
Rather than configuring a translated string in the toolbar manifest, the dopdown tries to translate the label itself to be more in line with the other configured toolbar items.

**How to verify it**
Enable formatting options on a node-type and try overwriting the translation
```
ui:
  inline:
    editorOptions:
      formatting:
        p: true
        h1: true
        h2: true
        h3: true
        h4: true
        h5: true
        h6: true
        pre: true
```

Regarding the translation workflow I found https://www.neos.io/community/participate/translating-neos.html but I'm not completely sure what to make of it.

### DIFF
--- a/Resources/Private/Translations/en/Main.xlf
+++ b/Resources/Private/Translations/en/Main.xlf
@@ -20,6 +20,30 @@
             <trans-unit id="imageCropper__aspect-ratio-placeholder" xml:space="preserve">
                 <source>Choose an Aspect Ratio</source>
             </trans-unit>
+            <trans-unit id="ckeditor__toolbar__style__p" xml:space="preserve">
+                <source>Paragraph</source>
+            </trans-unit>
+            <trans-unit id="ckeditor__toolbar__style__h1" xml:space="preserve">
+                <source>Headline 1</source>
+            </trans-unit>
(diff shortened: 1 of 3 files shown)


mhsdesign:
thanks ;)

"looks" already fine by reading... what is left over?

... well i saw the missing translations, but dont we manage them via weblate? (i have no clue ^^)

PRGfx:
Okay, seeing that this seems to be the workflow with translations, I think that should be it then.

mhsdesign (approved):
Nice thanks for adjusting that!

(by reading... ^^)

markusguenther:
As it is not a feature, I, personally, would like to see that as well for 7.3. What do @Sebobo and @mhsdesign think about that?

Sebobo:
Fine with me

mhsdesign:
Will this sudden label translation be unexpected for editors?

but fine with me lets target 7.3

mhsdesign:
teamwork :P

markusguenther:
Yeah, rebased the changes and you changed the base branch :D

markusguenther (approved):
Thanks for your contribution :)

PR #3229, opened 2022-10-14
FEATURE: Regenerate UriPathSegment in the backend

### LINKED ISSUE #3127: Regenerating uriPathSegment replaces "&" with "and"

PRGfx:
**What I did**
To achieve the same behavior in UriPathSegment generation between the UI and the backend (i.e. updating a path segment should use the same logic as when creating a new page), the `NodeUriPathSegmentGenerator` is used in a backend service endpoint.

This allows adjusting the slugification in a consistent way through a single decoration in the backend.

**How I did it**
A new endpoint is added to the BackendServiceController and exposed through the endpoint definitions in the UI.
The UriPathSegment editor then calls that backend to ask for a new slug for the current node title rather than using a javascript library to calculate it in the frontend.

**How to verify it**
This change is introduced against the problem described in #3127, so most notably it will now no longer regenerate `&` to `and`, but it will work better with internationalized transliteration as well, as this was not handled in the previously used JS module.

Maybe we could get rid of the previously used JS module, however I'm unsure about the implications.

This PR originates from #3167.
Resolves #3127


PRGfx:
@Sebobo you mentioned having adjusted the uri-path-segment generation before. Does this approach make sense to you/do you think customizing the behavior by decorating the `Neos\Neos\Utility\NodeUriPathSegmentGenerator::generateUriPathSegment` would work for such cases or is there another method that should be used?

Classes/Controller/BackendServiceController.php
```diff
@@ -545,4 +552,18 @@ function ($envelope) {
...
+     *
+     * @param NodeInterface $contextNode
+     * @param string $text
+     * @return string
+     * @throws \Neos\Neos\Exception
+     */
+    public function transliteratePathSegmentAction(NodeInterface $contextNode, string $text): string
+    {
+        $slug = $this->nodeUriPathSegmentGenerator->generateUriPathSegment($contextNode, $text);
+        return json_encode($slug);
```
Sebobo: Use the JSON view feature like this:

```suggestion
        $this->view->assign('value', $slug);
```

Classes/Controller/BackendServiceController.php
```diff
@@ -545,4 +552,18 @@ function ($envelope) {
...
+
+    /**
+     * Generates a new uri path segment for the given node and title
+     *
+     * @param NodeInterface $contextNode
+     * @param string $text
+     * @return string
+     * @throws \Neos\Neos\Exception
+     */
+    public function transliteratePathSegmentAction(NodeInterface $contextNode, string $text): string
```
Sebobo: Should be void as we use the JSON view, you can remove the param and returns parts in the doc comment.

```suggestion
    public function transliteratePathSegmentAction(NodeInterface $contextNode, string $text): void
```

Configuration/Routes.Service.yaml
```diff
@@ -84,3 +84,11 @@
     '@controller': 'BackendService'
     '@action': 'getAdditionalNodeMetadata'
   httpMethods: ['POST']
+
+-
+  name: 'Generate UriPathSegment'
+  uriPattern: 'uri-path-segment'
```
Sebobo: Describe the action, maybe like this(?):

```suggestion
  uriPattern: 'create-uri-path-segment'
```

packages/neos-ui-editors/src/Editors/UriPathSegment/index.js
```diff
@@ -1,8 +1,10 @@
 import React, {PureComponent} from 'react';
+import {connect} from 'react-redux';
 import PropTypes from 'prop-types';
 import unescape from 'lodash.unescape';
-import slugify from '@sindresorhus/slugify';
```
Sebobo, line 4: If we don't need the extension then anymore, we can get rid of it 🥳

Sebobo (changes requested):
Great change. Left some notes

packages/neos-ui-backend-connector/src/Endpoints/index.ts
```diff
@@ -638,6 +639,24 @@ export default (routes: Routes) => {
...
+            data.set('__csrfToken', csrfToken);
+            data.set('contextNode', nodeContextPath);
+            data.set('text', text);
+
+            return ({
+                url: routes.ui.service.transliteratePathSegment,
+                method: 'POST',
+                credentials: 'include',
+                body: data
+            });
```
mhsdesign: just to stay consistent with the rest of the ui ;)

```suggestion
            return ({
                url: routes.ui.service.transliteratePathSegment,
                method: 'POST',
                credentials: 'include',
                headers: {
                    "X-Flow-Csrftoken": csrfToken,
                    "Content-Type": "application/json",
                },
                body: JSON.stringify({
                    contextNode: nodeContextPath,
                    text
                })
            });
```

PRGfx:
For every consistency issue there was an example I went off of ^^ tbf I didn't evaluate the "most common" patterns.

Before I go and make these changes: are there preferences on commit history, i.e. so far I split my changes into separate commits based on concerns - should I rather clean these up or just push the suggested changes?

Sebobo:
Your commits look great so far. But IMO the main change should be marked as `FEATURE:` instead of `TASK:`.
Would be best to group all suggestions into one commit.

PRGfx:
As I understood you prefer it being squashed in the end, I just pushed it separately for now.

That changing dependencies includes the .yarn/cache/ items is correct?

Sebobo:
Yes that is correct.
We prefer multiple commits if they help understanding the individual steps during review and later on during debugging.
But if there are many "fixup" "fixed another bug" etc. those should be merged into one. I use Gittower and try to clean up, reorder and combine my commits in my PRs before finally giving them to review.

PRGfx:
I squashed the main commits into a "FEATURE" commit. To me the main feature of this PR does not work with only parts of that.

Thank you for your support.

mhsdesign:
Hi - i just tested it and it works fine ;)

https://user-images.githubusercontent.com/85400359/197387060-d9c666bf-af04-421a-bf8e-2423b265efc7.mp4

but since this is now async im wondering about the UX on slower connections / server ...
is this behaviour realistic?


https://user-images.githubusercontent.com/85400359/197387513-0dbc3a26-ffb6-4a7a-b654-44ddf931c636.mp4


i played with this:


https://user-images.githubusercontent.com/85400359/197387550-46d99c9b-61a2-441b-aab3-ad2608925bb7.mp4


but its not so cool for fast responses ... (or is it?)

https://user-images.githubusercontent.com/85400359/197387641-ed283787-59de-4a4b-b00c-a5b6a057257b.mp4

PRGfx:
There was a question in slack regarding refreshing anchor links and I thought about abstracting this editor into something that could refresh from an arbitrary data source.
However I'm not sure if that would be implemented with something like existing data-source options + client eval and preconfiguration or whether that would be some derived editor built with a specific callback.

I thought about the missing feedback on this editor as it is. Maybe it is enough to disable the button until resolved, if that has some distinctive look.
Regarding the spinner there is this idea to just start it after a few milliseconds in case it returns early, but then of course it might act the same, just a bit delayed ^^

Sebobo:
I think disabling the input and spinning is best. IMO this is an action that doesn't occur all the time and the editor waits for the result.
But we might add a transition to the background color, then it wouldn't hurt the eye.

PRGfx:
The button becomes disabled now and the icon spins while the process is busy.
Thank you for your changes @mhsdesign! I changed the the iconButton changes to accept iconProps instead of a complete icon and renamed the state variable to make it a bit more generic.

@Sebobo changing the style for inputs exceeds the scope of this issue I think. An animation would need to be set for the base background and thus drastically change the behavior for regular focusing of inputs I think.

Sebobo:
Thx!

One more thing we should clean up, you named the method in 3 different ways right now:

* generate-uri-path-segment
* regeneratePathSegment
* transliteratePathSegment

I think we should have only one variant. `transliterate` doesn't sound right to me, therefore I personally prefer `generate` or `regenerate`.

What do you guys think?

mhsdesign:
yes i stumbled on that too - but forgot to mention it ^^
just name everything from route to action something like `generate-uri-path-segment`

PRGfx:
Fair point. The editor method is a bit out of line now but I want to avoid the same name for different functions.

PRGfx:
Hey, are there any more changes you want done, anything I can/should do?

packages/react-ui-components/src/IconButton/iconButton.tsx
```diff
@@ -18,6 +18,11 @@ export interface IconButtonProps extends HTMLButtonElementAttributesExceptStyle
      */
     readonly icon: string | IconName;
 
+    /**
+     * Additional props passed to the icon
+     */
+    readonly iconProps?: React.ComponentProps<typeof Icon>;
```
mhsdesign, line 24: we discussed whether to use slots or additionalProps but it seems to boil down to personal preference - and i dont know if we decided in the ui already for the one or other pattern

im team "slots" but this is also fine with me (as its not such a critical change)

mhsdesign (approved):
Thx for the adjustments!

Sebobo (approved):

packages/react-ui-components/src/IconButton/iconButton.tsx
mhsdesign: another alternative would be `isIconSpinning` or sth ...

PR #3236, opened 2022-10-26
BUGFIX: Resolve JS menu warning

Sebobo:
The proptypes in the UI complained about the missing `skip18n` property for menu entries. Additionally there were a few typos in the proptypes.

```
react_devtools_backend.js:4026 Warning: Failed prop type: The prop `menu[1].children[0].skipI18n` is marked as required in `App`, but its value is `undefined`.
    in App (created by ConnectFunction)
    in ConnectFunction (created by Root)
    in Neos (created by Root)
    in Unknown (created by Root)
    in Provider (created by Root)
    in div (created by Root)
    in Root
```


mhsdesign (approved):
by reading ;)

... i think i encountered the same or similar error in `<App/>` sometime ...

markusguenther (approved):
Good catch :)

PR #3238, opened 2022-10-26
BUGFIX: Load neos backend path after restoring user impersonation

### LINKED ISSUE #3237: BUG: Switch back mode for impersonation is broken

- [X] I have looked for the existing issues

### Current Behavior

When I switch back to the user, I end up in an error.
The session has changed, but the redirect seems to run in Access Denied error.

When I manually go to `/neos` everything is fine, and I can use my old logged-in user.

### Expected Behavior

When I switch back, I don't get an error and can continue to work.

### Steps To Reproduce

1. Open User module in backend
2. Switch to another user
3. Open the Content Module
4. Switch back via User-Menu

### Environment

```markdown
- Flow:8.0
- Neos:8.0
```

### Anything else?

_No response_

markusguenther:
Instead of reloading the same page, which leads to an access denied error caused by the wrong workspace, we now load the /neos path like in the rest of the Neos backend.

Fixes: #3237

### DIFF
--- a/packages/neos-ui-sagas/src/UI/Impersonate/index.js
+++ b/packages/neos-ui-sagas/src/UI/Impersonate/index.js
@@ -38,7 +38,7 @@ export function * impersonateRestore({globalRegistry}) {
             } else {
                 yield put(actions.UI.FlashMessages.add('restoreUserImpersonateUser', errorMessage, 'error'));
             }
-            window.location.reload();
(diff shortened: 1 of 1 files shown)


grebaldi (approved):
LGTM :+1: 

I was also able to reproduce the issue locally and can verify that with your change switching back to the initial user works as expected now :)

PR #3239, opened 2022-10-26
BUGFIX: Adjust width of dropdown content

### LINKED ISSUE #3235: Long labels in the User-Setting Dropdown does not fit
### Description

Since the latest release (8.1.1) the User-Settings drop down cannot render long labels anymore and the layout is broken.
The reason seems to be this PR (https://github.com/neos/neos-ui/pull/3221)

In English and with not so long Names, everything is fine. With defaults in German, it already breaks.

Screenshot 2022-10-26 at 09 37 13

### Steps to Reproduce

1. Switch backend language to German
2. Open the user dropdown
3. `Benutzereinstellungen` is then too long

#### Expected behavior

Labels don't break the layout

#### Actual behavior

Long labels will be rendered in a new line

### Affected Versions

Neos: 8.1.0

UI:8.1.1

markusguenther:
The content of the user dropdown was limited in the width, and therefore it could happen that the space for the labels was not enough. Now the contents section can get bigger than the trigger button.

Fixes: #3235

Screenshot 2022-10-26 at 16 28 12

### DIFF
--- a/packages/neos-ui/src/Containers/PrimaryToolbar/UserDropDown/style.css
+++ b/packages/neos-ui/src/Containers/PrimaryToolbar/UserDropDown/style.css
@@ -6,7 +6,7 @@
 .dropDown {
     position: static;
     display: block;
-    overflow-x: auto
+    overflow-x: auto;
 }
 .dropDown__btn {
     position: relative;
--- a/packages/neos-ui/src/Containers/PrimaryToolbar/UserDropDown/style.css
+++ b/packages/neos-ui/src/Containers/PrimaryToolbar/UserDropDown/style.css
@@ -6,6 +6,7 @@
 .dropDown {
     position: static;
     display: block;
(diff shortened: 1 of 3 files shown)


packages/neos-ui/src/Containers/PrimaryToolbar/UserDropDown/style.css
```diff
@@ -29,6 +30,7 @@
 .dropDown__contents {
     position: absolute;
     min-width: 160px;
+    width: max-content;
```
grebaldi, line 33: :ok_hand:

grebaldi (approved):
I was able to reproduce the issue locally and can verify that your changes fix the problem :+1: 

Code LGTM :)

dlubitz:
@markusguenther This is also an issue in 7.3. Can we get this also in there?
image

PR #3240, opened 2022-10-27
TASK: Change the CircleCI image to PHP 7.4.32

markusguenther:
Updates the PHP container image to use PHP `7.4.32` with composer 2, as composer 1 is deprecated and the build pipeline did not succeed anymore.

**What I did**
Changed the PHP container image


markusguenther:
Screenshot 2022-10-27 at 08 45 05

PR #3243, opened 2022-11-03
BUGFIX: Perform commit of image crop only if crop adjustments have changed

### LINKED ISSUE #3242: BUG: Opening Image Cropper causes immediate change
https://user-images.githubusercontent.com/2522299/199695632-b74ce466-2988-463c-9fef-7d50994c2bb6.mp4

### Description

When opening the Image Cropper, the Inspector immediately recognizes a change (given a crop is already applied) even though the editor didn't do anything.

### Steps to Reproduce

1. Click on an image that has a crop applied
2. Open the Image Cropper via the Inspector
3. See the "Apply" and "Discard" buttons at the bottom of the Inspector light up

#### Expected behavior

Opening the Image Cropper without applying a new crop does not cause a change.

### Affected Versions

Neos: 7.3 and above

UI: 7.3 and above

grebaldi:
fixes: #3242 

**What I did**

The `handleMediaCrop` method of the Image editor now performs a check on the crop adjustments it receives. It will now only commit a change, if the received crop adjustments differ from the crop adjustments that are already present.

**How to verify it**

If you're able to reproduce the issue described in #3242, then check out this change and the issue should disappear. Otherwise, the Image Editor (and the Image Cropper in particular) should not behave any different than before.

### DIFF
--- a/packages/neos-ui-editors/src/Editors/Image/index.js
+++ b/packages/neos-ui-editors/src/Editors/Image/index.js
@@ -127,15 +127,25 @@ export default class ImageEditor extends Component {
 
         const imageWidth = $get('originalDimensions.width', image);
         const imageHeight = $get('originalDimensions.height', image);
-        const cropAdjustments = {
+        const currentCropAdjustments = $get(CROP_IMAGE_ADJUSTMENT, image);
+        const nextCropAdjustments = {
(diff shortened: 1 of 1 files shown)


markusguenther (approved):
Thanks for fixing that issue 💙

PR #3245, opened 2022-11-08
BUGFIX: Allow to copy node without edit privilege

### LINKED ISSUE #3244: User cannot copy nodes without editing permission
### Description

When a user has no edit permission on a node, removing, cutting or hiding a node using the toolbar (in the document or node tree) is not possible. But the copy button has the exact same logic, so nodes cannot be copied.

That would be very useful, though, e.g. to copy nodes from a shared "content hub" site to another site (which is working perfectly fine, if editing permissions are given.)

### Steps to Reproduce

1. Create a Neos setup with two sites, A and B.
2. Disallow editing in site A for a user.
3. With said user try to copy a document or content node from site A to paste it into site B.

#### Expected behavior

Copying works, as it's a *read* operation.

#### Actual behavior

The copy button in the tree toolbars is disabled.

### Affected Versions

UI: 2.x and up

### History of the problem

In https://github.com/neos/neos-ui/issues/2444 a bug was reported: When clicking the copy button of the content tree without having selected a node there, an exception was raised.

That was fixed with https://github.com/neos/neos-ui/pull/2498: Only enable the copy button, if the node can be edited. That fix "worked", as that is never true, unless a node is selected. But it tied copying to being allowed to edit, which was an unintended side effect.

kdambekalns:
This changes the logic for disabling the copy button to only consider read-only workspaces as blocking for a copy operation.

`canBeEdited` and `destructiveOperationsAreDisabled` have no relevance for copying, as that is a read operation.

Fixes #3244

### DIFF
--- a/packages/neos-ui/src/Containers/LeftSideBar/NodeTreeToolBar/index.js
+++ b/packages/neos-ui/src/Containers/LeftSideBar/NodeTreeToolBar/index.js
@@ -205,11 +205,7 @@ export default class NodeTreeToolBar extends PureComponent {
                             focusedNodeContextPath={focusedNodeContextPath}
                             onClick={this.handleCopyNodes}
                             isActive={isCopied}
-                            disabled={
-                                isWorkspaceReadOnly ||
-                                destructiveOperationsAreDisabled ||
-                                !canBeEdited
-                            }
+                            disabled={isWorkspaceReadOnly}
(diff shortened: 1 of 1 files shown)


kitsunet (approved):

grebaldi (approved):

markusguenther (approved):
Thanks @kdambekalns  :)

PR #3252, opened 2022-11-17
BUGFIX: Load content tree without waiting for guest frame

### LINKED ISSUE #3251: BUG: Content tree is not loaded when guest frame cannot be initialised
### Description

When the guest frame crashes due to a rendering bug or something else the content tree stay empty (except the document node)

### Steps to Reproduce

1. Create a rendering error by removing a brace or similar in Fusion

#### Expected behavior

The content tree should fully load as the metadata in the rendered document is not required anymore since #2927

#### Actual behavior

The content tree stays empty

### Affected Versions

Neos: all

UI: all

Sebobo:
**What I did**

The content tree loads even when the guest frame crashes.

**How I did it**

The information from the guest frame is not needed anymore since #2927 so the action required for the content tree to load is now immediately triggered when the document node context path is available instead of waiting for the guest frame to fully initialise.
This also make the UI more responsive as the content tree now loads at the same time as the document tree.

**How to verify it**

Open the Neos UI and the content tree should load before a page has been rendered. When breaking the rendering it should also still load.

Resolves: #3251

### DIFF
--- a/packages/neos-ui-guest-frame/src/initializeGuestFrame.js
+++ b/packages/neos-ui-guest-frame/src/initializeGuestFrame.js
@@ -69,14 +69,19 @@ export default ({globalRegistry, store}) => function * initializeGuestFrame() {
     // Remove the inline scripts after initialization
     Array.prototype.forEach.call(guestFrameWindow.document.querySelectorAll('script[data-neos-nodedata]'), element => element.parentElement.removeChild(element));
 
-    yield put(actions.CR.Nodes.setDocumentNode(documentInformation.metaData.documentNode, documentInformation.metaData.siteNode));
+    const state = store.getState();
+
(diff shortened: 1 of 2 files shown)


crydotsnake (approved):
Tested it, and works as expected.

Sebobo:
Have to check the failing e2e test

kdambekalns (approved):
Looks good by reading

Sebobo:
The failing test actually spotted an error I would have introduced. Yay for tests!
The content frame still needs to be able to override the current document node in some cases as the rendered node is defined via Fusion (and other stuff) and not necessarily what the UI expects. But in most cases this is not the case so I added a check to prevent setting the document node twice.

markusguenther (approved):
Thanks 🙇‍♂️

PR #3254, opened 2022-11-20
WIP: Feature esbuild jest tests

mhsdesign:
... i dont want to mess with the actual pr https://github.com/neos/neos-ui/pull/3200 and need some fresh space :P

dont be fooled by the green checkmark next to `ci/circleci: unittests` in #3200 - 0 unit tests are executed, and we need to fix that.

## A few notes:
- i needed to update jest to at least 24 (latest 29) so i could pass custom options to the transformer (this is broken with 23 :O)
  - if we use 29, there are a lot of console.errors about `requestAnimationFrame` not beeing existing - i guess jest removed some polyfills ... 
- i didnt check out the new .yarn cache, as its a lot of updates again
- neither https://www.npmjs.com/package/esbuild-jest (in use) nor https://www.npmjs.com/package/jest-esbuild seem to be really active projects ...
- if you feel like it do a `yarn jest --clearCache` ;)


packages/jest-preset-neos-ui/jest-preset.json
```diff
@@ -5,7 +5,7 @@
   "coveragePathIgnorePatterns": [
     "\\.((story|spec).(js|tsx?))$"
   ],
-  "transformIgnorePatterns": ["node_modules/(?!@neos-project)"],
+  "transformIgnorePatterns": ["node_modules\\/(?!@neos-project)"],
```
mhsdesign, line 8: transforming everything doesnt work either ...

mhsdesign:
Outdated notes:
<details>
## Status `esbuild-jest`:

some workspaces work already :D (those with not so complex tests ...)

### Problem `@neos-project/neos-ui`

but when i run `yarn workspace @neos-project/neos-ui run test` i get
`Test Suites: 1 failed, 39 passed, 40 total`

image

... and now it starts to be confusing ... `PropertyGroup/index.spec.js` doesnt contain IconProps, so i guess it all ends up bundled in one file.

Also the error `ReferenceError: Cannot access 'defaultProps' before initialization` doesnt match the below stacktrace and ascii art, so that wont help us.

I checked, and it is transformed correctly: [online demo](https://hyrious.me/esbuild-repl/?version=0.15.14&mode=transform&input=export+const+defaultProps%3A+DefaultProps+%3D+%7B%0A++++color%3A+%27default%27%2C%0A++++padded%3A+%27none%27%2C%0A++++size%3A+%27sm%27%0A%7D%3B%0A%0Aclass+Icon+extends+PureComponent%3CIconProps%3E+%7B%0A++++public+static+readonly+defaultProps+%3D+defaultProps%3B%0A%0A++++public+render%28%29%3A+JSX.Element+%7C%C2%A0null+%7B%0A++++++++const+%7Bicon%7D+%3D+this.props%3B%0A%0A++++++++if+%28icon+%26%26+icon.substr%280%2C+11%29+%3D%3D%3D+%27resource%3A%2F%2F%27%29+%7B%0A++++++++++++return+%3CResourceIcon+%7B...this.props+as+ResourceIconProps%7D+%2F%3E%3B%0A++++++++%7D+else+%7B%0A++++++++++++return+%3CFontAwesomeIcon+%7B...this.props%7D+%2F%3E%3B%0A++++++++%7D%0A++++%7D%0A%7D%0A&options=--loader%3Dtsx)

Question is why and which `defaultProps` now is accessed illegally - by looking at some code, it all seems fine (and its working in the browser)
So i did a quick find  and replace for `export const defaultProps` to `export var defaultProps` so i get further and can just skip this issue for know but then im stuck at:
image

so no real step further ...

## Status `jest-esbuild`:

im just trying `jest-esbuild` (which requires at least jest 27) instead of `esbuild-jest` 

mocks wouldnt be working with this transform (that might be okay?) https://github.com/hannoeru/jest-esbuild/issues/6

after patching `jest-esbuild` with a little feature (to compensate that we are writing tsx in js files) it works even better :D

current error is the same as upwards `ReferenceError: Cannot access 'defaultProps' before initialization`
</details>

mhsdesign:
Esbuild seems to rename variables sometimes and prefix them with `2` (as the esbuild bundled code runs in one js scope instead nested, so there cannot be any name clashing)

```
FAIL src/Dialog/dialog.spec.tsx
  ● <Dialog/> › Portal should render correctly.

    expect(received).toMatchSnapshot()

    Snapshot name: `<Dialog/> Portal should render correctly. 1`

    - Snapshot  - 4
    + Received  + 4

    @@ -1,9 +1,9 @@
    - <CloseOnEscape
    + <CloseOnEscape2
        onEscape={[Function]}
      >
    -   <Portal>
    +   <Portal2>
          <section
            autoFocus={true}
            className="dialogClassName wideClassName errorClassName"
            role="dialog"
            tabIndex={0}
    @@ -40,7 +40,7 @@
              type="error"
            >
              Foo children
            </clickOutside(DialogWithoutEscape)>
          </section>
    -   </Portal>
    - </CloseOnEscape>
    +   </Portal2>
    + </CloseOnEscape2>
```

This is related to why the magic react selector doesnt work ... :(

Sebobo:
What the hell?!?

markusguenther:
Thanks for all your effort 💙

mhsdesign:
So except 2 test everything is passing!!!
That means 220 tests are fine :D 

Well now and then we get a console.error about `Warning: Failed prop type: The prop "editor" is marked as required in "EditorEnvelope", but its value is "undefined".` but this doesnt seem to count as an error :P

those two things are failing:

```
FAIL src/index.spec.js
  ● should not be writable.

    expect(received).toThrow()

    Received function did not throw

      at Object.<anonymous> (src/<stdin>:82:16)
```

and 

```
FAIL src/createConsumerApi.spec.js (7.141 s)
  ● "createConsumerApi" should create a global, read-only object

    expect(received).toThrow()

    Received function did not throw

      at Object.<anonymous> (src/<stdin>:12:25)

  ● "createConsumerApi" should expose the passed api, with each key being read-only

    expect(received).toThrow()

    Received function did not throw

      at Object.<anonymous> (src/<stdin>:33:26)
```

mhsdesign:
idk why the unit tess dont run on ci since the yarn update ... see also the problem on 8.2 https://github.com/neos/neos-ui/pull/3256

mhsdesign:
CI works better, but now only checks the typescript interfaces package (the first package)

image

mhsdesign:
yess greeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeen ;)

this was the last bit: https://github.com/neos/neos-ui/pull/3254/commits/53805ba859452462ba4f562f08d3eb1d945a9f0c fixing https://github.com/neos/neos-ui/pull/3254#issuecomment-1322714718

PR #3260, opened 2022-11-23
TASK: Remove unused DTO

Sebobo:


markusguenther (approved):
Thanks for taking care and thanks to @kdambekalns for the hint.

Issue #3263, opened 2022-11-24
Task: Open (optional) todos and wishes after esbuild change


mhsdesign:
See also https://github.com/neos/neos-ui/pull/3200

Todos
- [x] Critical https://github.com/neos/neos-ui/issues/3262
- [x] extract webpack neos-ui-extensiblity adapter outside of workspaces (so we dont need babel installed anymore :P)
- [x] just for fun, as its currently a bit odd done: replace `./Build/createVersionFile.sh` with esbuild define - but test would fails if not defined in jest
- [x] no need for `postcss-import` as esbuild does it itself https://github.com/neos/neos-ui/pull/3266
- [x] do we still need postcss-hexrgba -> esbuild minifies a lot -> so it may not even be preserved. What is the usecase? https://github.com/neos/neos-ui/pull/3266
- [x] use postcss `composes` again instead of scss `@extends` (see notes at https://github.com/neos/neos-ui/pull/3200#issuecomment-1281229469)
- [x] replace style.css to style.module.css when used as module https://github.com/neos/neos-ui/pull/3200#issuecomment-1272485910
- [x] make yarn cache smaller - eg dont include all the variations of esbuild, darwin debian ...
- [x] analyze bundle (again) with https://bundle-buddy.com/ to find duplicate dependencies in bundle
- [x] remove odd proxy HostOnlyStyles.js file (note the "js") artefact of https://github.com/g45t345rt/esbuild-style-plugin
- [x] sourcemap: true https://github.com/neos/neos-ui/pull/3271
- [x] remove unneded legacy webpack dependencies
- [x] remove neos build package
- [x] update readme https://github.com/neos/neos-ui/pull/3273
- [x] set target to 2020 https://github.com/neos/neos-ui/issues/3267
- [x] remove env.rootPath
- [x] remove esbuild.js todo https://github.com/neos/neos-ui/blob/1f597ee18521442643c881d4201411d5af17d649/esbuild.js#L44
- [x] cleanup ../dist/ in neos-ui-extensibility needed for https://github.com/neos/neos-ui/issues/3215
- [x] exclude tsbuildinfo from npm
- [x]  check if we have ts checks on lint
- [x] restrict whats published to npm

Not really gonna happen:
- [ ] improve unit tests https://github.com/folke/esbuild-runner (experiment with sourcemap inline and copy other things so errors are better displayed ;))
- [ ] use pnp https://github.com/neos/neos-ui/pull/3218
[...]

PR #3266, opened 2022-11-24
TASK: esbuild adjustments

mhsdesign:
some leftovers from #3200 

This pr holds
- remove postcss auto prefixer (no need as we support es2020 #3274 )
- remove postcss hexrgba plugin which allowed to write `rgba(#000, .5)` instead of `rgba(0, 0, 0, .5)` (and its few usages it just seems wrong to pull in a plugin for 6usages)
- remove post css import plugin (esbuild will handle css imports already)
- use typescript parser also for linting js files (that way we can ged rid of babel)
- enable scss linting and fix minor problems

related: https://github.com/neos/neos-ui/issues/3263

die babel


packages/react-ui-components/src/Tree/node.js
```diff
@@ -181,7 +181,6 @@ export class Header extends PureComponent {
             isOver,
             isDragging,
             canDrop,
-            connectDragPreview,
             ...restProps
         } = this.props;
         const rest = omit(restProps, ['onToggle', 'isCollapsed', 'hasError', 'isDragging', 'dragForbidden']);
```
mhsdesign:
```suggestion
        const rest = omit(restProps, ['onToggle', 'isCollapsed', 'hasError', 'isDragging', 'dragForbidden', 'connectDragPreview']);
```

mhsdesign:
Id like to get that clean up also merged.
That where the commits i had laying around locally when #3200 was merged.

PR #3271, opened 2022-11-25
Task: Always generate sourcemap (ca. 14mb)

mhsdesign:
Like we did in the webpack production build


Sebobo (approved):
Complicated change, but fine by reading!

mhsdesign:
ping @markusguenther i know its not an easy one

markusguenther (approved):
💙

Issue #3214, opened 2022-10-08
Update Readme / Better dev setup guide


mhsdesign:
Wip

mhsdesign:
Technically it was greatly improved by @JamesAlias via: https://github.com/neos/neos-ui/pull/3369

we now need to tackle the readme a bit further ^^

PR #3273, opened 2022-11-25
Update README.md

mhsdesign:


mhsdesign:
as via https://github.com/neos/neos-ui/issues/3214 proposed there are many more things to do better. This is just a quick one that can be merged fast and easy to update the readme at all ;)

markusguenther (approved):

PR #3282, opened 2022-11-30
BUGFIX: Fix dropdown spacings for reset button

### LINKED ISSUE #3281: Spacing issue in the preset filter
### Description

The clear button is a bit to close to the dropdown caveat in the preset filter.

### Steps to Reproduce

1. Open Neos Demo for instance
2. Filter by preset 
3. hover the reset button

#### Expected behavior

Screenshot 2022-11-29 at 19 58 34

#### Actual behavior

Screenshot 2022-11-29 at 19 55 27

[What actually happened]

### Affected Versions

Neos: 8.2.0

UI: 8.2.0

markusguenther:
Since the esbuild change, the dropdown reset button is moved into the space from the chevron. This did not happen before because the styles from dropdown and the reset were combined with composes. Right now, we don’t use composes anymore and the rewrite had issues here so that padding clashed a bit.

Fixes: #3281

### DIFF
--- a/packages/react-ui-components/src/DropDown/style.scss
+++ b/packages/react-ui-components/src/DropDown/style.scss
@@ -23,6 +23,10 @@
     white-space: nowrap;
     line-height: var(--spacing-GoldenUnit);
     background: var(--colors-ContrastNeutral);
+
+    &.dropDown__btn--withChevron {
+        padding-right: var(--spacing-GoldenUnit);
+    }
 }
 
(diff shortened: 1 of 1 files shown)


markusguenther:
**Before:**

Screenshot 2022-11-30 at 20 18 49

**After:**

Screenshot 2022-11-30 at 20 25 40

Some for the Preset Filter for instance

Issue #3262, opened 2022-11-24
Task: All 8.2 neos npm packages need to be delayed released and checked if they still work after esbuild change


mhsdesign:
By changing the build system from webpack to esbuild in https://github.com/neos/neos-ui/pull/3200 and partly banning babel - a lot of internals change.

This issue holds left todos from #3200

Following sub packages of this monorepo are affected:
- https://www.npmjs.com/package/@neos-project/react-ui-components
- https://www.npmjs.com/package/@neos-project/neos-ui-extensibility
- https://www.npmjs.com/package/@neos-project/debug-reason-for-rendering
- https://www.npmjs.com/package/@neos-project/build-essentials
- https://www.npmjs.com/package/@neos-project/positional-array-sorter

Todos
- [x] re add `main` fields and build subpackages command for npm subsplits
- [x] check if neos-ui-extensiblity still works (bable stuff) did we update babel?
- [x] fix scss files in lib esm build of react-ui-components

mhsdesign:
Following things where changed:

packages/neos-ui-extensibility/package.json
- "main": "./dist/index.js", -> "module": "./src/index.ts",

packages/positional-array-sorter/package.json
- "main": "./dist/positionalArraySorter.js", -> "module": "./src/positionalArraySorter.js",


also in packages/react-ui-components/package.json we must make sure via esbuid to always use the "main" field instead of "module"

markusguenther:
We in the Jenkins release task we now change the module folder from src to dist and then we are using the babel transpiled code. So we can close this ...

Issue #3258, opened 2022-11-22
Task: Fix unittest console errors


mhsdesign:

mhsdesign:
Problem since using yarn ...

mhsdesign:
dub of https://github.com/neos/neos-ui/issues/3288

PR #3284, opened 2022-11-30
BUGFIX: Fix module path in package.json

Sebobo:


mhsdesign:
nenennenennennen das mach ich schon :)

Sebobo:
What do you mean?

mhsdesign:
ups hab den diff falsch gelesen

markusguenther:
English please 🙃

Sebobo:
So is this fix now correct or not?
The referenced file before does not exist and my IDE also complained about it.

mhsdesign:
hmm yes i tried to understand why ci fails ... it seems pretty much correct... odd that i missed this at all

(thanks for noticing and trying to fix it)

markusguenther:
> hmm yes i tried to understand why ci fails ... it seems pretty much correct... odd that i missed this at all

It is not just CI. You can see that locally as well!

mhsdesign:
I give up for now ... i couldnt fix it ...

i tried various typescript settings (https://www.typescriptlang.org/tsconfig#moduleResolution to Node16 Node12 NodeNext) upgraded typescript from 4.7 (now) to 4.9 and set  other useless options in the tsconfig ...

i think, we might be able to fix this one case manually by telling typescript where to find the array sorter, but that defeats the purpose of yarn workspaces.

So ultimately we want to fix
> typescript yarn workspaces cannot find module


hmmm wow:
> "main": "./src/positionalArraySorter.ts",
works too - but we want to distribute cjs and esm at the same time no? https://github.com/neos/neos-ui/issues/3262

markusguenther:
Thanks for this fix.

Tried it locally and the tests say that everything is fine, but we have some console errors inside the tests.
Checked the last 8.1 run and there were not console.errors. (https://app.circleci.com/pipelines/github/neos/neos-ui/2764/workflows/8c1dd7b7-268d-430f-9ccf-07eab162ff0c/jobs/12663)

For instance this one.
Screenshot 2022-12-02 at 22 30 27

markusguenther:
But these `console.errors` are not related to that change. So this helps to bring us forward :)
Also tested now with `"@neos-project/positional-array-sorter": "^8.2.1"` as dependency in the extensibility instead of the 8.1.4

markusguenther:
Created an issue for the errors, so that we don't miss it https://github.com/neos/neos-ui/issues/3288

mhsdesign:
thx for all the testing.
see #3258 for tge console errors ;)

PR #3291, opened 2022-12-08
BUGFIX: Fix yarn npm auth token

### LINKED ISSUE #3283: Don't require NPM_TOKEN locally
### Description

For publishing the packages to `npm` we are using now `yarn npm publish` therefore yarn needs configuration for NPM authentication. As we use 2FA, we use a token in the CI/Jenkins to be able to act. Sadly, the configuration is also used in the local environment when we execute yarn install.

#### Expected behavior

Locally, we don't need to define NPM_TOKEN.

#### Actual behavior
Screenshot 2022-11-30 at 20 57 30

### Affected Versions

Neos: 8.2.0

UI: 8.2.0

### Workaround
If you need to build the neos-ui, you have to comment out the line `npmAuthToken: "${NPM_TOKEN}"` in the `.yarnrc.yaml`

JamesAlias:
Fixes #3283 

**What I did**
Use empty fallback for `NPM_TOKEN` env variable.

**How I did it**
Live with @ahaeslich.

**How to verify it**
`yarn --version` should _not_ show missing NPM Token error.

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -27,6 +27,6 @@ changesetBaseRefs:
   - composer/8.2
 
 # NPM publish config
-npmAlwaysAuth: true
-npmAuthToken: "${NPM_TOKEN}"
+npmAlwaysAuth: false
+npmAuthToken: "${NPM_TOKEN-}"
 npmPublishAccess: "public"


markusguenther (approved):
Tested locally, and thanks for providing this patch and kudos to @Sebobo for bringing that up :)

PR #3294, opened 2022-12-13
FEATURE: custom node deletion text

pKallert:
**What I did**
Sometimes when deleting a node Editors should be warned that they should look out for certain things. 
For example, when using an nodetype as a reference, editors should be warned that they are potentially deleting an important element. 

**How I did it**
To show a custom delete confirm message I added a new option for the ui: deleteConfirmation. This message will be shown in the Delete Popup 

**How to verify it**
For any node type add deleteConfirmation in the ui section
```
'Neos.Demo:Content.Headline':
  ui:
    deleteConfirmation: 
        message: 'Deleting the headline has a bad SEO impact.'

```
Bildschirmfoto 2022-12-22 um 18 15 00
Bildschirmfoto 2022-12-22 um 18 20 57

**TBD**
1. Should we use a different key in the yaml-file so that we can add other messages later on?
2. Should we hide the "original" message if there is a custom delete message?


packages/neos-ui/src/Containers/Modals/DeleteNode/index.js
```diff
@@ -134,8 +137,12 @@ export default class DeleteNodeModal extends PureComponent {
...
                 >
                 <div className={style.modalContents}>
-                    <I18n id="Neos.Neos:Main:content.navigate.deleteNodeDialog.header"/>
-                    &nbsp; {nodesToBeDeletedContextPaths.length > 1 ? `${nodesToBeDeletedContextPaths.length} ${i18nRegistry.translate('nodes', 'nodes', {}, 'Neos.Neos.Ui', 'Main')}` : `"${$get('label', node)}"`}?
+                    <p>
+                        <I18n id="Neos.Neos:Main:content.navigate.deleteNodeDialog.header"/>
+                        &nbsp; {nodesToBeDeletedContextPaths.length > 1 ? `${nodesToBeDeletedContextPaths.length} ${i18nRegistry.translate('nodes', 'nodes', {}, 'Neos.Neos.Ui', 'Main')}` : `"${$get('label', node)}"`}?
+                    </p>
+                    {warnings.length > 0 ? <hr /> : ''}
+                    {warnings.map((warning, index) => <p key={index}>{warning}</p>)}
```
Sebobo: The warning should also be translated right?

Sebobo: And how about showing the `node.label` so you know the context when there are multiple?

Sebobo:
In your first example for referenced nodes it wouldn't really help so much IMO, as the editor doesn't know if there is one.
I wrote a UI plugin that shows references in each element for some customers. So it would be cool to have the option for some datasource that would respond with a more meaningful warning that could actually contain the number of references or similar. Wdyt? Of course this increase complexity a bit, but just messages which always appear don't help so much I think.

packages/neos-ui/src/Containers/Modals/DeleteNode/index.js
pKallert: node.label is very long sometimes, so it would look strange: 
Bildschirmfoto 2022-12-13 um 14 13 15

I was thinking about grouping by nodeType and then using the nodeType label, what do you think? 

Translation is possible now, but using i18n in the yaml file currently does not work. I will check in the core for that

pKallert:
Perhaps the reference stuff was not the best example. 
Another use case would be using the https://github.com/punktDe/nodereplicator .
When deleting Nodes across all dimensions warning the unsuspecting Editor would be nice.

packages/neos-ui/src/Containers/Modals/DeleteNode/index.js
Sebobo: How about a bullet list with items like this (untested Pseudocode):

`<NodeType.label>: <I>{String.cropAtWord(node.label, 100, '...'}</I>`

Sebobo:
Either way, it would be great to somehow have be possible to have a custom check.

If you define the `deleteConfirmation` as an object and for now only implement the `message` and support `i18n` as value it would be extendable later on. Then in the next step we could allow something like a datasource f.e. that would be called to fill the details of the dialog.

But when deleting lots of nodes this could quickly become a complicated dialog 😉

packages/neos-ui/src/Containers/Modals/DeleteNode/index.js
pKallert: I updated the label, it looks like this now: 

Bildschirmfoto 2022-12-22 um 18 15 00

pKallert:
Configuration and Labels are adjusted (see updated PR description). 

I am not sure how to make the message translatable with i18n, is there any documentation or good example I could have a look at?

Sebobo:
You can use the `i18nRegistry` which is already injected like this:

```
i18nRegistry.translate(value, fallback)
```

This should then already work with shorthand translation strings like this `Neos.NodeTypes:NodeTypes.Headline:ui.deleteConfirmation.message`.

To allow the `i18n` value you have to adjust the `NodeTypeConfigurationEnrichmentAspect` in Neos.Neos to convert the `i18n` string to the label matching the node and its package. So this would be a second PR.

Would be great if you could implement the first part though.

Sebobo:
Markus asked me and I just realised that you used the `I18n` component which should do the shorthand translation already. 
So if it works it's fine for me.

On thing about this component is that in my opinion it creates a lot of React overhead and you already have the `I18nRegistry` in your component which you can use directly instead of using the component.

PR #3299, opened 2022-12-15
TASK: Disable normal Dependabot updates

Sebobo:
They don't really have any value for us as we run updates manually when needed. Security updates are still enabled separately.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,20 +1,11 @@
 version: 2
 updates:
 
-  # Maintain dependencies for npm
-  - package-ecosystem: "npm"
-    directory: "/"
-    schedule:
-      interval: "daily"
-    target-branch: "5.3"
-    labels:
-      - "dependencies"
-
   # Maintain dependencies for Composer
   - package-ecosystem: "composer"
     directory: "/"
     schedule:
       interval: "daily"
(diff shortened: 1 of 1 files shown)


Sebobo:
I'm not sure what the target branch for composer means exactly 🤷🏻‍♂️

mhsdesign (approved):
👍

.github/dependabot.yml
```diff
@@ -1,20 +1,11 @@
 version: 2
 updates:
 
-  # Maintain dependencies for npm
-  - package-ecosystem: "npm"
-    directory: "/"
-    schedule:
-      interval: "daily"
-    target-branch: "5.3"
```
markusguenther, line 9: We can change that to 7.3 as well

.github/dependabot.yml
```diff
@@ -1,20 +1,11 @@
 version: 2
 updates:
 
-  # Maintain dependencies for npm
```
markusguenther: We can keep the block and just add `open-pull-requests-limit: 0`
That will lead to security updated only and disable regular version updates.

See: https://docs.github.com/en/enterprise-server@3.3/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates

PR #3300, opened 2022-12-19
TASK: Update plow-js to v3.0.0

grebaldi:
This is a rebased version of: https://github.com/neos/neos-ui/pull/3120

This PR contains an update for `plow-js` v3.0.0 (as tracked here: https://github.com/grebaldi/plow-js/pull/21).

`plow-js` v3.0.0 isn't released yet, that's why currently all packages point to `plow-js` v3.0.0-canary2.

I'd like to use this PR to gather some feedback. I did a smoke test in a limited Neos UI setup, but it would certainly be nice to get some feedback from 1 or 2 real-world projects.

If feedback's positive, I'll do a final release of `plow-js` v3.0.0 and update this PR accordingly.


mhsdesign:
Thx for the pr - im currently lacking the overview. Wasnt our goal to remove plow js? Im just asking if this 3.0 is a valuable release where we should invest into testing or should we better take care of replacing plow in the codebase (we know can use optional chaining f.x es2020)

grebaldi:
The long term idea was to remove plow-js (targetting 9.0). Additionally, we wanted to do this update for improved performance in 8.3.

Testing and the update itself don't take a lot of time, so it's probably worth it for the sake of performance (However, there's no way to measure either way).

Sebobo:
Yes that was the plan, thx @grebaldi !

Sebobo (approved):
I kinda trust the automatic tests for this

grebaldi:
**UPDATE:** I just released plow-js v3.0.0 (officially now, not a canary release) and updated this PR accordingly.

markusguenther (approved):
Your on fire :D 
Thanks!!!

PR #3302, opened 2022-12-20
FEATURE: adjust image editor to allow a custom upload screen component

andrehoffmann30:
**What I did**
I adjusted the image editor to allow a custom upload screen / dialog as an replacement option for the default upload behaviour. 

**How I did it**
I adjusted the `handleChooseFile` function to look for a `MediaUploadScreen` in the secondary editors registry. If a component is found , it is used and renderd as a secondary editor for the upload of a file. If no component is found, the previous behaviour is used for handeling the file upload.

**How to verify it**
For testing this behaviour you need to write a plugin and add a component to be used as `MediaUploadScreen` to the secondary editors registry in your plugin.

### DIFF
--- a/packages/neos-ui-editors/src/Editors/Image/index.js
+++ b/packages/neos-ui-editors/src/Editors/Image/index.js
@@ -216,12 +216,12 @@ export default class ImageEditor extends Component {
 
     handleChooseFile = () => {
         const {secondaryEditorsRegistry, options} = this.props;
-        const {component: MediaUploadScreen} = secondaryEditorsRegistry.get('Neos.Neos/Inspector/Secondary/Editors/MediaUploadScreen');
+        const {component: AssetUploadScreen} = secondaryEditorsRegistry.get('Neos.Neos/Inspector/Secondary/Editors/AssetUploadScreen');
 
-        if (MediaUploadScreen) {
+        if (AssetUploadScreen) {
             // set media type constraint to "image/*" if it is not explicitly specified via options.constraints.mediaTypes
             const constraints = {...options.constraints, mediaTypes: (options.constraints && options.constraints.mediaTypes) || ['image/*']};
-            this.props.renderSecondaryInspector('IMAGE_UPLOAD_MEDIA', () => <MediaUploadScreen type="images" constraints={constraints || {}} onComplete={this.afterUpload}/>);
+            this.props.renderSecondaryInspector('IMAGE_UPLOAD_MEDIA', () => <MediaUploadScreen type="images" constraints={constraints} onComplete={this.afterUpload}/>);
         } else {
             this.previewScreen.chooseFromLocalFileSystem();
             this.setState({isAssetLoading: true});
--- a/packages/neos-ui-editors/src/Editors/Image/index.js
+++ b/packages/neos-ui-editors/src/Editors/Image/index.js
@@ -215,8 +215,17 @@ export default class ImageEditor extends Component {
     }
 
     handleChooseFile = () => {
-        this.previewScreen.chooseFromLocalFileSystem();
-        this.setState({isAssetLoading: true});
+        const {secondaryEditorsRegistry, options} = this.props;
+        const {component: MediaUploadScreen} = secondaryEditorsRegistry.get('Neos.Neos/Inspector/Secondary/Editors/MediaUploadScreen');
+
+        if (MediaUploadScreen) {
+            // set media type constraint to "image/*" if it is not explicitly specified via options.constraints.mediaTypes
+            const constraints = {...options.constraints, mediaTypes: (options.constraints && options.constraints.mediaTypes) || ['image/*']};
+            this.props.renderSecondaryInspector('IMAGE_UPLOAD_MEDIA', () => <MediaUploadScreen type="images" constraints={constraints || {}} onComplete={this.afterUpload}/>);
+        } else {
+            this.previewScreen.chooseFromLocalFileSystem();
+            this.setState({isAssetLoading: true});
+        }
     }
 
     handleChooseFromMedia = () => {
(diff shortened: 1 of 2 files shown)


packages/neos-ui-editors/src/Editors/Image/index.js
```diff
@@ -215,8 +215,17 @@ export default class ImageEditor extends Component {
     }
 
     handleChooseFile = () => {
-        this.previewScreen.chooseFromLocalFileSystem();
-        this.setState({isAssetLoading: true});
+        const {secondaryEditorsRegistry, options} = this.props;
+        const {component: MediaUploadScreen} = secondaryEditorsRegistry.get('Neos.Neos/Inspector/Secondary/Editors/MediaUploadScreen');
```
Sebobo: IMO we should use the wording `AssetUploadScreen` here as all our models are called something `Asset...`

packages/neos-ui-editors/src/Editors/Image/index.js
```diff
@@ -215,8 +215,17 @@ export default class ImageEditor extends Component {
...
+
+        if (MediaUploadScreen) {
+            // set media type constraint to "image/*" if it is not explicitly specified via options.constraints.mediaTypes
+            const constraints = {...options.constraints, mediaTypes: (options.constraints && options.constraints.mediaTypes) || ['image/*']};
```
Sebobo, line 230: Thx for including the constraints with this

packages/neos-ui-editors/src/Editors/Image/index.js
```diff
@@ -215,8 +215,17 @@ export default class ImageEditor extends Component {
...
+            this.props.renderSecondaryInspector('IMAGE_UPLOAD_MEDIA', () => <MediaUploadScreen type="images" constraints={constraints || {}} onComplete={this.afterUpload}/>);
```
Sebobo: The fallback won't be any use here as constraints is always an object.

```suggestion
            this.props.renderSecondaryInspector('IMAGE_UPLOAD_MEDIA', () => <MediaUploadScreen type="images" constraints={constraints} onComplete={this.afterUpload}/>);
```

Sebobo (changes requested):
Thx! Left some comments

packages/neos-ui-editors/src/Editors/Image/index.js
andrehoffmann30: Thank you for the fast response. I adjusted the naming.

packages/neos-ui-editors/src/Editors/Image/index.js
andrehoffmann30: I removed th unneded fallback.

packages/neos-ui-editors/src/Editors/Image/index.js
```diff
@@ -215,8 +215,16 @@ export default class ImageEditor extends Component {
...
 
     handleChooseFile = () => {
-        this.previewScreen.chooseFromLocalFileSystem();
-        this.setState({isAssetLoading: true});
+        const {secondaryEditorsRegistry, options} = this.props;
+        if (secondaryEditorsRegistry.get('Neos.Neos/Inspector/Secondary/Editors/AssetUploadScreen')) {
+            // set media type constraint to "image/*" if it is not explicitly specified via options.constraints.mediaTypes
+            const constraints = {...options.constraints, mediaTypes: (options.constraints && options.constraints.mediaTypes) || ['image/*']};
+            const {component: AssetUploadScreen} = secondaryEditorsRegistry.get('Neos.Neos/Inspector/Secondary/Editors/AssetUploadScreen');
+            this.props.renderSecondaryInspector('IMAGE_UPLOAD_MEDIA', () => <AssetUploadScreen type="images" constraints={constraints} onComplete={this.afterUpload}/>);
```
Sebobo: Is setting the `isAssetLoading` state not also relevant in this branch?

andrehoffmann30: I don't know if showing the spinner is relavant, if a secondary editor with an upload dialog is shown. Addtional there is a problem when closing the secondary editor via the close button in the top right corner. If I set the `isAssetLoading` to true, I don't see a good way to reset the state to false when the secondary editor is closed via the close button. I think, a possible way could be an optional third parameter for the function `renderSecondaryInspector` for adding a `handleCloseSecondaryEditor` function. This function could than be used the adjust the behaviour of `onClose` by rewriting the `handleCloseSecondaryInspector` funtion in the inspector.

Sebobo (approved):

packages/neos-ui-editors/src/Editors/Image/index.js
grebaldi: It would be nice to pass along `propertyName`, `focusedNodePath`, `siteNodePath` and `metadata` ("Image" or "Asset") as well, so that all the information the `uploadAsset` endpoint needs is available in the secondary inspector: 
https://github.com/neos/neos-ui/blob/8.3/packages/neos-ui-backend-connector/src/Endpoints/index.ts#L224

grebaldi (changes requested):
Hi @andrehoffmann30,

thanks for your PR :) 

I was able to get it to work locally. But there's one small adjustment I'd like to suggest (see comment below). 

On a larger note:

Adding this feature would change the behavior of the `ImageEditor`, but would leave the behavior of the `AssetEditor` intact. Imho the upload mechanism should not differ between the two, so it would be necessary to adjust the `AssetEditor` as well.

packages/neos-ui-editors/src/Editors/Image/index.js
andrehoffmann30: Hallo @grebaldi , adjusted the asset editor to be able use the same custom upload dialog as the image editor. I have also adapted the code so that the additional data you mentioned is passed to the dialog. In the implementation of the dialog, you can then use the data accordingly if desired.

andrehoffmann30:
Hallo @mhsdesign I changed the merge goal to the 9.0 branch. Can you please adjust the labels accordingly? Thank you.

crydotsnake:
FYI: https://discuss.neos.io/t/feedback-wanted-for-neos-ui-features/6466

grebaldi (approved):
Hi @andrehoffmann30,

thanks a lot for your adjustments! I was able to test everything locally and it works perfectly :ok_hand:

PR #3304, opened 2022-12-21
BUGFIX: fix interdependent properties in node creation dialog

### LINKED ISSUE #2852: Enable depending properties in node creation dialog
You can use depending properties in the property inspector as described in https://docs.neos.io/cms/manual/content-repository/nodetype-properties#depending-properties
This behaviour does not work in the node creation dialog which is only a nuisance if you use it for properties that are later on accessible via the inspector, but a real problem if you want to use these properties in a node creation handler to e. g. create child node depending on which options were chosen before.

### Steps to Reproduce

If I have a config like:
```
My.Package:NodeType:
  properties:
    property:
      type: integer
      ui:
        showInCreationDialog: true
        label: i18n
        inspector:
          group: 'default'
          editorOptions:
            dataSourceIdentifier: 'data-source-1'
      validation:
    otherProperty:
      type: integer
      ui:
        showInCreationDialog: true
        label: i18n
        inspector:
          group: 'default'
          editorOptions:
            dataSourceIdentifier: 'data-source-2'
            dataSourceAdditionalData:
              argument: 'ClientEval:node.properties.property'
      validation:
```
it works in the inspector, but in the node creation dialog the `data-source-2` is called with the string `'ClientEval:node.properties.property'` as `$arguments['argument']` in the beginning and not refreshed, when the first select-box is chosen.

#### Expected behavior

I would expect, that at least the second select box is refreshed, when the first one is chosen. It would be perfect if I could even use sth. like `argument: 'ClientEval:data.anyField'` to use a field that is only defined in `ui.creationDialog.elements` and not saved as a property of the node afterwards. Also hiding elements of the creation dialog until others are chosen like you can do with elements in the inspector as described in https://docs.neos.io/cms/manual/content-repository/nodetype-properties#depending-properties would be nice.

### Tested on Versions

Neos: dev-master

UI: dev-master

Related: #1138

JamesAlias:
**What We did**
We added a transient _node_ and the _parentNode_ (based on _insertion mode_) to the _CreationDialog_ to be able to run basic `ClientEval:` operations known from the _Inspector_.
We also pulled the `prepocessing` of NodeConfigurations out of the _Inspector_ into a separate file.

We wrote tests for all the things we changed and added (unit & e2e).

**How I did it**
_Live_ with Neos community.

**How to verify it**
Add a Node with interdependent properties and creation dialog turned on. You can see an example NodeType added with this PR to the test setup.

Video:

https://user-images.githubusercontent.com/1615332/219017376-a03bb09a-0224-41a2-a4a4-d010d27db5d9.mp4

## TODO

- [x] get parentNode (depending on insert mode)
- [x] extract building transientNode and parentNode out of render method
- [x] build an abstraction for configuration processing to be used by both Insepector and CreationDialog
- [x] Clean up
- [x] Docs / Comments

Resolves: #2852


JamesAlias:
## Notes for Next LiveStream

- [x] extract preprocessing method Inspector (maybe rename to "processEvalsInViewConfiguration")
  - [x] write test for extracted method
- [x] Use preprocessing in CreationDialog
  - [x] use some transient node as node for clientEval (Inspector uses node from context)

JamesAlias:
# Next Steps
- get parentNode (depending on insert mode)
- extract building transientNode and parentNode out of render method

- maybe build an abstraction for configuration processing to be used by both Insepector and CreationDialog

Sebobo:
Thx!

If this is a bugfix, do you want to target older versions too? Or is it rather a feature for 8.3?

JamesAlias:
> Thx!
> 
> If this is a bugfix, do you want to target older versions too? Or is it rather a feature for 8.3?

Not sure. It touches more places than I am confortable with backporting.
I could try to rebase it to 8.0 but I'm not sure about LTS.

What do you think?

JamesAlias:
@Sebobo I can't even rebase to 8.0 without conflicts.

8.2 is possible without conflicts.
You can give it a try if you like, but I would settle for 8.2.

markusguenther:
Rebased the PR to 7.3 and made it PHP 7.4 compatible. Thanks for the fix and all the tests :)
Really, a role model of a bug fix. 👍

markusguenther (approved):

.gitignore
```diff
@@ -36,8 +36,14 @@ Resources/Public/Styles/Font*
...
+# System Files
+#
+.DS_Store
+
 #
 # editors / IDEs
 #
 .vscode/
 .idea/
+*.iml
```
mhsdesign, line 49: Thats your signature @JamesAlias xD

markusguenther: I did not know which tool uses that ;)

packages/neos-ui/src/preprocessNodeConfiguration.spec.js
```diff
@@ -0,0 +1,189 @@
+import preprocessNodeConfiguration from './preprocessNodeConfiguration';
+
+describe(`preprocessing NodeConfiguration`, () => {
```
mhsdesign, line 3: tests ❤️

.gitignore
mhsdesign: intellij

crydotsnake:
Was this upmerged also for 8.3?

PR #3314, opened 2023-01-01
BUGFIX: Adjust module path to distribution

markusguenther:
When the package is "compiled" the build will not handle typescript and expect a JavaScript module. As we create a distribution, we should target the dist folder.


markusguenther:
Ok seems we need to change the path while the release :/

PR #3322, opened 2023-01-04
BUGFIX: Ensure that `<DropDown/>` closes only after the current call-stack has been processed

### LINKED ISSUE #3305: BUG: Toggling Auto Publish seems not to be working 50% of the time
### Description

https://user-images.githubusercontent.com/2522299/209163074-54fe6b45-6ee8-432a-915e-92e22428e07a.mp4

Toggling the Auto-Publish checkbox seems not to be working 50% of the time.

### Steps to Reproduce

1. Go into any Neos backend (v8.2)
2. Try toggling the Auto-Publish Checkbox
3. See faulty behavior

#### Expected behavior

Toggling the Auto-Publish checkbox works every time :)

### Affected Versions

Neos: 7.3, 8.2

UI: 7.3, 8.2

grebaldi:
fixes: #3305 

**The Problem**

The behavior described in #3305 is a regression through https://github.com/neos/neos-ui/pull/3211. 

#3211 made sure that the `<Label/>` around the auto-publish checkbox is clickable. Normally, one would expect that clicking on the label would trigger a change in the corresponding checkbox input through event delegation.

To illustrate that idea:
```
      (User clicks on <Label/>)
                │
                │ // Click event is being delegated to the corresponding
                │ // <input type="checkbox"> element
                │
                ▼
(<CheckBox/> OnChange event happens)
```

Clicking inside the `<DropDown.Contents/>` also causes the outer `<DropDown/>` to close. Closing the `<DropDown/>` means, that the `<DropDown.Contents/>` are removed from the DOM.

This update mechanism is handled by `React` in a way that is non-deterministic (at least to us it is). This means that sometimes the `<DropDown.Contents/>`  are removed before the Click event from the `<Label/>` component reaches  the corresponding checkbox input:

```
 (User clicks on <Label/>)
            │
            💥 ◀─────── (DropDown.Contents are removed from DOM)
            ┊
            ▼
(<CheckBox/> is gone...)
```

**The Solution**

I lieu of `setImmediate`-support in browsers, I wrapped the toggling mechanism of the `<DropDown/>`-component in a `setTimeout(..., 0)` call.

This makes sure that components inside the `<DropDown.Contents/>` will receive all events that are currently scheduled, before the  `<DropDown.Contents/>` are removed from the DOM.

### DIFF
--- a/packages/react-ui-components/src/DropDown/wrapper.spec.tsx
+++ b/packages/react-ui-components/src/DropDown/wrapper.spec.tsx
@@ -33,25 +33,28 @@ describe('<DropDownWrapper/>', () => {
         expect(wrapper.state('isOpen')).toBe(false);
     });
 
-    it('should set the "isOpen" state value to opposite when calling the toggle method.', () => {
+    it('should set the "isOpen" state value to opposite when calling the toggle method.', async () => {
         const wrapper = shallow(<DropDownWrapper {...props}/>);
 
         // @ts-ignore
         wrapper.instance().handleToggle();
 
+        await new Promise(resolve => setTimeout(resolve, 0));
         expect(wrapper.state('isOpen')).toBe(true);
 
         // @ts-ignore
         wrapper.instance().handleToggle();
 
+        await new Promise((resolve) => setTimeout(resolve, 0));
         expect(wrapper.state('isOpen')).toBe(false);
     });
-    it('should set the "isOpen" state value to false when calling the close method.', () => {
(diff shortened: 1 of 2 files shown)


Sebobo (approved):
Thx Wilhelm!

I don't like that we need this kind of workaround but I can confirm that it resolves the issue.
We should have a different component for this kind of mechanism which doesn't do event magic.

crydotsnake (approved):
Thanks! Works as expected :)!

PR #3323, opened 2023-01-04
BUGFIX: Make sure that validation errors show up after clicking "Create New" in node creation dialog

### LINKED ISSUE #3319: BUG: Click on "Create New" does not trigger input validation in `CreationDialog`
### Description

Given a node type that configures several node creation dialog inputs with validation, said validation is not triggered when the "Create New" button of the creation dialog is clicked. Instead, the validation only gets triggered if at least one of the inputs receives at least one input event.

### Steps to Reproduce

Here's a node type that should enable reproduction:
```yaml
'Vendor.Site:Content.Example':
  superTypes:
    'Neos.Neos:Content': true

  ui:
    label: 'Validation example'
    inspector:
      groups:
        examples:
          label: 'Examples'

  properties:
    example1:
      type: string
      ui:
        label: 'Example #1'
        showInCreationDialog: true
        reloadIfChanged: true
        inspector:
          group: examples
      validation:
    example2:
      type: string
      ui:
        label: 'Example #2'
        showInCreationDialog: true
        reloadIfChanged: true
        inspector:
          group: examples
      validation:
    example3:
      type: string
      ui:
        label: 'Example #3'
        showInCreationDialog: true
        reloadIfChanged: true
        inspector:
          group: examples
```

1. Create a node with a node type like the one described above
2. Click on "Create New" in the node creation dialog
3. See nothing happen

As soon as you type something into any of the input fields, validation gets triggered immediately. This can be irritating however, since nothing tells the user that any of those fields are required. The only thing you see is that clicking on "Create New" apparently doesn't work.

#### Expected behavior

A click on "Create New" triggers the input validation, after which the button should disable itself (given that validation errors occurred).

#### Actual behavior

https://user-images.githubusercontent.com/2522299/210357939-e3c568da-76aa-4a1a-9b65-548279197efa.mp4

### Affected Versions

Neos: 7.3

UI: 7.3

grebaldi:
fixes: #3319

Turns out this could be achieved by simply setting the creation dialog's `isDirty` state to `true` when validation errors occur during `handleApply`.

On a side note, this made me wonder why there is an `isDirty` state for the creation dialog in the first place. It makes sense that the editor fields only show validation errors once they've been touched, but that is their concern. When "Create New" was clicked, all validation errors should show in any case.

### DIFF
--- a/packages/neos-ui/src/Containers/Modals/NodeCreationDialog/index.js
+++ b/packages/neos-ui/src/Containers/Modals/NodeCreationDialog/index.js
@@ -140,7 +140,7 @@ export default class NodeCreationDialog extends PureComponent {
         const validationErrors = this.getValidationErrorsForTransientValues(transient);
 
         if (validationErrors) {
-            this.setState({validationErrors});
+            this.setState({validationErrors, isDirty: true});
         } else {
             const {apply} = this.props;
             apply(transient);


Sebobo (approved):
Makes sense, smart solution. Thx!

crydotsnake (approved):
Thanks ! Works as expected. I like the solution :)

Issue #3224, opened 2022-10-12
Bug: ckeditor toolbar disappears in super rare cases while editing


mhsdesign:
### Description

Sorry, i hate to formulate such an issue without any real information.
But i talked to @NicoleNitschke at the sprint and she could confirm this issue.

So i will document this issue.

### Problem
While using the ckeditior the toolbar just dies in some rare cases -> and becomes black
after focusing another element and refocusing the previous ckeditor field - the toolbar is restored.

- im not sure anymore if i witnessed any console.errors but i think not.

Any further information would be great - if someone can reproduce this!!!

### Edit i could reproduce it:

UI version: v8.0.1

- you need to reload the Ui (to clear any js caches)
- then select the headline in the demo (directly)
- expect the toolbar to show
- click on the content collection - it resides in
- expect the toolbar to hide (as this is cust a content collection)
- click back to the headline
- expect the toolbar to show (again) -> FAILS!!!

[Bildschirmaufzeichnung vom 12.10.2022, 14:30:52.webm](https://user-images.githubusercontent.com/85400359/195343198-c342c04b-3d8a-4847-8cae-54d5af971127.webm)

### Affected Versions
UI:

i believe i saw it in:

... 7.1 - 8.0 ....

but its to expect this bug exists since all times

grebaldi:
What browser does this happen in?

I tested FF, Chrome and WebKit (all on Linux) with Neos UI 7.3 and 8.0 and cannot reproduce. But since I'm seeing a different toolbar for that headline, specifics of the setup you've been using cannot be ruled out.

I have however seen a different problem that may or may not be related to this one:

[Peek 2023-01-05 14-00 - #3224.webm](https://user-images.githubusercontent.com/2522299/210786672-eb0f0c13-0ffa-4742-9f58-7bf3a6949e2d.webm)

The SelectBox header label of the formatting SelectBox in the rich text editor toolbar remains empty on occasion, even though a value should be present.

**Update:** I seem to be onto something here, because `formattingUnderCursor` appears to lag when switching between nodes (which explains the Issue I described just then, and may also explain the initial issue).

mhsdesign:
thx will look into it ;)

markusguenther:
Should be fixed with 7.3.12

PR #3325, opened 2023-01-05
BUGFIX: Update `formattingUnderCursor` when CKEditor focus changes

grebaldi:
refs: #3224 

**The Problem**

[Peek 2023-01-05 14-00 - #3224.webm](https://user-images.githubusercontent.com/2522299/210801247-f63437d3-8172-41f6-ba8b-0128b1875229.webm)

While investigating #3224, I noticed that the `EditorToolbar` sometimes receives the wrong `formattingUnderCursor` when another node is being selected. In fact, in those cases, the `EditorToolbar` received the `formattingUnderCursor` corresponding to the previous selection.

By debugging the events triggered by CKEditor I was able to find out that in those cases a `change:isFocused` event got triggered, but no `change` event, which would have updated the `formattingUnderCursor`.

**The Solution**

This PR makes sure that `formattingUnderCursor` is properly updated when `change:isFocused` occurs through CKEditor. It does so by calling the `handleUserInteractionCallback` that is also called when the `change` event occurs.

Though I was unable to reproduce the original problem described in #3224, it would be very well explicable, if that problem was solved by this measure as well. It would be helpful, if someone who is able to reproduce the issue could verify this :)

### DIFF
--- a/packages/neos-ui-ckeditor5-bindings/src/ckEditorApi.js
+++ b/packages/neos-ui-ckeditor5-bindings/src/ckEditorApi.js
@@ -64,6 +64,7 @@ export const createEditor = store => options => {
                 if (event.source.isFocused) {
                     currentEditor = editor;
                     editorConfig.setCurrentlyEditedPropertyName(propertyName);
+                    handleUserInteractionCallback();
                 }
             });
 


Sebobo (approved):
Thx!

I was able to (randomly) reproduce the issue and couldn't anymore with this fix.

mhsdesign:
thx will look into it ;)

crydotsnake (approved):
Tested it and works as expected!

PR #3326, opened 2023-01-08
FEATURE: Allow to translate multiple id's in one field

sbruggmann:
…Neos.Demo:Main:translated.a (Neos.Demo:Main:translated.b)'

**What I did**
Labels can now contain multiple translation id's, not just one.

**How I did it**
If there's a regex match, it will be processed in a loop.
Otherwise it follows the original behaviour.

**How to verify it**
Create a NodeType label like:
```
'Neos.Demo:Document.Homepage':
  ui:
    label: 'Static Neos.Demo:Main:translated.a (Neos.Demo:Main:translated.b)'
```
Add the corresponding translation to "Packages/Sites/Neos.Demo/Resources/Private/Translations/en/Main.xlf":
```
<trans-unit id="translated.a" xml:space="preserve">
    <source>Translated-A</source>
</trans-unit>
<trans-unit id="translated.b" xml:space="preserve">
    <source>Translated-B</source>
</trans-unit>
```
Reload the UI and you will see "Static Translated-A (Translated-B)" as a result.

**Why?**
For time to time we wish to have a more flexible label translation, especially for dynamically generated NodeTypes we see a big  advantage in this enhancement.

### DIFF
--- a/packages/neos-ui-i18n/src/registry/I18nRegistry.js
+++ b/packages/neos-ui-i18n/src/registry/I18nRegistry.js
@@ -90,42 +90,42 @@ export default class I18nRegistry extends SynchronousRegistry {
 
         let translation = idOrig;
 
+        // eslint-disable-next-line max-params
+        const translateOne = (idOrig, fallbackOrig, params = {}, packageKeyOrig = 'Neos.Neos', sourceNameOrig = 'Main', quantity = 0) => {
+            const fallback = fallbackOrig || idOrig;
+            const [packageKey, sourceName, id] = getTranslationAddress(idOrig, packageKeyOrig, sourceNameOrig);
+            let translation = [packageKey, sourceName, id]
+                // Replace all dots with underscores
+                .map(s => s ? s.replace(/\./g, '_') : '')
+                // Traverse through translations and find us a fitting one
+                .reduce((prev, cur) => (prev ? prev[cur] || '' : ''), this._translations);
+
+            translation = getPluralForm(translation, quantity);
+            if (translation && translation.length) {
+                if (Object.keys(params).length) {
+                    return substitutePlaceholders(translation, params);
+                }
+                return translation;
+            }
+
+            if (!errorCache[`${packageKey}:${sourceName}:${id}`]) {
+                logger.error(`No translation found for id "${packageKey}:${sourceName}:${id}" in:`, this._translations, `Using ${fallback} instead.`);
+
+                errorCache[`${packageKey}:${sourceName}:${id}`] = true;
+            }
+
+            return fallback;
+        }
+
         translation = translation.replace(
             /([a-z\d.]+:[a-z\d.]+:[a-z\d.]+)/gi,
-            match => this.translateOne(match, match, params, packageKeyOrig, sourceNameOrig, quantity)
+            match => translateOne(match, match, params, packageKeyOrig, sourceNameOrig, quantity)
         );
 
         if (translation !== idOrig) {
             return translation;
         }
 
-        return this.translateOne(idOrig, fallbackOrig, params, packageKeyOrig, sourceNameOrig, quantity);
-    }
-
-    // eslint-disable-next-line max-params
-    translateOne(idOrig, fallbackOrig, params = {}, packageKeyOrig = 'Neos.Neos', sourceNameOrig = 'Main', quantity = 0) {
-        const fallback = fallbackOrig || idOrig;
-        const [packageKey, sourceName, id] = getTranslationAddress(idOrig, packageKeyOrig, sourceNameOrig);
-        let translation = [packageKey, sourceName, id]
-        // Replace all dots with underscores
-            .map(s => s ? s.replace(/\./g, '_') : '')
-            // Traverse through translations and find us a fitting one
-            .reduce((prev, cur) => (prev ? prev[cur] || '' : ''), this._translations);
-
-        translation = getPluralForm(translation, quantity);
-        if (translation && translation.length) {
-            if (Object.keys(params).length) {
-                return substitutePlaceholders(translation, params);
-            }
-            return translation;
-        }
-
-        if (!errorCache[`${packageKey}:${sourceName}:${id}`]) {
-            logger.error(`No translation found for id "${packageKey}:${sourceName}:${id}" in:`, this._translations, `Using ${fallback} instead.`);
-
-            errorCache[`${packageKey}:${sourceName}:${id}`] = true;
-        }
-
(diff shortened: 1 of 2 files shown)


jonnitto (approved):
Stefan shows me the use-case, and it makes totally sense.

manuelmeister:
I would love to see a test for this ;)

jonnitto (approved):

crydotsnake (approved):
From the functionality, it's working great! And, it's a nice improvement

packages/neos-ui-i18n/src/registry/I18nRegistry.js
```diff
@@ -84,6 +84,30 @@ export default class I18nRegistry extends SynchronousRegistry {
 
     // eslint-disable-next-line max-params
     translate(idOrig, fallbackOrig, params = {}, packageKeyOrig = 'Neos.Neos', sourceNameOrig = 'Main', quantity = 0) {
+        let translation = idOrig;
+
+        const regex = /([a-zA-Z0-9.]+:[a-zA-Z0-9.]+:[a-zA-Z0-9.]+)/g;
+        let m;
```
Sebobo: Please use full variable names. I assume you mean `match` here?

packages/neos-ui-i18n/src/registry/I18nRegistry.js
```diff
@@ -84,6 +84,30 @@ export default class I18nRegistry extends SynchronousRegistry {
 
     // eslint-disable-next-line max-params
     translate(idOrig, fallbackOrig, params = {}, packageKeyOrig = 'Neos.Neos', sourceNameOrig = 'Main', quantity = 0) {
+        let translation = idOrig;
+
+        const regex = /([a-zA-Z0-9.]+:[a-zA-Z0-9.]+:[a-zA-Z0-9.]+)/g;
```
Sebobo: The following is a bit cleaner `/([a-z\d.]+:[a-z\d.]+:[a-z\d.]+)/gi`

packages/neos-ui-i18n/src/registry/I18nRegistry.js
```diff
@@ -84,6 +84,30 @@ export default class I18nRegistry extends SynchronousRegistry {
...
+        let m;
+
+        while ((m = regex.exec(idOrig)) !== null) {
```
Sebobo: You can use the String.replace with a regex here and capturing groups or a replacer function see https://www.javascripttutorial.net/javascript-regex/replace/.
Imo then you don't need the loop and you don't need the `translateOne`.
Wdyt?

Sebobo (changes requested):
Thx for the change.

I'm not sure about the use cases as they are only mentioned indirectly.

Regarding the bigger picture: I think all our translation methods have to behave (mostly) the same. So `Neos.Neos/Resources/Public/JavaScript/Services/Localization.js` would need an adjustment too, and also the PHP translation methods. 
I translate NodeTypes properties, labels, etc. in PHP in some of my plugins and this would cause unexpected behaviour if the translation works in one place but not the other.

packages/neos-ui-i18n/src/registry/I18nRegistry.js
sbruggmann: True, I've simplified the the JavaSscript now.
But I would keep the `translateOne` for a simpler separation and better readability.

sbruggmann:
@Sebobo Regarding the PHP translation I've tought about too. I could add it in https://github.com/neos/flow-development-collection/blob/8.3/Neos.Flow/Classes/I18n/Translator.php#L126 afterwords too.

Use Case:

Given Vendor.Package:Main:defaults.firstname = "Firstname " and Vendor.Package:Main:defaults.required = "required"

Simple:
`Vendor.Package:Main:defaults.firstname` is used multiple times but in some cases we need to name it `Vendor.Package:Main:defaults.firstname Vendor.Package:Main:defaults.required` instead of creating a new translation and use `Vendor.Package:Main:defaults.firstname.required`.

Advanced with a custom NamespacedMixinsPostprocessor:
Dynamic property creation where the same base property (ex. "firstname") has the existing translation `Vendor.Package:Main:defaults.firstname` and a clone of it can modify the existing translation with `%s (Vendor.Package:Main:defaults.required)` which does get `Vendor.Package:Main:defaults.firstname (Vendor.Package:Main:defaults.required)` and is translated to "Firstname (required)".

I could demo it to you if you're interrested ;-)

packages/neos-ui-i18n/src/registry/I18nRegistry.js
```diff
@@ -84,6 +84,26 @@ export default class I18nRegistry extends SynchronousRegistry {
...
+
+        if (translation !== idOrig) {
+            return translation;
+        }
+
+        return this.translateOne(idOrig, fallbackOrig, params, packageKeyOrig, sourceNameOrig, quantity);
+    }
+
+    // eslint-disable-next-line max-params
+    translateOne(idOrig, fallbackOrig, params = {}, packageKeyOrig = 'Neos.Neos', sourceNameOrig = 'Main', quantity = 0) {
```
mhsdesign: hi ;) this should become a private method

mhsdesign:
Does our php translator support multiple labels? If not id like to have this feature first introduced and discussed in php land. As our ui code base is often just a copy of the php codebase and principles.

sbruggmann:
I'll do a PR for Flow over Easter.
As the standard functionality does stay the same, i'd say the change is just opening up for advanced use and can go further than "all other".

btw. here's an example where we generate nested NodeType labels: https://github.com/jvm-tech/JvMTECH.SelectiveMixins#example-a The only thing missing is translation support.

sbruggmann:
To reflect "multiple shorthand identifiers in one translation id string" in the backend part, I've introduced the following PRs:
- Flow: https://github.com/neos/flow-development-collection/pull/3014
- Neos: https://github.com/neos/neos-development-collection/pull/4180

markusguenther:
Thanks for all your efforts @sbruggmann, but sadly, we will not make it into 8.3.
So, a nice feature for the next version 💙

mhsdesign (changes requested):
With the translation refactoring https://github.com/neos/neos-ui/pull/3804 this change is now incompatible now and also its hard to maintain this kind of api where one translate call accepts multiple identifiers and more static template strings. The behaviour would already be undefined considering parameter substitution or fallbacks which are provided to the `translate()` api. I hope its okay with you as ill close this pr. The general feature can of-course be further discussed but its impossible to implement this correctly at the core of the translation function.

PR #3327, opened 2023-01-10
BUGFIX: Show relogin dialog when remote action results in redirect to login page

### LINKED ISSUE #3320: BUG: Re-Login Dialog does not appear when sessions are destroyed
### Description

After running `./flow session:destroyall` the UI does not show the Re-Login Dialog on remote actions, but instead shows a cryptic error message. The user needs to reload the page to be able to log in again.

### Steps to Reproduce

1. Change some content, so the publish button lights up
2. Run `./flow session:destroyall`
3. Click on the publish button
4. See error message appear

#### Expected behavior

After running `./flow session:destroyall` (and on login timeouts of which I don't know whether they're affected as well), the Neos UI shows the Re-Login Dialog, so the user is able to log in on the fly.

#### Actual behavior

https://user-images.githubusercontent.com/2522299/210363764-9adb70f8-6ec7-4fd1-a185-3e1abfaf6e04.mp4

### Affected Versions

Neos: 7.3

UI: 7.3

grebaldi:
solves: #3320 

**The Problem**

image

When the user is not authenticated, a remote action like "publish" answers with a `303` Redirect. The `fetchWithErrorHandling` mechanism on the client side however expects a `401` response (which is a reasonable expectation, semantically speaking).

**The Solution**

At first I thought it'd be best to make the UI API respond with correct status codes, but then I figured that the `303` redirect is actually more desirable in case someone hits the URL directly in the browser. 

So I adjusted `fetchWithErrorHandling`, so that it detects a redirect to the login page and enters the relogin-dialog routine as usual.

The way this works is by looking into the fetch response: If it has been redirected and the target URL ends in `"login"`, `fetchWithErrorHandling` assumes that this has been a redirect to the Neos login page. This is not quite safe, because in theory the redirect could have been configured differently. But I'm out of ideas as to how a more correct detection mechanism should look like in this case, and as a bugfix this should suffice to fix the problem for >99% of all setups.

### DIFF
--- a/packages/neos-ui-backend-connector/src/FetchWithErrorHandling/index.ts
+++ b/packages/neos-ui-backend-connector/src/FetchWithErrorHandling/index.ts
@@ -6,6 +6,20 @@ interface RequestQueueItem {
     reject: (reason?: any) => void;
 }
 
+const isAuthenticationRequiredAsPerResponse = (response: Response) => {
+    if (response.status === 401) {
+        // The response says explicitly that the request is unauthorized
+        return true;
+    }
+
+    if (response.redirected) {
+        // The request has been redirected to the login page
+        return response.url.endsWith('login');
+    }
+
+    return false;
+};
+
 class FetchWithErrorHandling {
     /**
      * The current CSRF token being used.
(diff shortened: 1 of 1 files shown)


Sebobo (approved):
Thx, works as described!

jonnitto (approved):

PR #3328, opened 2023-01-10
FEATURE: Apply link input value immediately after search term was cleared

### LINKED ISSUE #3311: UX: Once a link is set via the `LinkEditor`, it's confusing how to delete it
### Description

https://user-images.githubusercontent.com/2522299/209475564-2cb43f20-e4f3-4bc1-9f26-12877005619e.mp4

To delete a link in the `LinkEditor`, you have to first click on "delete" and then click on "apply". This is confusing, because the second action is non-obvious and also makes little sense, because there's no alternative to it.

I'd like to suggest to get rid of the second action and apply the deletion after the first click.

(An alternative would be of course to add another action, so the deletion can be reverted)

### Steps to Reproduce

1. Find yourself a LinkEditor and set a link
2. Apply the change
3. Reload
4. Try to delete the link

#### Expected behavior

A click on "delete" suffices to delete the link.

### Affected Versions

Neos: 8.2

UI: 8.2

grebaldi:
solves: #3311

When the search term gets cleared, the change is registered immediately:

[Peek 2023-01-10 16-45 - Link Editor after.webm](https://user-images.githubusercontent.com/2522299/211600025-44ea4d19-680f-485f-af17-da4319f0c7a8.webm)

### DIFF
--- a/packages/neos-ui-editors/src/Library/LinkInput.js
+++ b/packages/neos-ui-editors/src/Library/LinkInput.js
@@ -190,6 +190,12 @@ export default class LinkInput extends PureComponent {
                     }
                 });
         } else {
+            if (this.state.searchTerm) {
+                // The search term has been emptied. To avoid confusion, the emptied
+                // search term gets immediately applied.
+                this.props.onLinkChange(searchTerm);
+            }
+
             this.setState({
                 isLoading: false
             });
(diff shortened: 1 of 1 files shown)


Sebobo (approved):

Sebobo:
Thx, hated the old behaviour 😉

jonnitto (approved):

crydotsnake (approved):
Tested it and works as expected! Thanks for this improvement

PR #3334, opened 2023-01-16
BUGFIX: Hide inline toolbar if current editPreviewMode is not an edit mode

### LINKED ISSUE #3144: Floating inline toolbar in preview mode (desktop/tablet ...)
### Steps to Reproduce

1. Select a content element node (in edit mode) 
2. Click on the desktop-preview

#### Expected behavior

View the page in desktop-mode without Ui overlays

#### Actual behavior

image

The inline toolbar is still visible - and even functional (allows to create a new node, and one can even edit this node ^^)

### Affected Versions

UI version: v7.0.3

grebaldi:
fixes: #3144 

**The Problem**

The `<ContentCanvas/>` did not distinguish between edit- and preview-modes when rendering the `<InlineUI/>` component.

This lead to the `<InlineUI/>`  still being present, even though a preview-mode (like "desktop") had been selected.

**The Solution**

I added a condition to the `render` method of the `<ContentCanvas/>` component. If the currently selected `editPreviewMode` is not an edit-mode, the `<InlineUI/>` component will not be rendered any longer.

### DIFF
--- a/packages/neos-ui/src/Containers/ContentCanvas/index.js
+++ b/packages/neos-ui/src/Containers/ContentCanvas/index.js
@@ -87,6 +87,10 @@ export default class ContentCanvas extends PureComponent {
         });
         const InlineUI = guestFrameRegistry.get('InlineUIComponent');
         const currentEditPreviewModeConfiguration = editPreviewModes[currentEditPreviewMode] || editPreviewModes[Object.keys(editPreviewModes)[0]];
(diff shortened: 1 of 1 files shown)


crydotsnake (approved):
Thanks!
Works as expected

markusguenther (approved):
Thanks for all the fixes @grebaldi 💙

PR #3336, opened 2023-01-16
BUGFIX: Colocate react-ui-components esm-build declarations and sources

### LINKED ISSUE #3329: ESM build of react-ui-components source files in wrong place
### Description

Since Neos.UI 7.3.6 or even earlier the source are in a different folder than their declaration files from the react-ui-components ESM build, making it hard to use.
Only the d.ts files are contained are in the correct place.

#### Expected behavior

Source files are present and can be imported

#### Actual behavior

Source files are missing and import is not possible

### Affected Versions

Neos: 7+

UI: 7.3.6 or earlier. 5.3.13 is still working

Sebobo:
The relative import in the selectbox component forced typescript to create a different file structure for the source files than for the declaration files causing issues with importing the esm build.

By using the package style import the folder structure is correct again and by defining the rootDir in the tsconfig file an error will be thrown if a relative include outside of the components package is used.

Resolves: #3329

### DIFF
--- a/packages/react-ui-components/src/SelectBox/selectBox.js
+++ b/packages/react-ui-components/src/SelectBox/selectBox.js
@@ -5,7 +5,7 @@ import {$get} from 'plow-js';
 import SelectBox_Option_SingleLine from '../SelectBox_Option_SingleLine';
 import mergeClassNames from 'classnames';
 import isEqual from 'lodash.isequal';
-import isNil from '../../../utils-helpers/src/isNil';
+import {isNil} from '@neos-project/utils-helpers';
 
 // TODO: document component usage && check code in detail
 export default class SelectBox extends PureComponent {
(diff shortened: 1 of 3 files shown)


mhsdesign:
Fine by reading ;)

mhsdesign (approved):

packages/react-ui-components/src/SelectBox/selectBox.js
```diff
@@ -5,7 +5,7 @@ import {$get} from 'plow-js';
 import SelectBox_Option_SingleLine from '../SelectBox_Option_SingleLine';
 import mergeClassNames from 'classnames';
 import isEqual from 'lodash.isequal';
-import isNil from '../../../utils-helpers/src/isNil';
+import {isNil} from '@neos-project/utils-helpers';
```
grebaldi, line 8: The package manifest of `@neos-project/react-ui-components` is missing this dependency.

grebaldi (changes requested):
Hi @Sebobo,

I was able to reproduce the issue and can confirm that your change fixes it. There's a small remaining issue though (see code comment).

packages/react-ui-components/src/SelectBox/selectBox.js
Sebobo: Thx!

grebaldi (approved):

PR #3337, opened 2023-01-18
BUGFIX: Reset focused fusion path to `null` whenever the edit-/preview-mode changes

### LINKED ISSUE #3335: BUG: Inline Toolbar disappears when switching between edit-modes
### Description

When switching between edit-modes after focusing a content element in the guest frame, the Inline Toolbar disappears.

*Interestingly though, the Inline Toolbar re-appears when switching back to the original edit-mode.*

### Steps to Reproduce

1. Select a content element in the guest frame while in "In-Place" edit mode
2. Switch to "Raw Content" edit-mode
3. See the Inline Toolbar disappear

#### Expected behavior

The Inline Toolbar should attach itself to the focused content element when switching between edit modes.

#### Actual behavior

[Peek 2023-01-16 17-04 - Inline Toolbar vanishes.webm](https://user-images.githubusercontent.com/2522299/212725187-0ee407aa-7050-44a0-bc62-0f9668704e91.webm)

### Affected Versions

Neos: 7.3

UI: 7.3

grebaldi:
fixes: #3335 

**The Problem**

When focusing a node inside the `<ContentCanvas/>`, the associated `fusionPath` of that content element will be saved to the redux store. When the current edit-/preview-mode is then changed, said `fusionPath` is no longer valid (because the content element will be rendered through an entirely different path).

This confuses the `<InlineUI/>` component. It uses the stored `fusionPath` to locate the DOM node inside the `<ContentCanvas/>` that it attaches itself to. Since the `fusionPath` is invalid after the switch, the DOM node can longer be found and the `<InlineUI/>` becomes invisible.

**The solution**

The code responsible for locating nodes inside the `<ContentCanvas/>` can be found here:
https://github.com/neos/neos-ui/blob/b424a7adb4bb1844063c632b7c1d7e29a37c8bd3/packages/neos-ui-guest-frame/src/dom.js#L73-L77

It anticipates situations in which no `fusionPath` is set (which is the case when selecting a node in the `<ContentTree/>` component for example). Basically, if there's no `fusionPath` it selects the first node that matches the focused `contextPath`, which I found quite acceptable for switching between edit-/preview-modes.

Therefore, I altered the `CR.Nodes` reducer to set the focused `fusionPath` to `null` whenever the `UI.EditPreviewMode.SET` action occurs. With this, the  `<InlineUI/>` component no longer disappears when the edit-/preview-mode is switched.

### DIFF
--- a/packages/neos-ui-redux-store/src/CR/Nodes/index.ts
+++ b/packages/neos-ui-redux-store/src/CR/Nodes/index.ts
@@ -2,6 +2,7 @@ import produce from 'immer';
 import {defaultsDeep} from 'lodash';
 import {action as createAction, ActionType} from 'typesafe-actions';
 import {actionTypes as system, InitAction} from '@neos-project/neos-ui-redux-store/src/System';
+import {actionTypes as editPreviewMode, Action as EditPreviewModeAction} from '@neos-project/neos-ui-redux-store/src/UI/EditPreviewMode';
 
 import * as selectors from './selectors';
 import {calculateNewFocusedNodes, getNodeOrThrow} from './helpers';
@@ -391,7 +392,7 @@ const moveNodeInState = (
 //
 // Export the reducer
 //
-export const reducer = (state: State = defaultState, action: InitAction | Action) => produce(state, draft => {
+export const reducer = (state: State = defaultState, action: InitAction | EditPreviewModeAction | Action) => produce(state, draft => {
     switch (action.type) {
         case system.INIT: {
             draft.byContextPath = action.payload.cr.nodes.byContextPath;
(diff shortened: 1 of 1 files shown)


grebaldi:
Hmm... I do not understand why the E2E tests fail (but they do so consistently). From the log I take that moving multiple nodes in the Document Tree via Drag and Drop leads to an unexpected state, but I cannot reproduce the issue.

grebaldi:
Well... The nth run did pass :sweat_smile:

crydotsnake (approved):
Thx! Works as expected 🙂

markusguenther (approved):
Thanks @grebaldi

PR #3338, opened 2023-01-18
BUGFIX: Distribute height between inspector tabs header and content dynamically

### LINKED ISSUE #2683: The last element in a SelectBox cannot be reached in the Inspector in certain constellations
### Description
The last element in a SelectBox disappears behind the inspector action buttons and thus becomes inaccessible.

### Steps to Reproduce
1. Inspector-TabNavigation needs to be 2 lines
2. DropDown is last element
3. Browser window height is smaller than the Inspector tab panel
4. Open dropdown and scroll down to select the last element

#### Expected behavior
The last selection option can be reached without problems.
image

#### Actual behavior
- Browser height ok, all is visible and reachable
image

- Browser height to small
image

### Affected Versions
Neos: 4.3.11 and previous
UI: 3.8.1 and previous

### Simple bugfix could be:
`.style__tabs__content___OZ4nq {
    // height: calc(100% - 141px);
    height: calc(100% - 184px);
    ...
}`

or 
`.style__inspectorTabPanel___2z7Yl {
    ...
    // height: 100%;
    height: calc(100% - 43px);
    ...
}`

grebaldi:
fixes: #2683 

**The Problem**

The `<Inspector/>`-styles allocate height between the tabs header and the tabs content based on the assumption that there's only one tab header row. As soon as the tabs header breaks into a second row, the tabs content underflows the action buttons at the bottom of the `<Inspector/>` component.

**The Solution**

I've adjusted the `<Inspector/>`-styles using `display: flex` to distribute the height between the tabs header and the tabs content dynamically. This way, all of the tabs contents remain reachable, even if the tabs header breaks into a third row:

[Peek 2023-01-18 16-15 - Inspector spacing.webm](https://user-images.githubusercontent.com/2522299/213212353-8881b671-fb3e-4bc8-900f-80057fe75cea.webm)

### DIFF
--- a/packages/neos-ui/src/Containers/RightSideBar/Inspector/style.css
+++ b/packages/neos-ui/src/Containers/RightSideBar/Inspector/style.css
@@ -1,4 +1,6 @@
 .inspector {
+    display: flex;
+    flex-direction: column;
     height: 100%;
     width: 100%;
     overflow: hidden;
@@ -24,6 +26,9 @@
     }
 }
 .tabs {
+    display: flex;
+    flex-direction: column;
+    height: calc(100% - 101px); /* Accounting for <SelectedElement/> */
     margin-left: var(--spacing-Quarter);
     border-left: 1px solid var(--colors-ContrastDark);
     background: var(--colors-ContrastDarker);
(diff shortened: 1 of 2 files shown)


markusguenther (approved):
You are on fire at the moment ;)
Thank you 👏

crydotsnake (approved):
Tested and works as expected!
Thanks for taking care

PR #3341, opened 2023-01-25
BUGFIX: Stop click event propagation on selected element in select box header

### LINKED ISSUE #3057: AssetEditor: Cannot replace the Asset on clicking the Asset's title

grebaldi:
fixes: #3057 

**The Problem**

Hard to pin down, but I'll try: The issue described in #3057 relates to the `AssetEditor` which uses the `<SelectBox/>` component for input. The `<SelectBox/>`  component uses the `<DropDown/>` component for rendering. The `<DropDown/>` component understands a click on its header as a request to toggle its `expanded` state. When `expanded === true`, the `<SelectBox/>`  changes its header into a search input, so the user can search through the available select options.

Now, the `AssetEditor` kind of misuses the `<SelectBox/>` for its purpose. It narrows the available select options to the currently selected asset, so the `<SelectBox/>` displays the latter properly. This however means that the behavior described above leads to a search input, but only one available select option: the currently selected one. (Side note: On input, the search is still being performed, the `<SelectBox/>` simply does not show the resulting options)

**The Solution**

I've tried to correct the behavior of the `AssetEditor`, so that it shows the search results, but with no luck.

So, I ended up implementing the suggestion from #3057. Through `stopPropagation`, a click on the currently selected option won't toggle the underlying `<DropDown/>` component any longer.

### DIFF
--- a/packages/neos-ui-editors/src/Editors/AssetEditor/index.js
+++ b/packages/neos-ui-editors/src/Editors/AssetEditor/index.js
@@ -243,6 +243,7 @@ export default class AssetEditor extends PureComponent {
                 placeholder={this.props.i18nRegistry.translate(this.props.placeholder)}
                 options={this.props.value ? this.state.options : this.state.searchOptions}
                 value={this.getValue()}
+                onHeaderClick={() => { /* prevent toggling of select box dropdown */ }}
                 onValueChange={this.handleValueChange}
                 displayLoadingIndicator={this.state.isLoading}
                 showDropDownToggle={false}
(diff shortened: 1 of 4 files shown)


Sebobo (approved):
Great, thx!

PR #3344, opened 2023-01-26
FEATURE: add some ClientEval functionality in nodeCreationDialog

FahimNsr:
**What I did**

I bring some ClientEval functionality into nodeCreationDialog.

**How I did it**

Because in the nodeCreationDialog step, the node doesn't create yet I build an artificial Node Object in the State of nodeCreationDialog Modal, which helps us to get the value of each node.properties and then I used the  `preprocessViewConfiguration` function that's almost similar to the function that we used in Inspector, to set the right value and configuration in nodeCreationDialog.

**How to verify it**

I upload a video to help to better understand.

https://user-images.githubusercontent.com/52664771/216024761-464b6bea-cc69-4f03-a5af-382cf861d771.mov

### DIFF
--- a/packages/neos-ui/src/Containers/Modals/NodeCreationDialog/index.js
+++ b/packages/neos-ui/src/Containers/Modals/NodeCreationDialog/index.js
@@ -4,6 +4,7 @@ import {connect} from 'react-redux';
 import {$get, $set} from 'plow-js';
 import memoize from 'lodash.memoize';
 import cx from 'classnames';
+import {produce} from 'immer';
 
 import {neos} from '@neos-project/neos-ui-decorators';
 import {actions} from '@neos-project/neos-ui-redux-store';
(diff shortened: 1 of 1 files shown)


JamesAlias:
This has actually been worked on by the community in the Neos Coding Live Stream 😅 

see #3304 

I'm in the process of cleaning our MR up and then it would be ready to merge as well.

FahimNsr:
> This has actually been worked on by the community in the Neos Coding Live Stream 😅
> 
> see #3304
> 
> I'm in the process of cleaning our MR up and then it would be ready to merge as well.

Hi @JamesAlias ,

I saw your PR, and that was very nice that you add the dependency on the Parent node.

I would to mention that as you can see in the video I fixed the hidden property for NodeCreationDialog in my PR with some configuration in 2 different ways.

1. we can get the hidden property of Inspector and set the same value for NodeCreationDialog in this File. (lines 123-125)

> Packages/Neos/Neos.Neos/Classes/NodeTypePostprocessor/CreationDialogPostprocessor.php


Screenshot 2023-03-02 at 09 59 03


or

2. we can set this config in yaml for the _email_ property (as you see in the Video) (lines 8-13)

Screenshot 2023-03-02 at 09 57 20


@Sebobo  @grebaldi, I would to know your opinion too about this hidden Property.

mhsdesign:
Hi @FahimNsr we fixed https://github.com/neos/neos-ui/issues/3483 (that hidden doesnt work in the node creation dialog)

now with https://github.com/neos/neos-ui/pull/3507 and https://github.com/neos/neos-development-collection/pull/4297

so i think we can close your pr?

Thanks for your effort!

PR #3345, opened 2023-01-26
BUGFIX: Reset/Clear `<SelectNodeType/>`-dialog on apply

### LINKED ISSUE #3034: Clear node filter in node insert dialog when re-opening
### Description

If you filtered for a node while creating a node and re-open the node insert dialog, then the filter is still »active« and pre-filled with the values you just used.

### Steps to Reproduce

1. Open the insert node dialog
2. Filter for a node
3. Insert it
4. Re-open the dialog

#### Expected behavior

The filter input field is empty, nothing is filtered

#### Actual behavior

The filter input field holds the last value and is, therefore, applying the filter.

grebaldi:
fixes: #3034 

**The Problem**

The `<SelectNodeType/>`-dialog uses internal state to manage the filter search term (and toggling of the help message display for node types, which the aforementioned issue does not cover).

When the user presses `cancel`, said state gets cleared, so the dialog is in its initial state the next time it gets opened. 

When the user presses `apply`, the state remains untouched, so when the dialog gets opened again, the filter search term and a possibly open help message are still there.

This can be confusing, especially if the dialog gets opened in a different context, in which both items would lead to different results. 

For example, if the user tries to insert a document after just having inserted a content element (with a ndoe type filter search term applied), the `<SelectNodeType/>`-dialog may indicate that there's no document node type available (due to the still applied search term):
image

Likewise, the dialog may show a help message for a node type that isn't actually available:
image

**The Solution**

I ensured that both, the `cancel` and `apply` action lead to the same cleanup routine for the dialog (which means, both filter search term and active help message are being reset).

This solution has a couple of drawbacks however:

- When inserting lots of content nodes of the same type in sequence, some users may actually prefer the current behavior
- If there's a node creation dialog involved, clicking on `back` would appear to reset the  `<SelectNodeType/>`-dialog, which might be confusing to some

### DIFF
--- a/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
+++ b/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
@@ -115,10 +115,8 @@ export default class SelectNodeType extends PureComponent {
 
     handleCancel = () => {
         const {cancel} = this.props;
-        this.setState({
-            filterSearchTerm: ''
-        });
-        this.handleCloseHelpMessage();
+
+        this.handleClose();
         cancel();
     }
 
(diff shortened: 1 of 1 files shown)


Sebobo (approved):

Sebobo:
Thx! Looks good by reading

crydotsnake (approved):
Tested, and works as expected!

PR #3351, opened 2023-01-31
!!! TASK: Remove legacy editor name support

Sebobo:
This is breaking as old style editor paths will not be mapped anymore and would stop working with this change.

These were deprecated since Neos 4.3.

### DIFF
--- a/packages/neos-ui-editors/src/EditorEnvelope/index.js
+++ b/packages/neos-ui-editors/src/EditorEnvelope/index.js
@@ -53,9 +53,7 @@ export default class EditorEnvelope extends PureComponent {
 
     getEditorDefinition() {
         const {editor, editorRegistry} = this.props;
-        // Support legacy editor definitions
-        const editorName = editor.replace('Content/Inspector/Editors', 'Neos.Neos/Inspector/Editors');
-        return editorRegistry.get(editorName);
+        return editorRegistry.get(editor);
     }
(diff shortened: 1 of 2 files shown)


Sebobo:
A code migration should be added for this

ahaeslich:
@Sebobo would this be a migration we could include here: [neos/rector](https://github.com/neos/rector)?

Sebobo:
Uhm I don't know. I would have put it in here.

ahaeslich:
> Uhm I don't know. I would have put it in here.

It may be an additional help for all projects that accidentially use the legacy editor name in their yaml files. I just found some occurences it in one of my projects thanks to your PR neos/neos-development-collection#4023 🙈 

But then again I'm not sure if we will touch yaml files with the rector package.

markusguenther (approved):
Good catch :)
Thanks for the PR

markusguenther:
@ahaeslich Do we in 9.0 only use rector by now? Or also the 'old' migrations???

mhsdesign (changes requested):
we should imo use our current migration framework and work with simple string replace or sth across the yaml files and be done with it ;)

mhsdesign (changes requested):
we should imo use our current migration framework and work with simple string replace or sth across the yaml files and be done with it ;)

PR #3349, opened 2023-01-29
BUGFIX: Remove initiatingUserId from createNode command

ahaeslich:
If you add a content element to the page the following exception is thrown:

```
Exception in line 107 of 
  Packages/Application/Neos.Neos.Ui/Classes/Domain/Model/Changes/AbstractCreate.php: 
  Neos\ContentRepository\Core\Feature\NodeCreation\Command\CreateNodeAggregateWithNode::__construct(): 
  Argument #5 ($parentNodeAggregateId) must be of type Neos\ContentRepository\Core\SharedModel\Node\NodeAggregateId, 
  Neos\ContentRepository\Core\SharedModel\User\UserId given
```

See: https://neos-project.slack.com/archives/C3MCBK6S2/p1674926113660949 

**How to verify it**

Try to add a document or content node in a Neos 9 installation.

Relates: neos/neos-development-collection@4c9dd087324317c8b2823b2ee4a658b23d638892

PR #3352, opened 2023-01-31
BUGFIX: Remove initiatingUserId from node commands

ahaeslich:
- checked for all usages of `getInitiatingUserIdentifier` to remove the initiatingUserId from node commands
- removed function `getInitiatingUserIdentifier` because it is not needed anymore
- removed unused imported classes in touched files

Relates: neos/neos-development-collection@4c9dd087324317c8b2823b2ee4a658b23d638892
Relates: #3349


skurfuerst:
Thanks so much 😍

Issue #3357, opened 2023-02-11
BUG: spacing issues in some parts of the ui


crydotsnake:
### Description

I noticed that some distances in the neos-ui are no longer as they used to be. I think one of the last changes broke it. I haven't been able to find out which change caused this bug.

Here are a few comparisons

### demo.neos.io

![demoneosio](https://user-images.githubusercontent.com/39345336/218260043-8fd8eebb-8282-4047-b574-74e3fce22f54.png)

![demoneosio1](https://user-images.githubusercontent.com/39345336/218260047-81a500e9-4f5c-4abd-920c-0d8ad9793005.png)

![demoneosio2](https://user-images.githubusercontent.com/39345336/218260488-48cd4149-ab75-4ce6-ae94-56bdfc77e912.png)

![demoneosio3](https://user-images.githubusercontent.com/39345336/218272032-93232c01-2176-495f-93b0-b4a293534ed8.png)

### 8.2.x-dev

![local](https://user-images.githubusercontent.com/39345336/218260056-6d99ecea-d2b6-48a2-9e5e-acd8bfd3e617.png)

![local1](https://user-images.githubusercontent.com/39345336/218260059-c2ebd577-482d-4c96-baa0-f7b63e223f47.png)

![local2](https://user-images.githubusercontent.com/39345336/218260497-1e10dbfa-e5d9-43ef-9864-00f51d8e6a1d.png)

![local3](https://user-images.githubusercontent.com/39345336/218272035-7694e0d8-cd30-4c83-8c52-51b4db7a8426.png)

### Affected Versions

Neos:
8.2.x-dev

UI:
8.2.x-dev

You can get this information by running `composer show` or using the package management module within Neos.
The UI version can also be found inside the left sidebar after version 1.0.3.

Sebobo:
Can't really reproduce with the 8.2 branch. 
Did you rebuild, and if yes, did you do `make build` or `make build-production` and what is your browser, etc.

crydotsnake:
I used `make setup` which should also build the files. And my browser is chrome.

Sebobo:
Can this be closed now?

markusguenther:
We fixed this and if we find another parts with issues, feel free to reopen ;)
Thanks for reporting @crydotsnake

crydotsnake:
Hm. I updated my `8.3` branch to the latest changes, but it is still broken.
After i pulled the latest changes, i used `make clean && make setup` too have a fresh base.

[...]

PR #3358, opened 2023-02-11
Task: Back to css module composes and remove sass

mhsdesign:
With the esbuild change after lots of headaches me and @JamesAlias had to remove the "css modules composes" feature from the codebase and introduces another dependency: sass with `@extends` - this was at that time the only option to prevent the esbuild approach from failing.

But there are a few downsides to the sass `@extends`:
- code duplication in the bundle 82.8kb vs 116.1kb (but this is more theoretically ;))
- another dependeny: sass for such a small purpose -> this will make the react-ui-components standalone package harder to build for external people as they now need sass too (besides webpack, postcss and a lot of postcss plugins)
- i have the feeling not using composes but `@extends` messed with the specificity in certain places like here: https://github.com/neos/neos-ui/issues/3357 still to investigate if this would help.

Of course there are also lots of downsides to css modules `composes` for example all the problems we ran into until i could carefully hack it to work ;) But i think it is the right way, until we reconsider sometime completely our css style handling.

-------

It might be worth a though to merge this into 8.2, so we dont have to deal with sass in any branch ;) otherwise 8.2 will be special ^^

---

btw all this "post css module composes" thing plus esbuild was sooooo hard to figure out and i spend so much time of digging into the css modules source code and trying to think of a way to make it work ^^ But here we are finally and it should work now like in the webpack build ;)

todo:
- [ ] remove scss linting
- [ ] fix ci  codestyle
- [ ] dev build has duplicated reset classes that are confusion in the dev tools. -> solution: build custom css modules plugin that fixes this fx. use lightning css


mhsdesign:
superseded by https://github.com/neos/neos-ui/pull/3366

PR #3306, opened 2022-12-22
!!!TASK: Remove plow-js

grebaldi:
solves: #3118 

I'm moving from package to package. Until the last package is finished, this remains a WIP.

## TODOs

- [x] Remove plow-js from `@neos-project/neos-ui-sagas`
- [x] Remove plow-js from `@neos-project/neos-ui`
  -  [ ] This package registers `plow-js` as a runtime dependency for plugins. Removing plow from here will break every plugin that uses `plow-js` itself. <s>What do we do about that?</s> I think it's feasible to offer plugin vendors support with this. I con open PRs for the most common plugins that use plow in this manner. Everyone else needs to be warned in advance, while 9.0 needs to be released with proper instructions. If removing plow is not feasible, Plugin vendors basically only need to add `plow-js` as an explicit dependency and rebuild. 
     - [x] `Yoast.YoastSeoForNeos`: https://github.com/Yoast/Yoast-SEO-for-Neos/pull/92
     - [ ] `Psmb.FlatNav`: https://github.com/psmb/Psmb.FlatNav
     - [ ] `Psmb.SplitAdd`: https://github.com/psmb/Psmb.SplitAdd
     - [ ] `Breadlesscode.SimpleEditorExtend`: https://github.com/breadlesscode/neos-simple-editor-extend
     - [ ] `Visol.Neos.LinkClass`: https://github.com/visol/Visol.Neos.LinkClass
  - [ ] Don't know how to test node tree presets
  - [x] <s>User impersonation needs testing</s>
    - works properly
  - [x] `ClientEval:` in Inspector needs testing: https://github.com/neos/neos-ui/blob/1f4766ca77d63b4dd6a3f5f52e23db24191899e5/packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js#L155-L160 (This is a bit funny, tbh :slightly_smiling_face: )
    - This is fixed in #3304 
- [x] Remove plow-js from `@neos-project/neos-ui-backend-connector`
- [x] Remove plow-js from `@neos-project/neos-ui-ckeditor5-bindings`
  - [x] Tables aren't working atm (unrelated to this PR) -> this needs more testing once they work again (see: https://github.com/neos/neos-ui/pull/3321)
- [x] Remove plow-js from `@neos-project/neos-ui-contentrepository`
- [x] Remove plow-js from `@neos-project/neos-ui-editors`
[...]

PR #3364, opened 2023-02-14
BUGFIX: Lazy-Load focused node, if it's not fully loaded yet

### LINKED ISSUE #3219: Bug: `Selected element` select a parent node, which is a shortcut.
### Description

The inspector shows an infinite loading icon when a parent node was selected via `Selected element` which is a shortcut.

### Steps to Reproduce

This bug is not reproducible, when the shortcut was already loaded by the ui.

1. Have shortcut with a nested page (and publish them?)
2. Go to a page that is not the shortcut or a descendant of the shortcut
3. Reload the ui to make sure all JavaScript caches about nodes are flushed
4. expect the previous selected node to be still active
5. go to the subpage of the shortcut (without clicking the shortcut itself (gehe nicht über los ziehe keine 200 € ein usw.) )
6. use the `Selected element` dropdown to select the parent / eg the shortcut

[Bildschirmaufzeichnung vom 10.10.2022, 14:17:43.webm](https://user-images.githubusercontent.com/85400359/194864653-6d121aea-4cc3-4d4c-9ce9-14ce6a476a14.webm)

#### Expected behavior

The inspector should work as usually

#### Actual behavior

Infinitive loading to death

### What happened

the `<Inspector />` uses the `renderFallback` with spinner because `node.isFullyLoaded` is falsy or `state.viewConfiguration.tabs` is falsy 

https://github.com/neos/neos-ui/blob/7665b057e7f72fbd56026c106d3bab8528403469/packages/neos-ui/src/Containers/RightSideBar/Inspector/index.js#L344-L350

### Affected Versions

UI:  v8.0.1

grebaldi:
fixes: #3219 

**The Problem**

CR Node representations in the central Redux store have a flag `isFullyLoaded` that indicates whether the node has been loaded with all of its meta data. If the node has only been loaded for being displayed in the document- or content-tree, this flag is absent, otherwise it is set to `true`.

The intention behind flag this was to lazy-load the remaining metadata should it become relevant in a specific interaction context. Apparently though, this mechanism has never been implemented.

**The Solution**

I adjusted the Inspector-saga to perform the aforementioned lazy-load. This fixes #3219, but adds a `plow-js` call, which needs to be kept in mind for #3306.

### DIFF
--- a/packages/neos-ui-sagas/src/UI/Inspector/index.js
+++ b/packages/neos-ui-sagas/src/UI/Inspector/index.js
@@ -5,6 +5,8 @@ import {findNodeInGuestFrame} from '@neos-project/neos-ui-guest-frame/src/dom';
 
 import {actionTypes, actions, selectors} from '@neos-project/neos-ui-redux-store';
 
+import backend from '@neos-project/neos-ui-backend-connector';
+
 import {applySaveHooksForTransientValue} from '../../Changes/saveHooks';
 
 const getFocusedNode = selectors.CR.Nodes.focusedSelector;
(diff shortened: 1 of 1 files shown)


packages/neos-ui-sagas/src/UI/Inspector/index.js
```diff
@@ -38,6 +40,18 @@ export function * inspectorSaga({globalRegistry}) {
             //
             if (nextAction.type === actionTypes.CR.Nodes.FOCUS) {
                 yield put(actions.UI.Inspector.closeSecondaryInspector());
+
+                const focusedNode = yield select(getFocusedNode);
+                if (!$get('isFullyLoaded', focusedNode)) {
```
mhsdesign, line 45: cant we just use opt chaining in 7.3? ^^

grebaldi: we can't, sadly

mhsdesign: okay then thats fine with me ;)

markusguenther (approved):
Great that you tackled the issue. Could reproduce the bug and your solution works fine :)

PR #3365, opened 2023-02-14
TASK: Remove target branch in dependabot config

markusguenther:
We get non security pull requests from Dependabot and that in the wrong branch. After a chat with a GitHub product manager, the issue seems to be that the open-pull-requests-limit config does not work with target branch together.

So, we remove that now and hope that the docs and the pm are right, and we just get security PRs.

https://twitter.com/jhutchings0/status/1625254020188340224

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,6 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
-    target-branch: "7.3"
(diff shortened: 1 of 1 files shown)


mhsdesign (approved):
Lets try that ;)

mhsdesign:
Linting fails unrelated but will be fixed with https://github.com/neos/neos-ui/pull/3371

Issue #3330, opened 2023-01-14
BUG: Publish Dropdown has a large distance to the right


crydotsnake:
### Description

The publish dropdown has a padding to the right, which makes it doesn't really look nice because the chevron icon no longer looks centered.

### Neos 8.2
![SCR-20230114-sbx](https://user-images.githubusercontent.com/39345336/212492707-2c2553c1-d06b-41e3-b33b-bea7489918a1.png)

### Neos 7.3

![SCR-20230114-smn](https://user-images.githubusercontent.com/39345336/212493050-8c0c0981-f0ef-4aa0-b1d2-d866337903c0.png)

### Affected Versions

Neos:
8.2
UI:
8.2

crydotsnake:
This bug came up in #3282

mhsdesign:
Let’s see if it will be fixed by #3366 ;)

mhsdesign:
fixed in #3366

mhsdesign:
Bear in mind my fix will only be available for Neo 8.3 due to technical reasons.
But I figured that its okay since people will most likely use the LTS either way ;)

mhsdesign:
will close this as its fixed with https://github.com/neos/neos-ui/pull/3366

PR #3217, opened 2022-10-09
TASK: Script: Rename all `import style from "./style.css"` to `style.module.css`

mhsdesign:
Currently any build stack gets confused by our "normal" `style.css` files, as they are not named `style.module.css`.
Sure we can work against that but its clearly better to stick to the conventions

```
Renamed "108" files
Replaced "124" imports
```

FYI the build works perfectly ;)

### Show me the wizard who did this:

<details>

```php
<?php

/*
 * Script to rename all import style from "./style.css" to style.module.css
 *
 * !!! Operates on ./packages in the current __DIR__
 *
 * Build for linux - should work on Mac - don't even try it on windoofs
 */

function renameStrategy(string $filename, string $extension): string {
    return $filename . '.module.css';
}

(function () {
    die("enable me");

    $regex = <<<'REGEX'
    /import[ ]+[a-zA-Z0-9]+[ ]+from[ ]+["'](?P<imported>.*\.css)["']/
    REGEX;

    $fileIterator = recursiveIterator(__DIR__ . "/packages");

    /**
     * Indexed by realpath of 'from' to rename. Value is 'to' be renamed to.
     * @var array<string,string> $filesToRename
     */
    $filesToRename = [];

    $replacedImports = 0;

    foreach (iterateOverFilesAndSelectByFileEnding($fileIterator, [".js", ".ts"]) as $file) {
        $contents = file_get_contents($file);

        $currentFileDir = dirname($file);

        $newContents = preg_replace_callback($regex, function ($matches) use ($currentFileDir, $file, &$filesToRename, &$replacedImports) {
            $rawImported = $matches["imported"];

            if (str_starts_with($rawImported, "@neos-project/")) {
                // package import `@neos-project/neos-ui-guest-frame/src/style.css`
                $import = substr_replace($rawImported, __DIR__ . "/packages/", 0, \strlen("@neos-project/"));
            } else {
                // relative import `../../style.css`
                $import = $currentFileDir . '/' . $rawImported;
            }

            if (!file_exists($import)) {
                echo <<<MSG
                --------
                import '$rawImported';
                File $import does not exist or this simplified dependency resolution does not work.
                It was imported in $file
                --------
                MSG;
                return $matches[0];
            }

            [
                'basename' => $basename,
                'dirname' => $dirname,
                'extension' => $extension,
                'filename' => $filename
            ] = pathinfo($import);

            $newFileNameWithExtension = renameStrategy($filename, $extension);
[...]

Issue #3213, opened 2022-10-07
pre-build and pre-pack css & js of react-ui-components


JamesAlias:
### Description
Just for us to remind us about this possible issue.

The current way of distributing the `react-ui-components` requires a extensive build setup/tooling and exposes internal sources.
For example you have to build your css with post-css the exact way we do it in the neos-ui monorepo.

[Description of the bug or feature]

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

#### Expected behavior

[What you expected to happen]

#### Actual behavior

[What actually happened]

### Affected Versions

Neos:

UI:

You can get this information by running `composer show` or using the package management module within Neos.
The UI version can also be found inside the left sidebar after version 1.0.3.

mhsdesign:
My thought:

i find it odd to expose the postcss.rc and other internal things like the webpack.config

it would be great to prebuild the css modules and just include the css as normal import like:

```js
import style from "./style.module.css"

export const Button = () => (
    <button className={style.button} />
)
```

becomes

```js
import "./style-button.css"

const style = {button: "button21323"};

export const Button = () => (
    <button className={style.button} />
)
```

PR #3366, opened 2023-02-14
FEATURE: CSS build refactor (make react-ui-components easy to use)

mhsdesign:
solves https://github.com/neos/neos-ui/issues/3213
solves #3330

Puhh im happy to finally present you this pr.
I have been working on this concept since we started the esbuild pr but now everything just fell in place and ive got a fully working* improved css build and react-ui-components build.

*I currently cant find any style shifts in a side to side comparison :D

I took my previous refactoring at https://github.com/neos/neos-ui/pull/3358 even further, now with a better written cssModules plugin https://github.com/neos/neos-ui/blob/aa9f5497678225b8b9414b247f97042aaa165447/cssModules.js and an actual consumer (dev) friendly build of the `@neos-project/react-ui-components`

##  What I did

- **reduce the build speed by a third!!!** _esbuild + postcss + sass = 1500ms_ -> now **esbuild + lightningcss = 500ms**!!!
- replace sass `@extends` with css-modules `composes` https://github.com/css-modules/css-modules#composition
  - less code duplication in the css bundle 82.8kb vs 116.1kb :D
- replace post-css-modules with [lightningcss](https://lightningcss.dev/) (in Rust by parcel) as it offers an api that makes `composes from "./file.css"` usable as esbuild plugin: see: https://github.com/neos/neos-ui/blob/aa9f5497678225b8b9414b247f97042aaa165447/cssModules.js
  - the post-css-modules plugin offers no fully working way to build an integration with things like esbuild. The api is so cruel and unusable. In the webpack build it only worked by using a huge pile of shit - so im really happy that lightningcss offers a so simple api.
- ~Our`.reset` css modules compose mixin now uses the `:where()` selector, to force a low specificity at all time (just in case somehow the reset class lands at the bottom at the stylesheet, where it would have a higher specificity otherwise, which was our problem in the first place so we used sass. But after investigating if found out that since using my esbuild plugin the css class is as wanted at the top so it would be fine either way for now)~ edit i reverted to not using `:where` as we dont need it atm
- figure out to create a prepack build script for the `react-ui-components`
  - source maps
  - full build css (applied cssmodules, nesting and cssvariables)
  - typescript declarations
  - able to import only certain components
  - api surface should not change
- rewrote the css at some places to make it work with lightningcss
  - css grid template areas are hashed now and local to each css module, so we use composes from to use the template area in another css module: https://github.com/parcel-bundler/lightningcss/issues/144#issuecomment-1091821622
  -  use function syntax instead of `:global` pseudo class: https://lightningcss.dev/css-modules.html#unsupported-features
- make the build-essentials obsolete (the css variables reside now in top level as css file) 
  - infact the package is now only used by the webpack extensibility package
  - moved css var declaration from `packages/build-essentials/src/styles/styleConstants.js` to `cssVariables.css`
    - the file `cssVariables.js` creates a plugin for lightningcss which embeddes the css variables as usual
- renamed `packages/react-ui-components/src/index.js` to a typescript file so we have types when using `import {Button} from "index.js"`  

Todos
- [x] https://github.com/neos/neos-ui/pull/3217 (after this one ;))
- [x] remove any leftover .scss references like in linting
- [x] investigate and fix now occurring minor style issues in the neos ui
  - [x] maybe we need to revert some css styling changes again that happened due to the sass setup in 8.2 (i checked all ocurred css changes but none of them need to be reverted)
- [x] test the neos ui
- [x] figure out what to do with our legacy `packages/build-essentials/src/styles/styleConstants.js` -> it is currently used by the neos-ui-extensibility plugin build ... by accident?
  - the question is, whether its a usecase to use the neos ui css vars in a custom plugin (which currently seems to work)
  - also do we need the css vars in the backend modules anymore?: https://github.com/neos/neos-development-collection/blob/2790d168ab2bfd9ef0056e7c24a8e3c6d8e20169/Neos.Neos/Resources/Private/Styles/_CSSVariables.scss#L44
- [x] selectBox.js `import {isNil} from '@neos-project/utils-helpers';`: due to `isNil` beeing in `@neos-project/utils-helpers` and this package not beeing on npm we cant properly use the react-ui-components as declaring it a dependecy alone would solve it. Regression from https://github.com/neos/neos-ui/pull/3069. use value == null;
- [x] add comment for why `value == null`
- [x] ~extract css grid declaration in another better named css file~
- [x] https://github.com/mhsdesign/esbuild-composes-from-css-modules
  - [x] create an option for nesting and dont make it the default
  - [x] only import composes dependency once per file
  - [x] inline it instead
- [x] react-ui-component rewrite readme
- [x] react-ui-extensibility docs + demo (eg. how to get the font noto sans to work)
- [x] currently backendmodules like https://github.com/Flowpack/media-ui use the lib-esm build of the react-ui compoents and might use runtime css variables defined for each backend module: https://github.com/neos/neos-development-collection/blob/2790d168ab2bfd9ef0056e7c24a8e3c6d8e20169/Neos.Neos/Resources/Private/Styles/_CSSVariables.scss#L44 is there a problem when the lib-esm build comes now with hardcoded values? I dont think so...
- [x] rename lib-esm to dist and dont publish src files to npm ;)
- [x] clean up compileWithCssVariables
- [x] ~revert to [name]-[hash] css classes (for easier selecting in e2e (or place data-testid) )~
- [x] what if the media ui would use the new react-ui-components too? it would style crash as the css module names have the same names ... this is always problematic when two react-ui-components builds will be present ... -> solution modify prefix for react-ui-components
- [x] check date picker
- [x] https://github.com/neos/neos-ui/issues/3378
- [x] react-ui-components do we need to allow direct imports eg `@neos-project/react-ui-components/dist/Button/index.js` or even unthemend? `@neos-project/react-ui-components/dist/Button/button.js`
- [x] cleanup ts configs
- [x] Check on https://github.com/neos/neos-ui/pull/3282
- [x] Check on https://github.com/neos/neos-ui/issues/3330
- [x] Assert no extensions are used when importing

## How I did it

blood. sweat. tears.

## How to verify it

use the lib-esm build in the ui ^^

### react-ui-components build

see if it works correctly, as you have used before (by linking the package locally or some other hack)

or create a js and html file like:

```jsx
import React from "react"; 
import ReactDom from "react-dom";
import {Headline, Button, Label, IconButton} from "./lib-esm/index"

ReactDom.render(
    <div style={{background: "#2b2b2b"}}>
        <Headline type="h3">Hello</Headline>
        <Button>Click</Button>
        <Label>Hi</Label>
        <IconButton icon="neos">Hello</IconButton>
    </div>,
    document.getElementById("app"),
)

```
build it with `yarn esbuild test.jsx --outdir=dist --bundle` and view the html in the browser.

### Upgrade instructions for Neos Ui components [@neos-project/react-ui-components](https://github.com/neos/neos-ui/tree/8.3/packages/react-ui-components):

With this change the CSS build stack of the Neos Ui was modernized, due to this it became possible to make the package react-ui-components easier to use: Developers requiring the npm package will no longer need to worry about the required CSS preprocessing from the sources (lowering nesting and applying css-modules). A prepack script will build the css to native css and also take care of css modules. This means that for consumption of the components you will only need a simple bundler to load the vanilla css - nothing else. It is advised to read the new [readme](https://github.com/neos/neos-ui/tree/8.3/packages/react-ui-components) before updating so you can identify any other changes.


mhsdesign:
Is now completely ready for review ;)

See https://github.com/neos/neos-ui/blob/d8ab716a5753a00bec648cfd480bd0aa3e347abc/packages/react-ui-components/README.md for a new documentation of the react-ui-components ;)

markusguenther (approved):
Looks good to me, if we have bugs left, we need to fix them anyways.

PR #3368, opened 2023-02-15
BUGFIX: Adjust width of dropdown content

markusguenther:
The content of the user dropdown was limited in the width, and therefore it could happen that the space for the labels was not enough. Now the contents section can get bigger than the trigger button.

This was also an issue in 7.3 and not an issue related to the impersonation feature.
In the past, we fixed that only for the 8.1 and newer.

Thanks to @dlubitz for the report!

Fixes: #3235


dlubitz:
Tested. Thanks @markusguenther, works fine.

image

dlubitz (approved):

PR #3369, opened 2023-02-15
TASK: Add another way to run tests locally with a system under test in docker

JamesAlias:
**What I did**
I made it possible to start a neos test instance as _system under test_ (SUT) via docker. You are able to run testcafe tests on your machine to faster implement tests and features.

**How I did it**
Add a shell script to start the SUT.
Add a make command to do it fast.
Add some configs to make it possible.

We Mount the NeosUi root into the SUT.
Then we symlink the `[...]/Packages/Application/Neos.Ui` to this mounted volume.
We do the same for `[...]/Packages/Application/Neos.TestNodeTypes` -> `[...]/Tests/IntegrationTests/SharedNodeTypesPackage`.

This way we can build tests, features and bugfixes without the need to restart the SUT (in 99% of cases).
This speeds up the feedback loop by at least an order of magnitude. It's insane.

We don't even need to have a Neos Instance to develop the UI! I think this could be huge for new contributors.
Just start the SUT and you have a running Neos 🚀 

If you have to nuke it, just re-run the make command.

**Why I did it**
It's extremely slow to run tests on my local machine via docker because it completely wipes and reinstalls neos and all the things everytime I want to run 1 or 2 tests.
I decoupled the SUT and "running the tests". This makes writing tests and features so much faster for me.

**How to verify it**
- `make e2e-start-system-under-test`
- `yarn run testcafe chrome Tests/IntegrationTests/Fixtures -T 'sidebars'` - tests should not fail
- change something that breaks that feature (I just broke the LeftSideBar reducer)
- `make build-e2e-testing` - rebuild Ui
- `yarn run testcafe chrome Tests/IntegrationTests/Fixtures -T 'sidebars'` - tests should fail

### DIFF
--- a/Tests/IntegrationTests/start-system-under-test-docker.sh
+++ b/Tests/IntegrationTests/start-system-under-test-docker.sh
@@ -1,0 +1,80 @@
+#!/usr/bin/env bash
+
+set -e
+
+function dc() {
+    docker-compose -f ./Tests/IntegrationTests/docker-compose.system-under-test.yaml $@
+}
+
+echo "#############################################################################"
+echo "# Start docker environment...                                               #"
+echo "#############################################################################"
+dc down
+dc up -d
+dc exec -T php bash <<-'BASH'
+    rm -rf /usr/src/app/*
+BASH
+docker cp "$(pwd)"/Tests/IntegrationTests/. "$(dc ps -q php)":/usr/src/app
+sleep 2
+
+echo ""
+echo "#############################################################################"
+echo "# Install dependencies...                                                   #"
+echo "#############################################################################"
+dc exec -T php bash <<-'BASH'
+    cd /usr/src/app
+    sudo chown -R docker:docker .
+    sudo chown -R docker:docker /home/circleci/
+    cd TestDistribution
+    composer install
+BASH
+
+echo ""
+echo "#############################################################################"
+echo "# Initialize Neos...                                                        #"
+echo "#############################################################################"
+dc exec -T php bash <<-'BASH'
+    cd TestDistribution
+    # replace installed Neos Ui with local dev via sym link to mounted volume
+    # WHY: We want changes of dev to appear in system under test without rebuilding the whole system
+    rm -rf Packages/Application/Neos.Neos.Ui
+    ln -s /usr/src/neos-ui/ /usr/src/app/TestDistribution/Packages/Application/Neos.Neos.Ui
+    sed -i 's/host: 127.0.0.1/host: db/g' Configuration/Settings.yaml
+    ./flow flow:cache:flush
+    ./flow flow:cache:warmup
+    ./flow doctrine:migrate
+    ./flow user:create --username=admin --password=password --first-name=John --last-name=Doe --roles=Administrator || true
+BASH
+
+echo ""
+echo "#############################################################################"
+echo "# Start Flow Server...                                                      #"
+echo "#############################################################################"
+dc exec -T php bash <<-'BASH'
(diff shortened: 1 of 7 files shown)


JamesAlias:
If $someone is able to come up with a better name for the make command `e2e-start-system-under-test` I would appreciate it 😃 

Something like `start-dev-neos-instance-docker` or `start-test-neos-instance-docker`.. 
I'm not sure ^^ "naming things", right 😆

mhsdesign:
I just asked bing ki to fix this setup for linux lets see :D

Tests/IntegrationTests/start-system-under-test-docker.sh
```diff
@@ -0,0 +1,105 @@
...
+echo ""
+echo "#############################################################################"
+echo "# Create sym links to mounted Docker volumes...                             #"
+echo "#############################################################################"
+echo ""
+dc exec -T php bash <<-'BASH'
+    # replace installed Neos Ui with local dev via sym link to mounted volume
+    # WHY: We want changes of dev to appear in system under test without rebuilding the whole system
+    rm -rf Packages/Application/Neos.Neos.Ui
+    ln -s /usr/src/neos-ui/ /usr/src/app/TestDistribution/Packages/Application/Neos.Neos.Ui
```
mhsdesign: This works in generally, but not in this flow. Later i think due to the `composer reinstall neos/test-site` this linking will be gone and we have the original composer neos ui... so this step should come later?

Tests/IntegrationTests/start-system-under-test-docker.sh
```diff
@@ -0,0 +1,105 @@
...
+
+echo "#############################################################################"
+echo "# Start docker environment...                                               #"
+echo "#############################################################################"
+dc down
+dc up -d
+dc exec -T php bash <<-'BASH'
+    rm -rf /usr/src/app/*
+BASH
+docker cp "$(pwd)"/Tests/IntegrationTests/. "$(dc ps -q php)":/usr/src/app
```
mhsdesign: Todo: Lets use paths relative to the shellscript?

mhsdesign (changes requested):
Thank you for this :heart: i had some small issues running this on linux at first but that is fixed now ;)

After linking the neos-ui again the Host.js is found and i could continue to actually run a test.

For some reason the tests dindt work though ... nothing happened (besides opening the publish button dropdown ... and it failed)

mhsdesign (changes requested):
github is buggy ^^

Tests/IntegrationTests/start-system-under-test-docker.sh
JamesAlias: Fine by me, I didn't want to put too much time into it and copy pasted the most stuff.

It's clear, that we could build a better system for all of this but I needed a faster feedback loop fast 😬 
I want to build tests for Bugs I fix and wasting 2h just for building the test and another 2 for running them to see if the fix works and we don't mess up something else ist just too much ^^

Tests/IntegrationTests/start-system-under-test-docker.sh
JamesAlias: I used this setup and it worked for me.

Thanks for fixing the issues on Linux ❤️

mhsdesign: Well this is not fixed yet ... and I don't know why it would work on your machine ... I mean the problem is within docker after all

mhsdesign (approved):
Works *

Issue #3340, opened 2023-01-24
Minor build stack problem 


mhsdesign:
```
make build-react-ui-components-standalone
make build
```

fails, as "module" has a higher preference than "main"

```
  "module": "./lib-esm/index",
  "main": "./src/index",
```

```
✘ [ERROR] [plugin esbuild-style-plugin] Could not resolve "./style.scss"

    packages/react-ui-components/lib-esm/MultiSelectBox/index.js:3:18:
      3 │ import style from './style.scss';
         ╵                   ~~~~~~~~~~~~~~
```

PR #3371, opened 2023-02-15
BUGFIX: Fix npm releases & cleanup release process

mhsdesign:
fixes: https://github.com/neos/neos-ui/issues/3262 (https://github.com/neos/neos-ui/pull/3316)
fixes: https://github.com/neos/neos-ui/issues/3340
fixes: https://github.com/neos/neos-ui/issues/3374
fixes: https://github.com/neos/neos-ui/issues/3375

followup on: https://github.com/neos/neos-ui/pull/3200
related: https://github.com/neos/neos-ui/issues/3263

**What I did**
- speed up bump-version from 1m18s to 10s
- clean up a lot of release hacks (shellscripts)
- fixed our npm packages -> found a solution for this dilemma: https://stackoverflow.com/questions/55013953/pointing-the-main-field-in-package-json-conditionally
- !!! typescript transpiles now to es2020 for npm packages so no cjs anymore jippii

**How I did it**

i have no other hobbies

**How to verify it**

i tested "positional-array-sorter", "neos-ui-extensibility" and "react-ui-components" with `yarn pack` which is essentially a `yarn npm publish` dry run

lets make a 8.2.7-beta or sth? ^^ (only for npm stuff)


mhsdesign:
Hmm this is a nasty one ... this describes what we want: https://stackoverflow.com/questions/55013953/pointing-the-main-field-in-package-json-conditionally

and we cannot use the workaround to import directly from source in here https://github.com/neos/neos-ui/blob/5e7ae4a656717abbc7371c02e18bab21f341e54b/packages/neos-ui-extensibility/src/registry/SynchronousRegistry.ts#L2

like: `import positionalArraySorter from '@neos-project/positional-array-sorter/src/positionalArraySorter.ts';`

as the "neos-ui-extensibility" package is itself published also to the nom registry and would now contain wrong paths ...



maybe we can solve this better instead of using a shell script by using a custom yarn plugin and the hook "beforeWorkspacePacking": 
https://github.com/yarnpkg/berry/blob/fb77381410b795893d25321583bf9a6d25a758f0/packages/plugin-pack/sources/index.ts#L16

or use an existing plugin like https://github.com/nicolo-ribaudo/yarn-plugin-conditions

mhsdesign:
NOOOOOOOOOOOOOOOOOOICCCCCCCCCCCCCCCCEEEEEEEEEEEEE lets goo we dont need no own plugin, its a feature by yarn already... while digging though the yarn code a bit i found that they have the need for a similar feature here: https://github.com/yarnpkg/berry/blob/fb77381410b795893d25321583bf9a6d25a758f0/packages/plugin-npm-cli/package.json#L43


it seems that we can just use `publishConfig.main` to override the "main" field :D
https://yarnpkg.com/configuration/manifest#publishConfig

packages/neos-ui-extensibility/package.json
```diff
@@ -22,7 +25,7 @@
     "typescript": "^4.6.4"
   },
   "dependencies": {
-    "@neos-project/positional-array-sorter": "8.1.4",
+    "@neos-project/positional-array-sorter": "workspace:*",
```
mhsdesign, line 28: @markusguenther

mhsdesign:
i published a "8.2.7-beta1" to npm and it all looks fine ;) ... package.jsons seem to work as expected and also the dist folders contain now modern day javascript ;)


```sh
#!/usr/bin/env bash

set -xe

VERSION="8.2.7-beta1"
NPM_TOKEN="npm_123"

# go to root directory of Neos.Neos.Ui
DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
cd $DIR/../../

# build
make build-production

# publishing
VERSION=$VERSION make bump-version
VERSION=$VERSION NPM_TOKEN=$NPM_TOKEN make publish-npm
```

PR #3373, opened 2023-02-15
BUGFIX: Run GitHub actions also on PRs

markusguenther:
The new GitHub actions only run on Neos branches and do not work for PRs on forked repositories.

**What I did**
Adjusted the events that trigger the GH actions.

**How I did it**
In the docs we have a bunch of events. We use now PRs, pushes and workflow dispatches (for instance weblate)
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows

**How to verify it**
Check if this PR has unit tests and linting via GH action.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -1,6 +1,11 @@
 name: Build and test
 
-on: push
+on:
+  workflow_dispatch:
+  push:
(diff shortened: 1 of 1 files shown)


ahaeslich (approved):

Sebobo:
I don't understand the reason branch name version thing? Can we add some comment or link to some doc?
And why is workflow dispatch empty?

PR #3385, opened 2023-02-18
TAKS: Run typescript lint only once (speed up and DX)

mhsdesign:
**What I did**

previously each package started tsc to lint which will also check the whole project, so we were checking with ts unnecessary often. Im hoping to decrease lint time with this fix.

speed up make lint from 2:43 to 1:24 on CI

move tsconfig to root of project

export scss files for 8.2.x react-ui-components (they are broken, but know a little less - they will work properly with 8.3)

incremental=false for react-ui-components as the two tsc processes wont work on rebuild

**How I did it**

**How to verify it**


markusguenther (approved):
Tested that locally and it works 💙

mhsdesign:
Thanks ;)

Issue #3370, opened 2023-02-15
Remove unusefull tests `expect(true).toBe(true);`


mhsdesign:
i think one assertion is enough ;) currently we have 101

JamesAlias:
I think those tests were some kind of reminder to actually write _real_ tests ^^

mhsdesign:
yes but they might slow down testing and falsify our trust into them ;)
Also realistically - no one will anytime soon write the tests :P

JamesAlias:
agreed 😅

PR #3386, opened 2023-02-18
TASK: Remove empty tests

mhsdesign:
solves https://github.com/neos/neos-ui/issues/3370

ONLY tests were remove which content is EXACTLY

```js
test(`write tests`, () => {
    expect(true).toBe(true);
});
```


markusguenther (approved):
No all the valuable tests :P 

Thanks for the cleanup.

markusguenther:
We could also remove them in 7.3... would save time on all PRs ;)

mhsdesign:
hmm it sure is about a minute we can save ^^

PR #3387, opened 2023-02-19
TASK: Adding labels to PRs

markusguenther:
With this task, we add automatically labels to the PRs. That makes the maintaining easier, as we did that manually all the time.

**What I did**
We always add the target branch as label, so we see directly which version the PR targets. And if the Commit message is following our rules, we also add the `bugfix`, `task` or `feature` label.

**How I did it**
I use the runner from https://github.com/actions-ecosystem
They use node 12 which leads to a warning as actions are running node 16 but an issue is created and it runs anyways.

**How to verify it**
https://github.com/Unikka/neos-ui-new/actions/runs/4217847523

### DIFF
--- a/.github/workflows/add-pr-labels.yml
+++ b/.github/workflows/add-pr-labels.yml
@@ -1,0 +1,31 @@
+name: Add Labels to Pull Request
+
+on:
+  pull_request:
+    types: [opened, reopened]
+
+jobs:
+  add-labels:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add target branch label
+        uses: actions-ecosystem/action-add-labels@v1
+        with:
+          labels: ${{ github.event.pull_request.base.ref }}
+      - name: Add feature label
+        if: startsWith(github.event.pull_request.title, 'FEATURE:')
+        uses: actions-ecosystem/action-add-labels@v1
+        with:
+          labels: 'feature'
+      - name: Add bugfix label
+        if: startsWith(github.event.pull_request.title, 'BUGFIX:')
+        uses: actions-ecosystem/action-add-labels@v1
+        with:
+          labels: 'bugfix'
+      - name: Add task label
+        if: startsWith(github.event.pull_request.title, 'TASK:')
+        uses: actions-ecosystem/action-add-labels@v1
+        with:
+          labels: 'task'
+    env:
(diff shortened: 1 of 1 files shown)


markusguenther:
Guess will add this to the other repos (flow and neos) as well...
It saves much time...

markusguenther:
Ok that it works, is visible in the feed.
Screenshot 2023-02-19 at 21 00 41

ahaeslich:
What about changing the target branch afterwards? This we would do manually?

markusguenther:
> What about changing the target branch afterwards? This we would do manually?

Then I need to read the event docs of github actions again ;)
Currently it runs just on open new and reopen. So change basebranch and then close and reopen again would change the label... but then I also need to clean up 🙈

markusguenther:
Until the runner supports the remove all labels, I will skip that. And I don't want to maintain another custom library ;)
https://github.com/actions-ecosystem/action-remove-labels/issues/15

So when this will work anytime. We can remove all labels and readd the matching ones.

mhsdesign:
Can we also make a rename action so when I write `Task:` again instead of `TASK:` that it will be fixed automagically and you don't have to do it ^^

markusguenther:
Ok the removal works now but for some reason the runner for the labels is not triggered when you just change the branch.
On the next push, it will clean up the labels.

markusguenther:
Ok now it works :D

markusguenther:
Screenshot 2023-02-19 at 22 43 27
🥳

markusguenther:
> Can we also make a rename action so when I write `Task:` again instead of `TASK:` that it will be fixed automagically and you don't have to do it ^^

Work on that, but comes with another PR and actually not finished.

markusguenther:
Ok now the PR is working perfect and we cannot remove labels accidentally
Screenshot 2023-02-20 at 11 33 00
. As we now only remove the old base branch label when we change the branch :D

crydotsnake (approved):
Looks good to me!

ahaeslich:
Should we add this to the lowest maintained branch? Or are we fine with adding it to 8.3 and above?

markusguenther:
Hmm good point @ahaeslich

markusguenther:
Rebase is done ;) @ahaeslich

ahaeslich (approved):

PR #3395, opened 2023-02-23
BUGFIX: wrong hover color for DateInput component

crydotsnake:
**What I did**

When using the `DateTime` editor and hovering over the years, it has the wrong `:hover` color wich makes it unreadable.

### Before:

SCR-20230223-lfl

## After

SCR-20230223-pyl

**How I did it**
Adjusted the `style.css` for the `DateInput` component.

**How to verify it**
Create a NodeType with the `DateTime` editor, and hover over the Years.

### DIFF
--- a/packages/react-ui-components/src/DateInput/style.css
+++ b/packages/react-ui-components/src/DateInput/style.css
@@ -271,7 +271,7 @@
         }
         td.rdtMonth:hover,
         td.rdtYear:hover {
-            background: #eee;
+            background: var(--colors-PrimaryBlue);
         }
 
         .rdtCounters {


mhsdesign:
will be hard to upmerge to 8.3 (as this file completely changes) but the core fix should still apply.

Is it on purpose that these changes are made in the css section which is straight copied form the react-datetime source?
It could also be placed a bit downpage in "Overwrites for the calendar" but on the other hand i think in line 249 and 307 we already disobey our original idea.

So okayish with me - this whole inlining of css and tight coupling will just make it impossible to upgrade this datetime picker without rewrite, but as long as we never need to update the react-datetime picker this will be okay.

markusguenther (approved):
Thanks, @crydotsnake it works and the up merge is not the problem that should stop a change.

PR #3396, opened 2023-02-24
FEATURE: Improve node creation dialog

### LINKED ISSUE #3303: UX: Improve node creation dialog
### Problem
I am a Neos newcomer and would love to have more context if I add a new node.
Often after clicking the plus icon button (either in the content tree or directly in the layout), I often wonder If I've actually selected the correct node. The current selected node could be visible (but only if the content tree is open) in the background, but it is not so easy to read, with the backdrop.

If I've selected the wrong node (the one above or the one higher/lower in the hierarchy) I can still "salvage the situation" (this means that I don't have to press "cancel", because in my mind, I'm already looking for the node I wanted to create) with the `insert mode` control. Sadly for me the icons were not self-explanatory what these buttons actually do.

### Suggestion
Make the dialog title dynamic based on the insert position (Default is inside) and the current node label. This means that the title would be called `Add inside Section`. Below the title we change the label to `Insert position` and separate the `insert before / after` from `insert inside` as this often leads to different nodes being displayed. We change the icons of the button to reflect the direction and meaning of the action. `before` and `after` add something to the parent of this item, so the arrow is pointing to the left. `inside` is one level deeper as the current level, therefore we point the arrow to the right.

| before | after |
|-|-|
| Dialog is called "Create new" and button order is above, inside, below. Icons are not self explanatory| Dialog is called "Add inside" and the label name of the NodeType. The button order is above, below grouped together and inside separate.|
| Dialog is called "Create new" and button order is above, inside, below. Icons are not self explanatory|Dialog is called "Add above"  and the label name of the NodeType. The button order is above, below grouped together and inside separate. |

> **Note**
> (Not completely unrelated sidenote)
> I think, in the preview we could also introduce a separate button for inserting inside the current position.
>
> This comes naturally, as the plus already signifies that we add something at the position (I just realized this is not the case, it adds it below and not above. I think above should be the default if clicked in the preview, as the button lays between the current and the previous field. We could still add a separate below the node to add something below this node). 
>
> Bildschirmfoto 2022-12-20 um 17 15 02

pKallert:
closes #3303 

**What I did**
As discussed in the issue, I made the following changes to the Node Selection Dialog: 

- Add text to insert position buttons 
- Change the title of the Node Selection Dialog according to the current insert position 

There is another mention in the original issue about the inline toolbar in the preview that is still unresolved, thus I did not implement it. 

**How I did it**
I changed up the Node Selection Dialog according to the issue. 

I am unsure if I did the translation part of the feature right, suggestions are welcome 😄 
For the "Inside"- Icon I used the standard fontawesome icon. I could exchange it but perhaps this is good enough? 

**How to verify it**

With the changes in #3132  the layout of the Ui has already changed so this looks slightly different from the design in the original issue : 
Bildschirmfoto 2023-02-24 um 09 16 47

The changes influence also the copy and paste feature which now looks like this: 
Bildschirmfoto 2023-02-24 um 09 21 09

### DIFF
--- a/Resources/Private/Translations/en/Main.xlf
+++ b/Resources/Private/Translations/en/Main.xlf
@@ -23,13 +23,13 @@
             <trans-unit id="inside" xml:space="preserve">
                 <source>Inside</source>
             </trans-unit>
-            <trans-unit id="InsertModeTitleinto" xml:space="preserve">
+            <trans-unit id="InsertModeTitleInto" xml:space="preserve">
                 <source>inside</source>
             </trans-unit>
-             <trans-unit id="InsertModeTitlebefore" xml:space="preserve">
+             <trans-unit id="InsertModeTitleBefore" xml:space="preserve">
                 <source>above</source>
             </trans-unit>
-            <trans-unit id="InsertModeTitleafter" xml:space="preserve">
+            <trans-unit id="InsertModeTitleAfter" xml:space="preserve">
                 <source>below</source>
             </trans-unit>
             <trans-unit id="insertMode" xml:space="preserve">
--- a/Resources/Private/Translations/en/Main.xlf
+++ b/Resources/Private/Translations/en/Main.xlf
@@ -20,8 +20,17 @@
             <trans-unit id="below" xml:space="preserve">
                 <source>Below</source>
             </trans-unit>
-            <trans-unit id="below" xml:space="preserve">
+            <trans-unit id="inside" xml:space="preserve">
                 <source>Inside</source>
+            </trans-unit>
+            <trans-unit id="InsertModeTitleinto" xml:space="preserve">
+                <source>inside</source>
+            </trans-unit>
+             <trans-unit id="InsertModeTitlebefore" xml:space="preserve">
+                <source>above</source>
+            </trans-unit>
+            <trans-unit id="InsertModeTitleafter" xml:space="preserve">
+                <source>below</source>
             </trans-unit>
             <trans-unit id="insertMode" xml:space="preserve">
                 <source>Insert mode</source>
--- a/Resources/Private/Translations/en/Main.xlf
+++ b/Resources/Private/Translations/en/Main.xlf
@@ -14,6 +14,15 @@
             <trans-unit id="insert" xml:space="preserve">
                 <source>Insert</source>
             </trans-unit>
+            <trans-unit id="above" xml:space="preserve">
+                <source>Above</source>
+            </trans-unit>
+            <trans-unit id="below" xml:space="preserve">
+                <source>Below</source>
+            </trans-unit>
+            <trans-unit id="below" xml:space="preserve">
+                <source>Inside</source>
+            </trans-unit>
             <trans-unit id="insertMode" xml:space="preserve">
                 <source>Insert mode</source>
             </trans-unit>
--- a/packages/neos-ui-containers/src/InsertModeSelector/style.css
+++ b/packages/neos-ui-containers/src/InsertModeSelector/style.css
@@ -20,4 +20,9 @@
         border-top-right-radius: 2px;
         border-bottom-right-radius: 2px;
     }
+    button.intoButton{
+        margin-left: 8px;
+    }
 }
+
+
--- a/packages/neos-ui-containers/src/InsertModeSelector/style.css
+++ b/packages/neos-ui-containers/src/InsertModeSelector/style.css
@@ -20,7 +20,7 @@
         border-top-right-radius: 2px;
         border-bottom-right-radius: 2px;
     }
-    button.intoButton{
+    .intoButton{
         margin-left: 8px;
     }
 }
--- a/packages/neos-ui-containers/src/InsertModeSelector/style.css
+++ b/packages/neos-ui-containers/src/InsertModeSelector/style.css
@@ -12,6 +12,9 @@
 .buttonGroup {
     margin-left: 4px;
 
+    span {
+        margin-left: 7px;
+    }
     > button:first-child {
         border-top-left-radius: 2px;
         border-bottom-left-radius: 2px;
@@ -22,6 +25,9 @@
     }
     .intoButton{
         margin-left: 8px;
+    }
+    button{
+        padding: 0px 14px
     }
 }
 
--- a/packages/neos-ui-containers/src/InsertModeSelector/style.css
+++ b/packages/neos-ui-containers/src/InsertModeSelector/style.css
@@ -20,15 +20,17 @@
         border-top-right-radius: 2px;
         border-bottom-right-radius: 2px;
     }
-    .intoButton{
+    .intoButton {
         margin-left: 8px;
     }
-    button{
+    > button {
         padding: 0px 14px;
+
+        > span {
+            margin-left: 7px;
+        }
     }
-    span {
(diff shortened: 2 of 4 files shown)


markusguenther:
Need to investigate why the forked PR does not get the secrets. Because it is configured in CircleCI.

Resources/Private/Translations/en/Main.xlf
```diff
@@ -14,6 +14,24 @@
             <trans-unit id="insert" xml:space="preserve">
                 <source>Insert</source>
             </trans-unit>
+            <trans-unit id="above" xml:space="preserve">
```
crydotsnake, line 17: The translations are looking good! I would not think of anything better at the moment 😄 But it would be great, if you could add the translations also for german.

packages/neos-ui-containers/src/InsertModeSelector/style.css
```diff
@@ -20,4 +20,9 @@
         border-top-right-radius: 2px;
         border-bottom-right-radius: 2px;
     }
+    .intoButton{
+        margin-left: 8px;
+    }
 }
+
+
```
crydotsnake, line 36: It looks really good already!

The only thing I would add is the distance between the "Above" and the "Below" button. So it would look even, and also a bit prettier.

SCR-20230225-dtq


In Neos 8.2 it is not yet so completely in the eye stung, because here before only the icon was without text. With text, however, it is a small difference.

SCR-20230225-dzy

crydotsnake (changes requested):
Hello Paula!

Thanks you a lot, for working on this task ! 😃
I have added only a few suggestions, but all in all it looks very good !

Resources/Private/Translations/en/Main.xlf
mhsdesign: We can do that via crowdin ;)

crydotsnake: Yes, i know. But I am used to create the translations for EN and DE at the same time ;)

packages/neos-ui-containers/src/InsertModeSelector/style.css
mhsdesign: I would argue this was done on purpose in the original design from @manuelmeister in https://github.com/neos/neos-ui/issues/3303 and I really like this visual distinction! (And others did. as well in the thread)

Resources/Private/Translations/en/Main.xlf
mhsdesign: 😄

packages/neos-ui-containers/src/InsertModeSelector/style.css
crydotsnake: Yes, I just looked at it. But in my opinion it looks a bit squashed together. Why should I position two buttons next to each other, but separate the third button from the others. Then I'd rather make it uniform with an even spacing.

Resources/Private/Translations/en/Main.xlf
markusguenther: Weblate will take care then ;)

packages/neos-ui-containers/src/InsertModeSelector/style.css
mhsdesign: I think the distinction is important as "above" and "below" present oposites and thus belong closly together while "inside" is a third option which doesnt belong directly to the previous and should be separated.

Maybe we can settle on increasing the x-padding of each button ever so slightly? (but not the gap)

packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
```diff
@@ -189,10 +192,17 @@ export default class SelectNodeType extends PureComponent {
     }
 
     renderSelectNodeTypeDialogHeader() {
+        const {insertMode} = this.state;
+        const {i18nRegistry, nodeTypesRegistry, referenceNodeType} = this.props;
+        const nodeTypeLabel = $get('ui.label', nodeTypesRegistry.get(referenceNodeType))
+        const nodeTypeLabelText = i18nRegistry.translate(nodeTypeLabel, 'Node')
+        const addLabel = i18nRegistry.translate('Neos.Neos.Ui:Main:add', 'Add')
+
+        const insertModeLabel = i18nRegistry.translate('Neos.Neos.Ui:Main:InsertModeTitle' + insertMode, 'to')
```
mhsdesign: i would strongly recommend to not use string concatenation for labels, as its impossible to find them in the source code ^^ so im proposing a insertMode to label map or switch case (doesnt matter) but future us will thank us :D

packages/neos-ui-containers/src/InsertModeSelector/style.css
crydotsnake: > Maybe we can settle on increasing the x-padding of each button ever so slightly?

We can try

manuelmeister: Yes, the gap is intended, as often the content inside and beside is different but rarely (if ever) different between over and under.

manuelmeister (changes requested):
Thank you for your work!

I will send you (as soon as possible, maybe March 7., as I'm currently in asia on holiday) the custom icons, I designed for #3303.

packages/neos-ui-containers/src/InsertModeSelector/style.css
pKallert: I played around locally and I would suggest shrinking the distance between the icon and text and increasing the x-padding of the button: 
Bildschirmfoto 2023-03-03 um 11 09 21

Does this work for everyone?

mhsdesign: yes this looks great ;) :heart:

crydotsnake: Looks better.

packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
pKallert: I updated the PR with a switch case 👍

mhsdesign:
i see that you had trouble with css like this?
```css
.buttonGroup {
     margin-left: 4px;

     span {
         margin-left: 7px;
     }
}
```

the new 8.3 css pipeline wont like that. The css nesting spec requires to append an `&` to differentiate between elementName (selector) and propertyName:


```css
.buttonGroup {
     margin-left: 4px;

     & span {
         margin-left: 7px;
     }
}
```

> When a nested rule starts with an element selector, it may be ambiguous with a property. In these cases, the spec requires the selector be prefixed with a [nesting selector](https://drafts.csswg.org/css-nesting/#nest-selector) (&).

crydotsnake (approved):
Now its looking good! :) 

I also now understand the point from @manuelmeister regarding the non-space between the Above and Below buttons.

markusguenther (approved):
Thanks Paula... tested it locally and via CI https://app.circleci.com/pipelines/github/neos/neos-ui/3172/workflows/b101e549-b9f2-4d64-ac5c-af8cd5f41f2c

markusguenther:
Will merge it as it is and file an issue to replace the icon when @manuelmeister is back at home or has some spare time ;)

PR #3399, opened 2023-03-01
BUGFIX: `<Tabs>` add `onActiveTabChange` and use `activeTab` on initial render

mhsdesign:
**What I did**

My usecase is to use the `<Tabs>` component and save the currently selected item to the localstorage to restore the previously used state when the site is opened again.

The problem is currently its not possible to know the currently selected tab.
Thats why i introduced `onActiveTabChange`.
I also thought about adding an option to completely handle the state outside, but this would not make sense for this component. So thats why i keep it as a simple prop.

When experimenting, i also noticed that the `activeTab` prop will have no effect unless the above component is rerendered (which triggers `UNSAFE_componentWillReceiveProps`)

**How I did it**

**How to verify it**

### DIFF
--- a/packages/react-ui-components/src/Tabs/tabs.tsx
+++ b/packages/react-ui-components/src/Tabs/tabs.tsx
@@ -56,11 +56,8 @@ export default class Tabs extends PureComponent<TabsProps> {
     public static defaultProps = tabsDefaultProps;
 
     private updateActiveTab(activeTab: string | number): void {
-        if (this.props.onActiveTabChange) {
-            this.props.onActiveTabChange(activeTab);
-        }
-        this.setState({
-            activeTab
+        this.setState({ activeTab }, () => {
+            this.props.onActiveTabChange?.(activeTab);
         });
     }
 
--- a/packages/react-ui-components/src/Tabs/tabs.tsx
+++ b/packages/react-ui-components/src/Tabs/tabs.tsx
@@ -55,7 +55,7 @@ export default class Tabs extends PureComponent<TabsProps> {
 
     public static defaultProps = tabsDefaultProps;
 
-    private updateActiveTab(activeTab: string | number) {
+    private updateActiveTab(activeTab: string | number): void {
         if (this.props.onActiveTabChange) {
             this.props.onActiveTabChange(activeTab);
         }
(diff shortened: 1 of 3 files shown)


mhsdesign:
```
public constructor(props) {
    super(props)
    this.state.activeTab = props.activeTab
}
```

should do but ts doesnt like it ^^ (state is readonly)

Sebobo (changes requested):
Looks good to me. Can you add a test to verify the functionality with the string based id?
Seems all tests only use the numbers.

packages/react-ui-components/src/Tabs/tabs.tsx
```diff
@@ -45,19 +50,26 @@ interface TabsState {
...
     public state: TabsState = {
-        activeTab: 0,
+        activeTab: this.props.activeTab ?? 0
     };
 
     public static defaultProps = tabsDefaultProps;
 
+    private updateActiveTab(activeTab: string | number): void {
+        if (this.props.onActiveTabChange) {
+            this.props.onActiveTabChange(activeTab);
```
Sebobo: It would be better IMO to set the state before calling the callback to prevent the callback from interfering with the functionality. Maybe just a theoretical problem. Not sure.

mhsdesign: yeah i though about it too, most correct way would be:

```js
this.setState({activeTab}, () => {
 this.props.onActiveTabChange(activeTab)
})
```

(i decided against this, as there is no direct equivalent when using hooks  in the future ... or is there?
```js
useEffect(() => {
   props.onActiveTabChange(activeTab)
}, [activeTab])

```
)

mhsdesign: the order of 

```
this.props.onActiveTabChange(activeTab)
this.setState({activeTab})
```

```
this.setState({activeTab})
this.props.onActiveTabChange(activeTab)
```

shouldnt matter imo as the state is only updated after rerender or sth ...

Sebobo: There are various effects that can be used when we switch to hooks, or whatever is cool until then. 
I just wonder like throwing an error in the callback or running some operation and the UI would be blocked due to that. As I said, sounds a bit dumb, but 🤷🏻‍♂️

mhsdesign: okay this it is:
```js
this.setState({activeTab}, () => {
 this.props.onActiveTabChange(activeTab)
})
```

mhsdesign:
> Can you add a test to verify the functionality with the string based id?

Aba natürlich

Sebobo (approved):
Looks good by reading!

PR #3402, opened 2023-03-02
BUGFIX: `yarn version` your current branch contains multiple versioning files; this isn't supported

mhsdesign:
fixes https://github.com/yarnpkg/berry/issues/5109


markusguenther (approved):
Tested locally and seems to fix that. :)

PR #3403, opened 2023-03-05
BUGFIX: #3401 Update and Pin `"@friendsofreactjs/react-css-themr": "~4.2.0"`

### LINKED ISSUE #3401: BUG: Either update `@friendsofreactjs/react-css-themr` or declare correct peer dependency version
we require `"@friendsofreactjs/react-css-themr": "^4.1.0",` which includes the lates release 4.3 but there was a little breaking change in between and the types changed or something...

so people should not install the latest version of react-css-themr for ts support but need to require the same version as us: 4.1.0

solutions:

A: pin the peerDependency to 4.1.0 so people are not confused
B: update (in yarn.lock) react-css-themr and rebuild our type declarations for the react-ui components

mhsdesign:
fixes:  #3401

**What I did**
see #3401

**How I did it**

**How to verify it**


markusguenther:
This can solve the bug temporary, but I think it would be nice to fix it when it is broken. And we can release a 4.3.1 if needed ;)

markusguenther (approved):
Fine by now, but will add an issue on the repository and we should check that...

Issue #3393, opened 2023-02-23
Bug: Linting might be kaput


mhsdesign:
compare outputs of `make lint` vs `yarn eslint "packages/*/src/**/*.{js,jsx,ts,tsx}"`

a missing semicolon will not be found for example in 8.3 but in 7.3

Issue #3379, opened 2023-02-17
Move from tslint to typescript-eslint


JamesAlias:
### Description

We use `tslint` to lint our Typescript code and it is was deprecated in 2019 in favor of `typescript-eslint`.

I think it's time to make a transition and while we`re at it we can review some of the rules.

There are Rules like `interface-over-type-literal` which really bug me tbh.

Open for discussion. Maybe on a sprint.

mhsdesign:
see also https://github.com/neos/neos-ui/issues/2662 :)

i also dislike that one cannot write `foo == null` to also check for undefined.
And lately I got used to always using parens in arrow functions so ...

JamesAlias:
I searched for an open issue but didn't find it 🙈

PR #3404, opened 2023-03-05
Task: Cleanup Linting / Speedup / Fix / Remove Tslint

mhsdesign:
Solves: #3379
Solves: #3393

**What I did**

**How I did it**

**How to verify it**
search for tslint in codebase ^^


mhsdesign:
I removed the subpackages "lint" command and its handled now globally (which is better as we wont forget to add this command to packages, which happened.)

mhsdesign:
speeds up linting by a minute to 24 seconds 💯 🔥

markusguenther (approved):

PR #3411, opened 2023-03-06
WIP: HACK: 90 edit preview mode support

mhsdesign:
companion piece of https://github.com/neos/neos-development-collection/pull/4067

@grebaldi what should i do im scared

**What I did**

**How I did it**

**How to verify it**

todo:

revert https://github.com/neos/neos-ui/commit/ce395ca9e3614ec32f7af7a82d2759a74e68f62d once https://github.com/neos/neos-ui/pull/3569 is merged

### DIFF
--- a/packages/neos-ui-sagas/src/UI/ContentCanvas/index.js
+++ b/packages/neos-ui-sagas/src/UI/ContentCanvas/index.js
@@ -3,7 +3,7 @@ import {takeLatest, put, select, take, race} from 'redux-saga/effects';
 import {$get} from 'plow-js';
 import {getGuestFrameDocument} from '@neos-project/neos-ui-guest-frame/src/dom';
 
-import {actionTypes, actions} from '@neos-project/neos-ui-redux-store';
+import {actionTypes, actions, selectors} from '@neos-project/neos-ui-redux-store';
 
 /**
  * Load newly created page into canvas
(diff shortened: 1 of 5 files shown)


mhsdesign:
currently breaks other plugins like: https://github.com/sandstorm/MxGraph/blob/6e1b14bc63bdeb84effea2a687471e209db610b6/Resources/Private/JavaScript/MxGraph/src/DiagramEditor.js#L15-L29 as `'ui.contentCanvas.src'` is not the actual value of the iframe

> we discussed that this is fine ;)

mficzel:
Currently the review is triggered  very often on the ui side

packages/neos-ui-sagas/src/UI/ContentCanvas/index.js
```diff
@@ -44,7 +44,8 @@ export function * watchStopLoading({globalRegistry, store}) {
 export function * watchReload() {
     yield takeLatest(actionTypes.UI.ContentCanvas.RELOAD, function * (action) {
```
mhsdesign, line 45: idk but why is there no reducer of `actionTypes.UI.ContentCanvas.RELOAD` which updates the current iframe url (if in payload)

then we can use the same selector down in line 61 ... or we can put into the selector a modified temporal state ...

bwaidelich:
What's the state of this one? Is it still being worked on, @mhsdesign ?

mhsdesign:
currently untouched since dresden sprint. i dont know how https://github.com/neos/neos-development-collection/pull/4067 moved forward and what was discussed in berlin about it (i have the feeling things were discussed in this direction). but feel free to sign me up for a sync meeting about it ^^

mficzel:
Main change is that we switched to separate edit and preview actions and decide on the rendering-mode by looking at the main request. Once this is accepted in https://github.com/neos/neos-development-collection/pull/4067 it makes sense to proceed here.

packages/neos-ui-sagas/src/UI/ContentCanvas/index.js
```diff
@@ -55,7 +56,23 @@ export function * watchReload() {
             // If the new uri is provided in the action payload, use that
             //
             if (iframeWindow.location.href === currentIframeUrl) {
-                iframeWindow.location.href = uri || iframeWindow.location.href;
+                if (uri) {
+                    const state = yield select();
```
mhsdesign, line 60: as @grebaldi noted, yield is not allowed here and we have to get the state above / or convert it to a for of loop.

mficzel:
Close as outdated

PR #3423, opened 2023-03-10
FEATURE: add new node creation dialog icons

### LINKED ISSUE #3407: TASK: Change inside icon in create dialog
### Description

In the PR #3396 we have improved the create node dialog and the icon for inside was missing.
So when the custom Icon is available, we will replace the font awesome icon.

#### Actual look

Screenshot 2023-03-06 at 13 51 40

### Affected Versions

Neos: 8.3

UI: 8.3

manuelmeister:
**What I did**
Dry, implement the improved icons for the node creation dialog.

Fixes #3407


mhsdesign:
previously (font awesome)

image


custom icons: (still wip)
image

I fixed the resource path and styling a bit, but i have the feeling the icons and text are not vertically centered ... but cant fix that rn

... for some reason resource path icons and fa icons have different styling ... thats why the iconAlignmentHack class (to make it behave like fa)

manuelmeister:
I'll fix the styling for these icon buttons.

markusguenther:
I adjusted the Icons and the text alignment. We use now flex box.
Screenshot 2023-03-15 at 11 51 36

manuelmeister:
Perfect. I think, I would have also added a border radius to the edges between "below" and "inside"

markusguenther (approved):
Thanks for the custom icon and your first contribution to the Neos UI :)

markusguenther:
@manuelmeister do you want to adjust something more or can we remove the draft state?

markusguenther:
I also will take a look if the additional span is really needed for the icon. But that will end up in a new PR as we then need to target a lower branch.

markusguenther:
Rebased to the latest 8.3 states:
Screenshot 2023-03-15 at 12 33 28

crydotsnake (changes requested):
When i want to add a new content module to the site, It breaks the UI.

SCR-20230316-hvms

Just thought the PR would be ready for review because the status has changed ;)

manuelmeister:
Sorry, I've changed it, because I finished my changes (icons & radius). Maybe @markusguenther has not finished the integration.

markusguenther:
Tested that locally, and it works for Content and Documents. As the e2e Pipeline also not fails, because with this error it would, it seems to be an issue on your machine. Do you have more input what exactly lead to the error?

crydotsnake:
> Tested that locally, and it works for Content and Documents. As the e2e Pipeline also not fails, because with this error it would, it seems to be an issue on your machine. Do you have more input what exactly lead to the error?

I just checked out the branch, and builded the files:

```
make clean && make setup
```

I tryed it again with `make build` but this also doesent helps. If i switch back to the `8.3` branch and build everything again, the UI works as usuall.

markusguenther:
Talked with @mhsdesign and we could not reproduce this issue. If something comes up, we have some weeks left until release to fix it. Thanks for the feedback @crydotsnake, and thanks for the first PR @manuelmeister 💙

crydotsnake:
Will try it again when i'm home. Will come back to you 🙂

PR #3427, opened 2023-03-10
TASK: Fix order of node creation dialog SearchBar

### LINKED ISSUE #3419: Move filter in NodeCreation Dialog to the right
The node creation dialog was improved greatly with #3396 which is great. However we noticed a mistake we repeatedly ran into with the new design.

Since the filter is now on the left and has the cursor automatically placed in the field one can very easily confuse the select nodetype dialog with the create-document dialog and intuitively  starts typing the intended document title instead of selecting the nodetype. This quickly leads to hiding of all document types and can only be fixed by emptying the filter field again.

I confirmed the issue with some participants on the Neos sprint and we think that it can be fixed quite easily by moving the filter to the right and limiting the size.

#### Current behavior
Neos_news_-_Neos_Demo_Site_und___Marc_Henry_Schultz_-_Neos_-_Slack_und_sitegeist_Chat

#### Suggestion
Home_-_Neos_Demo_Site

### Affected Versions

Neos: 8.3

mhsdesign:
fixes: #3419

image

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
+++ b/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js
@@ -335,16 +335,16 @@ export default class SelectNodeType extends PureComponent {
                 id="neos-SelectNodeTypeDialog"
                 >
                 <div onKeyDown={this.handleKeyDown} role="searchbox" className={style.nodeTypeDialogHeader} key="nodeTypeDialogHeader">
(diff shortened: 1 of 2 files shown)


crydotsnake (approved):
Looks good! 😀

ahaeslich (approved):
by reading

PR #3429, opened 2023-03-12
TASK: consistently use public readonly properties in value objects, instead of getters

skurfuerst:
This is the compatibility change for https://github.com/neos/neos-development-collection/pull/4095


skurfuerst:
related: https://github.com/neos/neos-development-collection/pull/4095

PR #3433, opened 2023-03-17
BUGFIX: flowQueryAction (js endpoint for `q()`) handle non-existing nodes gracefully

mhsdesign:
if a plugin uses the flowquery backend api via
```ts

const {q} = backend.get();

q([siteNodeContextPath]).find("#123").getForTree()
```

but if the node-identifier `123` doesnt exist the NodeInfoHelper (`getForTree`) will fail.

image

**What I did**

we filter empty items out of the result of the flowquery, and only then pass it to the NodeInfoHelper, which will result in a nice empty json response.

**How I did it**

**How to verify it**

### DIFF
--- a/Classes/Controller/BackendServiceController.php
+++ b/Classes/Controller/BackendServiceController.php
@@ -533,13 +533,13 @@ class BackendServiceController extends ActionController
         $result = [];
         switch ($finisher['type']) {
             case 'get':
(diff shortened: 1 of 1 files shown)


nezaniel:
Looks good by reading.
Mainly out of curiosity: Where do the styleci issues come from?

mhsdesign:
style ci is going maniac since a few days, i already complained in slack https://neos-project.slack.com/archives/C04Q0TC15/p1679047298696289

Issue #3241, opened 2022-11-01
Feature: Document allow next previous navigation via browser history


mhsdesign:
### Steps to Reproduce

1. go to `https://localhost:8081/neos`
2. In the backend click on another documentNode 
3. expect he url to change
4. goto previous page via browser history

#### Expected behavior

goes to previous selected document node

#### Actual behavior

goes back to the login (as this is the site you entered first) `https://localhost:8081/neos`

[What actually happened]

### Affected Versions

Neos:

UI: 8.0.0

You can get this information by running `composer show` or using the package management module within Neos.
The UI version can also be found inside the left sidebar after version 1.0.3.

mhsdesign:
also middle mouse button / strg+click on document nodes ...

SPAs are never super accessibile https://github.com/neos/neos-ui/issues/1173

Issue #2938, opened 2021-08-09
Document Tree Keyboard Navigation


sbruggmann:
### Feature-Request

Navigating the Document Tree via Keyboard like PhpStorm would be very handy power users or useful for people who have difficulties with precise cursor handling.

- Arrow up: Previous sibling or previous sibling of the current parent
- Arrow down: Next sibling or next sibling of the current parent
- Arrow left: Close directory and jump out of them
- Arrow right: Open directory and jump into them
- Enter/Return: "Click" the current active document

PR #3440, opened 2023-03-22
FEATURE: Make document tree clickable

### LINKED ISSUE #3353: Feature: Make documents in pagetree into a link
### Description
This is based on feedback from Editors working with Neos.
 Editors sometimes want to work on multiple pages in parallel. For that, they open different pages in different browser tabs. 
Right now to open pages in the backend in new tabs, they have to open a new tab and navigate to the backend by hand. 

To make this workflow easier I would suggest making the documents in the page tree on the left side into links. 
This would allow backend users to open the pages with right click and "open page in new link".

pKallert:
**What I did**
I changed the document tree elements into links so that we can open them with right/middle click 

**How I did it**
Since I was not sure how to get the URL (neos/content) from the Node I tried around a little bit and changed the NodeInfoHelper to add the backendUri. I tried to adhere as much as possible to the currently existing "uri" property in the node. If there is a better place for this, I am ready to learn 😄 

Some googling told me that a tags without href are valid so I chose to exchange all of the spans into a-tag. Works still as expected. 

**How to verify it**

Links now show up in the left document tree: 

Bildschirmfoto 2023-03-22 um 16 15 50

closes #3353

### DIFF
--- a/Classes/Fusion/Helper/NodeInfoHelper.php
+++ b/Classes/Fusion/Helper/NodeInfoHelper.php
@@ -137,6 +137,7 @@ class NodeInfoHelper implements ProtectedContextAwareInterface
 
         if ($controllerContext !== null) {
             $nodeInfo = array_merge($nodeInfo, $this->getUriInformation($node, $controllerContext));
+            $nodeInfo = array_merge($nodeInfo, $this->getBackendUriInformation($node, $controllerContext));
             if ($controllerContext->getRequest()->hasArgument('presetBaseNodeType')) {
                 $presetBaseNodeType = $controllerContext->getRequest()->getArgument('presetBaseNodeType');
             }
@@ -204,6 +205,30 @@ class NodeInfoHelper implements ProtectedContextAwareInterface
         } catch (\Neos\Neos\Exception $exception) {
             // Unless there is a serious problem with routes there shouldn't be an exception ever.
             $nodeInfo['uri'] = '';
+        }
+
+        return $nodeInfo;
+    }
+
+    /**
+     * Get the "backendUri" for the given node
+     *
+     * @param NodeInterface $node
+     * @param ControllerContext $controllerContext
+     * @return array
+     */
+    protected function getBackendUriInformation(NodeInterface $node, ControllerContext $controllerContext): array
+    {
+        $nodeInfo = [];
+        if (!$node->getNodeType()->isOfType($this->documentNodeTypeRole)) {
+            return $nodeInfo;
+        }
+
+        try {
+            $nodeInfo['backendUri'] = $this->backendUri($node, $controllerContext);
+        } catch (\Neos\Neos\Exception $exception) {
+            // Unless there is a serious problem with routes there shouldn't be an exception ever.
+            $nodeInfo['backendUri'] = '';
         }
 
         return $nodeInfo;
@@ -414,6 +439,31 @@ class NodeInfoHelper implements ProtectedContextAwareInterface
 
         // Create an absolute URI
         return $this->linkingService->createNodeUri($controllerContext, $node, null, null, true);
+    }
+
+    /**
+     * Create a Backend uri (/neos/content[...])
+     *
+     * @param NodeInterface|null $node
+     * @param ControllerContext $controllerContext
+     * @return string
+     */
+    public  function backendUri(NodeInterface $node = null, ControllerContext $controllerContext)
+    {
+        if ($node === null) {
+            // This happens when the document node is not published yet
+            return '';
+        }
+
+        $request = $controllerContext->getRequest()->getMainRequest();
+        $uriBuilder = clone $controllerContext->getUriBuilder();
+        $uriBuilder->setRequest($request);
+
+        return $uriBuilder
+            ->reset()
+            ->setCreateAbsoluteUri(true)
+            ->setFormat('html')
+            ->uriFor('index', ['node' => $node], 'Backend', 'Neos.Neos.Ui');
     }
 
     /**
--- a/Classes/Fusion/Helper/NodeInfoHelper.php
+++ b/Classes/Fusion/Helper/NodeInfoHelper.php
@@ -137,7 +137,6 @@ class NodeInfoHelper implements ProtectedContextAwareInterface
 
         if ($controllerContext !== null) {
             $nodeInfo = array_merge($nodeInfo, $this->getUriInformation($node, $controllerContext));
-            $nodeInfo = array_merge($nodeInfo, $this->getBackendUriInformation($node, $controllerContext));
             if ($controllerContext->getRequest()->hasArgument('presetBaseNodeType')) {
                 $presetBaseNodeType = $controllerContext->getRequest()->getArgument('presetBaseNodeType');
             }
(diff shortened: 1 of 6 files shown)


mhsdesign:
Nice one, noticed that also here: https://github.com/neos/neos-ui/issues/3241#issuecomment-1298771076 

Im not sure why we need another uri (the backenduri) on the node (seems like a bad idea performance wise and also technically) but have to look into this deeper

Classes/Fusion/Helper/NodeInfoHelper.php
```diff
@@ -137,6 +137,7 @@ public function renderNodeWithMinimalPropertiesAndChildrenInformation(NodeInterf
 
         if ($controllerContext !== null) {
             $nodeInfo = array_merge($nodeInfo, $this->getUriInformation($node, $controllerContext));
+            $nodeInfo = array_merge($nodeInfo, $this->getBackendUriInformation($node, $controllerContext));
```
Sebobo: We have to be careful here to create more data. I imagine this could affect performance for large uncollapsed document trees?

Can we test/verify that?

Classes/Fusion/Helper/NodeInfoHelper.php
```diff
@@ -209,6 +210,30 @@ protected function getUriInformation(NodeInterface $node, ControllerContext $con
         return $nodeInfo;
     }
 
+    /**
+     * Get the "backendUri" for the given node
+     *
+     * @param NodeInterface $node
```
Sebobo: You don't need to add duplicate type hints

Classes/Fusion/Helper/NodeInfoHelper.php
```diff
@@ -416,6 +441,31 @@ public function uri(NodeInterface $node = null, ControllerContext $controllerCon
...
     }
 
+    /**
+     * Create a Backend uri (/neos/content[...])
+     *
+     * @param NodeInterface|null $node
+     * @param ControllerContext $controllerContext
+     * @return string
+     */
+    public function backendUri(NodeInterface $node = null, ControllerContext $controllerContext)
```
Sebobo: Return type missing and duplicate types in comment.

Sebobo (changes requested):

pKallert:
> Nice one, noticed that also here: [#3241 (comment)](https://github.com/neos/neos-ui/issues/3241#issuecomment-1298771076)
> 
> Im not sure why we need another uri (the backenduri) on the node (seems like a bad idea performance wise and also technically) but have to look into this deeper

@mhsdesign 
I thought adding it to the node would be cleaner and reusable but if there is concern about the impact on the performance I can also use the solution from the alt+click from here: https://github.com/neos/neos-ui/blob/8.3/packages/neos-ui/src/Containers/LeftSideBar/NodeTree/index.js#L62

markusguenther:
The end to end tests failed because the fluid version is not working. https://github.com/TYPO3/Fluid/commit/24f4494083c8d304680e4c9c38667dff33720dd4

But guess a new build would work since today 2.7.4 has been released.

packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js
```diff
@@ -380,6 +383,12 @@ export default class Node extends PureComponent {
         onNodeFocus(node.contextPath, metaKeyPressed, altKeyPressed, shiftKeyPressed);
         onNodeClick(srcWithBaseNodeType, node.contextPath, metaKeyPressed, altKeyPressed, shiftKeyPressed);
     }
+
+    createDirectNodeLink = () => {
+        const {node} = this.props;
+        const loc = window.location;
+        return loc.protocol + '//' + loc.hostname + (loc.port ? ':' + loc.port : '') + loc.pathname + '?node=' + node.contextPath;
```
mhsdesign:
```suggestion
        const uri = new URL(window.location.href);
        uri.searchParams.set("node", node.contextPath);
        return uri.toString();
```

i would suggest to use the url class when manipulating uris, makes things much easier and bulletproof ^^

packages/react-ui-components/src/Tree/node.js
```diff
@@ -222,16 +224,17 @@ export class Header extends PureComponent {
...
-                                <span
+                                <a
                                     {...rest}
                                     id={labelIdentifier}
                                     className={theme.header__label}
-                                    onClick={onLabelClick}
+                                    onClick={onLabelClick ? onLabelClick : (directLink ? this.onDirectLinkClick : undefined)}
                                     data-neos-integrational-test="tree__item__nodeHeader__itemLabel"
                                     role="treeitem"
+                                    href={directLink ? directLink : undefined}
```
mhsdesign:
```suggestion
                                    href={directLink}
```

packages/react-ui-components/src/Tree/node.js
```diff
@@ -222,16 +224,17 @@ export class Header extends PureComponent {
                                         <IconComponent icon={icon || 'question'} label={iconLabel} />
                                     }
                                 </div>
-                                <span
+                                <a
                                     {...rest}
                                     id={labelIdentifier}
                                     className={theme.header__label}
-                                    onClick={onLabelClick}
+                                    onClick={onLabelClick ? onLabelClick : (directLink ? this.onDirectLinkClick : undefined)}
```
mhsdesign: this line is a bit hard to read ... 

just a thought. what would you think about:
```js
const linkHandlingProps = directLink
  ? {
  href: directLink,
  onClick: (e) => e.preventDefault()
  }
  : {
  onClick: onLabelClick
  };

return <a
 other="props"
 {...linkHandlingProps}
/>

```

mhsdesign: @markusguenther ping pong

markusguenther:
Will check it later today

mhsdesign:
question is, what to do with the legacy onClick logic?

`onClick`
https://github.com/neos/neos-ui/blob/4497f9465f77ec9d083459ea77d4a2917ff95598/packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js#L363-L382

`onNodeClick`
https://github.com/neos/neos-ui/blob/4497f9465f77ec9d083459ea77d4a2917ff95598/packages/neos-ui/src/Containers/LeftSideBar/NodeTree/index.js#L59-L79

markusguenther:
While testing and fixing the End-to-end tests I stumbled over the issue that the multiple move of nodes is broken. Will try to fix that.

Only via Toolbar it is not working yet.
https://app.circleci.com/pipelines/github/neos/neos-ui/3370/workflows/48dac434-bd95-42bf-92bc-67948b9a2c53/jobs/15353

markusguenther:
Finally, the tests are green again :)

https://app.circleci.com/pipelines/github/neos/neos-ui/3376/workflows/d0c3291b-af25-4b80-8797-c82d65403d8b

markusguenther (approved):

markusguenther:
> question is, what to do with the legacy onClick logic?
> 
> `onClick`
> 
> https://github.com/neos/neos-ui/blob/4497f9465f77ec9d083459ea77d4a2917ff95598/packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js#L363-L382
> 
> `onNodeClick`
> 
> https://github.com/neos/neos-ui/blob/4497f9465f77ec9d083459ea77d4a2917ff95598/packages/neos-ui/src/Containers/LeftSideBar/NodeTree/index.js#L59-L79

This logic is still used on a higher layer for the multi operations on nodes. 
Have tested the feature and made the tests work again.

mhsdesign:
> > question is, what to do with the legacy onClick logic?
> > `onClick`
> > https://github.com/neos/neos-ui/blob/4497f9465f77ec9d083459ea77d4a2917ff95598/packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js#L363-L382
> > 
> > `onNodeClick`
> > https://github.com/neos/neos-ui/blob/4497f9465f77ec9d083459ea77d4a2917ff95598/packages/neos-ui/src/Containers/LeftSideBar/NodeTree/index.js#L59-L79
> 
> This logic is still used on a higher layer for the multi operations on nodes. Have tested the feature and made the tests work again.

This legacy second way now introduces a lot of complexity — and we must clean it up please 😅 (or share the logic, comment it and explain how things work and why this would be necessary) otherwise it’s really hard for future people to make sense of the code

This pr is only okay as intermediate step …

markusguenther:
> This legacy second way now introduces a lot of complexity — and we must clean it up please 😅 (or share the logic, comment it and explain how things work and why this would be necessary) otherwise it’s really hard for future people to make sense of the code
> 
> This pr is only okay as intermediate step …

Indeed ... we have an issue for making the tree accessible and as I watched into it some months ago I realized that it will need a bit further adjustments on the markup to achieve that. With the change from Paula, we have a first step.

https://github.com/neos/neos-ui/issues/2938

PR #3444, opened 2023-03-27
TASK: Restrict what is published to npm

mhsdesign:
**Why**

It is a good manner to only publish to NPM the dist files to reduce package size, but also since there is nothing to gain from using the source files over dist.


Sebobo:
Can you add a short description in the issue text for the "why"?

mhsdesign:
👍

PR #3451, opened 2023-03-30
FEATURE: Improve user experience for Dimension Switcher

### LINKED ISSUE #3413: FEATURE: Usability Improvements for Dimensions Menu
### Description

Currently, in Neos, the dimensions' menu can be somewhat cumbersome and difficult to use. Some editors have suggested several changes to improve the usability of the dimensions' menu in Neos 8.3.

Firstly, if there is only one dimension, users must click through two menus to open the dimensions list. This can be time-consuming and inefficient. To address this, we propose that the first and only dimension should either open by default when clicking the dimensions switcher, or that the secondary dropdown logic should be switched to only show the one list of the dimension. This would streamline the process of accessing dimensions and make it more efficient for users.

Secondly, to check if a document already exists in a different language, editors must click through all dimensions. This can be frustrating and time-consuming, especially for larger projects with many dimensions. We propose adding some kind of "active"/"inactive" color coding in the dimensions' dropdown to show if clicking on the link would create a new document. This would make it easier and more efficient for editors to navigate between dimensions and avoid creating duplicate documents.

Lastly, editors sometimes want to open a different dimension in a new tab. Currently, they must manually open a new tab and then switch to the desired dimension. We propose adding a feature where users can use right click/center click to open the document (if it exists) in a new tab. This would save time and reduce the number of clicks required to navigate between dimensions.

Additional specifications for these features will need to be discussed and defined, but we believe they would be valuable additions to Neos 8.3, improving the usability of the dimensions menu and enhancing the overall user experience.

This is a feature idea from the community brainstorming thread on discuss:
https://discuss.neos.io/t/attention-neos-community/6206/9

markusguenther:
**What I did**
To improve the UX, we want to have just one layer when we have just one dimension configured.
Currently, the Dimension Selector is inside a dropdown.

**How I did it**
I separated the DimensionSelector into an own component and added a property to render the Selector without the label.
The DimensionSwitcher now renders a single DimensionSelector if we have only one dimension and for multiple Dimensions we use the old way.

**How to verify it**
Test the Backend with one and multiple dimensions.

**Open To-dos**

- [x] Adjust the tests
- [x] Add a Dimension Icon for the single dimension
- [x] little CSS tweaks

Related: #3413

### DIFF
--- a/packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/DimensionSelector.js
+++ b/packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/DimensionSelector.js
@@ -1,11 +1,9 @@
 import React, {PureComponent} from 'react';
 import PropTypes from 'prop-types';
-import Icon from '@neos-project/react-ui-components/src/Icon/';
 import SelectBox from '@neos-project/react-ui-components/src/SelectBox/';
 import style from './style.module.css';
 import {$get, $transform} from 'plow-js';
 import mapValues from 'lodash.mapvalues';
-import I18n from '@neos-project/neos-ui-i18n';
 import sortBy from 'lodash.sortby';
 import {neos} from '@neos-project/neos-ui-decorators';
 import DimensionSelectorOption from './DimensionSelectorOption';
@@ -25,7 +23,7 @@ export default class DimensionSelector extends PureComponent {
         dimensionName: PropTypes.string.isRequired,
         isLoading: PropTypes.bool,
         onSelect: PropTypes.func.isRequired,
-        showOnlyDimensionDropdown: PropTypes.bool,
+        showDropDownHeaderIcon: PropTypes.bool,
 
         i18nRegistry: PropTypes.object.isRequired
     };
@@ -34,7 +32,7 @@ export default class DimensionSelector extends PureComponent {
         searchTerm: ''
     };
 
-    renderDimensionDropdown = () => {
+    render() {
         const {
             activePreset,
             isLoading,
@@ -42,7 +40,7 @@ export default class DimensionSelector extends PureComponent {
             dimensionName,
             onSelect,
             presets,
-            showOnlyDimensionDropdown
+            showDropDownHeaderIcon
         } = this.props;
 
         const presetOptions = mapValues(
@@ -72,7 +70,7 @@ export default class DimensionSelector extends PureComponent {
                 onValueChange={onPresetSelect}
                 value={activePreset}
                 allowEmpty={false}
-                headerIcon={showOnlyDimensionDropdown ? this.props.icon : null}
+                headerIcon={showDropDownHeaderIcon ? this.props.icon : null}
                 displaySearchBox={sortedPresetOptions.length >= 10}
                 searchOptions={searchOptions(this.state.searchTerm, sortedPresetOptions)}
                 onSearchTermChange={this.handleSearchTermChange}
@@ -83,27 +81,6 @@ export default class DimensionSelector extends PureComponent {
                 className={style.dimensionSwitcherDropDown}
             />
         )
-    }
-
-    render() {
-        const {
-            icon,
-            dimensionLabel,
-            dimensionName,
-            showOnlyDimensionDropdown
-        } = this.props;
-
-        const dimensionDropdown = this.renderDimensionDropdown();
-
-        return showOnlyDimensionDropdown === true ? dimensionDropdown : (
-            <li key={dimensionName} className={style.dimensionCategory}>
-                <div className={style.dimensionLabel}>
-                    <Icon icon={icon} padded="right" className={style.dimensionCategory__icon}/>
-                    <I18n id={dimensionLabel}/>
-                </div>
-                {dimensionDropdown}
-            </li>
-        );
     }
 
     handleSearchTermChange = searchTerm => {
--- a/packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/DimensionSelector.js
+++ b/packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/DimensionSelector.js
@@ -1,0 +1,110 @@
+import React, {PureComponent} from 'react';
+import PropTypes from 'prop-types';
+import Icon from '@neos-project/react-ui-components/src/Icon/';
+import SelectBox from '@neos-project/react-ui-components/src/SelectBox/';
+import style from './style.module.css';
+import {$get, $transform} from 'plow-js';
+import mapValues from 'lodash.mapvalues';
+import I18n from '@neos-project/neos-ui-i18n';
+import sortBy from 'lodash.sortby';
+import {neos} from '@neos-project/neos-ui-decorators';
+import DimensionSelectorOption from './DimensionSelectorOption';
+
+const searchOptions = (searchTerm, processedSelectBoxOptions) =>
+    processedSelectBoxOptions.filter(option => option.label && option.label.toLowerCase().indexOf(searchTerm.toLowerCase()) !== -1);
+
+@neos(globalRegistry => ({
+    i18nRegistry: globalRegistry.get('i18n')
+}))
+export default class DimensionSelector extends PureComponent {
+    static propTypes = {
+        icon: PropTypes.string.isRequired,
+        dimensionLabel: PropTypes.string.isRequired,
+        presets: PropTypes.object.isRequired,
+        activePreset: PropTypes.string.isRequired,
+        dimensionName: PropTypes.string.isRequired,
+        isLoading: PropTypes.bool,
+        onSelect: PropTypes.func.isRequired,
+        showOnlyDimensionDropdown: PropTypes.bool,
+
+        i18nRegistry: PropTypes.object.isRequired
+    };
+
+    state = {
+        searchTerm: ''
+    };
+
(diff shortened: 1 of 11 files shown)


markusguenther:
Eventually, also migrate to TS... depends on the amount of time as the Feature Freeze is close and we can do it also later.

markusguenther:
**Current State:**
Screenshot 2023-03-30 at 11 56 10

markusguenther:
**The tests work now:**

- Multi-Dimensional https://app.saucelabs.com/tests/20c08d23c08843d19c99d270e609f767
- Single Dimension https://app.saucelabs.com/tests/bcf885d4af5346cd886fbf5ad6e9b422

markusguenther:
Also added the Icon. So just some CSS glitches left, but the feature itself is done.
Screenshot 2023-03-30 at 14 46 52

markusguenther:
The only thing that puzzles me currently is the positioning of the icon. Sadly, the Selectbox can not only render the Icon in the header. When you add the icon to the options, all items will have the icon.

Maybe that would be a nice flag for the select box to render the icon just in the header.

But I think the CSS stuff can be perhaps also be a thing after feature freeze.

markusguenther:
Ok, just did it and added a new prop to be able to render just an icon inside the header.
If this is not wanted, we can remove the two commits again.

But that solves a lot of CSS issues and would be nice to have it in other places as well.

packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/index.js
```diff
@@ -142,6 +64,11 @@ export default class DimensionSwitcher extends PureComponent {
         loadingPresets: {}
     };
 
+    getDimensionIcon = (dimensionName, contentDimensionsObject) => {
+        const dimensionConfiguration = contentDimensionsObject[dimensionName];
+        return $get('icon', dimensionConfiguration) && $get('icon', dimensionConfiguration);
```
Sebobo: I don't "$get" this line?

packages/react-ui-components/src/SelectBox_Option_SingleLine/selectBox_Option_SingleLine.js
```diff
@@ -13,21 +13,27 @@ class SelectBox_Option_SingleLine extends PureComponent {
...
         const isDisabled = disabled || option.disabled;
 
         const finalClassNames = mergeClassNames({
             [className]: className
         });
 
+        // We want to show the icon of the option when it is defined
+        // If we have no option icon but showIcon is true, we want to show the icon of the SelectBox
+        // If we have no option icon and showIcon is false, we want to show no icon
+        const visibleIcon = option.icon ? option.icon : (showIcon ? icon : null);
```
Sebobo: What about icon is set but showIcon is false?
Sounds confusing?

packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/DimensionSelector.js
```diff
@@ -0,0 +1,112 @@
...
+export default class DimensionSelector extends PureComponent {
+    static propTypes = {
+        icon: PropTypes.string.isRequired,
+        dimensionLabel: PropTypes.string.isRequired,
+        presets: PropTypes.object.isRequired,
+        activePreset: PropTypes.string.isRequired,
+        dimensionName: PropTypes.string.isRequired,
+        isLoading: PropTypes.bool,
+        onSelect: PropTypes.func.isRequired,
+        showOnlyDimensionDropdown: PropTypes.bool,
```
Sebobo: The name of the prop doesn't really tell much about what is the alternative. 

Wouldn't it be better to leave the list item wrapping out of this component and decide the wrapping in the DimensionSwitcher?

Sebobo (changes requested):
Love the feature, left some comments

packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/index.js
markusguenther: It is there since 7 years... https://github.com/neos/neos-ui/blob/e4080d6bb1a14a2c5eb843109ee82af38b84c8cf/packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/index.js#L250

Just moved it out and did not think about it that much. But will check that and adapt it.

Sebobo: Yeah didn't check whether each part was new. But if we anyway touch it we can improve the internals a bit.

packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/DimensionSelector.js
markusguenther: Changed that now. Thanks for the feedback.

packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/index.js
markusguenther: This is gone now ;)

packages/react-ui-components/src/SelectBox_Option_SingleLine/selectBox_Option_SingleLine.js
markusguenther: Adjusted this part as well.

markusguenther:
> Love the feature, left some comments

Thanks for the valuable feedback. I introduced the `headerIcon` prop later in my local development and with that, I could drop the `showIcon` props now. So, it is less expanded and has the same value.

mhsdesign:
the spacing to the arrow was previously a bit prettier ;) (noticed it even without comparison)

image


image


thanks for this pr @markusguenther this was previously hard to use ^^

mhsdesign:
is it intentional that we still show the dropdown, if there is only one dimension?

image


image

crydotsnake:
> is it intentional that we still show the dropdown, if there is only one dimension?
> 
> image
> 
> image

As an idea, we could disable the dropdown if there is really only one dimension available.

crydotsnake (approved):
In functionality point of view it is working great!

markusguenther:
I already tried to remove the padding right today but was not successful yet. Try it later this evening when I am back home.

markusguenther:
> > is it intentional that we still show the dropdown, if there is only one dimension?
> > image
> > image
> 
> As an idea, we could disable the dropdown if there is really only one dimension available.


Don't get the point because this PR shows a single drop down when we have only one dimension, no dropdown for no dimension and the old behavior if we have more than one dimension.

**Single:**
Screenshot 2023-04-02 at 21 13 14

**Multi:**
Screenshot 2023-04-02 at 21 12 45

Screenshot 2023-04-02 at 21 12 38

markusguenther:
Ah now i see your point... hmm, guess if you have just one single dimension value you would normally create no dimension and then would see no drop down at all.

In your case it does not make so much sense.

crydotsnake:
> > > is it intentional that we still show the dropdown, if there is only one dimension?
> > > image
> > > image
> > 
> > 
> > As an idea, we could disable the dropdown if there is really only one dimension available.
> 
> Don't get the point because this PR shows a single drop down when we have only one dimension, no dropdown for no dimension and the old behavior if we have more than one dimension.
> 
> **Single:** Screenshot 2023-04-02 at 21 13 14
> 
> **Multi:** Screenshot 2023-04-02 at 21 12 45
> 
> Screenshot 2023-04-02 at 21 12 38

I think @mhsdesign's point was, that you can still use the dropdown menu if you have only one dimension available. And with only one dimension avaialble, i dont know if we still need the dropdown. Because we have no other dimension that we could switch too.

mhsdesign:
> Ah now i see your point... hmm, guess if you have just one single dimension value you would normally create no dimension and then would see no drop down at all.
> 
> In your case it does not make so much sense.

uups yes sorry i was brain dumb - too much sun

you can leave it as you proposed ^^

crydotsnake:
> Ah now i see your point... hmm, guess if you have just one single dimension value you would normally create no dimension and then would see no drop down at all.
> 
> In your case it does not make so much sense.

Good point.. I have not thought about that 😅

markusguenther:
Bit confused that the linter fails in a file I did not change 🤔

To be honest, your point is also valid, but I am uncertain if it were more confusion to configure a dimension and then see nothing in the UI. For this case, I would need to add another component that is no dropdown or button and just shows the dimension. So that you can see it but not change it. 

An easy workaround would be to disable the select box if we have just one item.
Then it would look like that.

Screenshot 2023-04-02 at 21 41 48

crydotsnake:
> An easy workaround would be to disable the select box if we have just one item.
Then it would look like that.

Hm. Wouldn't that be a bit confusing for the editor? Otherwise, this would be a good workaround IMO.

crydotsnake:
> > An easy workaround would be to disable the select box if we have just one item.
> > Then it would look like that.
> 
> Hm. Wouldn't that be a bit confusing for the editor? Otherwise, this would be a good workaround IMO.

@markusguenther If you still fine with that, we can use your idea. So if only one dimension is available, the checkbox will be disabled.

It shouldn't be confusing for the editor. Because in general the editor should know how many dimensions the website has where he is working on.

Issue #3097, opened 2022-04-05
ES-Module Plugin cant be loaded *


mhsdesign:
\* unless brute forced via AOP in [StyleAndJavascriptInclusionService](https://github.com/neos/neos-ui/blob/6ce277470258599c1d29f9ec2f5983e7e44d4338/Classes/Domain/Service/StyleAndJavascriptInclusionService.php#L100
)

### Description

Imagine you have a nice ESM Plugin and try to inject it via:

```yaml
Neos.Neos.Ui.resources.javascript.'My.EsmPlugin':
  resource: 'foo/bar.js'
```
now we get an error about `import` cannot be used because the script tag doesnt tell its an esm.

the problem is, that no attributes for the script tag are allowed except `defer`
```yaml
Neos.Neos.Ui.resources.javascript.'My.EsmPlugin':
  resource: 'foo/bar.js'
  defer: true # works
  type: 'module' # doesnt work
```

### Steps to Reproduce

write
`Neos.Neos.Ui.resources.javascript.'My.Plugin'.type: 'module'`

#### Expected behavior

a script tag with `<script type="module">` is rendered analog to the `defer` option

#### Actual behavior
 
the option is ignored

### Affected Versions

UI: since ever

### Solution

we would need to not only look out for the `defer` key but also for the key `type`
https://github.com/neos/neos-ui/blob/6ce277470258599c1d29f9ec2f5983e7e44d4338/Classes/Domain/Service/StyleAndJavascriptInclusionService.php#L100

also we could consider having an actual future proof `attributes` array which will be just rendered like in Neos.Fusion:Tag
i implemented this in https://github.com/mhsdesign/MhsDesign.ProposalNeosUiEsmPluginLoader

mhsdesign:
Fair enough if you ask, where did you forge an es module neos ui plugin? Via esbuild.
https://github.com/mhsdesign/esbuild-neos-ui-extensibility

mhsdesign:
I made a proposal package: https://github.com/mhsdesign/MhsDesign.ProposalNeosUiEsmPluginLoader

mhsdesign:
I would rather suggest to use native esm dynamic imports (supported with es2020 and neos 8.2 with esbuild 🎉)

PR #3456, opened 2023-04-02
TASK: #3097 allow module script plugins

mhsdesign:
**What I did**
solves #3097

Allows ES-Module Plugins to be loaded in the Neos.Ui, by allowing `<src type="module" />`

inlines the package: https://github.com/mhsdesign/MhsDesign.ProposalNeosUiEsmPluginLoader

for fully compatibility with type=esm plugins build via our new esbuild extension point

Usage:

`Neos.Neos.Ui.resources.javascript` and `stylesheets`
can make use of the proposed `attributes` array.

```yaml
Neos:
  Neos:
    Ui:
      resources:
        javascript:
          'My.Cool:Plugin':
            resource: 'resource://My.Cool/Public/NeosUserInterface/Plugin.js'
            # legacy first level attribute 'defer'
            # defer: true

            # NEW Attributes!
            attributes:
              type: "module"
              defer: true
              any: "thing"

        stylesheets:
          'My.Cool:Plugin':
            # NEW Attributes!
            attributes:
              any: "thing"
```

**How I did it**

**How to verify it**

### DIFF
--- a/Classes/Domain/Service/StyleAndJavascriptInclusionService.php
+++ b/Classes/Domain/Service/StyleAndJavascriptInclusionService.php
@@ -61,15 +61,15 @@ class StyleAndJavascriptInclusionService
 
     public function getHeadScripts()
     {
-        return $this->build($this->javascriptResources, function ($uri, $defer) {
-            return '<script src="' . $uri . '" ' . $defer . '></script>';
+        return $this->build($this->javascriptResources, function ($uri, $additionalAttributes) {
+            return '<script src="' . $uri . '" ' . $additionalAttributes . '></script>';
         });
     }
 
     public function getHeadStylesheets()
     {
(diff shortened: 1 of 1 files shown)


mhsdesign:
## Ausblick

yes i know our current loading of plugins and ui code (by injecting script tags an polluting the global namespace (and relying a lot on timing)) is say - not the yellow of the egg.

But before we offer a better / new way i think we should at least allow the plugin authors to create esm plugins already `<script type=module`


Thats also the reason why im targeting 7.3, so everyone can profit already

markusguenther:
This seems to me more a feature than a task or not?

mhsdesign:
Sure but a feature we don’t upmerge xD

mhsdesign:
Sure but a feature we don’t upmerge xD

PR #3457, opened 2023-04-02
TASK: Deprecate plow-js

### LINKED ISSUE #3425: TASK: 8.3 Deprecate `plow` (`$get`, `/\$[a-z]+\`) in plugins
### Description

PlowJS is being removed from the Neos Ui codebase, as the JS world moved on and now strict typing with Typescript is important and also modern JavaScript allows us to replace Plow's $get safe access on nullable values via optional chaining. This restructuring will affect Neos Ui plugins in the future, as we will remove plow JS from the Neos plugin api in Neos 9.0. For now we deprecate the use and log a warning in dev context.

Related: Extraction of plow js from the Neos Ui https://github.com/neos/neos-ui/pull/3306

### Migration in plugins:

To get rid of the console warnings:
image

You have two options, when using the plugin api directly for example with esbuild: [@neos-project/neos-ui-extensibility](https://github.com/neos/neos-ui/blob/8.3/packages/neos-ui-extensibility)

1: use the js-equivalent code like optional chaining etc see https://github.com/neos/neos-ui/pull/3306 for inspiration ;)
2: require `plow-js` as npm dependency and actually bundle it. To do that we must exclude it from the map:

```diff
- const extensibilityMap = require("@neos-project/neos-ui-extensibility/extensibilityMap.json");
+ const {'plow-js': _, ...extensibilityMap} = require("@neos-project/neos-ui-extensibility/extensibilityMap.json");
```

If youre using the webpack build stack: [@neos-project/neos-ui-extensibility-webpack-adapter](https://github.com/neos/neos-ui/tree/8.3/packages/neos-ui-extensibility-webpack-adapter) you currently have no option, to tell the bundler to use your plow version that you locally installed instead of the alias. Once 9.0 is released and we drop the alias in the build stack, the build system will pick up your local plow js package and bundle it. So just remember to rebuild the plugin once 9.0 is released.

### Affected Versions

Neos UI: 8.3

mhsdesign:
solves #3425 

to-do:

image

might generate too much noise ... (but only logging once won't help, as each plugin should get at leas one warning ... but it's not that easy to find out the caller (except looking into the stack trace which I didn't want to do for now))

### DIFF
--- a/packages/neos-ui/src/deprecated-plow.js
+++ b/packages/neos-ui/src/deprecated-plow.js
@@ -1,3 +1,4 @@
+/* eslint-disable */ // IDGAF
 import * as plowOriginal from 'plow-js';
 
 function augmentAndDeprecatedFunction(functionName, functionImplementation) {
--- a/packages/neos-ui/src/deprecated-plow.js
+++ b/packages/neos-ui/src/deprecated-plow.js
@@ -1,0 +1,24 @@
+import * as plowOriginal from 'plow-js';
+
+function augmentAndDeprecatedFunction(functionName, functionImplementation) {
+    return function() {
+        if (window?.neos?.systemEnv?.startsWith('Development')) {
+            console.warn(`Plow.js "${functionName}" is deprecated! See: https://github.com/neos/neos-ui/issues/3425`);
+        }
+        return functionImplementation.apply(this, arguments);
+    };
+}
+
+const deprecatedPlow = {};
+
+for (const functionName in plowOriginal) {
+    const functionImplementation = plowOriginal[functionName];
+
+    if (typeof functionImplementation === 'function') {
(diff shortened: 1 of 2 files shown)


markusguenther:
Nice Idea @mhsdesign 👍

packages/neos-ui/src/deprecated-plow.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable */ // IDGAF
+import * as plowOriginal from 'plow-js';
+
+function augmentAndDeprecatedFunction(functionName, functionImplementation) {
+    return function() {
+        if (window?.neos?.systemEnv?.startsWith('Development')) {
```
mhsdesign, line 6: fyi we do have some kind of log utils, but they are ~not~ rarely used and dont work 😂 

https://github.com/neos/neos-ui/issues/3458

markusguenther (approved):

PR #3459, opened 2023-04-03
BUGFIX: Some 9.0 fixes

bwaidelich:
Some more replacements of `NodeInterface` => `Node` and `getContextPath()` => `serializeForUri()`

### DIFF
--- a/Classes/Domain/Model/Feedback/Operations/Redirect.php
+++ b/Classes/Domain/Model/Feedback/Operations/Redirect.php
@@ -1,26 +1,27 @@
 <?php
 namespace Neos\Neos\Ui\Domain\Model\Feedback\Operations;
 
-use Neos\ContentRepository\Domain\Model\NodeInterface;
+use Neos\ContentRepository\Core\Projection\ContentGraph\Node;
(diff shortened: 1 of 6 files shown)


ahaeslich (approved):
Fine by reading + we shouldn't find any `NodeInterface` in this package anymore.

bwaidelich:
> we shouldn't find any NodeInterface in this package anymore.

There are still some. For example in the [NodeTreeBuilder](https://github.com/neos/neos-ui/blob/483461eb680c77fd8ea4070c2bc5c3e831092e50/Classes/Domain/Service/NodeTreeBuilder.php) – I tried to address those but I failed to understand the code or even what this class does exactly (i.e. where it is used in the backend)

ahaeslich:
> > we shouldn't find any NodeInterface in this package anymore.
> 
> There are still some. For example in the [NodeTreeBuilder](https://github.com/neos/neos-ui/blob/483461eb680c77fd8ea4070c2bc5c3e831092e50/Classes/Domain/Service/NodeTreeBuilder.php) – I tried to address those but I failed to understand the code or even what this class does exactly (i.e. where it is used in the backend)

I see 🤔

PR #3464, opened 2023-04-14
BUGFIX: Resolve console errors while jests test

markusguenther:
When we run our `jest` test we have plenty of console.errors in the output. That is distracting and should not happen.

We want to prevent errors like `Warning: Failed prop type: The prop `editorRegistry` is marked as required in `EditorEnvelope`, but its value is `null`.`

**What I did**
Walked through all packages and searched which packages were effected.

- [x] neos-ui
- [ ] neos-ui-editors
- [ ] neos-ui-sagas

### DIFF
--- a/packages/neos-ui-editors/src/_lib/testUtils.js
+++ b/packages/neos-ui-editors/src/_lib/testUtils.js
@@ -38,7 +38,9 @@ export const MockDataSourceDataLoader = {
         return this._currentPromise;
     }
 };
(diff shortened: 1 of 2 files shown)


markusguenther:
Will go thru all errors and hopefully resolve them.

markusguenther:
About the sagas, I am unsure if this can be removed ... there we test error handling.

mhsdesign:
Outdated needs to be redone

PR #3475, opened 2023-05-02
FEATURE: add ButtonEditor as new editor

### LINKED ISSUE #3474: FEATURE: add new inspector ButtonEditor
### Description

As an editor I'd like to have a simple button editor that behaves just like selectboxes.

This editor was published last year as [I13e.ButtonEditor](https://github.com/ideenstadtwerke/I13e.ButtonEditor/). We'd like to see this editor in the core.

#### Docs from Package:

> 
> ## Button editor for Neos CMS
> 
> This package provides a new inspector editor displaying it's options as buttons.
> 
> Buttons operates like check- or radioboxes.
> 
> Align
> 
> ## Installation
> 
> Run `composer require i13e/buttoneditor`.
> 
> ## Usage
> 
> Use the new editor in your properties
> 
> ```yaml
> Foo.Package:Content.Text:
>   properties:
>     align:
>       # ButtonEditor supports "string" and "array"
>       type: string
>       defaultValue: left
>       ui:
>         label: 'Align'
>         inspector:
>           editor: Neos.Neos/Inspector/Editors/ButtonEditor
>           editorOptions:
>             # Allow all buttons to be deselected
>             # boolean, default: false
>             allowEmpty: false
>             # Allow multiple buttons to be active
>             # If set to true also use `type` array
>             # boolean, default: false
>             multiple: false
>             # Disable all buttons
>             # boolean, default: false
>             disabled: false
>             values:
>               left:
>                 label: i18n
>                 icon: align-left
>               center:
>                 label: i18n
>                 icon: align-center
>               right:
>                 label: i18n
>                 icon: align-right
>               justify:
>                 label: i18n
>                 icon: align-justify
>                 # An alternative icon for the active state
>                 iconActive: align-justify
>                 # Disable a single option
>                 disabled: true
>             
>             # ---
>             # Optionally use a data source:
>             # 
>             # If `dataSourceIdentifier` or `dataSourceUri` is defined, the `values` from above will be ignored
>             dataSourceIdentifier: 'some-datasource'
>             dataSourceUri: 'some/custom-route'
>             dataSourceAdditionalData:
>               group: 'align'
>             dataSourceDisableCaching: false
> ```
> 
> As you may have realized, the configuration is nearly completely compatible with the default [SelectBoxEditor](https://neos.readthedocs.io/en/stable/References/PropertyEditorReference.html#property-type-string-array-string-selectboxeditor-dropdown-select-editor).
> 
> ## Examples
> 
> ### Basic example from above
> Align
> 
> ### Multiple with allowEmpty
> ```yaml
> editorOptions:
>   allowEmpty: true
>   multiple: true
>   values:
>     # define your own
> ```
> 
> Format
> 
> ### Multiple with allowEmpty:
> ```yaml
> editorOptions:
>   allowEmpty: true
>   values:
>     toggle:
>       label: i18n
>       icon: toggle-off
>       iconActive: toggle-on
> ```
> 
> Toggle
> 
>

reflexxion:
Add new `Neos.Neos/Inspector/Editors/ButtonEditor` as described in #3474

### DIFF
--- a/packages/neos-ui-editors/src/Editors/Button/index.js
+++ b/packages/neos-ui-editors/src/Editors/Button/index.js
@@ -1,0 +1,157 @@
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+import {Icon, IconButton} from '@neos-project/react-ui-components';
+import style from './style.module.css';
+
+import {neos} from '@neos-project/neos-ui-decorators';
+import {$transform} from 'plow-js';
+import {connect} from 'react-redux';
+import {selectors} from '@neos-project/neos-ui-redux-store';
+
+const getDataLoaderOptionsForProps = props => ({
+    contextNodePath: props.focusedNodePath,
+    dataSourceIdentifier: props.options.dataSourceIdentifier,
+    dataSourceUri: props.options.dataSourceUri,
+    dataSourceAdditionalData: props.options.dataSourceAdditionalData,
+    dataSourceDisableCaching: Boolean(props.options.dataSourceDisableCaching)
+});
+
+@neos(globalRegistry => ({
+    i18nRegistry: globalRegistry.get('i18n'),
+    dataSourcesDataLoader: globalRegistry.get('dataLoaders').get('DataSources')
+}))
+@connect($transform({
+    focusedNodePath: selectors.CR.Nodes.focusedNodePathSelector
+}))
+export default class ButtonEditor extends Component {
+    static propTypes = {
+        commit: PropTypes.func.isRequired,
+        value: PropTypes.any,
+        className: PropTypes.string,
+        options: PropTypes.shape({
+            allowEmpty: PropTypes.bool,
+            multiple: PropTypes.bool,
+            disabled: PropTypes.bool,
+            values: PropTypes.objectOf(PropTypes.shape({
+                label: PropTypes.string,
+                icon: PropTypes.string,
+                iconActive: PropTypes.string,
+                disabled: PropTypes.bool
+            })),
+
+            dataSourceIdentifier: PropTypes.string,
+            dataSourceUri: PropTypes.string,
+            dataSourceDisableCaching: PropTypes.bool,
+            dataSourceAdditionalData: PropTypes.objectOf(PropTypes.any)
+        }).isRequired,
+
+        i18nRegistry: PropTypes.object.isRequired
+    };
+
+    static defaultOptions = {
+        allowEmpty: false,
+        multiple: false,
+        disabled: false
+    };
+
+    state = {
+        active: [],
+        isLoading: false,
+        buttons: []
+    };
+
+    initialValueType = 'string';
+
+    constructor(props) {
+        super(props);
+
+        const {commit, options, value} = props;
+        this.initialValueType = Array.isArray(value) ? 'array' : 'string';
+        if (options.multiple && this.initialValueType === 'string') {
+            console.warn(`Misconfiguration in property "${props.identifier}". Multiple is activated but value type seems to be "string" but should be "array".`);
+        }
+
+        this.state = {
+            active: Array.isArray(value) ? value : (value ? [value] : []),
+            buttons: this.hasDataSource() ? [] : this.flattenValues(options.values)
+        };
+        this.commit = commit;
+    }
+
+    hasDataSource() {
+        return this.props.options.dataSourceIdentifier || this.props.options.dataSourceUri;
+    }
+
+    componentDidMount() {
+        if (this.hasDataSource()) {
+            this.loadOptions();
+        }
+    }
+
+    loadOptions() {
+        this.setState({isLoading: true});
+        this.props.dataSourcesDataLoader.resolveValue(getDataLoaderOptionsForProps(this.props), this.props.value)
+            .then(buttons => {
+                this.setState({
+                    isLoading: false,
+                    buttons
+                });
+            });
+    }
+
+    flattenValues(values) {
+        return Object.entries(values).map(([key, val]) => {
+            return {...val, value: key, label: this.props.i18nRegistry.translate(val.label)};
+        });
+    }
+
+    toggleActive(toggleValue) {
+        const {multiple, allowEmpty} = this.props.options;
+        // we'd like to have a copy instead of ref
+        let active = [...this.state.active];
+        if (multiple === false) {
+            if (active.includes(toggleValue)) {
+                active = [];
+            } else {
+                active = [toggleValue];
+            }
+        } else if (active.includes(toggleValue)) {
+            active.splice(active.indexOf(toggleValue), 1);
+        } else {
+            active.push(toggleValue);
+        }
+
+        // if allowEmpty is false but new active length will be 0, drop changes
+        if (!allowEmpty && active.length === 0) {
+            return;
+        }
+        this.setState({active});
+
+        this.commit(this.initialValueType === 'string' ? active[0] || '' : active);
+    }
+
+    isActive(val) {
+        return this.state.active.includes(val);
+    }
+
+    render() {
+        const options = Object.assign({}, this.constructor.defaultOptions, this.props.options);
+
+        return (<div className={style.buttonEditor}>
+            {
+                this.state.isLoading
+                    ? <Icon icon="spinner" size="lg" spin/>
+                    : this.state.buttons.map((button) => (<IconButton
+                            style="lighter"
+                            hoverStyle="darken"
+                            className={style.button}
+                            icon={button.iconActive ? (this.isActive(button.value) ? button.iconActive : button.icon) : button.icon}
+                            title={button.label}
+                            disabled={options.disabled || button.disabled}
+                            isActive={this.isActive(button.value)}
+                            onClick={this.toggleActive.bind(this, button.value)}
+                    />))
+            }
+        </div>);
+    }
+}
--- a/packages/neos-ui-editors/src/Editors/Button/index.js
+++ b/packages/neos-ui-editors/src/Editors/Button/index.js
@@ -2,9 +2,8 @@ import React, {Component} from 'react';
 import PropTypes from 'prop-types';
 import {Icon, IconButton} from '@neos-project/react-ui-components';
 import style from './style.module.css';
-
+import cx from 'classnames';
 import {neos} from '@neos-project/neos-ui-decorators';
-import {$transform} from 'plow-js';
 import {connect} from 'react-redux';
 import {selectors} from '@neos-project/neos-ui-redux-store';
 
@@ -20,12 +19,14 @@ const getDataLoaderOptionsForProps = props => ({
     i18nRegistry: globalRegistry.get('i18n'),
     dataSourcesDataLoader: globalRegistry.get('dataLoaders').get('DataSources')
 }))
-@connect($transform({
-    focusedNodePath: selectors.CR.Nodes.focusedNodePathSelector
+@connect(state => ({
+    focusedNodePath: selectors.CR.Nodes.focusedNodePathSelector(state)
 }))
+
 export default class ButtonEditor extends Component {
     static propTypes = {
         commit: PropTypes.func.isRequired,
+        highlight: PropTypes.bool,
         value: PropTypes.any,
         className: PropTypes.string,
         options: PropTypes.shape({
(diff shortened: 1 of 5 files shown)


packages/neos-ui-editors/src/Editors/Button/index.js
```diff
@@ -0,0 +1,157 @@
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+import {Icon, IconButton} from '@neos-project/react-ui-components';
+import style from './style.module.css';
+
+import {neos} from '@neos-project/neos-ui-decorators';
+import {$transform} from 'plow-js';
```
grebaldi:
```suggestion
```

plow-js will be removed in Neos 9.0 (see: https://github.com/neos/neos-ui/pull/3306).

packages/neos-ui-editors/src/Editors/Button/index.js
```diff
@@ -0,0 +1,157 @@
...
+    dataSourceDisableCaching: Boolean(props.options.dataSourceDisableCaching)
+});
+
+@neos(globalRegistry => ({
+    i18nRegistry: globalRegistry.get('i18n'),
+    dataSourcesDataLoader: globalRegistry.get('dataLoaders').get('DataSources')
+}))
+@connect($transform({
+    focusedNodePath: selectors.CR.Nodes.focusedNodePathSelector
+}))
```
grebaldi:
```suggestion
@connect(state => ({
    focusedNodePath: selectors.CR.Nodes.focusedNodePathSelector(state)
}))
```

as a replacement for `plow-js`'s `$transform` function.

packages/neos-ui-editors/src/Editors/Button/index.js
```diff
@@ -0,0 +1,157 @@
...
+    initialValueType = 'string';
+
+    constructor(props) {
+        super(props);
+
+        const {commit, options, value} = props;
+        this.initialValueType = Array.isArray(value) ? 'array' : 'string';
+        if (options.multiple && this.initialValueType === 'string') {
+            console.warn(`Misconfiguration in property "${props.identifier}". Multiple is activated but value type seems to be "string" but should be "array".`);
+        }
```
grebaldi: Imho this can be more aggressive. The editor should probably actually render this error message instead of the buttons :)

Though for the constraint check, it should not rely on analyzing the actual type of the value, but the configured type of the property (since it states "Misconfiguration").

An erroneous type of the property value should be handled gracefully by the editor.

AFAIK, the UI does not provide the configured type to the editors, but we should definitely change that. So hang on for a PR :smile:, I'll see what I can do.

packages/neos-ui-editors/src/Editors/Button/index.js
```diff
@@ -0,0 +1,157 @@
...
+
+    render() {
+        const options = Object.assign({}, this.constructor.defaultOptions, this.props.options);
+
+        return (<div className={style.buttonEditor}>
+            {
+                this.state.isLoading
+                    ? <Icon icon="spinner" size="lg" spin/>
+                    : this.state.buttons.map((button) => (<IconButton
+                            style="lighter"
```
grebaldi:
```suggestion
                    : this.state.buttons.map((button) => (<IconButton
                            key={button.value}
                            style="lighter"
```

to prevent the following error:

```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `ButtonEditor`. See https://fb.me/react-warning-keys for more information.
```

packages/neos-ui-editors/src/Editors/Button/index.js
```diff
@@ -0,0 +1,157 @@
...
+            return;
+        }
+        this.setState({active});
+
+        this.commit(this.initialValueType === 'string' ? active[0] || '' : active);
+    }
+
+    isActive(val) {
+        return this.state.active.includes(val);
+    }
```
grebaldi:
```suggestion
    isSelected(value) {
        return this.state.selected.includes(value);
    }
```

For clarity:
The button editor shares select box semantics, so imho a value is not `active`, but `selected`.

packages/neos-ui-editors/src/Editors/Button/index.js
```diff
@@ -0,0 +1,157 @@
...
+            });
+    }
+
+    flattenValues(values) {
+        return Object.entries(values).map(([key, val]) => {
+            return {...val, value: key, label: this.props.i18nRegistry.translate(val.label)};
+        });
+    }
+
+    toggleActive(toggleValue) {
```
grebaldi:
```suggestion
    handleSelect(selectedValue) {
```

Also, for clarity.

packages/neos-ui-editors/src/Editors/Button/style.module.css
```diff
@@ -0,0 +1,8 @@
+.buttonEditor {
+}
+
+.buttonEditor .button {
+    margin-right: var(--spacing-Quarter);
+    margin-top: var(--spacing-Quarter);
+    font-size: 20px;
+}
```
grebaldi:
```suggestion
.button.button {
    margin-right: var(--spacing-Quarter);
    margin-top: var(--spacing-Quarter);
    font-size: 20px;
}
```

No need for a base class. `.button.button` is for the desired specificity.

Also, I don't think the increased font-size is a good idea, because it's quite inconsistent with the rest of the UI:

|14px default|20px increased|
|-|-|
| image | image |

packages/neos-ui-editors/src/Editors/Button/index.js
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    isActive(val) {
+        return this.state.active.includes(val);
+    }
+
+    render() {
+        const options = Object.assign({}, this.constructor.defaultOptions, this.props.options);
+
+        return (<div className={style.buttonEditor}>
```
grebaldi:
```suggestion
        return (<div>
```

packages/neos-ui-editors/src/Editors/Button/index.js
```diff
@@ -0,0 +1,157 @@
...
+        this.props.dataSourcesDataLoader.resolveValue(getDataLoaderOptionsForProps(this.props), this.props.value)
+            .then(buttons => {
+                this.setState({
+                    isLoading: false,
+                    buttons
+                });
+            });
+    }
+
+    flattenValues(values) {
```
grebaldi:
```suggestion
    createButtonsFromConfiguration(configuredValues) {
```
For clarity:
This function does more than just flatten a hashmap of values to an array. The function name should reflect that.

packages/neos-ui-editors/src/Editors/Button/index.js
```diff
@@ -0,0 +1,157 @@
...
+        return Object.entries(values).map(([key, val]) => {
+            return {...val, value: key, label: this.props.i18nRegistry.translate(val.label)};
+        });
+    }
```
grebaldi:
```suggestion
        return Object.entries(configuredValues).map(([key, configuredValue]) => {
            return { value: configuredValue.value ?? key, label: this.props.i18nRegistry.translate(configuredValue.label), icon: configuredValue.icon, iconActive: configuredValue.iconActive};
        });
    }
```

A couple of things:

1. <s>`...val`</s> - No spread. Conversion from configuration should be very explicit, because spreads tend to cascade. This means that eventually meaningless configuration values end up being passed as props in unexpected places. This function should sanitize the input, so that doesn't happen.
2. `value: configuredValue.value ?? key` - to allow for explicit configuration of the value.

packages/neos-ui-editors/src/Editors/Button/index.js
```diff
@@ -0,0 +1,157 @@
...
+
+    loadOptions() {
+        this.setState({isLoading: true});
+        this.props.dataSourcesDataLoader.resolveValue(getDataLoaderOptionsForProps(this.props), this.props.value)
+            .then(buttons => {
+                this.setState({
+                    isLoading: false,
+                    buttons
+                });
+            });
```
grebaldi:
```suggestion
            .then(configuredValues => {
                this.setState({
                    isLoading: false,
                    buttons: this.createButtonsFromConfiguration(configuredValues)
                });
            });
```

To treat the data source result similarly to configured values.

grebaldi (changes requested):
Hi @reflexxion,

thanks a lot for adding this editor! I really like the idea. I reviewed the code and added a couple of comments.

I also did some manual tests with the new editor and like to discuss some general observations:

*What I'm writing here does not necessarily constitute actionable items within the scope of this PR. I'm just thinking out loud (by quietly typing words in a text box :D).*

**Defaults** 

For the `type: string` case, I had to configure `multiple: false` manually during my test. I think it should rather be the default. Likewise, I had to configure `allowEmpty: true` manually, which I would have expected to be the default intuitively (though I believe there are probably many opinions here).

Editor defaults are still being configured in `Neos.Neos` (see: https://github.com/neos/neos-development-collection/blob/93c5b77411b527fb99924b3eb8025e3653cdf4cb/Neos.Neos/Configuration/Settings.yaml#L223). We should probably change that. 

And there's another conceptual problem here: We can configure defaults for a specific editor and we can configure defaults for a specific dataType. We can't configure (yet) a default for the combination of both, which would become relevant in this case.

**Labels**

If buttons are configured with a magic `i18n` label, those labels are not covered by the automatic translation id mechanism. This is due to missing editor specifics in the `NodeTypeConfigurationEnrichmentAspect` over at `Neos.Neos` (see: https://github.com/neos/neos-development-collection/blob/93c5b77411b527fb99924b3eb8025e3653cdf4cb/Neos.Neos/Classes/Aspects/NodeTypeConfigurationEnrichmentAspect.php#L167).

I think it is high time the `NodeTypeConfigurationEnrichmentAspect` moves over to `Neos.Neos.Ui`.

**Dirty-State indication**

As of right now, the editor lacks dirty-state indication. When a value changes in the inspector it is important for the user to have visual feedback telling them, which value has changed. 

Inspector Editors have to provide their own styling for this state. Convention is to have an orange outline around the most representative part of the editor. (see also https://github.com/neos/neos-ui/pull/3504, which came up recently)

packages/neos-ui-editors/src/Editors/Button/index.js
mhsdesign: yes good catch. new feature code should not reintroduce plow and should be writting in native javascript. So `node.properties?.someProp` instead of `$get`.

mhsdesign:
@grebaldi i actually like to discuss the naming of the new editor. When i read "button editor" in the pr title i didnt really understand what the editor might do until i saw the screenshots. (I imagined 1 button which can be clicked or edited idk it didnt make sense in my head ^^)

maybe `ChoiceEditor` or something the like is better?

reflexxion:
> @grebaldi i actually like to discuss the naming of the new editor. When i read "button editor" in the pr title i didnt really understand what the editor might do until i saw the screenshots. (I imagined 1 button which can be clicked or edited idk it didnt make sense in my head ^^)

Look what I've put in the packages readme :-D
Toggle

> 
> maybe `ChoiceEditor` or something the like is better?

Yeah, I also wasn't really convinced by `ButtonEditor`.
But choice seems unconvincing too. Because selectbox or boolean editors are "choice editors" in some way?
Somehow it is a two state toggling editor but toggling many states...

Probably something like `ToggleEditor` or `SwitchEditor`? They can even be prefixed by `List`, `Multiple`, ... to clarify it? (`ListToggleEditor`)

@grebaldi thanks for your review. I'll update the PR soon.
But https://github.com/neos/neos-ui/pull/3475#discussion_r1239679057 seems a bit blocking. The editor has to know the property type (currently by checking it's value) to commit the value?

Sebobo:
I like `ChoiceEditor`. And in many cases it would probably replace the SelectBoxEditor used for a small list of values.

packages/neos-ui-editors/src/Editors/Button/style.module.css
reflexxion: I've removed the font-size and added the "dirty state highlighting"

packages/neos-ui-editors/src/Editors/Button/index.js
```diff
@@ -0,0 +1,165 @@
...
+
+    options = {};
+
+    initialValueType = 'string';
+
+    constructor(props) {
+        super(props);
+        const {commit, options, value} = props;
+        this.options = Object.assign({}, this.constructor.defaultOptions, options)
+        // TODO: get actual property data type
```
reflexxion, line 72: I've added that TODO line, to keep that in mind:
https://github.com/neos/neos-ui/pull/3475#discussion_r1239679057 
> AFAIK, the UI does not provide the configured type to the editors, but we should definitely change that. So hang on for a PR 😄, I'll see what I can do.

packages/neos-ui-editors/src/Editors/Button/index.js
reflexxion: Kept that one for highlighting

packages/neos-ui-editors/src/Editors/Button/index.js
mhsdesign: Seems to go into the direction of https://github.com/neos/neos-ui/pull/3580?

grebaldi:
> But https://github.com/neos/neos-ui/pull/3475#discussion_r1239679057 seems a bit blocking. The editor has to know the property type (currently by checking it's value) to commit the value?

Yeah, you're right. I this change isn't all that necessary. We can change this later on :)

mhsdesign:
@Sebobo proposed `OptionsEditor`

but we imo should avoid any naming confusion with our selectbox editor

Sebobo (changes requested):
We decided for `ChoiceEditor` from what I see, so the code still needs to be adjusted. As we are in feature freeze, this can be targeted to 9.1 and anyone can take over.

PR #3482, opened 2023-05-03
BUGFIX: Properly handle object identity DTOs in select box editor

### LINKED ISSUE #2553: SelectBoxEditor with ModelDataSource cannot be set to "null"
### Description

When using the SelectBoxEditor with a implementation of AbstractModelDataSource like it is the default for type `Neos\Media\Domain\Model\Tag` the selected option cannot be unselected again after the changes where applied (see screencast).

### Steps to Reproduce

1. Configure Nodetype with a tag property:
```
filterTag:
      type: 'Neos\Media\Domain\Model\Tag'
      defaultValue: null
      ui:
        label: 'Filter by Tag'
        reloadIfChanged: TRUE
        inspector:
          group: 'filterOptions'
          position: 30
          editorOptions:
            dataSourceAdditionalData:
              firstEmpty: true
```
2. Select a Tag in the backend and apply changes 
3. Try to deselect the tag again

#### Expected behavior

It should be possible to deselect the option / select a null value

#### Actual behavior

It switches back to the currently saved option. So the property cannot be set to `null` again.

[Screencast from Slack Channel](https://neos-project.slack.com/files/U050FRJ0S/FMV9ZCSSW/screen_recording_2019-09-11_at_09.22.35.mov)

### Affected Versions

Neos: 
4.3.4 

UI:
3.8.1

grebaldi:
fixes: #2553 

**The Problem**

A node property value may be persisted as an object identity DTO:
```json
{
  "__identity": "bef64ff9-4c58-4104-bee6-1395eff1e303",
  "__type": "Vendor\\Package\\Type"
}
```

If this is the case and the property is configured to be edited via select box, the persisted serialized object identity DTO is not recognized by the selectbox.

**Reproduction**

You'll need to prepare the following files:

*[Insert Path To Package]/Configuration/NodeTypes.[Insert Some Content Node Type Name].yaml:*
```yaml
'Vendor.Site:Content.SomeContent':
  # ...
  properties:

    tag:
      type: Neos\Media\Domain\Model\Tag
      defaultValue: null
      ui:
        label: 'Tag'
        reloadIfChanged: true
        inspector:
          group: example
          editorOptions:
            allowEmpty: true

    heckLetsAddABunchOfMoreTags:
      type: array<Neos\Media\Domain\Model\Tag>
      ui:
        label: 'Too many tags 😱'
        inspector:
          group: example
          editorOptions:
            multiple: true
```
☝️ This one's to have an example with two select box editors covering both the single-select and the multi-select case. You may add this to any existing node type or create a new one (can be a document as well).

*[Insert Path To Package]/Configuration/Settings.Neos.Neos.yaml:*
```yaml
Neos:
  Neos:
    userInterface:
      inspector:
        dataTypes:
          Neos\Media\Domain\Model\Tag:
            typeConverter: Neos\Neos\TypeConverter\EntityToIdentityConverter
            editorOptions:
              dataSourceIdentifier: 'tag-data-source'
          array<Neos\Media\Domain\Model\Tag>:
            typeConverter: Neos\Flow\Property\TypeConverter\TypedArrayConverter
            editorOptions:
              dataSourceIdentifier: 'tag-data-source'
```
☝️ This one's to achieve default type conversion and editor assignment for `Neos\Media\Domain\Model\Tag` and it's array counterpart. Funny side note: #2553 seems to suggest that this is actually supposed to be the default. During my research I was unable to confirm this, but it would make sense :)

*[Insert Path To Package]/Classes/Application/DataSource/TagDataSource.php:*
```php
<?php

/*
 * This script belongs to the package "Vendor.Site".
 *
 * This package is Open Source Software. For the full copyright and license
 * information, please view the LICENSE file which was distributed with this
 * source code.
 */

declare(strict_types=1);

namespace Vendor\Site\Application\DataSource;

use Neos\Flow\Annotations as Flow;
use Neos\Neos\Service\DataSource\AbstractDataSource;
use Neos\Flow\Persistence\Doctrine\PersistenceManager;
use Neos\Media\Domain\Repository\TagRepository;

#[Flow\Scope("singleton")]
final class TagDataSource extends AbstractDataSource
{
    /**
     * @var string
     */
    static protected $identifier = 'tag-data-source';

    /**
     * @Flow\Inject(lazy=false)
     * @var TagRepository
     */
    protected $tagRepository;

    /**
     * @Flow\Inject(lazy=false)
     * @var PersistenceManager
     */
    protected $persistenceManager;

    /**
     * Get data
     *
     * {@inheritdoc}
     */
    public function getData(NodeInterface $node = NULL, array $arguments = [])
    {
        $tags = $this->tagRepository->findAll();
        $result = [];

        foreach ($tags as $tag) {
            $result[] = [
                'value' => $this->persistenceManager->getIdentifierByObject($tag),
                'icon' => 'tag',
                'label' => $tag->getLabel()
            ];
        }

        return $result;
    }
}

```

☝️ This is the data source that provides our select boxes from above with options straight from the Media package's tag repository. 

For reproduction, just add a node of the type from above and see what happens if you set the select box to a value and then reload.

**The solution**

It's a bit of a lazy solution, but I couldn't think of anything more sophisticated that wouldn't break the select box editor. The gist is: I added a function that pre-processes incoming select box editor values and converts any detected object identity DTO down to it's string-based identity.

### DIFF
--- a/packages/neos-ui-editors/src/Editors/SelectBox/DataSourceBasedSelectBoxEditor.js
+++ b/packages/neos-ui-editors/src/Editors/SelectBox/DataSourceBasedSelectBoxEditor.js
@@ -7,6 +7,7 @@ import MultiSelectBox from '@neos-project/react-ui-components/src/MultiSelectBox
 import {selectors} from '@neos-project/neos-ui-redux-store';
 import {neos} from '@neos-project/neos-ui-decorators';
 import {shouldDisplaySearchBox, searchOptions, processSelectBoxOptions} from './SelectBoxHelpers';
+import {createSelectBoxValueStringFromPossiblyStrangeNodePropertyValue} from './createSelectBoxValueStringFromPossiblyStrangeNodePropertyValue';
 import PreviewOption from '../../Library/PreviewOption';
 
 const getDataLoaderOptionsForProps = props => ({
@@ -30,7 +31,17 @@ export default class DataSourceBasedSelectBoxEditor extends PureComponent {
         className: PropTypes.string,
         value: PropTypes.oneOfType([
             PropTypes.string,
-            PropTypes.arrayOf(PropTypes.string)
+            PropTypes.arrayOf(PropTypes.string),
+            PropTypes.shape({
+                __identity: PropTypes.string.isRequired,
+                __type: PropTypes.string
+            }),
+            PropTypes.arrayOf(
+                PropTypes.shape({
+                    __identity: PropTypes.string.isRequired,
+                    __type: PropTypes.string
+                })
+            )
         ]),
         options: PropTypes.shape({
             allowEmpty: PropTypes.bool,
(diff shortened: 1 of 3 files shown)


Sebobo:
Very nice, will take a look. 
I had actually hoped to add DataSources for Tags and Collections already in 8.3 but wasn't able to finish it on time :( 
But it would be great to add this to the Media UI as an interim solution. Together with this fix and the new AssetsHelper in 8.3 it should be much nicer to work with assets based on their tags and collections.

mhsdesign:
What is needed for this to be merged? Is it ready for review?

grebaldi:
It's ready :) Just needs review

Sebobo (approved):
Thx, works as expected with your example 🙂

I'm fine with refactoring the amazingly called new method at a later point.

markusguenther (approved):

Issue #3480, opened 2023-05-02
BUG: Regression node info table layout broken


mhsdesign:
8.0
<img width="318" alt="image" src="https://user-images.githubusercontent.com/85400359/235751100-4fe409b4-f9b2-44b9-bdbb-8ca701995d58.png">

8.3
<img width="311" alt="image" src="https://user-images.githubusercontent.com/85400359/235751150-c394610a-567c-416d-b1ca-be000581d853.png">

grebaldi:
I think this should actually be a description list (`<dl/>`) rather than a table: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl

grebaldi:
This seems to occur on 8.2 as well: https://neos-project.slack.com/archives/C0U0KEGDQ/p1683116589047559

ahaeslich:
> I think this should actually be a description list (`<dl/>`) rather than a table: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl

Yes, this would be way better. Actually it currently is a `<ul/>`. As seen in this change: #3478.

Sebobo:
Let's refactor the markup for 9.0, I fixed the style issue in #3484

PR #3484, opened 2023-05-05
BUGFIX: InfoTableLayout style broken

Sebobo:
Resolves: #3480


ahaeslich (approved):

mhsdesign:
Thanks ;)

We dindt think of the class concatenation when switching to esbuild ^^
I just checked the codebase and this was the only place we used it as far as i can tell

Sebobo:
@mhsdesign yes, I came to the same conclusion and decided its not worth to tough our build system because of that one file.

PR #3485, opened 2023-05-10
BUGFIX: Add primitive Root ErrorBoundary around whole UI

mhsdesign:
FYI, im currently developing on 8.3 but will cherry-pick this to 7.3 once ready... (webpack and me are never going back together)

The idea is that in case of fatal errors, a fallback error boundary is set at root level, so that less technical people dont report: "my neos just turned white" but can give directly informations about the problems. Users can also easily differentiate problems and if one error keeps turning up.

image

**What I did**

**How I did it**

**How to verify it**

ever


packages/neos-ui/package.json
```diff
@@ -58,6 +58,7 @@
     "react-dnd": "^10.0.0",
     "react-dnd-html5-backend": "^10.0.0",
     "react-dom": "^16.12.0",
+    "react-error-boundary": "^4.0.4",
```
Sebobo: Why don't you use the `ErrorBoundary` of React itself?

packages/neos-ui/src/Containers/ErrorBoundary/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+                {props.error instanceof Error &&
+                    <>
+                        <p>
+                            Name: {props.error.name || '-'}
+                        </p>
+                        <p>
+                            Message: {props.error.message || '-'}
+                        </p>
+                        <p>Stacktrace:</p>
+                        <pre className={styles.stackTrace}>{props.error.stack}</pre>
```
Sebobo: Here we should add a copy to clipboard button, I can send a snippet if you need.

packages/neos-ui/src/Containers/ErrorBoundary/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+                    </>
+                }
+                <p>For more information about the Error please refer to the JavaScript console.</p>
+            </div>
```
Sebobo: A reload button would also be nice, so users can use it to continue.

Sebobo (changes requested):
Love it!

We can make some improvements, like the copy button you mentioned but also a reload button.

Why didn't you use the native React ErrorBoundary? Does the library do something it doesn't?

packages/neos-ui/package.json
mhsdesign: i was unsure at first too but then i remembered that @grebaldi did use this lib also in Sitegeist.Archaeopteryx https://github.com/sitegeist/Sitegeist.Archaeopteryx/blob/a003751455ccd6759cd6968dea11c08481ee4296/Neos.Ui/error-handling/src/application/ErrorBoundary.tsx and i thought it is good practice.

This lib will allow dismissing of errors to and stuff like this in the future

packages/neos-ui/src/Containers/ErrorBoundary/index.tsx
mhsdesign: Added ;)
image

packages/neos-ui/package.json
Sebobo: I still don't see what's the benefit to the default ErrorBoundary?

packages/neos-ui/src/Containers/ErrorBoundary/index.tsx
mhsdesign: agreed and added ;)

mhsdesign:
Added the fancy buttons ^^ with @jonnitto (thanks!)

packages/neos-ui/package.json
grebaldi: IIRC React didn't have a default Error Boundary when I built Archaeopteryx. Error boundaries were a new concept back then.

I'd definitely opt for official API rather than dependency - so if React provides something, then I'd recommend removing the dependency to `react-error-boundary`.

crydotsnake:
Love that! ❤️

packages/neos-ui/package.json
mhsdesign: > There is no direct equivalent for static getDerivedStateFromError in function components yet. If you’d like to avoid creating class components, write a single ErrorBoundary component like above and use it throughout your app. Alternatively, use the [react-error-boundary](https://github.com/bvaughn/react-error-boundary) package which does that.

hu interesting ^^

Sebobo: Yes, that's the one I built for the Media.UI https://github.com/Flowpack/media-ui/blob/main/Resources/Private/JavaScript/media-module/src/components/ErrorBoundary.tsx

mhsdesign: adjusted ;) THanks

mhsdesign:
todo:
- find out why the CI fails
- ~backport to 7.3 (i just developed it on 8.3 :P)~

jonnitto:
I would prefer not to have the "fancy" Logo. In slack it is ok, but I would not use it on an official document

mhsdesign:
Okay no problemo ^^

Resources/Private/Translations/en/Main.xlf
```diff
@@ -326,6 +326,24 @@
...
+                <source>Copy technical details</source>
+            </trans-unit>
+            <trans-unit id="errorBoundary.technicalDetailsCopied" xml:space="preserve">
+                <source>Technical details copied</source>
+            </trans-unit>
+            <trans-unit id="errorBoundary.reloadUi" xml:space="preserve">
+                <source>Reload Neos UI</source>
+            </trans-unit>
+            <trans-unit id="errorBoundary.title" xml:space="preserve">
+                <source>Sorry, but the Neos UI could not recover from this Error.</source>
```
jonnitto:
```suggestion
                <source>Sorry, but the Neos UI could not recover from this error.</source>
```

Resources/Private/Translations/en/Main.xlf
```diff
@@ -326,6 +326,24 @@
...
+            </trans-unit>
+            <trans-unit id="errorBoundary.description" xml:space="preserve">
+                <source>Please reload the application, or contact your system administrator with the given details.</source>
+            </trans-unit>
+            <trans-unit id="errorBoundary.footer" xml:space="preserve">
+                <source>For more information about the Error please refer to the JavaScript console.</source>
```
jonnitto:
```suggestion
                <source>For more information about the error please refer to the JavaScript console.</source>
```

mhsdesign:
@Sebobo ping ^^

crydotsnake (approved):

Sebobo (approved):
Looks fine now by reading <3

mhsdesign:
Will add the translations via weblate ^^

PR #3487, opened 2023-05-11
BUGFIX: Perform additional search for links that look like externals or emails

### LINKED ISSUE #3016: BUG: looksLikeExternalLink() too unspecific
### Description

We have several Product-Nodetypes defined as Document-Nodetypes with a title of for example 12.233 (the SKU). When entering this sku in the link feature of the inline editable editor, `Neos.Neos.Ui:packages/neos-ui-editors/src/Library/LinkInput.js` thinks, this is a url as the Regex `^[\w.-]{2,}\.[\w]{2,10}$` in `looksLikeExternalLink()` is too specific in my opinion. `\w` contains digits, too even though TLDs AFAIK must not contain digits. This causes NEOS to stop searching the node tree and showing the `Format as HTTP Link?` message. Is there any reason to keep this regex that unspecific?

### Steps to Reproduce

1. Create a node with the title `23.233`.
2. Add an inline editable text and search for this node to link it using the link feature of ckeditor.

#### Expected behavior

It will search the node tree and display `23.233` as a suggestion in the dropdown.

#### Actual behavior

It will stop searching the node tree when typing the `23.23` because it considers the string an http link. It then shows `Format as HTTP Link?`.

### Affected Versions

All

grebaldi:
fixes: #3016 

|before|after|
|-|-|
|Screenshot_2023-05-11_16-44-44 3016 - before|Screenshot_2023-05-12_12-22-58 3016 - after 2 |

**The problem**

The link input will recognize certain strings as E-Mail addresses or probable external links (with very broad coverage there). If it does so, it won't perform a search on nodes or assets but will immediately show a suggestion like "Format as external link?".

It is however quite possible that the user is actually searching for a document or asset that is titled after an E-mail address or anything that may look like a domain. #3016 mentions an example of a document with an SKU as a title that was recognized as an external link and thus could not be found by the link input.

**The solution**

@RobReckham suggested in #3016 to adjust how `looksLikeExternalLink` recognizes external links. I figured however that it may be too difficult to find an actually satisfactory regex that would cover all possible scenarios properly.

So, I decided to go a different route and leave the algorithm as it is. Instead, the link input now performs the search for nodes and assets even if the input looks like an E-Mail address or external link. The actions "Format as external link?" or "Format as e-mail?" are still being displayed if `looksLikeExternalLink` recognizes the input as such, but they're now entirely optional.

### DIFF
--- a/packages/neos-ui-editors/src/Library/LinkInput.js
+++ b/packages/neos-ui-editors/src/Library/LinkInput.js
@@ -150,41 +150,41 @@ export default class LinkInput extends PureComponent {
     handleSearchTermChange = searchTerm => {
         this.setState({searchTerm});
 
-        if (looksLikeExternalLink(searchTerm)) {
-            this.setState({
-                isLoading: false,
-                searchOptions: [{
-                    label: this.props.i18nRegistry.translate('Neos.Neos.Ui:Main:ckeditor__toolbar__link__formatAsHttp', 'Format as https link?'),
-                    loaderUri: `https://${searchTerm}`
-                }]
-            });
-        } else if (isEmail(searchTerm)) {
-            this.setState({
-                isLoading: false,
-                searchOptions: [{
-                    label: this.props.i18nRegistry.translate('Neos.Neos.Ui:Main:ckeditor__toolbar__link__formatAsEmail', 'Format as email?'),
-                    loaderUri: `mailto:${searchTerm}`
-                }]
-            });
-        } else if (isUriOrInternalLink(searchTerm)) {
+        if (isUriOrInternalLink(searchTerm)) {
             this.setState({
                 isLoading: false,
                 searchOptions: []
             });
         } else if (searchTerm) {
-            this.setState({isLoading: true, searchOptions: []});
             // We store the searchTerm at the moment lookup was triggered, and only update the options if the search term hasn't changed
             const searchTermWhenLookupWasTriggered = searchTerm;
+            const groupedSearchOptions = [];
+
+            if (looksLikeExternalLink(searchTermWhenLookupWasTriggered)) {
+                groupedSearchOptions.push({
+                    label: this.props.i18nRegistry.translate('Neos.Neos.Ui:Main:ckeditor__toolbar__link__formatAsHttp', 'Format as https link?'),
+                    loaderUri: `https://${searchTermWhenLookupWasTriggered}`
+                });
+            } else if (isEmail(searchTermWhenLookupWasTriggered)) {
+                groupedSearchOptions.push({
+                    label: this.props.i18nRegistry.translate('Neos.Neos.Ui:Main:ckeditor__toolbar__link__formatAsEmail', 'Format as email?'),
+                    loaderUri: `mailto:${searchTermWhenLookupWasTriggered}`
+                });
+            }
+
+            this.setState({isLoading: true, searchOptions: groupedSearchOptions});
+
             this.props.linkLookupDataLoader.search(this.getDataLoaderOptions(), searchTerm)
                 .then(searchOptions => {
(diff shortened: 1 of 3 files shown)


Sebobo:
Awesome, thx!

Should we adjust the `Without Group` label to something else? I think it's a bit confusing right now. 
Maybe `Format options`?

grebaldi:
@Sebobo Yep, that makes sense. But I'm not sure whether additional translations would still allow this to be a bugfix. If not, I'd opt for adding the label in an additional PR for 9.0 post-upmerge.

Sebobo:
IMO that is fine, we had bugfixes for missing translations too, so it would just cause the next bug issue ;)

crydotsnake (approved):
Tested and works as expected!

mhsdesign (approved):
Works splendidly!

comparison:

| before | after |
| -------| ----- |
| image | image |
| image | image |
| image | image |
| image | image |

mhsdesign:
I added the missing translation definition in en and de ;)

packages/neos-ui-editors/src/Library/LinkInput.js
```diff
@@ -150,41 +150,43 @@ export default class LinkInput extends PureComponent {
...
+                        searchOptions.forEach(searchOption => {
                             searchOption.group = 'assetSourceLabel' in searchOption ? searchOption.assetSourceLabel : this.props.i18nRegistry.translate('Neos.Neos:Main:document');
-                            return searchOption;
+                            groupedSearchOptions.push(searchOption);
                         });
+
                         this.setState({
                             isLoading: false,
-                            searchOptions: groupedSearchOption
+                            searchOptions: [...groupedSearchOptions]
```
mhsdesign, line 189:
```suggestion
                            searchOptions: groupedSearchOptions
```

nitpick: is there a reason for this copy?

bwaidelich (commented):
I can't judge the code (due to my js ignorance) but I love this change, thank you! :)

packages/neos-ui-editors/src/Library/LinkInput.js
grebaldi: Yep. At line 177 you'll see that `groupedSearchOptions` is already put into state. State then holds a reference to `groupedSearchOptions` which gets mutated in line 184. So to make sure that the changes made to `groupedSearchOptions` after the search results have been fetched also lead to a changed reference, `groupedSearchOptions` needs to be copied at this point.

packages/neos-ui-editors/src/Library/LinkInput.js
```diff
@@ -150,41 +150,43 @@ export default class LinkInput extends PureComponent {
...
+
             this.props.linkLookupDataLoader.search(this.getDataLoaderOptions(), searchTerm)
                 .then(searchOptions => {
                     if (searchTermWhenLookupWasTriggered === this.state.searchTerm) {
-                        const groupedSearchOption = searchOptions.map(searchOption => {
+                        searchOptions.forEach(searchOption => {
                             searchOption.group = 'assetSourceLabel' in searchOption ? searchOption.assetSourceLabel : this.props.i18nRegistry.translate('Neos.Neos:Main:document');
-                            return searchOption;
+                            groupedSearchOptions.push(searchOption);
                         });
```
mhsdesign, line 185:
```suggestion
                        });
```

packages/neos-ui-editors/src/Library/LinkInput.js
mhsdesign: ahh i see, because there would be no diff in setState to the previous one... so technically it would properly cleaner to make a copy already before? But fine with me. Thanks for explaining - i imagined something like this, but its not always super obvious ^^ (i oversaw that 177 happens always and not conditionally ^^)

Issue #2925, opened 2021-07-18
Remove click-outside handling for Dialog


grebaldi:
### Motivation

I've come across multiple scenarios (namely: https://github.com/sitegeist/Sitegeist.Archaeopteryx and https://github.com/Flowpack/media-ui) in which UI extensions need to stack dialogs on top of each other.

This interferes with our current click-outside handling for Dialogs, which makes the assumption that such stacking never happens.

The effect of that can be seen here:

https://user-images.githubusercontent.com/2522299/126071348-86f89542-a69e-40ca-b5bb-d4edd5a3b8c5.mp4

You see that the new Media UI is embedded in the Node Creation Dialog (Dialog 1). Then we click on "Upload", which opens the Upload Modal of the new Media UI (Dialog 2). If we click anywhere inside Dialog 2, both dialogs close. When we click ouside Dialog 2, both dialogs close as well.

https://user-images.githubusercontent.com/2522299/126071499-f95ad2ee-90cc-4a2e-b034-901ee4aa3f5f.mp4

The reason is, that after opening Dialog 2, the entire screen is now an area that is considered to be outside Dialog 1. So if we click anywhere, Dialog 1 receives its request to close. Since Dialog 1 is removed from VDOM, and Dialog 2 is a child of Dialog 1, Dialog 2 has to leave as well.

### Possible Solution

Imho, we don't need such aggressive click-outside handling for our Dialogs. It would be sufficient to just react to clicks on the semi-transparent overlay, since it covers (almost) the entire screen.

I say almost, because the top bar of the ui is not covered:

![Screenshot_2021-07-18_16-47-08](https://user-images.githubusercontent.com/2522299/126071668-43f99131-8eb9-4fce-a494-3f6e57ead8b3.png)

This can potentially be worked around, because Dialogs live in an isolated part of the DOM tree, but I don't know if this behvaior is even intentional in the first place :sweat_smile:. (At least it doesn't seem right to me :monocle_face:)

I'd suggest to implement a solution as a bugfix for 5.3+.

### Affected Versions

Neos: >=3.3
UI: >=1.0

Sebobo:
Sounds good to me!

PR #3491, opened 2023-05-12
BUGFIX: Let unapplied changes overlay cover everything except inspector

### LINKED ISSUE #2670: Editor is not warned about unsave changes in inspector (on publish, on page close)
### Description

One of the editors of our customer stumbled upon the following issue:

### Steps to Reproduce

1. Do changes in the content area - now you have unpublished changes
2. Open the inspector on the right and change something there, but "forget" to click the apply button
3. You are used to the system warning you about unsaved changes in the inspector
4. You click the publish button and get the feedback, that all was successfully published
5. The change in the inspector is still unsaved, but you do not notice it, because in former versions of Neos you got a warning, if you left the inspector without hitting the "apply" button.
6. You log our or close the window and your change is lost without you even realizing it.

#### Expected behavior

If you click on the top frame when there is an not-applied change in the inspector, you should get the same warning about unsaved changes as when you click in the content area.

### Affected Versions

Neos: 5.x
UI: 5.x

This is not a big issue, but might be fixed easily and would avoid frustration for the editors.

### LINKED ISSUE #3174: Escape "Unapplied changes" when using "Selected element" to navigate to other Node
### Description
Funnely it works astonishing well and is not buggy at all ^^ - i just found an escape hatch.

### Steps to Reproduce

1. Make changes on a Content Element (in the Inspector)
2. use the Selected element selectbox to move upward (Content Collection / Page)

#### Expected behavior

Idk - but for some reason its not allowed to change the node by using the Content Tree with unapplied changes - so i assumed the same for the "Selected element"  select box.

#### Actual behavior

You moved to the other node ^^ 
now you are free ;)  - unless you go back to the edited node.

### Affected Versions

UI version: v8.0.1

grebaldi:
fixes: #2670

**The problem**

The unapplied changes overlay is supposed to prevent all editing actions as long as there are pending changes in the inspector, so that the user is forced to make a decision of whether to apply or discard their changes and prevent them from accidentally losing their unsaved work.

Unfortunately, the unapplied changes overlay does not cover the entire user interface, but only the content canvas area:

Screenshot_2023-05-12_12-35-38 - 2670 before

This (unintentionally) allows the user to perform all kinds of actions (like changing the edit-/preview-mode, discarding changes, etc.) and potentially lose their progress.

**The solution**

The first part of the solution is the attempt to let the unapplied changes overlay cover the entire user interface (minus the inspector area):
Screenshot_2023-05-12_13-43-29 - 2670 after

This is done in three steps:

1. Move the overlay directly underneath the document body via `ReactDOM.createPortal`, so that it competes for z-index globally
2. Adjust the positioning of the `&::after`-element, so it covers the entire top bar
3. Set the `--zIndex-Dialog` variable to a value that allows only flash messages to be drawn on top of the overlay

Step 3 turned out to be a bit tricky, because of the logic manifested in `packages/build-essentials/src/styles/styleConstants.js`. I was unable to achieve the desired effect without changing the variables name from `--zIndex-UnappliedChangesOverlay-Context` to `--zIndex-UnappliedChangesOverlay`, which is possibly breaky to anyone who relies on the former (this isn't official API though and is very unlikely to have any noticable effect).

I also adjusted `--zIndex-UnappliedChangesOverlay` in `cssVariables.css`, expecting that this will be necessary for later versions.

The second part of the solution has to do with the unapplied changes *dialog*. After I've tested the overlay, I noticed that the semi-transparent dialog overlay also did not cover the entire user interface:

Screenshot_2023-05-12_13-20-43 - 2670 zIndex Problem

I had noticed this before and documented the faulty behavior in #2925. To make the unapplied changes mechanism work as intended, I decided to fix this z-index issue in the scope of this PR as well.

Similar to the unapplied changes overlay, I had to rename `--zIndex-Dialog-Context` to `--zIndex-Dialog`.

*P.s.: The blurs you've undoubtedly noticed in the screenshots are being addressed over here: https://github.com/neos/neos-ui/pull/3489*

**EDIT:** E2E-tests revealed that I needed to adjust the z-index configuration for select box dropdowns as well. Hooray for E2E-tests! :smile:


mhsdesign:
> I also adjusted --zIndex-UnappliedChangesOverlay in cssVariables.css, expecting that this will be necessary for later versions.

1000% wow thanks, Yes this is due to dropping the `styleConstants.js` in the future (8.3) https://github.com/neos/neos-ui/pull/3366

and here is my description, as to why we have them also in the 7.3 branch ^^ https://github.com/neos/neos-ui/pull/3438

crydotsnake (approved):
Tested and works as expected!

mhsdesign (changes requested):
Thanks ;) works as described (tested on 8.3 to make sure the new cssVariables work)

what do you think about making the overlay also cover the "select element" box as well as the publishing dropdown:

image

this would then also resolve: https://github.com/neos/neos-ui/pull/3501

currently one can publish changes and also escape to the workspaces backend module via the selector.


another little detail i noticed on 8.3, is that the mouse becomes a pointer and the hover effect is triggered for the language select box
image

the language selectbox will also open, their items not be selectable:

image


probably an z-index issue

grebaldi:
@mhsdesign:
> what do you think about making the overlay also cover the "select element" box as well as the publishing dropdown:

Good point. I'd find that quite sensible (also think that `<SelectedElement/>` has actually no structural point in being part of the Inspector in the first place tbh.). It'd be also less risky compared to #3501.

If I get a couple of :+1: on this plan, I'll change it here and close #3501 :D

mhsdesign:
Yes I believe it’s much easier to maintain than to have #3501 in our code base. Sorry that you went through the hassle of creating #3501 if we discard your work

grebaldi:
Done :)

@mhsdesign:
> Yes I believe it’s much easier to maintain than to have https://github.com/neos/neos-ui/pull/3501 in our code base. Sorry that you went through the hassle of creating https://github.com/neos/neos-ui/pull/3501 if we discard your work

No worries! I'm all for it: Less entropy, fewer troubles :)

crydotsnake (commented):
It works as expected!

I remembered one more tiny detail. If you follow the steps in the issue to reproduce this bug, even though the inspector is dimmed, it is still possible for me to publish my changes even though I still have changes in the inspector that I haven't saved.


SCR-20230612-osed

Would it be a big effort to disable the publish button in such a case? Or should we discuss that maybe for Neos 9? Does it even make sense? 😅

grebaldi:
@crydotsnake:
> I remembered one more tiny detail. If you follow the steps in the issue to reproduce this bug, even though the inspector is dimmed, it is still possible for me to publish my changes even though I still have changes in the inspector that I haven't saved.

I can't reproduce that... In my setup, the publish button lights up orange, but if I click it, I get the "Unapplied Changes"-dialog as expected.

But then again I think it makes sense to give more visual indication here... I'll see what I can do :)

mhsdesign:
@grebaldi thank you so much for your adjustments.

I think its fine to not show any additional visual indication in the publish button, and either way it could be done in a separate pr (if at all ^^)

I will try this out and see if it as expected overlays with the publish button and then id say this is ready to be merged!

@crydotsnake are you sure that you can still click the publish button? Could you please try to debug this if so, maybe by checking with other browsers and investigating the css styles in the dev tools.

mhsdesign:
https://github.com/neos/neos-ui/assets/85400359/17cc6f5c-da7c-4ad3-83a3-cef0f409027a

@crydotsnake _it works on my mashine_ (well i cheated again as i tested it cherry-cheryy-lady picked on 8.3 :P

mhsdesign (approved):
Thanks it works now perfectly ;) And should survive the upmerge ^^

crydotsnake (approved):
Tested it again, and its working great!

grebaldi:
I've just noticed a regression :(

The "Unapplied changes" overlay now covers the secondary inspector. This makes it impossible to crop images. I will come up with a fix soon.

mhsdesign:
> The "Unapplied changes" overlay now covers the secondary inspector. This makes it impossible to crop images. I will come up with a fix soon.


oh well good catch 😅 i was about to release it the next days - we def need e2e test for that??? Should have affected every secondary inspector right? - I forgot to test that as well ^^

PR #3496, opened 2023-05-19
BUGFIX: Consider content role of reference node while calculating enabled state of add button

### LINKED ISSUE #2929: Node Creation dialog can be invoked, even though constraints don't allow any NodeTypes to be inserted
### Description

Given there's a content collection on the page that does not allow any content to be inserted, it is still possible to invoke the node creation dialog. This is a minor issue, since it doesn't make much sense to have such a content collection in the first place, but nevertheless, the "Add Node" button should be disabled in this scenario.

This behavior is possibly related to  #2917.

### Steps to Reproduce

1. Render a content collection on the page that doesn't allow any NodeTypes to be inserted
2. Select that content collection in Neos UI
3. You'll notice that the "Add Node" button can be clicked.
4. When you click on the button, an empty NodeCreationDialog shows up.

#### Expected behavior

The "Add Node" button is disabled when the selected collection does not allow for any NodeTypes to be inserted.

#### Actual behavior

The "Add Node" button is enabled and leads to an empty NodeCreationDialog.

Screenshot_2021-07-25_13-57-09

Screenshot_2021-07-25_13-56-39

### Affected Versions

Neos: 5.3

UI: 5.3

grebaldi:
fixes: #2929 

**The problem**

We have three toolbars through which a node can be created: The document tree toolbar, the content tree toolbar and the inline UI toolbar. When using the "add node" button on the latter two, the selection of node types is being calculated in a slightly different way than in the document tree toolbar.

The problem is that an editor would expect the content tree toolbar and the inline UI toolbar to exclusively deal with *contents* on the currently active document. The content repository however knows no difference between *content* nodes and *document* nodes. So in theory, when I select a content collection, the node type selection dialog should offer me all kinds of document nodes as possible siblings.

This is why we entered the concept of node type roles. A given node type can either have the role "document" or the role "content". When the editor hits the "add node" button on the content tree toolbar or the inline UI toolbar, the list of node types in the resulting node type selection dialog will be stripped down to only include node types with the role "content".

This distinction is missing from the algorithm that calculates the enabled state of the respective "add node" button, which results in possible scenarios in which the "add node" button can be clicked, but the resulting node type selection dialog is empty. (as described in #2929)

**The solution**

I added the appropriate node type role for the reference node to the respective `isAllowedToAddChildOrSiblingNodesSelector` calls in both the content tree toolbar and the inline UI toolbar containers.

### DIFF
--- a/packages/neos-ui-guest-frame/src/InlineUI/NodeToolbar/Buttons/AddNode/index.js
+++ b/packages/neos-ui-guest-frame/src/InlineUI/NodeToolbar/Buttons/AddNode/index.js
@@ -15,8 +15,13 @@ import {neos} from '@neos-project/neos-ui-decorators';
 
     return state => {
         const focusedNodeContextPath = selectors.CR.Nodes.focusedNodePathSelector(state);
+        const getNodeByContextPathSelector = selectors.CR.Nodes.makeGetNodeByContextPathSelector(focusedNodeContextPath);
+        const focusedNode = getNodeByContextPathSelector(state);
+
+        const role = focusedNode ? (nodeTypesRegistry.hasRole(focusedNode.nodeType, 'document') ? 'document' : 'content') : null;
         const isAllowedToAddChildOrSiblingNodes = isAllowedToAddChildOrSiblingNodesSelector(state, {
-            reference: focusedNodeContextPath
+            reference: focusedNodeContextPath,
+            role
         });
 
         return {
--- a/packages/neos-ui/src/Containers/LeftSideBar/NodeTreeToolBar/index.js
+++ b/packages/neos-ui/src/Containers/LeftSideBar/NodeTreeToolBar/index.js
@@ -311,8 +311,10 @@ const makeMapStateToProps = isDocument => (state, {nodeTypesRegistry}) => {
 
         const isHidden = $get('properties._hidden', focusedNode);
 
+        const role = focusedNode ? (nodeTypesRegistry.hasRole(focusedNode.nodeType, 'document') ? 'document' : 'content') : null;
         const isAllowedToAddChildOrSiblingNodes = isAllowedToAddChildOrSiblingNodesSelector(state, {
-            reference: focusedNodeContextPath
+            reference: focusedNodeContextPath,
+            role
         });
         const isHiddenContentTree = $get('ui.leftSideBar.contentTree.isHidden', state);
 


crydotsnake (approved):
Tested and works as expected. And totally makes sense IMO.

crydotsnake:
_Just as an idea_

would it make sense, too change the hover title of the button in this case? for example:

SCR-20230603-tvbp

Because "Create new" could be a bit irritating for the editor if the editor cant actually place any content elements in the content collection.

packages/neos-ui-guest-frame/src/InlineUI/NodeToolbar/Buttons/AddNode/index.js
```diff
@@ -15,8 +15,13 @@ import {neos} from '@neos-project/neos-ui-decorators';
 
     return state => {
         const focusedNodeContextPath = selectors.CR.Nodes.focusedNodePathSelector(state);
+        const getNodeByContextPathSelector = selectors.CR.Nodes.makeGetNodeByContextPathSelector(focusedNodeContextPath);
+        const focusedNode = getNodeByContextPathSelector(state);
+
+        const role = focusedNode ? (nodeTypesRegistry.hasRole(focusedNode.nodeType, 'document') ? 'document' : 'content') : null;
```
mhsdesign, line 21: thank you so much for fixing yet another odd edge case ! 

but i have a quick question

why dont we use `nodeTypesRegistry.getRole(focusedNode.nodeType)`, as the role could not only be `document` `content`, but also

`ignored` or `contentCollection` 


see

frontendConfiguration.nodeTypeRoles

also i truly dislike nested ternaries ^^

mhsdesign (changes requested):

mhsdesign:
> _Just as an idea_
> 
> would it make sense, too change the hover title of the button in this case? for example:
> 
> SCR-20230603-tvbp
> 
> Because "Create new" could be a bit irritating for the editor if the editor cant actually place any content elements in the content collection.

we discussed, that its not worth the effort to come up with a good heuristic what went actually wrong. So we will just display the create new in disabled state.

packages/neos-ui-guest-frame/src/InlineUI/NodeToolbar/Buttons/AddNode/index.js
grebaldi: > why dont we use `nodeTypesRegistry.getRole(focusedNode.nodeType)`, as the role could not only be `document` `content`, but also
> 
> `ignored` or `contentCollection`

The calculation of `role` in this instance replicates the implementation of `SelectNodeType`:

https://github.com/neos/neos-ui/blob/06cf2eb4ea12ebfc07716c87d6bfe3a2383b606f/packages/neos-ui/src/Containers/Modals/SelectNodeType/index.js#L65

I agree that `nodeTypesRegistry.hasRole(focusedNode.nodeType, 'document') ? 'document' : 'content'` is one heck of a confusing statement, but it is what `SelectNodeType` is doing already. I think that fixing the semantics would be preferable. That would however require fixing it at all sites, which imho would leave the scope of this bugfix... Not sure... What's your opinion?

> also i truly dislike nested ternaries ^^

I feel similarly. Yet, if I were to resolve the nested statements into blocks, this would roughly double the footprint of this PR. I think that's not worth the trouble :shrug:

mhsdesign: Okay thank you for explaining this. I didn't knew we already use it like this ^^

ForGet SeMaNTIcs :D

So yes considering this, i rather like the change to be small and fix the semantics - dare i say it - later*.

*later = never

mhsdesign (approved):

PR #3497, opened 2023-05-19
BUGFIX: Sync frame component location after in-frame navigation happened

### LINKED ISSUE #2954: Navigating to documents hidden in the node-tree breaks the link to the current page
### Description

When navigating to a document hidden in the node-tree (i.e. through links in the guest-frame), the link to the previously selected document in the node-tree does no longer work.

### Steps to Reproduce

1. Setup neos-base-distribution
2. Site-Package and site: https://gist.github.com/PRGfx/a6f33ddc2a53bd1e8b91f821aed6b6dc
    ```
    home (Broken.NodeTree:Document.Home)
    + directory (Broken.NodeTree:Document.Page)
    | + special 1 (Broken.NodeTree:Document.Special)
    | + special 2 (Broken.NodeTree:Document.Special)
    + directory 2 (Broken.NodeTree:Document.Page)
    ```
    - `Page` lists all pages of type `Special` on the site and links to them
    - rendering only extends `Neos.Neos:Page` and only outputs the document title
    - the node-tree is set to render `Neos.Neos:Document,!Broken.NodeTree:Document.Special`
3. Navigate
    - `directory` > `special 1` > `directory` fails
    - `directory` > `special 1` > `directory 2` works
    - `directory 2` > `special 1` > `directory 2` fails

#### Expected behavior

The directory page should load again

#### Actual behavior

- The page does not load, a `/neos/preview?node[__contextNodePath]=…` request is not fired
- the inspector properties are loaded for the correct page

Navigating back to the parent page fails

https://user-images.githubusercontent.com/3868790/133447742-e4337fe9-4873-4b0f-9353-93fc58e250fe.mp4

Navigating back to a different page works as expected

https://user-images.githubusercontent.com/3868790/133447913-e6ac5fe9-5dbc-49c1-91ef-4ee8a2220519.mp4

Navigating back to the previous page fails; the parent/sub-page relation does not matter

https://user-images.githubusercontent.com/3868790/133447936-9dd2e457-eac9-450f-906f-f42bce805032.mp4

### Affected Versions
tested with

Neos: 7.1
UI: 7.1

grebaldi:
fixes: #2954 

**The problem**

The `<Frame/>`-component tracks its own window location through its internal `location` state. Whenever it receives a new `src` prop, it checks against its `location` state to decide whether or not it needs to actually change its window location.

This tracking mechanism covers every navigation event triggered by the document tree or dimension selector, but it misses all navigation events that happen inside the frame (e.g. through click on a link). This leads to strange behavior (as described in #2954) when an in-frame navigation happens and afterwards you try to navigate back to the document that the `<Frame/>`-component happens to have stored as its latest internal `location` state.

***NOTE:** #2954 mentions that this behavior occurs only if the navigated-to documents are hidden (via `baseNodeType` in node tree preset setting) in the document tree. I found that this actually happens regardless of those documents being hidden in the document tree.*

**The solution**

A adjusted the `onLoad`-handler of our `<Frame/>`-component, so that it stores the latest window location inside the internal `location` state. This way the internal state tracks in-frame navigation events as well.

### DIFF
--- a/packages/react-ui-components/src/Frame/frame.tsx
+++ b/packages/react-ui-components/src/Frame/frame.tsx
@@ -132,10 +132,25 @@ export default class Frame extends PureComponent<FrameProps> {
             transitioning: false
         });
 
-        const {onLoad} = this.props;
+        if (this.ref) {
+            const {onLoad} = this.props;
 
-        if (typeof onLoad === 'function' && this.ref) {
-            onLoad(e);
+            if (typeof onLoad === 'function') {
+                onLoad(e);
+            }
+
+            // Sync with iframe window location, in case the latest navigation
+            // happened in-frame (e.g. via click on link)
+            try {
+                const win = this.ref.contentWindow;
+                if (win && win.location.href !== this.state.location) {
+                    this.setState({
+                        location: win.location.href
+                    });
+                }
+            } catch {
+                // This almost definitely means we've lost access to the iframe
+            }
         }
     }
 


crydotsnake (approved):

mhsdesign (approved):
Fine by reading

PR #3498, opened 2023-05-22
BUGFIX: Properly pause mousetrap during inline editing

### LINKED ISSUE #3099: BUG: shorcuts/hotkeys dont work in guest-iframe
### Description

shorcuts/hotkeys dont work when focus is set to the guest-iframe (last mouse click was there)

### Steps to Reproduce

1. open the NeosUI Demo
2. klick into the iframe (not an editable element, but just somewhere 'safe' into the body)
3. type any shorcut like `t f`

#### Expected behavior

shortcut is executed

#### Actual behavior

shortcut does only work, if one previously clicked into the main frame someplace

### Code
looking quickly over `handleHotkeys`, this behaviour should be allowed (unless one is inline editing) ...
https://github.com/neos/neos-ui/blob/6ce277470258599c1d29f9ec2f5983e7e44d4338/packages/neos-ui-sagas/src/UI/Hotkeys/index.js#L6

### Affected Versions

UI: 8.0 but also 7.3 and maybe other versions

Chromium Edge 100.0.1185.36  & Firefox  97.0.1

grebaldi:
fixes: #3099 

**The problem**

We have a `Hotkeys` saga that initializes a `Mousetrap` (see: https://www.npmjs.com/package/mousetrap). The saga does so once for the host frame. For the guest frame it then initializes a separate `Mousetrap` every time, a document has been loaded.

In theory this should ensure that keyboard shortcuts work both in the host and in the guest frame. There's one specific scenario though in which we do not want keyboard shortcuts to work at all and that is: inline editing (basically, because typing something into CKEditor would otherwise trigger all kinds of unintended actions).

This is why the `Hotkeys` saga tracks a `mousetrapPaused` flag. The idea is to pause the mousetrap every time the content canvas signals that there's inline editing going on:

https://github.com/neos/neos-ui/blob/ef72c79b8cbc94cdaf58a3941953aedae8ea8f7c/packages/neos-ui-sagas/src/UI/Hotkeys/index.js#L39-L41

#3099 describes that the hot keys do not work inside the guest frame at all. I found that this is due to `mousetrapPaused` always being `true`.

Looking at the declaration of the action creator for `SET_CURRENTLY_EDITED_PROPERTY_NAME`:

https://github.com/neos/neos-ui/blob/ef72c79b8cbc94cdaf58a3941953aedae8ea8f7c/packages/neos-ui-redux-store/src/UI/ContentCanvas/index.ts#L59

I could see that the payload for `SET_CURRENTLY_EDITED_PROPERTY_NAME` is not an object with a key `propertyName`, but a string that represents just the property name. So, the condition `action.payload.propertyName !== ''` is always true, because `action.payload.propertyName` is always `undefined`.

**The solution**

I changed the condition for `mousetrapPaused` to

```js
mousetrapPaused = action.payload !== '';
```

to match the intended semantics with the action creator definition for `SET_CURRENTLY_EDITED_PROPERTY_NAME`. Now keyboard shortcuts work as intended in the guest frame, but are ignored if there's inline editing going on.

### DIFF
--- a/packages/neos-ui-sagas/src/UI/Hotkeys/index.js
+++ b/packages/neos-ui-sagas/src/UI/Hotkeys/index.js
@@ -37,7 +37,7 @@ export function * handleHotkeys({globalRegistry, store}) {
 
         // Pause mousetrap during inline editing
         if (action.type === actionTypes.UI.ContentCanvas.SET_CURRENTLY_EDITED_PROPERTY_NAME) {
-            mousetrapPaused = action.payload.propertyName !== '';
+            mousetrapPaused = action.payload !== '';
         }
     }
 }


mhsdesign (approved):
Thank you for this well written description and fix ❤️ 

seems like a classic js refactoring bug ^^

crydotsnake (approved):

PR #3501, opened 2023-05-22
BUGFIX: Check for unapplied inspector changes before allowing focus change via `<SelectedElement/>`

### LINKED ISSUE #3174: Escape "Unapplied changes" when using "Selected element" to navigate to other Node

grebaldi:
fixes: #3174 

**The problem**

#3174 describes a way to bypass the "Unapplied Changes"-dialog that is supposed to prevent the user from losing pending inspector changes.

The user can focus a different node by using the `<SelectedElement/>` container that is always at the top of the inspector:
image

This works, because the "Unapplied Changes"-dialog is exclusively triggered by an overlay that appears, when there are pending changes, and covers the entire UI (after https://github.com/neos/neos-ui/pull/3491 has been merged that is) except the inspector. Since the `<SelectedElement/>` container is part of the inspector, it is not covered by the overlay and thus allows the user to bypass the "Unapplied Changes"-dialog.

**The solution**

The first part of the solution introduces a new redux action to the inspector state called `ELEMENT_WAS_SELECTED`. *(The naming is up for discussion, since it introduces past-tense as opposed to imperatives - which is more correct but runs counter the established convention. It may or may not be a confusing name anyway though - idk)*

The inspector saga is then used to handle `ELEMENT_WAS_SELECTED` actions so that it first checks if there are pending changes in the inspector, triggers the "Unapplied Changes"-dialog if so, and - depending on the result of the latter - focuses the selected node or leaves the state untouched.

The second part of the solution concerns the handling of the `Neos.Neos.Ui:ReloadContentOutOfBand` server feedback. I noticed that if I selected an element and then clicked on "apply" in the "Unapplied Changes"-dialog, the node I've selected would be focused for a brief moment, after which the UI jumped back to the previously focused node. 

The reason for this was, that the server feedback handler for `Neos.Neos.Ui:ReloadContentOutOfBand` wanted to re-focus the node it just reloaded, without checking whether it was still *supposed* to be focused. I added a check for that, so that no jumping occurs anymore.

### DIFF
--- a/packages/neos-ui-redux-store/src/UI/Inspector/index.ts
+++ b/packages/neos-ui-redux-store/src/UI/Inspector/index.ts
@@ -41,6 +41,7 @@ export enum actionTypes {
     DISCARD = '@neos/neos-ui/UI/Inspector/DISCARD',
     ESCAPE = '@neos/neos-ui/UI/Inspector/ESCAPE',
     RESUME = '@neos/neos-ui/UI/Inspector/RESUME',
+    ELEMENT_WAS_SELECTED = '@neos/neos-ui/UI/Inspector/ELEMENT_WAS_SELECTED',
 
     //
     // Actions to control the secondary inspector window
@@ -61,6 +62,10 @@ const apply = () => createAction(actionTypes.APPLY);
 const discard = (focusedNodeContextPath: NodeContextPath) => createAction(actionTypes.DISCARD, {focusedNodeContextPath});
 const escape = () => createAction(actionTypes.ESCAPE);
 const resume = () => createAction(actionTypes.RESUME);
+const selectElement = (selectedElementContextPath: NodeContextPath) =>
+    createAction(actionTypes.ELEMENT_WAS_SELECTED, {
+        selectedElementContextPath,
+    });
 
 const openSecondaryInspector = () => createAction(actionTypes.SECONDARY_OPEN);
 const closeSecondaryInspector = () => createAction(actionTypes.SECONDARY_CLOSE);
@@ -76,9 +81,10 @@ export const actions = {
     discard,
     escape,
     resume,
+    selectElement,
     openSecondaryInspector,
     closeSecondaryInspector,
-    toggleSecondaryInspector
+    toggleSecondaryInspector,
 };
 
 export type Action = ActionType<typeof actions>;
--- a/packages/neos-ui/src/Containers/RightSideBar/Inspector/SelectedElement/index.js
+++ b/packages/neos-ui/src/Containers/RightSideBar/Inspector/SelectedElement/index.js
@@ -23,22 +23,22 @@ import style from './style.css';
         focusedNodeParentLine: selectors.CR.Nodes.focusedNodeParentLineSelector(state)
     };
 }, {
-    focusNode: actions.CR.Nodes.focus
+    selectElement: actions.UI.Inspector.selectElement
 })
 export default class SelectedElement extends PureComponent {
     static propTypes = {
         focusedNode: PropTypes.object.isRequired,
         focusedNodeParentLine: PropTypes.array.isRequired,
 
-        focusNode: PropTypes.func.isRequired,
+        selectElement: PropTypes.func.isRequired,
         nodeTypesRegistry: PropTypes.object.isRequired
     };
 
     handleSelectNode = selectedNodeContextPath => {
-        const {focusNode, focusedNode} = this.props;
+        const {selectElement, focusedNode} = this.props;
 
         if (selectedNodeContextPath && selectedNodeContextPath !== $get('contextPath', focusedNode)) {
-            focusNode(selectedNodeContextPath);
+            selectElement(selectedNodeContextPath);
         }
     };
 
--- a/packages/neos-ui/src/manifest.js
+++ b/packages/neos-ui/src/manifest.js
@@ -565,8 +565,14 @@ manifest('main', {}, globalRegistry => {
                 nodes
             })
         );
-        store.dispatch(actions.CR.Nodes.focus(contextPath, fusionPath));
-        store.dispatch(actions.UI.ContentCanvas.requestScrollIntoView(true));
+
+        const currentlyFocusedNodeContextPaths =
+            selectors.CR.Nodes.focusedNodePathsSelector(store.getState());
+
+        if (currentlyFocusedNodeContextPaths.includes(contextPath)) {
+            store.dispatch(actions.CR.Nodes.focus(contextPath, fusionPath));
+            store.dispatch(actions.UI.ContentCanvas.requestScrollIntoView(true));
+        }
     });
 
     //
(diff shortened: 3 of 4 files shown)


packages/neos-ui-redux-store/src/UI/Inspector/index.ts
```diff
@@ -61,6 +62,10 @@ const apply = () => createAction(actionTypes.APPLY);
 const discard = (focusedNodeContextPath: NodeContextPath) => createAction(actionTypes.DISCARD, {focusedNodeContextPath});
 const escape = () => createAction(actionTypes.ESCAPE);
 const resume = () => createAction(actionTypes.RESUME);
+const selectElement = (selectedElementContextPath: NodeContextPath) =>
```
crydotsnake, line 65: I would name it: `selectedElement` to keep it consistent with `selectedElementContextPath`.

packages/neos-ui-redux-store/src/UI/Inspector/index.ts
```diff
@@ -76,9 +81,10 @@ export const actions = {
     discard,
     escape,
     resume,
+    selectElement,
```
crydotsnake, line 84: `selectedElement`

packages/neos-ui/src/Containers/RightSideBar/Inspector/SelectedElement/index.js
```diff
@@ -23,22 +23,22 @@ import style from './style.css';
         focusedNodeParentLine: selectors.CR.Nodes.focusedNodeParentLineSelector(state)
     };
 }, {
-    focusNode: actions.CR.Nodes.focus
+    selectElement: actions.UI.Inspector.selectElement
```
crydotsnake, line 26:
```suggestion
    selectedElement: actions.UI.Inspector.selectedElement
```

packages/neos-ui/src/Containers/RightSideBar/Inspector/SelectedElement/index.js
```diff
@@ -23,22 +23,22 @@ import style from './style.css';
...
 })
 export default class SelectedElement extends PureComponent {
     static propTypes = {
         focusedNode: PropTypes.object.isRequired,
         focusedNodeParentLine: PropTypes.array.isRequired,
 
-        focusNode: PropTypes.func.isRequired,
+        selectElement: PropTypes.func.isRequired,
```
crydotsnake, line 33:
```suggestion
        selectedElement: PropTypes.func.isRequired,
```

packages/neos-ui/src/Containers/RightSideBar/Inspector/SelectedElement/index.js
```diff
@@ -23,22 +23,22 @@ import style from './style.css';
...
         nodeTypesRegistry: PropTypes.object.isRequired
     };
 
     handleSelectNode = selectedNodeContextPath => {
-        const {focusNode, focusedNode} = this.props;
+        const {selectElement, focusedNode} = this.props;
```
crydotsnake, line 38:
```suggestion
        const {selectedElement, focusedNode} = this.props;
```

packages/neos-ui/src/Containers/RightSideBar/Inspector/SelectedElement/index.js
```diff
@@ -23,22 +23,22 @@ import style from './style.css';
...
 
         if (selectedNodeContextPath && selectedNodeContextPath !== $get('contextPath', focusedNode)) {
-            focusNode(selectedNodeContextPath);
+            selectElement(selectedNodeContextPath);
```
crydotsnake, line 41:
```suggestion
            selectedElement(selectedNodeContextPath);
```

crydotsnake (changes requested):
Just a few naming nitpicks. Otherwise this works fine!

packages/neos-ui-redux-store/src/UI/Inspector/index.ts
grebaldi: It is an action creator though, so it describes the *action* of selecting an element - not the result. `selectElement` is imho consistent with the the other action creators above (`discard`, `escape`, `resume` - these are all imperatives).

crydotsnake: Okay. Understood! 👍🏽

crydotsnake (approved):

packages/neos-ui-redux-store/src/UI/Inspector/index.ts
```diff
@@ -41,6 +41,7 @@ export enum actionTypes {
     DISCARD = '@neos/neos-ui/UI/Inspector/DISCARD',
     ESCAPE = '@neos/neos-ui/UI/Inspector/ESCAPE',
     RESUME = '@neos/neos-ui/UI/Inspector/RESUME',
+    ELEMENT_WAS_SELECTED = '@neos/neos-ui/UI/Inspector/ELEMENT_WAS_SELECTED',
```
mhsdesign, line 44: im not sure that i would get without having this image in mind





that we are talking about the little dropdown up top ^^ - its easily forgettable. But then again the naming is scoped to the inspector so i think its still fine.

ill try to think of another name ;)

packages/neos-ui/src/manifest.js
```diff
@@ -565,8 +565,14 @@ manifest('main', {}, globalRegistry => {
...
-        store.dispatch(actions.CR.Nodes.focus(contextPath, fusionPath));
-        store.dispatch(actions.UI.ContentCanvas.requestScrollIntoView(true));
+
+        const currentlyFocusedNodeContextPaths =
+            selectors.CR.Nodes.focusedNodePathsSelector(store.getState());
+
+        if (currentlyFocusedNodeContextPaths.includes(contextPath)) {
+            store.dispatch(actions.CR.Nodes.focus(contextPath, fusionPath));
+            store.dispatch(actions.UI.ContentCanvas.requestScrollIntoView(true));
+        }
```
mhsdesign, line 575: that seems to be part of the second fix you mentioned that i dont really understand at first glance. But i will try some debugging and see if i can then make sense of those lines plus your description ;)

mhsdesign (changes requested):
Wow. i never expected that to be actually fixed, thank you for going through all the odd edge cases!!!

I will need to continue this review another time :)

grebaldi:
Closed in favor of https://github.com/neos/neos-ui/pull/3491

PR #3505, opened 2023-05-25
BUGFIX: Fix accessing nodeAggregateId during content reload

ahaeslich:
**How to verify it**

Change a property which has the setting `reloadIfChanged: true`. You will get the flash message "Object of class Neos\ContentRepository\Core\SharedModel\Node\NodeAggregateId could not be converted to string".


bwaidelich (approved):
<3

PR #3519, opened 2023-06-02
!!! FEATURE: Overhaul `NodeCreationHandlerInterface`

### LINKED ISSUE #3509: BUG: `title` label changes in node not reflected in creationDialog
if you have a node with the following config (to override the label name of the title)

```yaml
Foo.Bar:NodeType:
  superTypes:
    Neos.Neos:Document: true
  title:
    ui:
      label: "hello"
```

it will still show the translated label "Title" in the creation dialog, and in the inspector the correct label "hello"

to have the wanted effect, we need to specify it explicitly:

```yaml
Foo.Bar:NodeType:
  superTypes:
    Neos.Neos:Document: true
  ui:
    creationDialog:
      elements:
       title:
         ui:
           label: "hello"
  title:
    ui:
      label: "hello"
```

to understand why i dug deeper:

In Neos.Neos properties.title.label is set to `i18n` in 
https://github.com/neos/neos-development-collection/blob/ef5c5738550774d32f591dffcf87f799f0d54fde/Neos.Neos/Configuration/NodeTypes.Document.yaml#L45
(this we want to override)

when showInCreationDialog is set to true, the i18n will still contain the translation path which is equal to the one from when its used as property: `Neos.Neos:NodeTypes.Document:properties.title`
(as the label generation is done before the postprocessors are applied: https://github.com/neos/neos-development-collection/issues/2867)

in Neos.Neos.Ui we specify as title label as `i18n` [here](https://github.com/neos/neos-ui/blob/f2d18d59c3dafa0a9a69a8eff95550c0898f503e/Configuration/NodeTypes.yaml#L9), i think, so it can be targeted with a different translation (even though the translation lies in Neos.Neos https://github.com/neos/neos-development-collection/blob/ef5c5738550774d32f591dffcf87f799f0d54fde/Neos.Neos/Resources/Private/Translations/en/NodeTypes/Document.xlf#L17) and the translation is the same: "Title"

Due to the explicit specification in Neos.Neos.Ui it seems logically why we need to override this one too, but it is at least counter intuitive.

If there are more people expecting the behaviour like me than people expecting to be able to override the translation
`creationDialog.title` i think we should remove the line https://github.com/neos/neos-ui/blob/f2d18d59c3dafa0a9a69a8eff95550c0898f503e/Configuration/NodeTypes.yaml#L9 and with it the additional translation.

### LINKED ISSUE #3567: re-write NodeTemplates // extend API
Implementation https://github.com/neos/neos-ui/pull/3519

Rewrite NodeTemplates https://github.com/Flowpack/Flowpack.NodeTemplates/pull/51

### LINKED ISSUE #3615: BUG: reference type in creation dialog doesn't work

### Description

Using a property with `reference` type in the creation dialog in Neos 9 does not create the module and shows a red popup message with 
`Given property "myProperty" is declared as "reference" in node type "My.Site:Foo.Bar" and must be treated as such.`
The selection of a page in the CD works as desired.

YAML
```   
    myProperty:
      type: reference
      ui:
        label: 'My Reference Property'
        showInCreationDialog: true
        inspector:
          editor: 'Neos.Neos/Inspector/Editors/ReferenceEditor'
          editorOptions:
            nodeTypes: [ 'My.Site:Document.Page ]
      validation:
        Neos.Neos/Validation/NotEmptyValidator: {}
```

### Steps to Reproduce

1. define a property of the `reference` type with `showInCreationDialog: true`
2. create the module in Neos and select a page
3. click 'Create New'

#### Expected behavior

- module is created with the correct selected page as value for this property

#### Actual behavior

- module isn't created
- error popup appears

### Affected Versions

Neos 9 (not sure if the bug appeared since Neos 9 or even earlier)
PHP 8.2

neos/neos-ui - 9.0.x-dev 684bb2f
neos/neos-ui-compiled -  9.0.x-dev 9115a2e

Relates: neos/neos-development-collection#4549

mhsdesign:
Neos.Neos adjustment https://github.com/neos/neos-development-collection/pull/4630

Replaces: https://github.com/neos/neos-ui/pull/3720 for Neos9
Resolves: https://github.com/neos/neos-ui/issues/3615
Resolves: https://github.com/neos/neos-ui/issues/3509

For now the api will be internal, Flowpack.NodeTemplates provides a good abstraction for this.

### The `NodeCreationHandlerInterface`

Contract to hook into the process before the node creation command is handled by the content repository
You can add additional steps to the node creation.
For example adding initial properties `NodeCreationCommands::withInitialPropertyValues()`,
or queuing additional commands like to create a child via `NodeCreationCommands::withAdditionalCommands()`
The node creation handlers factory can be registered on a NodeType:

```yaml
Vendor.Site:Content:
  options:
    nodeCreationHandlers:
      myHandler:
        factoryClassName: 'Vendor\Site\MyHandlerFactory'
        position: end
```

The factory must implement the `NodeCreationHandlerFactoryInterface` and
return an implementation with this `NodeCreationHandlerInterface` interface.
The current content-repository or NodeType-manager will be accessible via the content repository.

The new signature of the interface is as follows:

```php
interface NodeCreationHandlerInterface extends ContentRepositoryServiceInterface
{
    /**
     * @param NodeCreationCommands $commands original or previous commands,
     *                                       with the first command being the initial intended node creation
     * @param NodeCreationElements $elements incoming data from the creationDialog
     * @return NodeCreationCommands the enriched node creation commands,
     *                              to be passed to the next handler or run at the end
     */
    public function handle(NodeCreationCommands $commands, NodeCreationElements $elements): NodeCreationCommands;
}
```

Additionally the interface was relocated to `Neos\Neos\Ui\Domain\NodeCreation\NodeCreationHandlerInterface`.

### The `NodeCreationCommands`

A collection of commands that describe a node creation from the Neos Ui.
The node creation can be enriched via a node creation handler.
The first command points to the triggered node creation command.
To not contradict the users intend it is ensured that the initial node
creation will be mostly preserved by only allowing to add additional properties.
Additional commands can be also appended, to be run after the initial node creation command.
All commands will be executed blocking.
You can retrieve the subgraph or the parent node (where the first node will be created in) the following way:

```php
$subgraph = $contentRepository->getContentGraph()->getSubgraph(
    $commands->first->contentStreamId,
    $commands->first->originDimensionSpacePoint->toDimensionSpacePoint(),
    VisibilityConstraints::frontend()
);
$parentNode = $subgraph->findNodeById($commands->first->parentNodeAggregateId);
```

### The `NodeCreationElements`

Holds the deserialized elements of the submitted node creation dialog form

Elements are configured like properties or references in the schema,
but its up to the node-creation-handler if they are handled in any way or just left out.

Elements that are of simple types or objects, will be available according to its type.
For example myImage will be an actual image object instance.

```yaml
Vendor.Site:Content:
  ui:
    creationDialog:
      elements:
        myString:
          type: string
        myImage:
```

Elements that refer to nodes are of type `references` or `reference`.
They will be available as NodeAggregateIds collection.

```yaml
Vendor.Site:Content:
  ui:
    creationDialog:
      elements:
        myReferences:
          type: references
```

The naming `references` in the `element` configuration does not refer to the content repository reference edges.
Referring to a node will just denote that an editor will be used capable of returning node ids.
The node ids might be used for setting references but that is up to a node-creation-handler.
              
To promoted properties / references the same rules apply:

```yaml
Vendor.Site:Content:
  properties:
    myString:
      type: string
      ui:
        showInCreationDialog: true
```


Classes/NodeCreationHandler/NodeCreationCommands.php
```diff
@@ -0,0 +1,58 @@
+<?php
+declare(strict_types=1);
+
+namespace Neos\Neos\Ui\NodeCreationHandler;
+
+use Neos\ContentRepository\Core\CommandHandler\CommandInterface;
+use Neos\ContentRepository\Core\Feature\NodeCreation\Command\CreateNodeAggregateWithNode;
+use Neos\ContentRepository\Core\Feature\NodeModification\Dto\PropertyValuesToWrite;
+
+class NodeCreationCommands implements \IteratorAggregate
```
mhsdesign: i dislike `CreateNodeAggregateWithNodeAndAdditionalCommands`

thats the one i like `NodeAggregateWithNodeCreationRecipe`

mhsdesign:
@bwaidelich suggested, that instead of passing the `ContentRepository` into the `NodeCreationHandlerInterface::handle()` we could also consider using the new domain services from the ESCR.

Those domain services can be  build via `ContentRepositoryRegistry::getService($crId, $serviceFactory)` https://github.com/neos/neos-development-collection/blob/f3d553315045c23b9deb7d9b897c84e39d399305/Neos.ContentRepositoryRegistry/Classes/ContentRepositoryRegistry.php#L107 


This would in turn mean, that the `NodeCreationHandlerInterface` would extend the `ContentRepositoryServiceInterface` (the domain service)

```php
interface NodeCreationHandlerInterface extends ContentRepositoryServiceInterface
{
    public function handle(NodeCreationCommands $commands, array $data): NodeCreationCommands;
}
```

and we wouldn't register the `NodeCreationHandlerInterface` directly in the configuration anymore, but its factory (the above mentioned `$serviceFactory`):

```yaml
'Neos.Neos:Content':
  options:
    nodeCreationHandlers:
      fooBar:
        factoryClassName: 'Some\NodeCreationHandlerFactoryInterface'
```

the factory would look like:
```php
interface NodeCreationHandlerFactoryInterface extends ContentRepositoryServiceFactoryInterface
{
    public function build(ContentRepositoryServiceFactoryDependencies $serviceFactoryDependencies): NodeCreationHandlerInterface;
}
```

and the `$serviceFactoryDependencies` would hold among others the ContentRepository:
https://github.com/neos/neos-development-collection/blob/01f7f8f6e5d7daa50a8bc68bdc280ddf3ccdb7f9/Neos.ContentRepository.Core/Classes/Factory/ContentRepositoryServiceFactoryDependencies.php#L45

Classes/Domain/NodeCreation/NodeCreationElements.php
```diff
@@ -0,0 +1,87 @@
...
+ *       elements:
+ *         myString:
+ *           type: string
+ *
+ * while reference-like elements are of type `references` or `reference`:
+ *
+ *     creationDialog:
+ *       elements:
+ *         myReferences:
+ *           type: references
```
mhsdesign: maybe we find a better name than `PropertyLike` and `ReferenceLike`

related: https://github.com/neos/neos-development-collection/pull/4677

Classes/Domain/NodeCreation/NodeCreationElements.php
```diff
@@ -0,0 +1,105 @@
...
+{
+    /**
+     * @param array<string, mixed> $propertyLikeValues
+     * @param array<string, NodeAggregateIds> $referenceLikeValues
+     * @param array<int|string, mixed> $serializedValues
+     * @internal you should not need to construct this
+     */
+    public function __construct(
+        private array $propertyLikeValues,
+        private array $referenceLikeValues,
```
mhsdesign:
```suggestion
        private array $elementValues,
```

array<string, NodeAggregateIds|mixed>

Classes/Domain/Model/Changes/AbstractCreate.php
```diff
@@ -48,6 +65,11 @@ abstract class AbstractCreate extends AbstractStructuralChange
      */
     protected ?string $name = null;
 
+    /**
+     * An (optional) node aggregate identifier that will be used for testing
+     */
+    protected ?NodeAggregateId $nodeAggregateId = null;
```
bwaidelich, line 71: a field that is only used for testing? suspicious :)

Classes/Domain/Model/Changes/AbstractCreate.php
```diff
@@ -88,6 +110,16 @@ public function getName(): ?string
...
 
+    public function setNodeAggregateId(string $nodeAggregateId): void
+    {
+        $this->nodeAggregateId = NodeAggregateId::fromString($nodeAggregateId);
+    }
+
+    public function getNodeAggregateId(): ?NodeAggregateId
+    {
+        return $this->nodeAggregateId;
+    }
```
bwaidelich, line 121: I suppose this is all just used for testing..
Can't we create some `AbstractCreateForTesting` (or similar) class that wraps this one or is that over the top?

Classes/Domain/NodeCreation/NodeCreationElements.php
bwaidelich: as far as I understand it, this DTO is an object that can be used to prime the creation of nodes with some parameters and we (currently) use it to map the values of the NodeCreation dialog.

As mentioned in our weekly today, I wonder why we need to separate the values already at this point. If we _do_ I would suggest `$propertyValues` and `$referencedNodeAggregateIds` but I would personally prefer to keep this as a simple key-value DTO for arbitrary parameters.

That would mean that the code below would instead of
```php
if (!$elements->hasReferenceLike($referenceName)) {
    continue;
}
$setReferencesCommands[] = SetNodeReferences::create(
    $commands->first->contentStreamId,
    $commands->first->nodeAggregateId,
    $commands->first->originDimensionSpacePoint,
      ReferenceName::fromString($referenceName),
      NodeReferencesToWrite::fromNodeAggregateIds($elements->getReferenceLike($referenceName))
);
```
be s.th. like:
```php
if (!$parameters->has($referenceName)) {
    continue;
}

$setReferencesCommands[] = SetNodeReferences::create(
    $commands->first->contentStreamId,
    $commands->first->nodeAggregateId,
    $commands->first->originDimensionSpacePoint,
      ReferenceName::fromString($referenceName),
      NodeReferencesToWrite::fromNodeAggregateIds(NodeAggregateIds::fromArrray($parameters->get($referenceName)),
);
```

(potentially with some error handling if the expected format does not match)

And I would even leave out the deserialization at this point..

bwaidelich (commented):
I only skimmed over it so far and left some comments.
I might be missing context, so feel free to ignore/overrule me :)

Classes/Domain/Model/Changes/AbstractCreate.php
mhsdesign: no yes :D all these classes are super ugly and hacky. I discussed adding the `nodeAggregateId` thing with sebastian and he was okay with it. But thanks for keeping an eye on this.

Classes/Domain/Model/Changes/AbstractCreate.php
mhsdesign: yes but acceptable: https://github.com/neos/neos-ui/pull/3519#discussion_r1500682880

Classes/Domain/NodeCreation/NodeCreationElements.php
mhsdesign: Yes thanks i had a little blockade in my brain as i somehow wanted to port the idea of separating properties and references https://github.com/neos/neos-development-collection/pull/4677 into the elements as well.

I also talked to Wilhelm about that and we found that `references` might be a little misleading as type here but introducing `nodeAggregateIds` or anything else is not worth it.
Its just important to note that the `type references` thingies have at that stage nothing in common with real reference edges. There is just a node editor that allows to "references" a node by its id.

I documented it like 

> The naming `references` in the `element` configuration does not refer to the content repository reference edges.
> Referring to a node will just denote that an editor will be used capable of returning node ids.
> The node ids might be used for setting references but that is up to a node-creation-handler.

So yes i simplified the vo and now it contains all elements in one blob.

> And I would even leave out the deserialisation at this point.

Here i have a strong objection. The data in a format only the neos ui understands. If every handler serialises it differently it leads to bugs like https://github.com/neos/neos-ui/issues/3719
By encapsulating it we make it an implementation detail and are later-on easier able to change the neos ui serialisation format.

Classes/Domain/NodeCreation/NodeCreationCommands.php
```diff
@@ -0,0 +1,130 @@
...
+     * @return \Traversable<int, CommandInterface>
+     */
+    public function getIterator(): \Traversable
+    {
+        yield $this->first;
+        foreach ($this->additionalCommands as $command) {
+            // if yield from is used, the default setting of iterator_to_array, "preserve_keys: true"
+            // would cause the first command to be overridden.
+            yield $command;
+        }
```
grebaldi:
```suggestion
        yield from [$this->first, ...$this->additionalCommands];
```

for brevity

Classes/Domain/NodeCreation/NodeCreationElements.php
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * @internal returns values formatted by the internal format used for the Ui
+     * @return \Traversable<int|string, mixed>
+     */
+    public function serialized(): \Traversable
+    {
+        yield from $this->serializedValues;
+    }
```
grebaldi, line 87:
```suggestion
```

This is marked `@internal`, but I couldn't find any references.

Classes/Domain/NodeCreation/NodeCreationElements.php
```diff
@@ -0,0 +1,96 @@
...
+    public function has(string $name): bool
+    {
+        return isset($this->elementValues[$name]);
+    }
+
+    /**
+     * Returns the type according to the element schema
+     * For elements that refer to a node {@see NodeAggregateIds} will be returned.
+     */
+    public function get(string $name): mixed
```
grebaldi, line 75: It would be great if `NodeCreationElements` would implement `ArrayAccess` and return `ValueAccessor`s (as introduced in https://github.com/neos/flow-development-collection/pull/3149). This would allow consumers to write:

```php
$myField = $elements['myField']->stringOrNull();
```

Otherwise, implementers would have to validate this stuff on their own (and likely in a very similar fashion).

But this is just a nice-to-have :innocent:

Classes/Domain/NodeCreation/NodeCreationHandlerInterface.php
```diff
@@ -0,0 +1,50 @@
...
+ *             position: end
+ *
+ * The factory must implement the {@see ContentRepositoryServiceFactoryInterface} and
+ * return an implementation with this {@see NodeCreationHandlerInterface} interface.
+ *
+ * The current content-repository or NodeType-manager will be accessible via the factory dependencies.
+ *
+ * @api
+ */
+interface NodeCreationHandlerInterface extends ContentRepositoryServiceInterface
```
grebaldi: I know, I'm probably opening a can of worms, but I have to say, that the whole `ContentRepositoryServiceFactoryInterface` -> `ContentRepositoryServiceInterface` chain makes me feel very uneasy.

I would much prefer to have a `NodeCreationHandlerInterface` that looks like this:

```php
interface NodeCreationHandlerInterface
{
    public function handle(EverythingTheNodeCreationHandlerNeeds $nodeCreationStuff): NodeCreationCommands;
}
```

Implying that `EverythingTheNodeCreationHandlerNeeds` consists of:
```php
final readonly class EverythingTheNodeCreationHandlerNeeds
{
    public function __construct(
        public NodeCreationCommands $commands, 
        public NodeCreationElements $elements,
        public NodeTypeManager $nodeTypeManager,
    ) {}
}
```
(possibly other stuff as well)

Pros:
- This way, `NodeCreationHandler`s can be `#[Flow\Scope("singleton")]` and utilize DI in all the familiar ways
- We retain full control over what is exposed as API to every `NodeCreationHandler`, without leaking all of `ContentRepositoryServiceFactoryDependencies` 

Cons:
- The hypothetical `EverythingTheNodeCreationHandlerNeeds` is probably hard to mock, which would render even simple use cases of `NodeCreationHandler`s hard to unit test (Though I believe that something's in the works to ease this kind of stuff)

Of course we cannot just call it `EverythingTheNodeCreationHandlerNeeds`. I think we have a broader problem here, which at least I seem to encounter rather frequently when designing (or just thinking about) extensibility-related APIs. It always feels like a lot of effort in order to avoid calling it `NodeCreationContext`, or `WhateverContext`, even when it is clearly the most appropriate naming scheme. I know the reason for our avoidance of this particular name. This avoidance however definitely feels paralyzing at times and leads to, dare I say, overengineered FactoryFactories and such.

I wouldn't want to block this PR for this reason, but I'd be interested in considering and discussing alternatives.

Classes/Domain/Service/NodePropertyConversionService.php
```diff
@@ -43,17 +45,13 @@ class NodePropertyConversionService
      *
      * @param string|array<int|string,mixed>|null $rawValue
      */
-    public function convert(NodeType $nodeType, string $propertyName, string|array|null $rawValue): mixed
+    public function convert(string $propertyType, string|array|null $rawValue): mixed
```
grebaldi, line 48: Nice one! :)

Classes/Domain/Service/NodePropertyConversionService.php
```diff
@@ -113,6 +111,35 @@ public function convert(NodeType $nodeType, string $propertyName, string|array|n
...
+            }
+            $propertyType = $elementConfiguration['type'] ?? 'string';
+            if ($propertyType === 'references' || $propertyType === 'reference') {
+                $nodeAggregateIds = [];
+                if (is_string($rawValue) && !empty($rawValue)) {
+                    $nodeAggregateIds = [$rawValue];
+                } elseif (is_array($rawValue)) {
+                    $nodeAggregateIds = $rawValue;
+                }
+                $convertedElements[$elementName] = NodeAggregateIds::fromArray($nodeAggregateIds);
```
grebaldi, line 134:
```suggestion
                $convertedElements[$elementName] = $this->convertReferenceOrReferences($rawValue);
```

Just a refactoring nitpick :innocent:

grebaldi (commented):
Hi @mhsdesign,

thanks so much for taking care of ESCRing the node creation extensibility mechanism. I love your idea of having a declarative change set via `NodeCreationCommands`. As I was testing your change manually, I also came across your execllent code documentation and error handling - which turned out to be very helpful indeed :) All in all, node creation handling via your design works like a charm :+1: 

Also, :+1: for continuing the Layered-Architecture namespacing :)

The code comments I have left touch on some difficult design aspects. All of my statements represent nothing but my own two cents on the matter and I believe that we require broader consensus on how the design of the API should be. I do not intend to block this PR because of it - it'd be fine for me, if we stick to the design as it is, but I wanted to leave some room for discussion, that's why I didn't click "approve" right away.

Finally, I would like to offer a thought on what this change may mean going forward: Because `NodeCreationHandlerInterface::handle` returns the `NodeCreationCommands` object, we're now capable invoking an (almost) arbitrary set of commands via the node creation dialog. It is even possible to replace the original intent of creating a node of the selected node type (correct me if I'm wrong).

This is by no means a problem. I rather think that it points us towards a more generic task-based interface, that is not NodeType-centric. This isn't relevant for the PR at hand, but I can imagine that this may lead to projects eventually using pseudo NodeTypes to emulate it (because NodeTypes are the only way to initiate the NodeCreation mechanism). At some point we would need to decouple NodeTypes from the NodeCreationDialog (and rename the entire "NodeCreation" concern to something more akin to "PerformContentTask" or similar).

Now, that's enough from me :sweat_smile: Thanks again, @mhsdesign for putting in so much effort to make this stuff work :heart:

Classes/Domain/NodeCreation/NodeCreationElements.php
mhsdesign: Yes, but it might be used by plugins as legacy behaviour like NodeTemplates: https://github.com/Flowpack/Flowpack.NodeTemplates/blob/8671cd6b6d62bbe957cb0a9038c2c2caa0e81fe8/Classes/Domain/TemplateNodeCreationHandler.php#L61-L62

I actually am thinking about deprecating this in node templates as well, but still lack some input ;) https://github.com/Flowpack/Flowpack.NodeTemplates/issues/77#issuecomment-1951363663

Classes/Domain/NodeCreation/NodeCreationHandlerInterface.php
mhsdesign: I do agree that this makes creating a node creation handler a bit more boiler platy, but i wanted to try out the extensibility mechanism of "core domain content repository services" which Bastian pointed out to me. (see https://github.com/neos/neos-ui/pull/3519#issuecomment-1584046648 where i elaborated the approach)

But i would very much like to discuss this with you in person, i think it comes down to personal preference.
I for one think its cool that all services that do something strongly related to the content repository implement the interface and use that pattern as it makes them easy to identify and it automatically avoids this logic being placed in a controller directly.

Classes/Domain/NodeCreation/NodeCreationElements.php
mhsdesign: Hmmm i somehow still find the naming and handling of the `ValueAccessors` a little strange and mostly like to validate, or gracefully validate (string cast or no error) which this cannot, on my own.

I think we should put the `ValueAccessors` first under a real stress test where we heavily read configuration (like the content repository registry, or the site configuration), so that we can be sure that it is now final api and feature complete ;)

But lets discuss this in person as well ;)

mhsdesign:
Thank you so much for your kind words and the extensive feedback! ❤️ 

I could answer a few questions already: 

> It is even possible to replace the original intent of creating a node of the selected node type (correct me if I'm wrong).

It _is_ technically, but 100% marked as footgun and not to do, as it contradicts the intention, and will crash if the ui cant find the node id afterwards ;)

> This is by no means a problem. I rather think that it points us towards a more generic task-based interface, that is not NodeType-centric. 

Hmm i think as well that this would be moving to fast and maybe we can implement this for a `NodeCreationDialog` version 3
For now i would say strong coupling to the content repository (via the cr services https://github.com/neos/neos-ui/pull/3519#discussion_r1513491762) is already a good step forward.

mhsdesign:
As discussed with Wilhelm and Christian we will make the node creation handlers internal for now, as there is not much demand for the api and the main consumers are the NeosUi itself or Flowpack.NodeTemplates. Currently its not worth to draft out the perfect api and declare it as such, even though the implementation will unlikely change in the next time (hopefully ^^)

Additionally using the `ContentRepositoryServiceInterface` as discussed in https://github.com/neos/neos-ui/pull/3519#issuecomment-1584046648 was reverted in favour of a much simpler factory interface which will not leak the hardcore `ContentRepositoryServiceFactoryDependencies`:


```php
interface NodeCreationHandlerFactoryInterface
{
    public function build(ContentRepository $contentRepository): NodeCreationHandlerInterface;
}
```

but as its internal now either way, it shouldnt matter too much ^^

grebaldi (approved):

Tests/IntegrationTests/TestDistribution/composer.json
```diff
@@ -32,6 +32,9 @@
     },
     "extra": {
       "patches": {
+        "neos/neos-development-collection": {
+          "TASK: Remove declaration of ui nodeCreationHandlers": "https://github.com/neos/neos-development-collection/pull/4630.patch"
```
grebaldi, line 36: @mhsdesign:

We missed this one.

mhsdesign: ups 😅

PR #3521, opened 2023-06-10
TASK: improve document title node creation handler

### LINKED ISSUE #3567: re-write NodeTemplates // extend API

mhsdesign:
depends on:

https://github.com/neos/neos-development-collection/pull/4324

in Neos 8.3, we could use the `NodeUriPathSegmentGenerator::generateUriPathSegment`  but now we dont have actually the node at hand.

Previously `generateUriPathSegment` was partially inlined but by creating the lower level helper `generateUriPathSegmentFromTextForDimension` we can use it again.

### DIFF
--- a/Classes/NodeCreationHandler/NodeCreationCommands.php
+++ b/Classes/NodeCreationHandler/NodeCreationCommands.php
@@ -1,0 +1,58 @@
+<?php
+declare(strict_types=1);
+
+namespace Neos\Neos\Ui\NodeCreationHandler;
+
+use Neos\ContentRepository\Core\CommandHandler\CommandInterface;
+use Neos\ContentRepository\Core\Feature\NodeCreation\Command\CreateNodeAggregateWithNode;
+use Neos\ContentRepository\Core\Feature\NodeModification\Dto\PropertyValuesToWrite;
+
+class NodeCreationCommands implements \IteratorAggregate
+{
+    /** @var array<int|string, CommandInterface> */
+    public readonly array $additionalCommands;
+
+    /**
+     * @interal to guarantee that the initial create command is mostly preserved as intended.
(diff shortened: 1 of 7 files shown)


Classes/NodeCreationHandler/NodeCreationCommands.php
```diff
@@ -0,0 +1,62 @@
...
+        /**
+         * The original node creation command.
+         * It is only allowed to mutate the properties via {@see self::withInitialPropertyValues()}
+         */
+        public readonly CreateNodeAggregateWithNode $initialCreateCommand,
+        /**
+         * Add a list of commands that are executed after the initial created command was run.
+         * This allows you to create child-nodes and other operations.
+         */
+        CreateNodeAggregateWithNode|SetNodeProperties|DisableNodeAggregate|EnableNodeAggregate|SetNodeReferences ...$additionalCommands
```
mhsdesign, line 33: @jonnitto just told me about his usecase to copy nodes on creation from a "template", so we should allow copy too

Classes/NodeCreationHandler/NodeCreationCommands.php
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * @internal to guarantee that the initial create command is mostly preserved as intended.
+     * You should use {@see self::withInitialPropertyValues()} and {@see self::withAdditionalCommands()} instead.
+     */
+    public function __construct(
+        /**
+         * The original node creation command.
+         * It is only allowed to mutate the properties via {@see self::withInitialPropertyValues()}
+         */
+        public readonly CreateNodeAggregateWithNode $initialCreateCommand,
```
mhsdesign, line 28: we should name it $first instead

mhsdesign:
Outdated

PR #3522, opened 2023-06-11
TASK: abstractCreate Make NodeName Obsolet

mhsdesign:
we should discuss, if we need to set a nodeName at all, as it would resolve todo:

```php
        // TODO: the $name=... line should be as expressed below
        // $name = $this->getName() ?: $this->nodeService->generateUniqueNodeName($parent->findParentNode());
        $nodeName = NodeName::fromString($this->getName() ?: uniqid('node-', false));
```

@nezaniel what do you say? I could not find a similar functionality to `generateUniqueNodeName` in the ESCR. Also when looking at the soft constraint in the NodeCreationCommandHandler one can see that it is not that super simple to find out if a nodeName has already been occupied, so we might as well dont introduce this again?

I have the feeling we really need to discuss nodeNames, currently people mostly use it for small unique identifiers on the same page. Like for building up html ids to link to. In case the nodeName would be null it would be unexpected.

**What I did**

**How I did it**

**How to verify it**


bwaidelich (approved):
Looks good by reading – I wonder why we used the previous version at first?

PR #3523, opened 2023-06-11
9.0: Use site detection result instead of the default site in hostframe

nezaniel:
In a 9.0 multisite installation, the hostframe always initialized the default site

**What I did**
**How I did it**
I replaced the default site with the site detection result

**How to verify it**
Have a Neos9 multisite installation and call any homepage but the default one


skurfuerst (approved):
Good catch, makes perfect sense :)

PR #3524, opened 2023-06-12
BUGFIX: #3495 selectBoxHeaderWithSearchInput inputContainer has max-height

### LINKED ISSUE #3495: BUG: Dimension Switcher closes on option hover
regression from https://github.com/neos/neos-ui/pull/3451

### Description

In Chrome or Brave and Neos 8.3 the new dimension switcher closes itself when an unselected option gets highlighted by hovering over it and the drop down cannot be reopened. It is working in Firefox though.

### Steps to Reproduce

1. Open the Dimension Switcher in the top bar
2. Without clicking, just hover over an unselected dimension option. The Dimension Switcher immediately closes itself
3. Click on the Open/Close icon trying to reopen the Dimension Switcher. This also does not work
4. Click somewhere else to finally "reset" the Dimension Switcher to be able to repeat step 1.

The video is recorded on Ubuntu 22.04.2 LTS and Chrome Version 113.0.5672.63 (Official Build) snap (64-bit).  
This behaviour is also reported from colleagues running Chrome or Brave on Mac OS.

#### Expected behavior

Able to select and click a dimension to switch to the dimension.

#### Actual behavior

[Screencast from 2023-05-17 14:46:33.webm](https://github.com/neos/neos-ui/assets/19683930/7d1627a0-3759-4f89-84b9-44843240fd02)

Note: In the video, I clicked on the Dimension Switcher open/close icon, multiple times to demonstrate that it does not reopen again.

### Affected Versions

Neos: 8.3.2

UI: 8.3.0

Related: #3413, #3451

gjwnc:
This PR fixes #3495 . Thanks to @Liioooo for finding and fixing the issue!  
Tbh I don't actually understand why this fixes the bug, but I can confirm dimension switcher now works in chrome too.

**How to verify it**

Just open the dimension switcher with multiple languages in chrome and select a language

### DIFF
--- a/packages/react-ui-components/src/SelectBox_HeaderWithSearchInput/style.module.css
+++ b/packages/react-ui-components/src/SelectBox_HeaderWithSearchInput/style.module.css
@@ -5,6 +5,7 @@
 
 .selectBoxHeaderWithSearchInput__inputContainer {
     width: 100%;
+    max-height: 40px;
 }
 
 .selectBoxHeaderWithSearchInput__input {


crydotsnake (changes requested):
Hm. I don't really like the solution. Because now we have the problem that the language switcher can no longer be closed via the icon when it is unfolded. And on the other hand, moving the mouse over the language changes the width of the elements. Which has a rather unsightly effect.

https://github.com/neos/neos-ui/assets/39345336/91e82d0b-e5a6-47ec-ac0f-d6e06297dbde

And also, the magnifying glass icon is a little too close to the language switcher.

SCR-20230612-tity

## Possible Solution

My suggestion would be that the dropdown with the languages gets by default the same width as the dropdown field above. So that there is virtually a transition, and the lines are exactly matching.

Because if we would do it the other way around: The dropdown field at the top would be as wide as the list with the languages, I'm not sure if it wouldn't be too narrow with the distance in the toolbar.

SCR-20230612-tjjp


Afterwards, it would be necessary to check whether the icon with the magnifying glass is still too close to the language changer.

What do you think? @markusguenther

gjwnc:
@crydotsnake I understand your very valid concerns. Our goal here was to get the dimension switcher to at-least work. E.g. the topic of the bug report #3495, was that the dimension switcher auto-closes itself in Chrome. And this PR fixes this.  

What you described in your video is the current Neos 8.3 behaviour and this PR does not interfere with this afaik.  
Imho, your concerns mentioned here should be moved in a separate bugreport. Everything you describe is current standard behaviour in NEOS 8.3 and is unrelated to this fix.  
I think the current released dimension switcher is broken. And we tried to get it usable with this fix for all browser. So my suggestion is to apply this fix and maybe refactor the dimension switcher afterwards to fix your concerns.

crydotsnake:
Hello @gjwnc 

> What you described in your video is the current Neos 8.3 behaviour and this PR does not interfere with this afaik.
Imho, your concerns mentioned here should be moved in a separate bugreport. 

The issues I mentioned in my comment above and which are also visible in my video, have only occurred with the changes of this pull request. Therefore, I think that my comments from above belong here in this pull request. 🙂

> Everything you describe is current standard behaviour in NEOS 8.3 and is unrelated to this fix. I think the current released dimension switcher is broken. And we tried to get it usable with this fix for all browser.

That's not quite right. The dimension switcher only breaks, when more than 10 languages are configured. But for example with only 3 languages configured, the dimensions switcher behaves like usual in 8.3:

https://github.com/neos/neos-ui/assets/39345336/b259c67d-86e7-4700-b9dd-5fd9296b34d4

crydotsnake:
## Comparison

> The dimension switcher only breaks, when more than 10 languages are configured. 

https://github.com/neos/neos-ui/assets/39345336/e748a60c-2065-479b-b086-ab3befbfc8a7

gjwnc:
To avoid confusion, I try to summarize the observations @crydotsnake mentioned from my perspective. It seems the behavior depends on browser and number of languages.

My and company team member observations are these:
Given a project with more than 10 languages and without this PR and default Neos 8.3 UI we have the following
* `language switcher can no longer be closed via the icon when it is unfolded` This happens on all tested browser FF, Chrome, Brave and Safari
* `moving the mouse over the language changes the width of the elements` Also happens on all tested browser FF, Chrome, Brave, Safari
* `magnifying glass icon is a little too close to the language switcher` Same as above
* `the language menu disappears on hover` Happens only in Brave and Chrome on Ubuntu and Mac. This can be seen in Video 3 at second 23.

The last mentioned issue - which is the main problem - disappears  on all tested browser for our project with the fix from this PR, that is adding `max-height: 40px` to `.selectBoxHeaderWithSearchInput__inputContainer`. This might be a hint on what is going on to someone else :shrug:    
Since there are now a bunch of other issues mentioned here, you're right @crydotsnake, that this PR does not fix this all and this needs more time to fix.

I'm not able to fix the dimension switcher with everything mentioned because of lack of know-how in frontend dev.   
@crydotsnake Maybe you could also elaborate on your setup Mac/Win/*nix or which browser you tested to get a better picture on what is working under which conditions and what not? And since I cannot fix this, I would suggest to comment in the original issue #3495 or creating a new one with more details?  
What is very confusing to me is that, if I get you right, this PR is causing the disappearing behavior in Video 3 second 23 instead of fixing it, right? We have the exact opposite observations here.   
How do we proceed with this PR and the issue itself?

mhsdesign (changes requested):
i found another hotfix: https://github.com/neos/neos-ui/pull/3565

as we still need to investigate why the selectbox fails, this might be a good start so id like this pr to stay open and we can prepare a proper fix ;)

crydotsnake:
Closed since #3565 is merged.

Issue #2918, opened 2021-06-15
Bug: Copy Paste Ckeditor Invalid HTML code in DB properties


markuspfeifenberger:
### Description

If a property is configured as inlineEditable (true), without autoparagraph (false) and removeFormat (true), the HTML code sent to Neos and saved in the database is not valid, when the Code is pasted, eg using copy-paste of a formated browser content.
            
### Steps to Reproduce

1. NodeType configuration
```
properties:
    headline:
      type: string
      ui:
        inlineEditable: true
        inline:
          editorOptions:
            autoparagraph: false
            formatting:
              removeFormat: true
            placeholder: i18n
```
2. Edit the headline in Neos and copy-paste HTML from a browser, e.g.:

```
<p><b>This text is bold</b></p>
<p><i>This text is italic</i></p>
<p>This is<sub> subscript</sub> and <sup>superscript</sup></p>
```
(The code can be copied from https://www.w3schools.com/html/html_formatting.asp - first white box, all three lines)

#### Expected behavior

The code should be stored as correct HTML

```
This text is bold This text is italic This is subscript and superscript
```

#### Actual behavior

Request URL: `/neos/ui-services/change` sends the code, where the payload is changed and stored in the DB
```
This text is bold</span><span>This text is italic</span><span>This is subscript and superscript
```
### Affected Versions

Neos: > 5

UI: 7.0.3

![Screen Shot 2021-06-15 at 14 58 45](https://user-images.githubusercontent.com/7107293/122056952-68d89d80-cdea-11eb-91f3-4f2ff296f91a.png)

Sebobo:
Can confirm this bug in my customers project. I'll try to get some budget to fix it.

PR #3532, opened 2023-06-21
BUGFIX: Fix ckeditor inline mode `autoparagraph: false`

mhsdesign:
Bug from https://github.com/neos/neos-ui/pull/2018
Resolves: #2918

Since there is still no native support for a inline ckeditor mode. https://github.com/ckeditor/ckeditor5/issues/762 we will continue our hacky road ;)

This fixes several issues:
- Inserting text with newlines will now correctly insert soft breaks (br)
- We use our own pseodo `<neos-inline-wrapper>` tag to avoid the issue of having to parse html with regex and doing it wrong #2918

i also added a test for some inline editing behavior. (Like that there are no outer span or p tags)
But writing a tests for #2918 is currently impossible, as our tests currently dont use `https` which means we cant access the `navigator.clipboard` and test that ckeditor works correctly when content is pasted. For that reasons you can use the new manual testing environment to check everything is working correctly.

-------------

We remove the outer tags in the first place, so they are not included in the final html.

image

Without processing:

image

**What I did**

**How I did it**

**How to verify it**


Sebobo:
We now have to decide to still use this hack or adjust the CK plugin to fix this in the model....

crydotsnake (approved):
Tested and works as expected

mhsdesign:
We must also test this carefully with other ck plugins ^^

mhsdesign:
I just found a regression in 7.3 which doesn't appear in 8.3:
The heading select box doesnt work once you selected "paragraph" - you can never go back to a h1 h2 heading.

This made no sense in my head, as the node type definition were exactly the same and i though the code base should have changed much and that we should both have ckeditor v16 installed.

Turns out wrong. This 2020 update (Neos 5.2) https://github.com/neos/neos-ui/pull/2624/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R97-R106 - which supposedly updated ckeditor from 11/12 to 16 actually only changed the double declared dependencies in the root package json, but the "real" dependencies on the ckeditor-bindings package https://github.com/neos/neos-ui/blob/14e1a0b4fab75c10b437e31880b856b54ff290e4/packages/neos-ui-ckeditor5-bindings/package.json#L13-L24 were forgotten. That means the update dindt happen at this suspected time. Only lately while cleaning up the codebase in Neos 8.2 we dropped the duplicated imports declaring "12" and were left with the "16" version. That means we accidentally did the real ckeditor update in https://github.com/neos/neos-ui/pull/3116


I will revert the "optional" `isLimit: true` part as it doesnt seem to work well in ckeditor v 12 and will target 8.2 for this special fix.


## Edit 2

the issue `isLimit` caused could also be reproduced in Neos 8.3. I tracked the "issue" down to that the heading command is not enabled, because `getSelectedBlocks` will stop searching if it encounters a limit element: https://github.com/ckeditor/ckeditor5-engine/blob/28c6f90d62ad33fd1b1199f7a45821ddc546c2d3/src/model/selection.js#L856

I believe that this is not an issue, but we simply use "[isLimit](https://ckeditor.com/docs/ckeditor5/16.0.0/api/module_engine_model_schema-SchemaItemDefinition.html)" wrong ^^.

> `isLimit` – It can be understood as whether this element should not be split by `Enter`. Examples of limit elements: `$root`, table cell, image caption, etc. In other words, all actions that happen inside a limit element are limited to its content. **Note**: All objects (`isObject`) are treated as limit elements, too.

So while it makes sense to not be able to set the heading like `h1` - as we are indeed in the INLINE mode, this would be to breaky in a bugfix.

Also `isLimit` will only prevent from multiple paragraphs being manually entered -> eg by pasting or entering, but when the contents are already two p tags `<p>1</p><p>2</p>` or a heading `<h1>1</h1>` the editor will not forbid this.

packages/neos-ui-ckeditor5-bindings/tests/manual/index.html
```diff
@@ -0,0 +1,21 @@
+<html lang="en">
```
mhsdesign, line 1: i added some manual tests, which can be used to test our ckeditor integration without a fully fledged neos

mhsdesign:
We actually dont need `isLimit: true` to solve that newlines wont be preserved on paste

To enable inserting `br` when content with newlines is pasted, it seems enough to remove this upcast `editor.conversion.for('upcast').elementToElement({model: 'paragraph', view: 'span', converterPriority: 'high'});` - i did this accidentally (by only specifying "editingDowncast" and "dataDowncast")

Sebobo (approved):
❤️

PR #3533, opened 2023-06-22
TASK: PHP 8.1 deprecations compatibility

kdambekalns:
This fixes deprecation warnings. We cannot specify the return types, as Neos 7.3 needs to be compatible with PHP 7.4. As of Neos 8.0 the proper types are added.

**What I did**

**How I did it**

**How to verify it**

Run this on PHP 8.1, no more deprecation warnings should appear (`PHP Deprecated:  Return type of Neos\Neos\Ui\Domain\Model\RenderedNodeDomAddress::jsonSerialize() should either be compatible with JsonSerializable::jsonSerialize(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in …`)

### DIFF
--- a/Classes/Domain/Model/FeedbackCollection.php
+++ b/Classes/Domain/Model/FeedbackCollection.php
@@ -62,6 +62,7 @@ class FeedbackCollection implements \JsonSerializable
      *
      * @return array
      */
+    #[\ReturnTypeWillChange]
     public function jsonSerialize()
     {
         $feedbacks = [];
--- a/Classes/Domain/Model/RenderedNodeDomAddress.php
+++ b/Classes/Domain/Model/RenderedNodeDomAddress.php
@@ -97,6 +97,7 @@ class RenderedNodeDomAddress implements \JsonSerializable
      *
      * @return array
      */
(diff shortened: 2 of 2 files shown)


grebaldi (approved):
Hi @kdambekalns,

this makes total sense, of course. 

Yet, I'm a bit mystified by not being able to see the deprecation messages in the first place... Since this affects the `FeedbackCollection` class, those messages should light up all over the place, but my PHP 8.1 reproduction environment stays silent. (If I run the affected scripts through PHP directly however, I *do* see the messages - which is how I was able to verify the effectiveness of your change.)

I suspect therefore that there's some suppression mechanism in place somewhere, which should probably be removed as well.

Unless of course it's just my environment behaving strangely :sweat_smile: 

Did you see deprecation warnings in the UI somewhere?

kdambekalns:
No, for the UI they came up when running Behat tests. In `Testing` context, we handle those warnings as exceptional – in other contexts (probably) not.

With Flow and Neos the warnings are clearly visible if you just run unit or functional tests…

PR #3535, opened 2023-06-22
BUGFIX: Ensure that only numerical property values are converted to integer

### LINKED ISSUE #3157: Unsetting an integer property isn't possible
### Description

You can't set the value of an integer property to `null` after setting a valid integer value before. I think this should be possible, because sometimes `null` as value for the property might be handy. E.g. if your valid integer value range is -∞ to +∞ (including 0).  So, just entering no value before submitting the changes could accidentally result in a valid value (which is `0`).

### Steps to Reproduce

1. Create an int property

```'V.S:NodeType':
  properties:
    integerValue:
      type: integer
      ui:
        label: 'Test Integer'
        reloadIfChanged: true
```

2. Go to the backend and enter any value, apply and publish
3. Delete the value, apply and publish
4. Check the property value in the Eel shell or with Neos.Fusion:Debug prototype it will be `0`

#### Expected behavior

Deleting the value should IMHO result in `null` value for the property.

#### Actual behavior

The value will be set to `0` instead of `null`.

### LINKED ISSUE #3158: Backend transforms an entered string to 0 for integer properties.
### Description

If you enter a string into the input filed of an integer property the backend turns it into a `0` when saving. If your valid integer value range is -∞ to +∞ (including 0), then entering an invalid value before submitting the changes could accidentally result in a valid value (which is `0`).

### Steps to Reproduce

1. Create an int property

```'V.S:NodeType':
  properties:
    integerValue:
      type: integer
      ui:
        label: 'Test Integer'
        reloadIfChanged: true
```

2. Go to the backend and enter a string value (`foo`), apply and publish
4. Check the property value in the Eel shell or with Neos.Fusion:Debug prototype it will be `0`

#### Expected behavior

Entering an invalid value should IMHO result in `null` as value for the property.

#### Actual behavior

The property will be set to `0`. This is unexpected/unwanted behaviour IMHO because 0 might be a valid value and an invalid value was entered. So, currently the backend might transform an invalid value in a valid one.

grebaldi:
fixes: #3158, #3157 

**The problem**

The UI's `NodePropertyConversionService` over-generalizes when converting incoming values to integer. It just casts any incoming `string` to `int` via `(int) $someValue`.

Now, PHP throws no error if the value we're casting to `int` actually cannot be meaningfully converted. So, pretty much anything it cannot make sense of ends up being `0`:

```
php > echo (int) "foo";
0
```

**The solution**

I adjusted the method `convertInteger` of `NodePropertyConversionService` to only convert strings to integers that pass PHP's `is_numerical` check.

In case the string doesn't pass, `convertInteger` now defaults to `null`.

Because of this, I needed to adjust the `TextFieldEditor` component, so that it now handles the occurrence of `null`-values correctly.

As a result, if I enter a non-numerical string into a text field for a property of type `integer`, the property will receive `null` as a value. The same thing happens, if I enter an empty string.

### DIFF
--- a/Classes/Domain/Service/NodePropertyConversionService.php
+++ b/Classes/Domain/Service/NodePropertyConversionService.php
@@ -155,11 +155,15 @@ class NodePropertyConversionService
      * Convert raw value to integer
      *
      * @param mixed $rawValue
-     * @return integer
+     * @return null|integer
      */
     protected function convertInteger($rawValue)
     {
-        return (int)$rawValue;
+        if (is_numeric($rawValue)) {
+            return (int) $rawValue;
+        }
+
+        return null;
     }
 
     /**
--- a/packages/neos-ui-editors/src/Editors/TextField/index.js
+++ b/packages/neos-ui-editors/src/Editors/TextField/index.js
@@ -41,7 +41,7 @@ export default class TextField extends PureComponent {
             className={className}
             id={id}
             autoFocus={finalOptions.autoFocus}
-            value={value}
+            value={value === null ? '' : value}
             onChange={commit}
             placeholder={placeholder}
             onKeyPress={onKeyPress}
--- a/packages/neos-ui-editors/src/Editors/TextArea/index.js
+++ b/packages/neos-ui-editors/src/Editors/TextArea/index.js
@@ -34,7 +34,7 @@ export default class TextAreaEditor extends PureComponent {
 
         return (<TextArea
             id={id}
-            value={value}
+            value={value === null ? '' : value}
             className={className}
             onChange={commit}
             disabled={finalOptions.disabled}


packages/neos-ui-editors/src/Editors/TextField/index.js
```diff
@@ -41,7 +41,7 @@ export default class TextField extends PureComponent {
             className={className}
             id={id}
             autoFocus={finalOptions.autoFocus}
-            value={value}
+            value={value === null ? '' : value}
```
mhsdesign, line 44: i must say i dont fully understand this line 😂

mhsdesign (approved):
Thanks you ;)
Looks good for know. In the future (Neos 9) id like to have strict validation via ajax directly in the editor. (And we could make our NodePropertyConversion thing more strict by throwing exceptions ...)

packages/neos-ui-editors/src/Editors/TextField/index.js
bwaidelich: I suppose this errored for a value of `null` and since `convertInteger` now has a nullable return value, this became an issue.

Classes/Domain/Service/NodePropertyConversionService.php
```diff
@@ -155,11 +155,15 @@ protected function convertDateTime($rawValue)
...
+     * @return null|integer
      */
     protected function convertInteger($rawValue)
     {
-        return (int)$rawValue;
+        if (is_numeric($rawValue)) {
+            return (int) $rawValue;
+        }
+
+        return null;
```
bwaidelich, line 166: I'm not sure where this is called etc. but is turning `"notANumber"` into `null` better than 0?
And could other parts of the system rely on this return type to be not nullable (like the one below)?

I wonder if an exception would be feasible here?

mhsdesign: Exceptions have to be caught or at least nicely presented to the editor. 

I plan it as said above for 9 with https://github.com/neos/neos-development-collection/issues/4304

> Looks good for know. In the future (Neos 9) id like to have strict validation via ajax directly in the editor. (And we could make our NodePropertyConversion thing more strict by throwing exceptions ...)

but I agree that this sudden change in behaviour might not be cool as well.

I could also think of keeping this unfixed for 7.3…8.3 and only fix it properly in 9.0 - I mean the issue was open since 2018 not super critical I suppose;)

packages/neos-ui-editors/src/Editors/TextField/index.js
grebaldi: Passing `null` as a value to an `<input/>` element will lead React to treat it as "uncontrolled": https://react.dev/reference/react-dom/components/input#controlling-an-input-with-a-state-variable

This has all kinds of undesirable implications.

Like @bwaidelich  said, before the bugfix, `value` just wasn't `null`. So, with the bugfix I needed to introduce this conversion as well.

Classes/Domain/Service/NodePropertyConversionService.php
grebaldi: I think the more annoying issue (that ought to be fixed for LTS imho) is https://github.com/neos/neos-ui/issues/3157 (unsetting an optional integer value is impossible).

mhsdesign:
Tested it, and after fixing also the `TextAreaEditor` to handle `null` as value it works as expected:

before

https://github.com/neos/neos-ui/assets/85400359/fa2316c0-7133-4913-b097-ae30bf6ee1f4

after

https://github.com/neos/neos-ui/assets/85400359/89e1b964-ece7-40f0-bb2a-bef2af3d59c9

packages/neos-ui-editors/src/Editors/TextField/index.js
mhsdesign: btw one of the "implications" being that even if the value changed from "foo" to null, it would not update the textarea

bwaidelich (approved):
> I think the more annoying issue (that ought to be fixed for LTS imho) is https://github.com/neos/neos-ui/issues/3157 (unsetting an optional integer value is impossible).

Very good point, take my vote :)

PR #3536, opened 2023-06-23
BUGFIX: Disable asset search through asset editor when media browser is disabled

### LINKED ISSUE #1571: Setting mediabrowser to false in File Selection Editor still enables search for existing files
The upload form still allows to search through the existing assets, even though the configuration is set to mediabrowser:false. The actual Media Browser button is no longer there, but the functionality essentially remains, if the editor can still pick from the pool of assets by searching for them. 

Or... Did I misunderstand the feature?

grebaldi:
fixes: #1571 

**The problem**

As stated in #1571, even if the media browser has been deactivated via feature-toggle in the asset editor configuration, it is still accessible through search-as-you-type, because the search results are sourced from the media module.

I can very well be argued that by configuring `mediaBrowser: false`, the integrator's intent is to restrict the asset editor solely to its upload function. At the moment this intent could be bypassed.

**The solution**

I added code to dynamically disable the asset search in the asset editor depending on the value of the `mediaBrowser` feature toggle. If `mediaBrowser` is set to `false`, search will be disabled as well.

Despite the quick fix, I think we should discuss if we want to consider this the correct way to interpret the feature toggle. (Maybe also use :+1: / :-1:  on this description)

### DIFF
--- a/packages/neos-ui-editors/src/Editors/AssetEditor/index.js
+++ b/packages/neos-ui-editors/src/Editors/AssetEditor/index.js
@@ -238,7 +238,7 @@ export default class AssetEditor extends PureComponent {
             <SelectBox
                 optionValueField="identifier"
                 loadingLabel={this.props.i18nRegistry.translate('Neos.Neos:Main:loading')}
-                displaySearchBox={true}
+                displaySearchBox={this.isFeatureEnabled('mediaBrowser')}
                 ListPreviewElement={AssetOption}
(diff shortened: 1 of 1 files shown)


jonnitto (approved):

markusguenther (approved):

markusguenther:
Found another little issue while testing. The select box of the multi-select box for the search is not removed for a disabled feature set.
Screenshot 2023-12-18 at 20 25 47

But as this another issue, I will create one and provide an extra fix.

PR #3538, opened 2023-06-24
BUGFIX: Ensure that SecondaryInspector is always above UnappliedChanges-overlay

grebaldi:
refs: https://github.com/neos/neos-ui/issues/2670

**The Problem**

This one fixes a regression introduced in #3491 (see also: https://github.com/neos/neos-ui/pull/3491#issuecomment-1605394390).

There, the global z-Index of the "unapplied changes"-overlay has been increased, so that it covers the entire UI (minus the inspector). By accident, the "unapplied changes"-overlay now also covers the secondary inspector.

This, of course, leads to quite irritating problems. For example:

When I try to crop an Image, I will succeed at dragging the crop handle at first. Any additional click inside the secondary inspector however will trigger the "unapplied changes"-dialog.

**The solution**

I adjusted the "image editor" E2E test, so that it now captures the interference between the secondary inspector and the "unapplied changes"-overlay. Then I changed all relevant z-Indices, so that the interference stops.


crydotsnake (approved):
Fine by reading

mhsdesign (approved):

PR #3539, opened 2023-06-25
BUGFIX: Ensure that UI renders updated content after node type change

### LINKED ISSUE #1910: Node renders old state after changing node info/node type
### Description

When changing a node's node type ion the nodeinfo panel, the node is re-rendered which is of course absolutely correct. unfortunately it renders the state of the node as it was on the initial pageload, without the changes taking effect. However, the node info returned from the server represent the updated date.

### Steps to Reproduce

1. Change node type of a node

#### Expected behavior

The changes should take effect when the node is re-rendered.

#### Actual behavior

image

image

auswahl_001

### Affected Versions

Neos: 4.0.4

UI: 1.1.0

grebaldi:
fixes: #1910 

**The problem**

#1910 should have been fixed by #2892. However, I was able to find some edge-cases:

1. When a content element is rendered trough a `Neos.Neos:ContentCase`-case that is not the `default`-case, but still redirects to a fusion prototype that depends on the node type name, the wrong content element will be rendered when the node type changes.
2. Even when the correct content shows up, CKEditor placeholders may still reflect the old node type, because they are populated from stale node type configuration.

**The solution**

To fix 1) I made the regular expression in `RenderedNodeDomAddress` that captures the fusion path of the closest `Neos.Neos:ContentCase` a bit more lenient on the case-part. This covers more cases (sorry for the inflationary and ambiguous  use of the word "case" here - but I don't know what to do about it :D), but will only serve as a temporary solution. OutOfBand-Rendering requires more love in general.

To fix 2) I made sure that the `UpdateNodeInfo` feedback is delivered before the `ReloadContentOutOfBand` feedback, so that the feedback handler for the latter will operate on the updated node type.

### DIFF
--- a/Classes/Domain/Model/Changes/Property.php
+++ b/Classes/Domain/Model/Changes/Property.php
@@ -223,6 +223,11 @@ class Property extends AbstractChange
 
             $this->updateWorkspaceInfo();
 
+            // This might be needed to update node label and other things that we can calculate only on the server
+            $updateNodeInfo = new UpdateNodeInfo();
+            $updateNodeInfo->setNode($node);
+            $this->feedbackCollection->add($updateNodeInfo);
+
             $reloadIfChangedConfigurationPath = sprintf('properties.%s.ui.reloadIfChanged', $propertyName);
             if (!$this->getIsInline() && $node->getNodeType()->getConfiguration($reloadIfChangedConfigurationPath)) {
                 if ($this->getNodeDomAddress() && $this->getNodeDomAddress()->getFusionPath()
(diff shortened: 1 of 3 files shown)


mhsdesign (approved):
Fine by reading ;)

> OutOfBand-Rendering requires more love in general.

i totally agree - and i think we should definitely make the neos ui work with any rendering engine and dont tie it to fusion. So wed need a clean api here (and neos.neos could provide the glue for fusion) - BSC component view will need it as well

bwaidelich (approved):
+1 by reading, thanks!

Issue #3542, opened 2023-06-26
TASK: E2E Testing cleanup


mhsdesign:
Issues i found:

The homepage node is not clean but has already one node due to the import:

<img width="771" alt="image" src="https://github.com/neos/neos-ui/assets/85400359/f3f0e6ea-e766-4394-b211-0e108f551cc5">

This is a problem as any test running on the homepage has to deal with another content element already being there. Currently it works through mere luck.

Instead we should put this to be deleted node on another subpage like "Node to delete", wait we exactly have it like this but dont utilize it?

![image](https://github.com/neos/neos-ui/assets/85400359/7b584ead-cb33-48d9-9bd6-10c915155c7f)

Additionally the more tests should run on their own site to avoid any collision. Like the `createNewNodes` tests should run on their respective site, which is already there but unused?

PR #3543, opened 2023-06-26
TASK: More isolated e2e tests (on their own page)

mhsdesign:
resolves: #3542

I tried to start with my proposed changes. For now i want to run the `createNewNodes.e2e.js` on their own site. But for some reason i can't create new nodes via the inline ui if there is no existing node in the content collection. (At least thats the difference between when the test is run on the home page. And removing the existing node on the homepage yields the same error)

So i tried to debug this test `yarn run testcafe chrome Tests/IntegrationTests/Fixtures -T 'Can create content node from inside InlineUI' --debug-on-fail`

It testcafe just says `1) Uncaught object "[object Object]" was thrown. Throw Error instead.`. I could narrow it down via a try catch to the following lines. And after refining i found out that the line `click(Selector('.neos-contentcollection'))` was failing.

https://github.com/neos/neos-ui/blob/7ea5dd72e8c621c830e72ce1432663854fc92f67/Tests/IntegrationTests/Fixtures/1Dimension/createNewNodes.e2e.js#L159-L169

After putting it in a try catch, it logged 

```js
{
  callsite: CallsiteRecord {
    filename: 'Neos.Neos.Ui/Tests/IntegrationTests/Fixtures/1Dimension/createNewNodes.e2e.js',
    lineNum: 166,
    callsiteFrameIdx: 5,
    stackFrames: [
      [Object],    [Object],
      [Object],    [Object],
      [Object],    [Object],
      CallSite {}, [Object],
      [Object],    [Object]
    ],
    isV8Frames: true
  }
}
```

which is indeed not an error but a normal object ...

if i dont wrap it in a try catch, testcafe will just fail:

image

And without the try catch testcafe doesnt find the inline add node button ...

The tests `Check the nodetype help in create dialog` and `Check that nodetype without help has no help button` should currently fail due to the same behaviour ...

### DIFF
--- a/Tests/IntegrationTests/Fixtures/1Dimension/createNewNodes.e2e.js
+++ b/Tests/IntegrationTests/Fixtures/1Dimension/createNewNodes.e2e.js
@@ -1,6 +1,6 @@
 import {Selector, RequestLogger} from 'testcafe';
 import {ReactSelector} from 'testcafe-react-selectors';
-import {beforeEach, subSection, checkPropTypes} from '../../utils';
+import {createBeforeEach, subSection, checkPropTypes} from '../../utils';
 import {Page} from '../../pageModel';
 
 /* global fixture:true */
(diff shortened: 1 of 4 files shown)


markusguenther:
@mhsdesign Is that still a thing you want to work on?

markusguenther:
@mhsdesign Is that still a thing you want to work on?

PR #3553, opened 2023-06-29
FEATURE: Serious ckeditor inlineMode `isInline`

### LINKED ISSUE #3545: FEATURE: Proper inline mode (`autoparagraph "false"`)
### Description

CKeditor doesnt offer a "real" inline mode currently https://github.com/ckeditor/ckeditor5/issues/762 but we can improve our implementation by https://stackoverflow.com/questions/48896019/ckeditor5-inline-build-puts-a-p-tag-inside-of-an-h1-tag-how-do-i-disable-this/49130552#49130552

This would fix following issues:

- Edit: might be fixed with https://github.com/neos/neos-ui/pull/3532: When you paste text with newlines, they will be turned currently into a bunch of spans but there are no breaks (`br`)
  - input `foo\nbar` -> expected `foo<br>bar` actualy `<span>foo</span><span>bar</span>`
- it is possible to have multiple root elements like after pasting or when the ckeditor is booted with `<p>foo</p><p>bar</p>`
  - this was previously for a long time really buggy and you would get invalid html https://github.com/neos/neos-ui/pull/3532
- you use the heading formatting, which makes no sense on an inline editor, as it should produce contents without outer tags.
- our super hacky code to not produce outer tags https://github.com/neos/neos-ui/pull/3532 can be cleaned up
- we can fix https://github.com/neos/neos-ui/issues/3223 while we are at it (because we need to extend the editor and will do our own construction phase)
- autoparagraph "false" is currently not the best naming for what we actually need, so this will be for legacy behaviour but we can deprecate it and add a `isLimit` or `isInlineMode` like this...
- further we should define who determines the inline mode. Is it the yaml config or the `block: false` from the `Neos.Fusion:Editable`?

mhsdesign:
resolves #3545

**What I did**

**How I did it**

**How to verify it**

todo:
- pasting newlines should insert brs


grebaldi:
E2E tests:
image

Didn't know it was supposed to :smile:

Tests/IntegrationTests/SharedNodeTypesPackage/NodeTypes/Content/InlineHeadline.yaml
```diff
@@ -21,7 +21,7 @@
         inlineEditable: true
         inline:
           editorOptions:
-            autoparagraph: false
+            isInline: true
```
mhsdesign, line 24: - inlineMode?
- inlineContentMode?
- block = false

or decide in fusion?

```
Neos.Fusion:Editable {
   block = false
}
```

mhsdesign: We decided to use the the same naming as in fusion: `block` false.

And its allowed to either specify it in the nodeType or via the rendering.

Following logic would apply:

```js
    const isInlineMode = editorOptions?.block === false ||
         propertyDomNode.tagName === 'SPAN' ||
         propertyDomNode.tagName === 'H1' ||
         propertyDomNode.tagName === 'H2' ||
         propertyDomNode.tagName === 'H3' ||
         propertyDomNode.tagName === 'H4' ||
         propertyDomNode.tagName === 'H5' ||
         propertyDomNode.tagName === 'H6' ||
         propertyDomNode.tagName === 'P';
```


happy cases:

`editorOptions: { } | { block: true }` + `Neos.Fusion:Editable { } | { block: true }` -> `<div>{isInlineMode: false}</div>` 
`editorOptions: { }` + `Neos.Fusion:Editable { block: false }` -> `<span>{isInlineMode: true}</span>` 
`editorOptions: { block: false }` + `Neos.Fusion:Editable { block: false }` -> `<span>{isInlineMode: true}</span>` 

but this would also lead to cases like this:

`editorOptions: { block: false }` + `Neos.Fusion:Editable { } | { block: true }` -> `<div>{isInlineMode: true}</div>` 
`editorOptions: { block: true }` + `Neos.Fusion:Editable { block: false }` -> `<span>{isInlineMode: false}</span>` 


So the the notation of `block: false` will always overrule the one from the `Neos.Fusion:Editable`. But the rendering will still be affected via `Neos.Fusion:Editable` 
(i hope i have no logic error in my equations ^^)

grebaldi (changes requested):
Hi @mhsdesign,

thank you for tackling this problem :)

I tested your changes locally and I noticed a problem with the placeholders:

| before | after |
|-|-|
| Peek 2024-01-15 16-39 | Peek 2024-01-15 16-41 |

Editor initialization and editing itself are working fine. Only the placeholder is missing - which makes the editor virtually unusable.

I used this setup for reproduction:
<details>
<summary>
NodeType
</summary>

```yaml
'Vendor.Site:Content.Headline':
  superTypes:
    'Neos.Neos:Content': true

  ui:
    icon: header
    label: 'Headline'

  properties:
    type:
      type: string
      ui:
        label: 'Headline Type'
        reloadIfChanged: true
        inspector:
          editorOptions:
            values:
              h1:
                label: 'H1'
              h2:
                label: 'H2'
              h3:
                label: 'H3'
              h4:
                label: 'H4'
              h5:
                label: 'H5'
              h6:
                label: 'H6'
    text:
      type: string
      ui:
        inlineEditable: true
        inline:
          editorOptions:
            placeholder: '- Insert Text here -'
            isInline: true
            linking:
              anchor: true
              title: true
              relNofollow: true
              targetBlank: true
            formatting:
              strong: true
              em: true
              u: true
              sub: true
              sup: true
              del: true
              p: false
              h1: false
              h2: false
              h3: false
              h4: false
              h5: false
              h6: false
              pre: false
              underline: true
              strikethrough: true
              removeFormat: true
              left: true
              right: true
              center: true
              justify: true
              table: false
              ol: false
              ul: false
              a: true
```

</details>

<details>
<summary>Fusion</summary>

```neosfusion
prototype(Vendor.Site:Content.Headline) < prototype(Neos.Neos:ContentComponent) {
  type = ${q(node).property('type') || 'div'}

  text = Neos.Neos:Editable {
    property = 'text'
    block = false
  }

  renderer = Neos.Fusion:Tag {
      tagName = ${props.type}
      content = ${props.text}
    }
}
```

</details>

Another issue:
> We decided to use the the same naming as in fusion: block false.

:point_up: this is not yet reflected in code afaict. The `editorOptions` configuration still says `isInline`.

Sebobo (changes requested):
Let's target 9.1 with CK 44 instead as we have multiple changes to the CK integration there.

PR #3555, opened 2023-06-29
BUGFIX: 3223 ck plugin `editor.neos` not initialized

### LINKED ISSUE #3223: Bug: Ckeditor Plugin: `this.editor.neos.contextPath` is not available during plugin `init()`
### Description

this line should make it easy to find out inside a Ckeditor plugin the current `contextPath` and `propertyName` (among others things)

https://github.com/neos/neos-ui/blob/7665b057e7f72fbd56026c106d3bab8528403469/packages/neos-ui-ckeditor5-bindings/src/ckEditorApi.js#L75-L76

But `editor.neos` is initialized only after the editor has been created (see the async `then`)

This results in not beeing able to access `neos` at first but we need to wait for the next iteration in the eventloop

as shown in this plugin `Prgfx.Neos.DynamicPlaceholder` by @PRGfx 
https://github.com/PRGfx/Prgfx.Neos.DynamicPlaceholder/blob/c301d04a399ca1b85cdbca1418f3370a9e8540f3/Resources/Private/src/manifest.js#L40-L49

i can also confirm this behaviour, in my case i used `setTimeOut(() => console.log(this.editor.neos), 0)`

Fyi - this is purely an extension point and not used in the codebase.

### Proposal to fix this:

i dislike the "hack" that we attach something to `this.editor` -> as that could break any minute -> if ck decides to make editor immutable.

i also dislike god objects -> and the editor just has become one. (even the global registry is available via editor.neos)

a fine solution would be to use the editor identifier / instance and have some registry to resolve this identity back to `contextPath` and `propertyName`

mhsdesign:
resolves: #3223

**What I did**

**How I did it**

**How to verify it**

insert into a ckeditor plugin inside `init`
`console.log(this.editor.neos)`

and check it outputs the object

### DIFF
--- a/packages/neos-ui-ckeditor5-bindings/src/ckEditorApi.js
+++ b/packages/neos-ui-ckeditor5-bindings/src/ckEditorApi.js
@@ -43,6 +43,16 @@ export const createEditor = store => async options => {
         propertyDomNode
     });
 
+    class NeosEditor extends DecoupledEditor {
+        constructor(...args) {
+            super(...args);
+            // We attach all options for this editor to the editor DOM node, so it would be easier to access them from CKE plugins
+            // this has to be done after / in the constructor as `create` is async and plugins accessing .neos have to account for this
+            // https://github.com/neos/neos-ui/issues/3223
+            this.neos = options;
+        }
+    }
+
     return DecoupledEditor
         .create(propertyDomNode, ckEditorConfig)
         .then(editor => {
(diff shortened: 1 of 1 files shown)


packages/neos-ui-ckeditor5-bindings/src/ckEditorApi.js
```diff
@@ -43,6 +43,16 @@ export const createEditor = store => async options => {
...
+        constructor(...args) {
+            super(...args);
+            // We attach all options for this editor to the editor DOM node, so it would be easier to access them from CKE plugins
+            // this has to be done after / in the constructor as `create` is async and plugins accessing .neos have to account for this
+            // https://github.com/neos/neos-ui/issues/3223
+            this.neos = options;
+        }
+    }
+
     return DecoupledEditor
```
mhsdesign:
```suggestion
    return NeosEditor
```

grebaldi (approved):
:+1: by reading

Issue #3557, opened 2023-06-30
8.3.0 -> 8.3.1 breaks CKE placeholder functionality for elements without autoParagraph


patricekaufmann:
There is no placeholder being rendered when autoParagraph is set to `false`. The attribute `data-neos-placeholder` and the ::before Tag rendering the placeholder are both missing.

8.3.0:
![image](https://github.com/neos/neos-ui/assets/41791606/d7c2640a-e020-4c16-ae32-6e2819b4c71d)

8.3.1:
![image](https://github.com/neos/neos-ui/assets/41791606/fc914b16-5e50-4375-ae43-a099cdf77198)

mhsdesign:
Thank you so much for reporting it that fast. We tried to make sure that https://github.com/neos/neos-ui/pull/3532 doesn't break anything accidentally but it seems the placeholder needs some extra love.

mhsdesign:
fixed ;) and will be released soon with 7.3.19 ;)

PR #3558, opened 2023-06-30
BUGFIX: 3557 placeholder plugin with autoparagraph false

mhsdesign:
resolves: #3557
as the placeholder plugin was a bit hacky and broke in a regression from https://github.com/neos/neos-ui/pull/3532
Because `editor.getData` now returned the new internal  `<neos-inline-wrapper>` tags. The preprocessing to strip them is now applied earlier, so other calls to `editor.getData` now get the expected results.

also a problem is resolved that no placeholder is shown, if one chooses `h1` tag when `autoparagraph false` is enabled

additionally it cleans up the codebase and uses the native ckeditor placeholder plugin instead of our hacky home-brewed solution.
 

before and after:

https://github.com/neos/neos-ui/assets/85400359/c9429fc7-c250-4ab9-acaa-6815acbaa484

https://github.com/neos/neos-ui/assets/85400359/4de4ed14-e603-471b-92e3-9821fab8a3cf

https://github.com/neos/neos-ui/assets/85400359/15c4db9a-156b-4942-a567-76afc4293619

**What I did**

**How I did it**

**How to verify it**


mhsdesign:
So i was wondering why we even have our more or less hacky placeholder plugin.

When we initially integrated CKEditor 5 https://github.com/neos/neos-ui/pull/1942 its latest version was v10. And only with v12 the "ckeditor5-editor-decoupled" had native support for placeholders.
https://github.com/ckeditor/ckeditor5-editor-decoupled/blame/3a6ddfe6644a539f8cff892d19f2d3f7b578317e/src/decouplededitorui.js#L132-L149

see docs: https://ckeditor.com/docs/ckeditor5/16.0.0/api/module_core_editor_editorconfig-EditorConfig.html#member-placeholder and https://ckeditor.com/docs/ckeditor5/16.0.0/features/editor-placeholder.html


luckily we use v12 in 7.3 and v16 in 8.3 so we can utilize this more stable native placeholder support as of right now.

packages/neos-ui-ckeditor5-bindings/src/manifest.config.js
```diff
@@ -147,5 +144,21 @@ export default ckEditorRegistry => {
             ]}
     }));
 
+    //
+    // @see https://ckeditor.com/docs/ckeditor5/16.0.0/api/module_core_editor_editorconfig-EditorConfig.html#member-placeholder
+    //
+    config.set('placeholder', (config, {globalRegistry, editorOptions}) => {
```
Sebobo: Would be helpful to have a note here how we integrate the native placeholder. For me it was directly visible.

Sebobo (changes requested):
Looks good and works in my tests, just one nitpick regarding a comment.

kdambekalns (approved):
Looks good to me…

packages/neos-ui-ckeditor5-bindings/src/manifest.config.js
mhsdesign: Adjusted ;)

Sebobo (approved):

packages/neos-ui-ckeditor5-bindings/src/manifest.config.js
```diff
@@ -81,13 +81,18 @@ export default ckEditorRegistry => {
...
+    // - configuration of language
+    // - and placeholder feature see https://ckeditor.com/docs/ckeditor5/16.0.0/api/module_core_editor_editorconfig-EditorConfig.html#member-placeholder
     //
     config.set('baseConfiguration', (ckEditorConfiguration, {globalRegistry, editorOptions, userPreferences}) => {
         const i18nRegistry = globalRegistry.get('i18n');
-        return Object.assign(ckEditorConfiguration, {
-            language: String($get('interfaceLanguage', userPreferences)),
-            neosPlaceholder: unescape(i18nRegistry.translate($get('placeholder', editorOptions) || ''))
-        });
+        const placeholder = $get('placeholder', editorOptions);
```
grebaldi, line 89: This will affect https://github.com/neos/neos-ui/pull/3306 - leaving this here as a reminder :)

grebaldi (approved):
Works like a charm! :+1:

packages/neos-ui-ckeditor5-bindings/src/manifest.config.js
mhsdesign: Thx i know ^^

mhsdesign: fixed in upmerge ;)

skill level over 99999

PR #3559, opened 2023-07-04
TASK: cleanup disabled autoparagraph mode

mhsdesign:
While the fix at https://github.com/neos/neos-ui/pull/3532 was the most correct one in terms of behaviour its also the hardest one to maintain over time. And brought a regression see https://github.com/neos/neos-ui/pull/3558.
@grebaldi had the idea to simplify the optional cleanup.

In the disabled paragraph mode (which is now also named in the code like this, in preparation for https://github.com/neos/neos-ui/pull/3553) we still trim the outer span, but in case there was another span somewhere we return the original content (so we dont ever run into this: #2918)

**What I did**

**How I did it**

**How to verify it**


crydotsnake (approved):
Looks good to me..

packages/neos-ui-ckeditor5-bindings/src/cleanupContentBeforeCommit.spec.js
```diff
@@ -8,3 +8,28 @@ test('remove empty nbsp', () => {
...
+        assertCleanedUpContent('<span>foo</span>', 'foo');
+    })
+
+
+    test('cleanup multiple root <span>', () => {
+        assertCleanedUpContent('<span>foo</span><span>bar</span>', '<span>foo</span><span>bar</span>');
+    })
+
+    test('cleanup <span> root after other root', () => {
+        // in the case you had multiple paragraphs and a headline and switched to autoparagrahp: false
```
grebaldi:
```suggestion
        // in the case you had multiple paragraphs and a headline and switched to autoparagraph: false
```

grebaldi (approved):
LGTM :+1: (aside from one typo nitpick)

packages/neos-ui-ckeditor5-bindings/src/cleanupContentBeforeCommit.spec.js
mhsdesign: i was waiting for it to spell it wrong ^^

crydotsnake (approved):
Looked into this again. Tested it and it works as expected.

Issue #3546, opened 2023-06-29
EPIC: E2E Tests for Neos 9.0


grebaldi:
This issue covers everything related to E2E Tests on Neos 9.0. There have already been some issues open with regards to 9.0, but we need to fix some problems that exist in the 8.x-branches first.

## 8.x-related

- [x] https://github.com/neos/neos-ui/issues/3301
- [x] https://github.com/neos/neos-ui/issues/3540

## 9.0-related

- [x] https://github.com/neos/neos-ui/issues/3406
- [ ] https://github.com/neos/neos-ui/issues/3542
- [x] https://github.com/neos/neos-ui/issues/3541

Issue #3301, opened 2022-12-19
BUG: E2E fail to run locally via `make test-e2e-docker`


grebaldi:
Running `make test-e2e-docker` on a fresh checkout of Neos.Neos.Ui 8.3 leads to the following error message on my machine:

```
An exception occurred in driver: SQLSTATE[HY000] [2002] php_network_getaddresses: getaddrinfo failed: Temporary failure in name resolution

  Type: Doctrine\DBAL\Exception\ConnectionException
  File: Packages/Libraries/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractMySQLDriv
        er.php
  Line: 112
```

Docker version: 20.10.12, build 20.10.12-0ubuntu4
Docker-compose version: 1.29.2

That's all I can tell so far. I'll investigate further...

**UPDATE:**
The issue was introduced here: https://github.com/neos/neos-ui/pull/3253

The db container is now ARM-based. It won't run on amd64 machines.

I'll see, if there's a possible workaround.

markusguenther:
Hmm, is it possible to have kind of universal images? I just tested that with an ARM-based Mac and Sebastian used it but has a similar computer :/

Sorry for that.

grebaldi:
> Sorry for that.

No worries :)

This has made me think about how to make the E2E-setup more accessible in general (for local development). I think it would be nice, if adding E2E test cases and running the tests was a bit easier than it is right now. I find the prospect of the E2E tests turning red in CircleCI always a bit scary, because debugging with the E2E setup adds a lot of headache.

I was wondering, if a DDEV-supported setup is possible, because this is what I usually use for local development. It takes away most of the headache that comes with a custom docker-compose setup. And most importantly for this issue: It comes with support for ARM64: https://github.com/drud/ddev/issues/1966

There is also precedent for running DDEV on a CI Server: https://dev.to/tomasnorre/codeception-tests-in-travis-ci-with-ddev-and-selenium-1607

So this may be worth a try :)

markusguenther:
Some years ago, I tried to create a test distribution for the UI. Because I thought the "look into the circle ci yaml" to understand how it works was too complicated.

But today I am unsure if this improved it ;)

https://github.com/neos/neos-ui-testing-distribution

mhsdesign:
we should use the db image `mysql:8` instead of specifying only the arm version (`arm64v8/mysql:8`)... we had the same issue with #3369 @JamesAlias

PR #3560, opened 2023-07-05
BUGFIX: Ensure that docker E2E tests run locally on 8.0 & 8.1

### LINKED ISSUE #3562: E2E Flaky Tests ` Discarding: delete a content node and then discard deletion`
### Description

it is just super flaky, it succeeds locally 1 out of 5 times and similar or worse on CI

image

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

#### Expected behavior

[What you expected to happen]

#### Actual behavior

[What actually happened]

### Affected Versions

Neos:

UI:

grebaldi:
refs: #3301, #3546
fixes: #3562

This is the first in a series of PRs leading up to Neos 9.0. I want to make sure that our E2E tests can be executed locally on every maintained branch.

**The problem**

There were a couple of issues on 8.0:

1.) The `docker-compose` setup used `thecodingmachine/php:7.4-v4-cli` as the PHP image, but Neos 8.0 requires at least PHP 8.0.

2.) Locally, the E2E tests persistently failed due to console errors. 

As per https://github.com/neos/neos-ui/blob/4eb1e72b6a1b451eb8460606844bf10abeb7a2f4/Tests/IntegrationTests/utils.js#L24-L26 any occurring console error makes the tests fail.

The errors that were logged were in the form of 
```
uncaught at watchReloadState
  at takeLatest
  at reloadState
    TypeError: Failed to fetch
[...]
```

or
```
uncaught at watchCurrentDocument
  at takeLatest
  at loadContentForDocument
    TypeError: Failed to fetch
[...]
```

I could narrow down the moment at which those errors were logged to right after login via https://github.com/neos/neos-ui/blob/4eb1e72b6a1b451eb8460606844bf10abeb7a2f4/Tests/IntegrationTests/utils.js#L11-L16

The reason is that Testcafé will reload the document that the login was redirected to. This reload interrupts several fetch requests that the UI has already issued, leading to those saga-errors from above. As per https://github.com/DevExpress/testcafe/issues/5042 the reload is necessary for Testcafé's role concept to work properly and cannot be prevented.

3.) The `2Dimension` fixture failed to load due to missing files in `resource://Neos.TestSite/Private/...`

I had occasions at which the site import failed, because the `resource://Neos.TestSite/Private/Content/...` was missing the image resources referenced in `Sites.xml`.

Even with those resources added, the site failed to load due to a missing `Root.fusion`. Neos only complains in the `2Dimension` case, even though the `1Dimension` case has no `Root.fusion` as well. I was unable to figure out why that is.

**The solution**

For 1.) I updated the docker image to `thecodingmachine/php:8.0-v4-cli-node16`, which is used for the other 8.x branches as well.

For 2.), since it is impossible to circumvent the reload, I added some code to the `adminUser` role function that makes sure that the UI is fully loaded before Testcafé can do anything else. The performance impact of this is negligible, because the role function is only executed once per fixture-run.

For 3.) I added the missing files (including an empty `Root.fusion`).

With these changes, I managed to have consistently green E2E test runs on my machine.

To confirm this, check out this branch and run:
```bash
make test-e2e-docker browser=chromium # or browser=chrome
```

The tests should run successfully.

**EDIT:** Just tested it, and these changes fix the E2E tests for  8.1 as well


grebaldi:
Ironically enough, E2E tests fail on Circle CI now. This is however unrelated to this PR, as it happens on 8.0 regardless: https://app.circleci.com/pipelines/github/neos/neos-ui/3622/workflows/32ce411a-c9f9-4922-9eeb-4fd65fe03ba3/jobs/16321

I'll look into it.

**EDIT:** This looks like a regression through https://github.com/neos/neos-development-collection/pull/3618. The failing test relies on the `.neos-inline-editable` class which has been removed by that PR in Neos.Neos.

I also discovered that our E2E tests are using completely weird versions of `neos/neos` and other neos-development-collection members (`dev-master` is being pulled). But this is another issue entirely...

**EDIT 2:** https://github.com/neos/neos-ui/pull/3560/commits/35e282f68e2f1bdc6bc812f33c3c9cc948053604 fixes the issue.

crydotsnake (commented):
I run the tests like you said with:

```
make test-e2e-docker browser=chrome
```

I noticed that not every test has a green mark. So i assume this test was skipped or couldnt be applied?:

SCR-20230706-mnae

Also one test is failing. I dont know if this has something to do with your PR. Or if this test has been failing already before:

SCR-20230706-mmlv

mhsdesign:
this is imo just bad UX

> I noticed that not every test has a green mark. So i assume this test was skipped or couldnt be applied?:

the yellow things represent sub steps and its totally normal for them to be yellow (had to figure that out as well)


yes i can confirm

> Also one test is failing. I dont know if this has something to do with your PR. Or if this test has been failing already before:

the `SelectBox opens above in creation dialog if there's not enough space below.` tests fails for me as well.


recording of the failure (actual):

https://github.com/neos/neos-ui/assets/85400359/585c7d5d-b121-4e0a-8c28-98d38c8f8dff

expected:

https://app.saucelabs.com/tests/bc53b20428f1449096520a2c1cc01663

at minute 6:25 - for some reason all the node creation dialog options are missing (except the title)

mhsdesign:
Well i have absolutely no clue, why the ci has items in the node creation dialog:

ci:


| ci | locally |
| ----- | ----- |
|  image      |  image       | 


i ensured at first that its not a build problem by using `neos-ui-compiled` but even then the problem occurred (in every browser). 

<details>

<summary>Holzweg - i just noted the wrong nodeType is used</summary>

I debugged via react the props of the node creation dialog and could also only find the element `title` declared.
That pointed me to the nodeType declaration (where i should have looked in the first place) and astonishingly i found out that we dont declare any node creation items (only the inherited title form the document)

LOL?? https://github.com/neos/neos-ui/blob/c2a7e0867d2c9fb26b75d4f9bbb82d75364633ec/Tests/IntegrationTests/SharedNodeTypesPackage/NodeTypes/Document/SelectBoxTestPage/OpensAboveInInspector.yaml#L16 

so after adding the 
```yaml
ui:
  showInCreationDialog: true
```

the test passes locally 😂 

</details>

i noticed way to late that the labels are different which suggest that we deal with different nodeTypes 😂 and not only this, but the order in the node creation dialog is consistently different ??? 

| ci | locally |
| ----- | ----- |
|   image     |   image  | 


The order of the nodeTypes seems to be directly influenced by how they were registered in the NodeTypesRegistry. (At least we use [`NodeTypesRegistry.getGroupedNodeTypeList`](https://github.com/neos/neos-ui/blob/c2a7e0867d2c9fb26b75d4f9bbb82d75364633ec/packages/neos-ui-contentrepository/src/registry/NodeTypesRegistry.ts#L127) in the `<SelectNodeType />` modal.



locally the endpoint with the nodeTypes shemas delivers the nodes in the order as expected (from the ci)

image

(also the parsed json is correct:

```js
neos.endpoints.getJsonResource("http:\/\/127.0.0.1:8081\/neos\/schema\/node-type?version=198476b89c5a9d878b2975a96d891b79_1688671802").then(console.log)
```

image
)


when dumping them from the NodeTypesRegistry (access the globalRegistry in the dev console by logging the `<Root />`  props via the react dev tools into the console and save them as global variable `nodeTypesRegistry = temp1.globalRegistry.get('@neos-project/neos-ui-contentrepository')`)

the raw registries content `nodeTypesRegistry._registry` shows also the wrong order:

image
 

`nodeTypesRegistry._registry` is simply filled with `nodeTypesRegistry.set` in a loop in the bootstrap: https://github.com/neos/neos-ui/blob/c2a7e0867d2c9fb26b75d4f9bbb82d75364633ec/packages/neos-ui/src/index.js#L105
which calls  `nodeTypesRegistry._registry.push` on the registries storage. (Which should imo preserve ordering no?)


(note that as far as i can remember, all tests run locally fine on 8.3 - i have never run 7.3...8.2 tests before locally)

### edit as found out here: https://github.com/neos/neos-ui/pull/3560#issuecomment-1626920642

the json from the server is actually in the "wrong" order

image

The node type configuration order seems to be different on every system. And matches with the order of the schema endpoint and the node creation dialog.

image

grebaldi:
Hi @mhsdesign,

I cannot reproduce this. For debugging purposes, would you mind logging `nodeTypeName` inside `forEach` at this point?:

https://github.com/neos/neos-ui/blob/c2a7e0867d2c9fb26b75d4f9bbb82d75364633ec/packages/neos-ui/src/index.js#L104C1-L105

(Trying to figure out when exactly the order of node type names gets scrambled - `Object.keys` may not be entirely deterministic here)

JamesAlias:
@mhsdesign the test fails because `testcafe` has an issue with resizing the window (wich is necessary to run that test).

if we upgrade `testcafe` to latest 2.X the issue is fixed (currently `2.6.2`)

mhsdesign:
On 8.3 i get:

image


and the registry shows:
image

(again the json is delivered in the correct order)

the console logs are actually also in the wrong order:


image


->

Actually i just found out that logging an object to the console or using the parsed json inspector in the network tab doesnt preserve the order 😂 , so it seems al the time the json delivered from the server is in the wrong order:

image


The node type configuration order seems to be different on every system. And matches with the order of the schema endpoint and the node creation dialog.

image

This order is determined by the nodetypes loader: https://github.com/neos/neos-development-collection/blob/699b50d23f85c612e392b697ff68a00f257e82df/Neos.ContentRepository/Classes/Configuration/NodeTypesLoader.php#L46C19-L46C19

and the files are just "found" in this order:

image

but this doesnt match with the order they appear when listing the directory:

image

mhsdesign:
It seems that the `RecursiveDirectoryIterator` used for loading the NodeType definition (from the new `NodeTypes` folder) is just unordered on linux. (Every project will have its own order, the order is consistent but only for the one project)

> On Windows, you will get the files ordered by name. On Linux they are not ordered.

https://www.php.net/manual/en/class.recursivedirectoryiterator.php#120971

mhsdesign:
> @mhsdesign the test fails because `testcafe` has an issue with resizing the window (wich is necessary to run that test).
> 
> if we upgrade `testcafe` to latest 2.X the issue is fixed (currently `2.6.2`)

we are using ^1.7 (1.8) on neos 8.0 so i didnt test the update for now. 2.x is imo used only in never branches like 8.3

grebaldi:
@mhsdesign:
> It seems that the `RecursiveDirectoryIterator` used for loading the NodeType definition (from the new `NodeTypes` folder) is just unordered on linux. (Every project will have its own order, the order is consistent but only for the one project)

:scream: 

So, the order of node types is non-deterministic in general? Well then I suppose it's best to adjust the test to use a different selector strategy in node type selector dialog... Wdyt?

mhsdesign:
> So, the order of node types is non-deterministic in general?

only when the feature https://github.com/neos/neos-development-collection/pull/3332 is used ;)

and hopefully this will be fixed via https://github.com/neos/neos-development-collection/pull/4394

> Well then I suppose it's best to adjust the test to use a different selector strategy in node type selector dialog... Wdyt?

yeah sure, that anyways. Its bad to rely on the order - i think its a specificity problem. `Opens Above` appears in both nodeType names, and thus both might be clicked but testcafe decides for the first occurrence.

mhsdesign:
> yeah sure, that anyways. Its bad to rely on the order - i think its a specificity problem. Opens Above appears in both nodeType names, and thus both might be clicked but testcafe decides for the first occurrence.


yes that was it, fixed in https://github.com/neos/neos-ui/pull/3560/commits/9fddf45bab27b07555c6642c5eaecae1027484eb ;)

Now all tests pass locally ❤️ thank you again for the efforts!!!

mhsdesign (approved):

mhsdesign:
Do i have to discard the docker container after i already run the tests for 8.3? Because i currently have an odd set of dependencies:

```
 Neos.Cache                          8.3.x-dev      
 Neos.ContentRepository              dev-master     
 Neos.Diff                           dev-master     
 Neos.Eel                            8.3.x-dev      
 Neos.Flow                           8.3.x-dev      
 Neos.Flow.Log                       8.3.x-dev      
 Neos.FluidAdaptor                   8.3.x-dev      
 Neos.Fusion                         dev-master     
 Neos.Fusion.Afx                     dev-master     
 Neos.Fusion.Form                    2.1.1          
 Neos.Http.Factories                 8.3.x-dev      
 Neos.Imagine                        3.3.2          
 Neos.Media                          dev-master     
 Neos.Media.Browser                  dev-master     
 Neos.Neos                           dev-master     
 Neos.Neos.Ui                        8.0.x-dev      
 Neos.Neos.Ui.Compiled               8.0.x-dev 
```

the dependencies are just really odd implicitly through the ui required:

https://github.com/neos/neos-ui/blob/81664ee77cd143e54d5a7da052e8b09ad786f67f/Tests/IntegrationTests/TestDistribution/composer.json#L12-L17

grebaldi:
@mhsdesign Yeah, dependencies are weird. That needs to be fixed on 7.3 already though, so this will be another set of PRs :/

mhsdesign:
Lets merge this so the CI is green before i do a release?

gjwnc:
> only when the feature [neos/neos-development-collection#3332](https://github.com/neos/neos-development-collection/pull/3332) is used ;)

None of the configs is or was sorted. Not the nodeTypes or Settings or Policy. Before [neos/neos-development-collection#3332](https://github.com/neos/neos-development-collection/pull/3332), the config was mainly loaded using Flow's [`YamlSource::load`](https://github.com/neos/flow-development-collection/blob/5.0/Neos.Flow/Classes/Configuration/Source/YamlSource.php#L87). This one is using `glob()` which, regarding sorting, has the same behavior.  
Regarding Settings or other yaml files you just often don't notice the difference between your various installations. This is because your package order introduces a "good enough" hierarchy which, after merging the yaml files, results in, apart from sorting, the same yaml config - e.g. you likely find the same values using the same path injecting a setting.  
A very simple example to show the unsortedness can be `./flow configuration:show` itself. If you have many packages, it is likely that there is a difference in the results from your local dev and production. I don't mean a difference of existing/not existing packages because of require-dev, I mean that the order of settings is just different also for packages existing on both installations.  

A quick check on a project of ours showed, that on production I get
```
PunktDe:
    EditConflictPrevention:
...
Sandstorm:
    TemplateMailer:

```
and on my local dev it is reversed - Sandstorm before PunktDe.

mhsdesign:
Thank you @gjwnc for your helpful insights, I reposted them here https://github.com/neos/neos-development-collection/pull/4394#issuecomment-1629254790

PR #3561, opened 2023-07-05
BUGFIX: Ensure that docker E2E tests run locally on 8.2 & 8.3

### LINKED ISSUE #3301: BUG: E2E fail to run locally via `make test-e2e-docker`

### LINKED ISSUE #3540: TASK: Dev experience fail if e2e tests are executed without `build-e2e-testing`
as described [here](https://github.com/neos/neos-ui/pull/3200/commits/a273b9e0e20347f8cf1295b96a93530b1a249858), we need to use keepNames so that react magic selectors can work.

This is done only if the build is made for e2e testing, as this safes us 100Mb in the bundle:
https://github.com/neos/neos-ui/blob/9585ca792616d52963e69f71c12c1915dfd5cea4/esbuild.js#L29

So without this options the e2e tests will fail locally.

So we should definitely document this more prominent and also check inside our end to end tests if this option was set to guard against unnecessary debugging and headaches ^^

Neos 8.2 and up affected

TODO
- [ ] Also it would be nice to fail when the user interface language is not correct :)

grebaldi:
includes: https://github.com/neos/neos-ui/pull/3560
refs:  #3546
fixes: #3301, #3540

This is the second in a series of PRs leading up to Neos 9.0. I want to make sure that our E2E tests can be executed locally on every maintained branch.

**The problem**

8.2 shares all problems described in https://github.com/neos/neos-ui/pull/3560.

Beyond those, there were the following issues:

1.) As described in #3301, an platform-specific docker image for the database container was introduced through https://github.com/neos/neos-ui/pull/3253. (See also: https://docs.docker.com/build/building/multi-platform/)

2.) As described in #3540, contributors must execute `make build-e2e-testing` before a local E2E test run. This is not really a bug, but can be quite frustrating due to its counter-intuitiveness.

**The solution**

This PR includes all commits from https://github.com/neos/neos-ui/pull/3560, (https://github.com/neos/neos-ui/commit/35e282f68e2f1bdc6bc812f33c3c9cc948053604 has been cherry-picked to https://github.com/neos/neos-ui/pull/3561/commits/21c90972d2e13857a35f7812fe1cd49f4eb4d363)

For 1.): In #3301, @mhsdesign commented:
> we should use the db image `mysql:8` instead of specifying only the arm version (`arm64v8/mysql:8`)... we had the same issue with #3369 @JamesAlias 

I did just that. As far as I understand the docker documentation on this (https://docs.docker.com/build/building/multi-platform/), docker will pull the appropriate platform-specific image if available.

The docker hub page for mysql (https://hub.docker.com/_/mysql/) states
> * Supported architectures: ([more info](https://github.com/docker-library/official-images#architectures-other-than-amd64))
> [amd64](https://hub.docker.com/r/amd64/mysql/), [arm64v8](https://hub.docker.com/r/arm64v8/mysql/)

Which leads me to believe that this solution should work for `arm64v8` machines as well. I cannot verify that independently though.

For 2.), I just added `build-e2e-testing` as a direct dependency to the `test-e2e-docker` make target. This way, it is automatically performed before a local E2E test run. The esbuild performance is so good, that the build actually doesn't add noticable delay when compared to the entire rest of the test ramp-up.

With these changes, I managed to have consistently green E2E test runs on my machine.

To confirm this, check out this branch and run:
```bash
make test-e2e-docker browser=chromium # or browser=chrome
```

The tests should run successfully.

**EDIT:** Just tested it and these changes fix the E2E tests for 8.3 as well. On to 9.0! :D


crydotsnake (commented):
## Edit

It works now. Maybe i had a typo or something 😅 

------------

I cant run the tests:

```
This command can use a lot of memory when importing sites with many resources.
If the import is successful, you will see a message saying "Import of site ... finished".
If you do not see this message, the import failed, most likely due to insufficient memory.
Increase the memory_limit configuration parameter of your php CLI to attempt to fix this.
Starting import...
---
Import of site "Neos Test Site" finished.
Publishing resources of collection "static"
Publishing resources of collection "persistent"
Browserslist: caniuse-lite is outdated. Please run:
  npx browserslist@latest --update-db
  Why you should do it regularly: https://github.com/browserslist/browserslist#browsers-data-updating
ERROR Cannot find the browser. "chromium" is neither a known browser alias, nor a path to an executable file.

Type "testcafe -h" for help.
make: *** [test-e2e-docker] Error 1
```

With chrome i get the same error.

crydotsnake:
SCR-20230706-slmw

Here the test fails also. Can you reproduce it? is that normal? @mhsdesign

mhsdesign:
> Here the test fails also. Can you reproduce it? is that normal? @mhsdesign

yes interestingly i can reproduce that.
It seems the assertion strategy just doesnt work (and it is also especially odd - i refactored it to something more clean).
The badge itself shows up correctly and shows the correct count.

fixed and locally working with https://github.com/neos/neos-ui/pull/3561/commits/6d1da0aabf9aaf41ab1fc75e0a65461d9cb7495a

i also cherry-picked https://github.com/neos/neos-ui/pull/3560#issuecomment-1627747477
(dont ask why i dindt merge - its too warm ^^ xD)

mhsdesign (approved):
Works now on mac too ;)

For a second two select box test failed, but i couldnt reproduce it and they were never flaky again (4 successive full runs)

mhsdesign:
Lets merge this so the CI is green before i do a release?

PR #3504, opened 2023-05-25
BUGFIX: Add dirty-state highlighting to RangeEditor

grebaldi:
fixes: #3499, #3500

**The problem**

#3499 describes that the `RangeEditor` is lacking dirty-state highlighting. The reason is that the highlighting-style is usually forwarded through the `<EditorEnvelope/>`-component, which in the case of the `RangeEditor` isn't happening.

#3500 describes that hitting "Discard" at the bottom of the Inspector does not reset the inputs of the `RangeEditor`. There are two inputs, one with `type="range"` and another one with `type="text"` that is rendered at the bottom and reflects the same (numerical) value. 

Both inputs receive the value without fallback. When the edited property is empty, they receive `null` which confuses React and leads to the following console warning:

image
> Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.

When the value is edited any non-null value is reflected through the inputs. On "Discard" however, the value is reset to `null`, which React then ignores.

**The solution**

For #3499 I added a simple highlight-style:
image

For #3500 I added fallbacks to empty string for the value assignments on the inputs (`value={value || ''}`), which leads React to correctly treat them as controlled inputs.

PR #3564, opened 2023-07-06
BUGFIX: Use better fallback mechanism for input value of RangeEditor

### LINKED ISSUE #3563: Value 0 is not possible anymore on Range Slider
Since the PR #3504 the value `0` isn't possible anymore.

If the `min` is `0` and I try to set it to this, the range slider jumps back to an empty string

https://github.com/neos/neos-ui/assets/4510166/3e34e184-9fae-4744-b4a9-01bd406714e3

grebaldi:
fixes: #3563 

**The problem**

The range editor used a fallback to ensure that its input receives a string value. This was implemented as an OR-based fallback, e.g.:
```js
value={value || ''}
```

Unfortunately, this has lead to `0` being treated as an empty string, which was not intended.

This is a regression through https://github.com/neos/neos-ui/pull/3504.

**The solution**

I have replaced the fallback mechanism to handle the value `0` explicitly:
```js
value={value === 0 ? '0' : (value || '')}
```

### DIFF
--- a/packages/neos-ui-editors/src/Editors/Range/index.js
+++ b/packages/neos-ui-editors/src/Editors/Range/index.js
@@ -63,6 +63,7 @@ class RangeEditor extends PureComponent {
     render() {
         const options = {...this.constructor.defaultProps.options, ...this.props.options};
         const {value, highlight} = this.props;
+        const valueAsString = value === 0 ? '0' : (value || '');
 
         return (
             <div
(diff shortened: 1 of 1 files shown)


mhsdesign (approved):
Makes sense 👍

jonnitto (approved):
Jep, now it works. Sorry that I missed that in the preview review. Thanks for the quick fix

PR #3569, opened 2023-07-07
TASK: Make E2E tests work for Neos 9.0

grebaldi:
refs: #3546, #3406
~includes: #3571~
~includes: https://github.com/neos/neos-ui/pull/3618~

This is the third and final in a series of PRs leading up to Neos 9.0. I want to make sure that our E2E tests can be executed locally on every maintained branch.

**How to test this**

You should of course be able to run the E2E tests locally via
```sh
make test-e2e-docker browser=chromium # or browser=chrome
```

The E2E tests use `cweagans/composer-patches` to include patches from the development collection. Documentation can be found here: https://github.com/cweagans/composer-patches/blob/1.x/README.md
We discussed to keep this dependency to make developing across repos easier ;)

**Remaining TODOs**

- [x] Make anything run at all
- [x] Fixture: 1Dimension
  - [x] Create event export from site export
  - [x] Fix all occurring E2E test failures
    - [x] 1Dimension/createNewNodeInContainer.e2e.js Line 33: "AssertionError: expected false to be truthy"
      - Root cause is here: https://github.com/neos/neos-development-collection/blob/9.0/Neos.Neos/Resources/Private/Fusion/Prototypes/ContentCollection.fusion#L30
      - I suppose that both `Neos.Node.inBackend` and `node.context.currentRenderingMode.edit` can no longer be relied upon that easily
      - Known as: https://github.com/neos/rector/issues/23
      - [x] Most likely fixed by: https://github.com/neos/neos-development-collection/pull/4067
    - [x] 1Dimension/discarding.e2e.js Line 41: "AssertionError: After discarding we are back to the main page: expected 'user-admin__eyJsYW5ndWFnZSI6ImVuX1VTIn0=__d7315383-adb8-45a5-8124-ffabeee
bc51a' to deeply equal '/sites/neos-test-site@user-admin;language=en_US'"
       - [x] Find proof that `/sites/neos-test-site@user-admin;language=en_US` and `user-admin__eyJsYW5ndWFnZSI6ImVuX1VTIn0=__d7315383-adb8-45a5-8124-ffabeeebc51a` represent the same node
         - They don't. The expected node address should rather be `user-admin__eyJsYW5ndWFnZSI6ImVuX1VTIn0=__f676459d-ca77-44bc-aeea-44114814c279`. `d7315383-adb8-45a5-8124-ffabeeebc51a` presumably refers to the node that has just been discarded in this test. Redirect to the home page ought to happen, but fails.
[...]

PR #3571, opened 2023-07-09
TASK: Replace usage of ```Neos.Node.inBackend(node)```

mficzel:
Resolves partly https://github.com/neos/neos-development-collection/issues/4396

This pr replaces usage of `Neos.Node.inBackend(node)` which is removed here https://github.com/neos/neos-development-collection/pull/4067 with `Neos.Backend.isEditMode(request) || Neos.Backend.isPreviewMode(request)`

To verify you can run the ui with the branch https://github.com/neos/neos-development-collection/pull/4067 be aware that the changes to adjust the ui to the separation of the edit and preview endpoints are not part of this pr.

### DIFF
--- a/Resources/Private/Fusion/Prototypes/Page.fusion
+++ b/Resources/Private/Fusion/Prototypes/Page.fusion
@@ -19,7 +19,7 @@ prototype(Neos.Neos:Page) {
 
             @process.json = ${Json.stringify(value)}
             @process.wrapInJsObject = ${'<script>window[\'@Neos.Neos.Ui:DocumentInformation\']=' + value + '</script>'}
-            @if.inBackend = ${Neos.Node.inBackend(documentNode)}
+            @if.inBackend = ${Neos.Backend.isEditMode(request) || Neos.Backend.isPreviewMode(request)}
 
             // We need to ensure the JS backend information is always up to date, especially
             // when child nodes change. Otherwise errors like the following might happen:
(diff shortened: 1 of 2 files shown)


mhsdesign:
even if this is more explicit, i wonder if its enough to either A: mostly use `Neos.Backend.isEditMode(request)` or reduce code written by introducing `Neos.Backend.isBackend(request)`

mficzel:
@mhsdesign i think asking for "are you edit or preview mode" is much more explicit. I also thought about a separate inBackend method that returns (self::isEditMode(request) || self::isPreviewMode()) but abstained for now as i do not like the semantic of inBackend as beeing either edit or preview. 

BTW: The previous semantic of Neos.Node.inBackend was even worse as it only checked that the node was not live.

grebaldi (approved):
I was able to test this successfully with https://github.com/neos/neos-ui/pull/3571.

I included this change over there.

I'm not sure how to go about this. This PR cannot be merged without https://github.com/neos/neos-development-collection/pull/4067 over at `neos-development-collection` having been merged first. https://github.com/neos/neos-development-collection/pull/4067 however cannot be merged without further changes in the UI (see: https://github.com/neos/neos-ui/pull/3569/commits/ce395ca9e3614ec32f7af7a82d2759a74e68f62d).

I would therefore suggest to close this PR in favor of https://github.com/neos/neos-ui/pull/3571.

mficzel:
Close this here as this needs to be redone.

Issue #3574, opened 2023-07-10
BUG: Neos UI crashes after applying references with `Neos.NodeTypes.ContentReferences:ContentReferences`


dlubitz:
### Description

If I add references to the `Neos.NodeTypes.ContentReferences:ContentReferences` and apply them, some errors are thrown to the console. After that, I don't get any feedback on the UI to further action. E.g if I click the "publish" button afterward I don't get any feedback, but the publish is happen in the background.

I've this issue with Neos 7.3, but the "Key must be a string" errors are also shown in Neos UI 8.3, but not the last one at `watchServerFeedback`.

### Steps to Reproduce

1. Create node with type `Neos.NodeTypes.ContentReferences:ContentReferences`
2. Select a random reference in ReferenceEditor
3. Click apply

#### Actual behavior

![image](https://github.com/neos/neos-ui/assets/13046100/6bb7dff0-4745-4092-81e3-927e7cc20801)

### Affected Versions

Neos: 7.3.13
UI: 7.3.17

dlubitz:
I'll add the server response as this might help to debug.
```json
{
    "timestamp": "2023-07-10T14:49:54+0000",
    "feedbacks": [
        {
            "type": "Neos.Neos.Ui:UpdateWorkspaceInfo",
            "description": "New workspace info available.",
            "payload": {
                "name": "user-lubitz",
                "publishableNodes": [
                    {
                        "contextPath": "\/sites\/vendor\/node-55a7c2fd8b3d8\/node-5756db9b91547\/node-95am5850s6je3\/node-8f5uhbnmzv9b4\/main\/node-574ef2e2dd5f0\/column0\/node-99wviknw8bfmq@user-lubitz",
                        "documentContextPath": "\/sites\/vendor\/node-55a7c2fd8b3d8\/node-5756db9b91547\/node-95am5850s6je3\/node-8f5uhbnmzv9b4@user-lubitz"
                    }
                ],
                "baseWorkspace": "live",
                "readOnly": false
            }
        },
        {
            "type": "Neos.Neos.Ui:ReloadContentOutOfBand",
            "description": "Rendering of node \"\/sites\/vendor\/node-55a7c2fd8b3d8\/node-5756db9b91547\/node-95am5850s6je3\/node-8f5uhbnmzv9b4\/main\/node-574ef2e2dd5f0\/column0\/node-99wviknw8bfmq\" required.",
            "payload": {
[...]

PR #3575, opened 2023-07-10
BUGFIX: Prevent error when initialising a node that is not in the store yet

Sebobo:
This can happen when a node is added via content references, as the backend doesn’t have any nodedata yet for the nodes that are rendered in the guest frame.

Resolves: #3574

### DIFF
--- a/packages/neos-ui-guest-frame/src/initializeContentDomNode.js
+++ b/packages/neos-ui-guest-frame/src/initializeContentDomNode.js
@@ -12,6 +12,13 @@ import style from './style.css';
 
 export default ({store, globalRegistry, nodeTypesRegistry, inlineEditorRegistry, nodes}) => contentDomNode => {
     const contextPath = contentDomNode.getAttribute('data-__neos-node-contextpath');
+
+    if (!nodes[contextPath]) {
+        // Node is not available in the store yet, so we can't initialize any interaction
+        console.warn(`Node with context path "${contextPath}" is not available in the store yet.`);
+        return;
+    }
+
     const isHidden = $get([contextPath, 'properties', '_hidden'], nodes);
     const hasChildren = Boolean($count([contextPath, 'children'], nodes));
     const isInlineEditable = nodeTypesRegistry.isInlineEditable($get([contextPath, 'nodeType'], nodes));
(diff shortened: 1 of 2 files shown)


Sebobo:
An alternative would be to return the necessary node data before the html is rendered. Not sure if that is viable though.

dlubitz (approved):
I tested the fix locally and it solves at the issue.

mhsdesign (changes requested):
I can confirm that this works, but i think it should be easily doable to fetch the required node informations.

In the case https://github.com/neos/neos-ui/issues/3574 we dont need any inline interaction, but one can never know if some one expects it or if we would currently silently ignore other edge cases (what will be hard to debug)

calling `neos.endpoints.getAdditionalNodeMetadata` after the out of band rendering should do ...

mhsdesign:
Hmm this doesnt do it ...

to be pasted after:

https://github.com/neos/neos-ui/blob/e89cc22ddb67d900e557430d300392e11f61152a/packages/neos-ui/src/manifest.js#L544

```js
        const {getAdditionalNodeMetadata} = backend.get().endpoints;

        const uninitializedNodes = [];

        for (const el of children) {
            const contextPath = el.getAttribute('data-__neos-node-contextpath');
            if (!selectors.CR.Nodes.byContextPathSelector(contextPath)(store.getState())) {
                uninitializedNodes.push(contextPath);
            }
        }

        if (uninitializedNodes.length) {
            const additionalNodeMetadata = await getAdditionalNodeMetadata(uninitializedNodes);

            if (additionalNodeMetadata) {
                store.dispatch(actions.CR.Nodes.merge(additionalNodeMetadata));
            }
        }
```

maybe the information of `getAdditionalNodeMetadata` is not that helpfull, it only seems to return partial nodes ...


solved it - alternative fix here: https://github.com/neos/neos-ui/pull/3576

Sebobo:
@mhsdesign I now added warnings, so when debugging something we would at least have some information.

dlubitz:
This is the PR for fusion side: https://github.com/neos/neos-development-collection/pull/4399

mhsdesign (approved):

PR #3576, opened 2023-07-10
BUGFIX: 3574 initialize foreign referenced nodes

### LINKED ISSUE #3574: BUG: Neos UI crashes after applying references with `Neos.NodeTypes.ContentReferences:ContentReferences`

mhsdesign:
resolves: #3574

actual fix instead of hotfix: #3575 and https://github.com/neos/neos-development-collection/pull/4399

similar to @grebaldi solution at: #3364

**What I did**

**How I did it**

**How to verify it**

revert: https://github.com/neos/neos-development-collection/pull/4399 and try to follow #3574

### DIFF
--- a/packages/neos-ui/src/manifest.js
+++ b/packages/neos-ui/src/manifest.js
@@ -544,7 +544,7 @@ manifest('main', {}, globalRegistry => {
 
         const children = findAllChildNodes(contentElement);
 
-        // in case there are foreign refernced nodes, we need to put them into the store:
+        // in case there are foreign referenced nodes, we need to put them into the store:
         const uninitializedReferencedNodes = [];
         for (const el of children) {
             const contextPath = el.getAttribute('data-__neos-node-contextpath');
--- a/packages/neos-ui/src/manifest.js
+++ b/packages/neos-ui/src/manifest.js
@@ -20,6 +20,7 @@ import {
 import initializeContentDomNode from '@neos-project/neos-ui-guest-frame/src/initializeContentDomNode';
 
 import style from '@neos-project/neos-ui-guest-frame/src/style.css';
+import backend from '@neos-project/neos-ui-backend-connector';
 
 manifest('main', {}, globalRegistry => {
     //
@@ -513,7 +514,7 @@ manifest('main', {}, globalRegistry => {
     // When the server advices to replace a node (e.g. on property change), put the delivered html to the
     // correct place inside the DOM
     //
-    serverFeedbackHandlers.set('Neos.Neos.Ui:ReloadContentOutOfBand/Main', (feedbackPayload, {store, globalRegistry}) => {
+    serverFeedbackHandlers.set('Neos.Neos.Ui:ReloadContentOutOfBand/Main', async (feedbackPayload, {store, globalRegistry}) => {
         const {contextPath, renderedContent, nodeDomAddress} = feedbackPayload;
         const domNode = nodeDomAddress && findNodeInGuestFrame(
             nodeDomAddress.contextPath,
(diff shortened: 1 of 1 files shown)


packages/neos-ui/src/manifest.js
```diff
@@ -513,7 +514,7 @@ manifest('main', {}, globalRegistry => {
     // When the server advices to replace a node (e.g. on property change), put the delivered html to the
     // correct place inside the DOM
     //
-    serverFeedbackHandlers.set('Neos.Neos.Ui:ReloadContentOutOfBand/Main', (feedbackPayload, {store, globalRegistry}) => {
+    serverFeedbackHandlers.set('Neos.Neos.Ui:ReloadContentOutOfBand/Main', async (feedbackPayload, {store, globalRegistry}) => {
```
mhsdesign, line 517: idk, if its wise to make this async ... sure the whole feedback is async, but now from the feedback collection this will be handled last despite of the order - i think

Sebobo:
Should we go with both PRs just to be one the safe side?

mhsdesign:
No I rather not - I like to be more explicit and if we just ignore any unexpected input wrong behaviour will later be worse to debug then if there was an exception.

we can still write a safeguard and log an error to the console if you want…

Sebobo:
Yes then I will adjust my PR to do that. The code there just relies too much on data being there.

packages/neos-ui/src/manifest.js
mhsdesign: no the execution starts immediately, so there should be no change of behavior for the "happy" case.

```js
console.log('main 1');

const asyncAwait = async () => {
    console.log('asyncAwait 2');

    await new Promise(r => r());

    console.log('asyncAwait 3');
}

asyncAwait();

const promiseCallback = () => {
    console.log('promiseCallback 4');

    new Promise(r => r()).then(
        () => console.log('promiseCallback 5')
    );
}

promiseCallback();

console.log('main 6');
```

logs:
```
[Log] main 1
[Log] asyncAwait 2
[Log] promiseCallback 4
[Log] main 6
[Log] asyncAwait 3
[Log] promiseCallback 5
```

dlubitz:
This is the PR for fusion side: https://github.com/neos/neos-development-collection/pull/4399

dlubitz (approved):
I like! ❤️  I can't evaluate the technical solution, but it looks reasonable.

packages/neos-ui/src/manifest.js
```diff
@@ -543,6 +544,27 @@ manifest('main', {}, globalRegistry => {
 
         const children = findAllChildNodes(contentElement);
 
+        // in case there are foreign refernced nodes, we need to put them into the store:
```
grebaldi:
```suggestion
        // in case there are foreign referenced nodes, we need to put them into the store:
```

grebaldi (commented):
LGTM code-wise.

However, I was unable to reproduce the issue as per description in #3574.

mhsdesign:
> However, I was unable to reproduce the issue as per description in https://github.com/neos/neos-ui/issues/3574.

Maybe because we have the hotfix https://github.com/neos/neos-ui/pull/3575 already merged? (So the console output should still complain, but less so)

Also for future reviewers: Make sure to revert https://github.com/neos/neos-development-collection/pull/4399 once its merged ;)

grebaldi:
@mhsdesign 
> Maybe because we have the hotfix #3575 already merged? (So the console output should still complain, but less so)

Oh yes, didn't see that :)

PR #3573, opened 2023-07-10
BUGFIX: 3531 node creation dialog prevent accidental close

mhsdesign:
resolves: #3531

discussion happening in the issue ;)

https://github.com/neos/neos-ui/assets/85400359/ab29bdeb-5fc2-4995-ac00-b4c238ad1a38

**What I did**

**How I did it**

**How to verify it**

PR #3579, opened 2023-07-11
BUGFIX: 3531 Prevent closing of node creation dialog

### LINKED ISSUE #3531: BUG: Click outside of the node creation dialog closes the dialog
I think making this decision is to simple and a bit unexpected user behavior. If you fill out a more complex dialog and accidentally click into the darkened space around, it is immediately closed without asking what to do with the pending changes.

I would rather:
A: Do nothing (one should use the escape key or the "back" button)
B: Show an popup warning that we will now proceed to close

... Fun fact, due to my nervous clicking, this bit me in a customer presentation ^^

mhsdesign:
resolves: #3531

(alternative fix than https://github.com/neos/neos-ui/pull/3573)

and show shaking animation like in the neos login via DialogWithOverlay.preventClosing

we currently always prevent any accidental closing independent of pending changes

https://github.com/neos/neos-ui/assets/85400359/b81ce4a8-8f12-4465-8573-3077212654f5

(And yes escape is covered / prevented as well)

thanks @grebaldi for the help ;) ❤️ 

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/packages/neos-ui/src/Containers/Modals/NodeCreationDialog/index.js
+++ b/packages/neos-ui/src/Containers/Modals/NodeCreationDialog/index.js
@@ -307,7 +307,7 @@ export default class NodeCreationDialog extends PureComponent {
                 actions={[this.renderBackAction(), this.renderSaveAction()]}
                 title={this.renderTitle()}
                 onRequestClose={this.handleCancel}
-                preventClosing={true}
+                preventClosing={this.state.isDirty}
                 type="success"
                 isOpen
                 style={this.state.secondaryInspectorComponent ? 'jumbo' : 'wide'}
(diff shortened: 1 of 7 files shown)


crydotsnake (changes requested):
Looks good!

But i would make the border orange in this case :)

Edit:

This makes the visual feedback more understandable for the editor. It probably is. But an orange colored border simply fits better here.

mhsdesign:
@crydotsnake adjusted see https://github.com/neos/neos-ui/issues/3531#issuecomment-1632040173 ;)

crydotsnake (approved):
Works great! 💙

grebaldi (approved):
Code looks :+1: and I was able to verify manually that both closing the dialog by clicking outside and closing the dialog by  escape key are prevented when there's input in the node creation dialog.

Very nice work, @mhsdesign! :)

Sebobo (approved):
Could be breaking if someone used the exported DialogWithoutOverlay which we don't explicitly expose in the api, but let's assume this didn't happen.

PR #3581, opened 2023-07-13
BUGFIX: Show resize icon on safari for document tree

### LINKED ISSUE #3556: BUG: Document and Content tree not resizable on safari
resizing was introduced with https://github.com/neos/neos-ui/pull/3227

### Description

Resize of the document tree in Safari with Neos 8.3 is not possible as the icon is invisible

### Steps to Reproduce

Try to resize the document tree in Neos 8.3

#### Expected behavior

Resize icon is visible and can be dragged

#### Actual behavior

Icon is not visible

### Affected Versions

Neos: 8.3

UI: 8.3

Sebobo:
**What I did**

Both resize icons for the left sidebar are now visible (functionality was always there only invisible)

Resolves: #3556

**How I did it**

Removed the background color from the tree component itself and instead set the background colours of the parent component that define the resize option.

**How to verify it**

Check in all browsers whether the resize icons all properly show up.

### DIFF
--- a/packages/neos-ui/src/Containers/LeftSideBar/NodeTree/style.module.css
+++ b/packages/neos-ui/src/Containers/LeftSideBar/NodeTree/style.module.css
@@ -3,7 +3,6 @@
     /* Node Tree is expected to be rendered inside a display: flex container - and it will take up all remaining space */
     flex: 1 0 0px;
     transition: var(--transition-Slow) ease height, overflow-y var(--transition-Slow);
-    background: var(--colors-ContrastDarker);
     border-bottom: 1px solid var(--colors-ContrastDark);
     border-right: 1px solid var(--colors-ContrastDark);
 }
--- a/packages/neos-ui/src/Containers/LeftSideBar/NodeTree/style.module.css
+++ b/packages/neos-ui/src/Containers/LeftSideBar/NodeTree/style.module.css
@@ -1,5 +1,4 @@
 .pageTree {
-    overflow-y: auto;
     /* Node Tree is expected to be rendered inside a display: flex container - and it will take up all remaining space */
     flex: 1 0 0px;
     transition: var(--transition-Slow) ease height, overflow-y var(--transition-Slow);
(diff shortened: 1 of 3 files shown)


crydotsnake (commented):
The resize icon is visible. But for a reason i still cant resize the document tree. I dont think I'm too stupid too catch it 🤔 

But maybe this is something for another PR since this was only about the icon (?)

https://github.com/neos/neos-ui/assets/39345336/fe1a03c2-abbf-421d-a19a-8aceb0a15ae6

Sebobo:
@crydotsnake can you give more details about your browser and version?

crydotsnake:
> @crydotsnake can you give more details about your browser and version?

Safari Version 16.5.1

Sebobo:
Ah thx, I can reproduce the issue when the window height is quite small. I will check. This fix doesn't make sense without fixing this too.

markusguenther:
I could make the resize work in safari, but the Icons are still pretty small, and it is no good UX to resize the trees in Safari. In other browsers, it still works fine.

markusguenther:
Ok, acceptance tests now also work :)

Sebobo:
Will test again

Sebobo:
> I could make the resize work in safari, but the Icons are still pretty small, and it is no good UX to resize the trees in Safari. In other browsers, it still works fine.

But now the whole tree incl. the topbar scrolls? That's also wrong

Sebobo:
So I tried a few things. The main problem is that Safari uses a higher z-index for the scrollbar of the page tree subcomponent than for the resize icon of the parent container and they overlap. I tried the trick from https://stackoverflow.com/questions/21684101/css-vertical-scrollbar-padding-left-right-in-ul-possible but that didn't work properly either.

What works is to add a padding to `leftSideBar__top` which makes the icon work, but scrolling looks strange then... not sure if we can fix this without Safari being fixed 🤷🏻‍♂️

mhsdesign:
Hi whats the status on this one? We found out that the autoprefixer for safari wasn't working so maybe this is related?

PR #3588, opened 2023-08-07
BUGFIX: Image cropper has incorrect height with 150% zoom

mhsdesign:
The `containerHeight` should be (on my pc) `532px` but instead is higher with `617px` as the `ReactCrop` component will already be rendered and falsifies the result (stretches the inner height). To get the correct height of an empty SecondaryEditor, we either need to calculate `100vh - 41px - 41px` (the size of the upper toolbars)

An alternative approach would be, to `display: none` everything in the secondary editor (especially the `ReactCrop` component) to be able to calculate the correct height without any inner influence, and later show the inner contents again.

actual:

image

expected:

image

**What I did**

**How I did it**

**How to verify it**

Set your resolution (on windows in the settings to 150%)

or on mac, just use the browser zoom to get to 150%

### DIFF
--- a/packages/neos-ui-editors/src/SecondaryEditors/ImageCropper/index.js
+++ b/packages/neos-ui-editors/src/SecondaryEditors/ImageCropper/index.js
@@ -86,7 +86,9 @@ export default class ImageCropper extends PureComponent {
     componentDidMount() {
         //
         // Calculate and set maximum height for the cropped image
-        const secondaryEditorHeight = window.innerHeight - 41 - 41;
+        // The upper toolbars (the publish tool bar and ckeditor bar) are each 41px
+        const upperToolbarHeights = 41 + 41;
+        const secondaryEditorHeight = window.innerHeight - upperToolbarHeights;
         const toolbarStyles = getComputedStyle(this.toolbarNode);
         const toolbarFullHeight = parseInt(toolbarStyles.height, 10) + parseInt(toolbarStyles['margin-top'], 10) + parseInt(toolbarStyles['margin-bottom'], 10);
         const spacing = 32;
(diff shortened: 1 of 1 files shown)


packages/neos-ui-editors/src/SecondaryEditors/ImageCropper/index.js
```diff
@@ -86,11 +86,12 @@ export default class ImageCropper extends PureComponent {
     componentDidMount() {
         //
         // Calculate and set maximum height for the cropped image
-        const containerHeight = this.containerNode.parentElement.clientHeight;
+        const secondaryEditorHeight = window.innerHeight - 41 - 41;
```
mhsdesign: While this solution is a little hacky (as that we hardcode where it will be used) but i think this is fine for the current usecase?

Another solution would be to use bare css to fit the react cropper somehow. Or is there a way to get the minHeight of a container - the height it would have without any children (without hiding its contents)

mhsdesign: Just tried all my css wizardry but in the version we've got, there is no way (at least for me) we need to use max-width with hard coded values.

but also note we are using 1.2 while version !!! 10 !!! is out there - might as well be fixed now

JamesAlias: Add a comment (in the code), why we hardcoded it like this.
Then it's absolutely fine for me 👍 

What I like to do is to give "magic numbers" a name so I have at least an idea where they're coming from.
E.g. 

```suggestion
        // The upper toolbars (the publish tool bar and ckeditor bar) are each 41px
        const upperToolbarHeights = 41 + 41;
        const secondaryEditorHeight = window.innerHeight - upperToolbarHeights;
```

grebaldi (changes requested):
Hi @mhsdesign,

I'm unable to reproduce this on Linux with Firefox, but the general idea in-code looks good to me. I agree with @JamesAlias' comment though :)

grebaldi (approved):
Thanks :)

mhsdesign:
> I'm unable to reproduce this on Linux with Firefox

did you zoom in 150% or more? ^^ (Or tried to set the desktop resolution to 150%?)

JamesAlias (approved):

PR #3596, opened 2023-08-15
TASK: Add moment locales to have a localized DatePicker

JamesAlias:
**What I did**
A customer notized that the DatePicker Editor is not localized (anymore?).
So I added the missing moment locales.

**How I did it**
Our current DatePicker is localized via `moment` locales, so I had to import them.
This sadly adds ~200kb to the bundle.

We could also only import specific locales (e.g. `de`, etc.) to keep the bundle size impact smaller.

**How to verify it**
Change language in user setting and open the DatePicker Editor in the Neos Inspector.

### DIFF
--- a/packages/react-ui-components/src/DateInput/dateInput.tsx
+++ b/packages/react-ui-components/src/DateInput/dateInput.tsx
@@ -10,6 +10,8 @@ import {PickDefaultProps} from '../../types';
 import Button from '../Button';
 import Icon from '../Icon';
 
+// WHY: Because momentJs locales are not bundled automatically, we have to explicitly add them.
+import 'moment/locale/de';
 
 export interface DateInputProps {
     /**


skurfuerst (approved):
looks good to me

Sebobo:
I would prefer to not include **all** locales and make this more easily configurable so we can extend whenever additional ones are really required. 

I guess the 200KB contain many many more format types than we even display. Is it somehow possible to just select the ones we actually need for the picker?

And if it's a bugfix, then we should apply it to an older version, no?

mhsdesign (changes requested):
If we really add ~200kb to the minified bundle id say this is really a lot. I dont know how to justify this 😭 

Is this fix related to https://github.com/neos/neos-ui/issues/2467? 

and speaking of the date picker: to solve https://github.com/neos/neos-ui/issues/3391 we need either way a roundtrip to the server. Maybe we can generally remove some date calculating logic from js and move it to the server? (and thus save lots of js? ^^)

Sebobo:
Also something to look at https://dockyard.com/blog/2020/02/14/you-probably-don-t-need-moment-js-anymore

skurfuerst:
Hey, thanks for your feedback!

Hm, I feel we are coming from the "100s" to the "1000s".

We are right now looking at fixing the issue of the customer that the weeks start at Sundays in the date picker - which he finds confusing. I am totally fine with doing a bit more to get this into the core, but I would NOT right now remove moment.js or anything like that.

If that's a blocker for you, then we won't push this forward anymore but look how to solve this otherwise, but only in the project then @Sebobo @mhsdesign. I'd be curious about your take on this :)

I see the "pain" of the added 200kb; and I am totally fine reducing this a bit if this is feasible.

All the best,
Sebastian

Sebobo:
Sorry, I didn't mean we have to remove momentjs now, that can be another task. But we spend so much time on reducing the bundle and I don't want to make it bigger after a 5 minute discussion.

If we for now add a few locales that cover 99% of our users and only slightly increase the bundle, I'm fine. Then we should have a follow-up task for Neos 9.x to fix this properly.
So many projects have problems mit the size of momentjs and try to get rid of it.

skurfuerst:
Awesome - thanks @Sebobo for your clarification ❤️

JamesAlias:
All good points.

I can only add the german locale for now, as it fixes my current issue.

Moving away from momentjs would be nice, but it's a peer dependency of the DatePicker component.

mhsdesign:
I would also be fine as a fix to only additionally include the "de" locale. I was also just making sure that we dont take any rushed steps after having optimized the bundle size so long.

mhsdesign:
Does this bug also occur in 7.3? If so id like to have it back-ported and declared as `BUGFIX`

bwaidelich:
On the risk of missing the point: Maybe it's an option to use built-in [`Intl.*`localizations](https://dockyard.com/blog/2020/02/14/you-probably-don-t-need-moment-js-anymore) of the browser and only fall back to custom locales for older browsers (if required)

JamesAlias:
@mhsdesign the bug seems to be introduced with our move from webpack to esbuild.
Looks like webpack just bundled the whole momentJs bundle (including locales).

simonschaufi:
Moment is heavy. I use date-fns instead in customer projects. Maybe this is a good alternative?

JamesAlias:
date-fns is also my go-to replacement for momentJs but the DatePicker Component is depending on momentJs.
So we would have to change the DatePicker component as well.

skurfuerst:
@mhsdesign needing your second review :)

mhsdesign:
> the bug seems to be introduced with our move from webpack to esbuild.

then we should target 8.2 ;)

baschny:
Maybe the "Datepicker" should be replaced altogether? UX-wise it is still "painful", see also #1503 where this has been discussed before. Is there still no better alternative in react, which at the same time is not dependant on momentJS?

🤔

mhsdesign (approved):
lgtm ;)

PR #3600, opened 2023-08-23
Update Main.xlf: Add 'add' to the translation of the ui

yvirtp:
I added a language label for Neos.Neos.Ui:Main:add that is called in the host.js: n.translate(“Neos.Neos.Ui:Main:add”,“Add”) and always uses its fallback value.
Therefore it isn't translatable.


dlubitz (approved):
:tada: Thank you for your contribution!

PR #3601, opened 2023-08-23
BUGFIX: 3594 reduce deprecation logs

### LINKED ISSUE #3594: BUG `plow` deprecated log spamming
regression from https://github.com/neos/neos-ui/issues/3425

image

i feared this already, but dindt know it was that bad ^^

mhsdesign:
Closes #3594

**What I did**

Reduce maximal logs to 20. (This will most likely then only warn about the first loaded plugin instead all affected plugins, but its hard to determine the plugin who called)

Also console.warn was replaced with error to have also stack straces in safari and firefox:

(you can expand the arrow)

image

**How I did it**

**How to verify it**

### DIFF
--- a/packages/neos-ui/src/deprecated-plow.js
+++ b/packages/neos-ui/src/deprecated-plow.js
@@ -1,10 +1,27 @@
 /* eslint-disable */ // IDGAF
 import * as plowOriginal from 'plow-js';
 
+const MAX_WARNINGS = 20;
+
+let warningCount = 0;
+
+let isDev = undefined;
+
 function augmentAndDeprecatedFunction(functionName, functionImplementation) {
     return function() {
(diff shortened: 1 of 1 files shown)


Sebobo:
I think even 5 or 10 entries would be enough. Isn't there a way to look at the call stack and see which file caused it? 
This way we could show it once for each JS file.

mhsdesign:
Hmm yes i thought about this too, but the string output of `(new Error).stack` is not standardized - the whole property isnt. And it looks on every Browser different and must be parsed differently - idk if its worth it?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stack?retiredLocale=de

Sebobo (approved):
Don't like it, but have to better idea 🤷🏻‍♂️

JamesAlias (approved):

Issue #3470, opened 2023-04-29
BUG: Async Neos Ui Plugins: `regeneratorRuntime is not defined`


pKallert:
### Description

When using neos expensibility to develop a plugin, using async throws errors: 
[Description of the bug or feature]

<img width="860" alt="Bildschirmfoto 2023-04-21 um 14 53 04" src="https://user-images.githubusercontent.com/91674611/235312441-122e47d5-84ff-4245-918e-579c6e09606f.png">

### Steps to Reproduce

1. Use Neos extensibility with the following dependencies: 
```json
devDependencies": {
        "@neos-project/build-essentials": "^8.2.0",
        "@neos-project/neos-ui-extensibility": "^8.2.0"
    },
```
2. Use async (for example for getting data from a dataSource):
```js
async function getData(globalRegistry) {
	const dataLoadersRegistry = globalRegistry
		.get('dataLoaders')
		.get('DataSources');
	return dataLoadersRegistry.resolveValue({
		contextNodePath: null,
		dataSourceIdentifier: 'test-identifier'
	});
}
```

#### Expected behavior
I should be able to use async and Promises with neos-ui-extensibility without adding additional packages. It should not throw additional errors. 

#### Actual behavior
I need to add an additional dependency "regenerator runtime" for no error to show up
Package: https://www.npmjs.com/package/regenerator-runtime

### Affected Versions

Neos-ui-extensibility: 8.2

UI: 8.2

mhsdesign:
Alternatively i can recommend using the new esbuild api: https://github.com/neos/neos-ui/tree/8.3/packages/neos-ui-extensibility

Sebobo:
Hm shouldn't this be solved by a correct browserslist config? 
The regenerator runtime is only formlos browser targets if I remember correctly.

mhsdesign:
I did also notice this problem in one of my plugins - but in relation with generators `yield` instead of like you with `async await`?

This is a regression from the EsBuild change in 8.2 as we removed the polyfills (`polyfill.js`) and thus also the regenerator runtime polyfill (`import 'regenerator-runtime/runtime';`). We removed the polyfills as we were from then on targeting only modern Browsers (at least ES2020) but i didnt think of the fact that build plugins might transpile down to an older es version and thus require `regeneratorRuntime`.

[...]

PR #3604, opened 2023-08-30
BUGFIX: Neos 8.3 plugins dont lower yield in plugins

mhsdesign:
Resolves: #3470

This resolves a regression from the EsBuild change in 8.2 as we removed the polyfills (`polyfill.js`) and thus also the regenerator runtime polyfill (`import 'regenerator-runtime/runtime';`). _We removed the polyfills as we were from then on targeting only modern Browsers (at least ES2020) but i didnt think of the fact that build plugins might transpile down to an older es version and thus require `regeneratorRuntime`._

The problem is obscured because other older Neos Ui Plugins might already ship the global `regeneratorRuntime`  and then other plugins can profit from this invisible dependency and work. But as soon as no other plugin is installed anymore or the settings loading order changes (due to a deployment [on another unix system where settings might be picked up in a different order](https://github.com/neos/neos-development-collection/pull/4394#issuecomment-1629254790)) your plugin might stop working.

Plugins using the modern webpack extensibility (v8.2 or 8.3) should simply at no time build plugins which require the legacy `regeneratorRuntime`. This should be the default - and its currently not.

Removing the `babel-plugin-transform-regenerator` from our webpack config works and doesnt transform the code. `yield` stays a js-generator and will not be rewritten to a state machine which depends on `regeneratorRuntime`.

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/packages/neos-ui-extensibility-webpack-adapter/scripts/helpers/webpack.config.js
+++ b/packages/neos-ui-extensibility-webpack-adapter/scripts/helpers/webpack.config.js
@@ -45,7 +45,6 @@ module.exports = function (neosPackageJson) {
                                 require.resolve('babel-plugin-transform-es2015-block-scoping'),
                                 require.resolve('babel-plugin-transform-es2015-typeof-symbol'),
                                 require.resolve('babel-plugin-transform-es2015-modules-commonjs'),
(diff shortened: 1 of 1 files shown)


Sebobo (approved):
Fine by reading

markusguenther (approved):

PR #3607, opened 2023-09-04
FEATURE: Introduce URL Validator

crydotsnake:
**What I did**

I added a new validator that checks, if a URL begins with: `https`, or `https`.
I'm still not quite if this would be really usefull for someone. So I'm open for any opinions and ideas if you have one! :)

Please use this pull request in combination with:
https://github.com/neos/neos-development-collection/pull/4486

**How I did it**

**How to verify it**

### DIFF
--- a/packages/neos-ui-validators/src/URL/index.spec.js
+++ b/packages/neos-ui-validators/src/URL/index.spec.js
@@ -1,0 +1,20 @@
+import urlValidator from './index';
+
+const httpsFormat = /^https?:\/\//;
+const httpFormat = /^http?:\/\//;
+
+test('URL without https', () => {
+   expect(urlValidator('example.com without https')).toBe(null);
+});
+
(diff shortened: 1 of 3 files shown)


mhsdesign:
Hmm i think people are currently using the regex validator to archive the same behaviour. A bigger gain would be imo to actually validate a full uri by spec? But then again a simple Regex validator works also in most cases …

crydotsnake:
Closed because I'm not sure if this would be really useful. Since we already have a Validator for Regular Expressions. And there i could also say as a pattern: `/^https?:\/\//` or `/^http?:\/\//`

Sebobo:
If we have a validator like that it should have a regex like defined by the spec and not just the protocol.
Until then this could also be solved by an example on docs.neos.io and improved by a PR that allows custom error messages for the regular expression validator.

PR #3610, opened 2023-09-05
Catch empty node names

nezaniel:
This resolves an obsolete dependency on node names for tethered checks


mhsdesign:
I found out about this one too - we discussed that we actually want to rather check the classification.

Also we don't want the nodeName to be set at all for new nodes. All those changes are incorporated here: https://github.com/neos/neos-ui/pull/3515

nezaniel:
okay :)
I just need the branch for the starship right now; This seems to be the only place where the UI has issues with unnamed nodes

mhsdesign:
Okay but i would prefer if you could test my branch https://github.com/neos/neos-ui/pull/3515 ^^ And approve it :D


3515 should also work fine in outer space

nezaniel:
yes, sure, will do :)

PR #3619, opened 2023-09-16
BUGFIX: fix hover on toggle icon not changing color to blue, add hove…

Devclaim:
**What I did**

*Problem 1: Hover States*

Looking at the CSS it looks as if the toggle button of a node in the Content and Page NodeTrees is supposed to change their color to blue when hovering over them.

While at it I also replicated the behavior from the top left burger menu. The Label and the icon now appear blue when they are either hovered over or selected (While selected the grey Background is still present).

*Problem 2: Collapsing all children*

Since I am new to Reselect  I wasn't able to create a new selector to get all descendants of a node. Also I didn't want to break anything accidentally. Instead I added the function to optionally collapse the children of a node by shift-clicking the toggle icon of a node.

*Problem 3: Make it easier to drop a node into a dropTarget*

**How I did it**

*Problem 1: Hover States*

* Styles are now applied to the toggle  SVG
* Added a hover state to the nodes

*Problem 2: Collapsing all children*

1. The toggle action now has more arguments
    - ```collapseChildren: boolean``` true if the shift key is pressed on the click event
    - ```childrenContextPaths: NodeContextPath[]``` the context paths of all children of the clicked node 
      - children that cant be toggled are filtered out
      - checks if children are loaded at all
   - ```childrenCollapsedByDefault: boolean```
     - checks if children are collapsedbydefault (unlike a normal node these nodes must be removed from the redux store to be appear collapsed

2. To integrate a simple collapse all button the toggle buttons could be selected and the click events could be fired including the shiftKey event via simple JS. I wonder if a solution like this would be up to the standards? Since the redux actions would be fired, I don't think it would create any unwanted side effects.

*Problem 3: Make it easier to drop a node into a dropTarget*

I added extra height to the dropTarget but only after a node is already being dragged.

**How to verify it**

Before: No hover effects

After:
New Hover Effects and collapsing children nodes in action:

### DIFF
--- a/packages/neos-ui-redux-store/src/UI/ContentTree/index.ts
+++ b/packages/neos-ui-redux-store/src/UI/ContentTree/index.ts
@@ -33,7 +33,7 @@ export enum actionTypes {
     SET_AS_LOADED = '@neos/neos-ui/UI/ContentTree/SET_AS_LOADED',
 }
 
-const toggle = (contextPath: NodeContextPath) => createAction(actionTypes.TOGGLE, contextPath);
+const toggle = (contextPath: NodeContextPath, collapseChildren: boolean, childrenContextPaths: NodeContextPath[], childrenCollapsedByDefault: boolean) => createAction(actionTypes.TOGGLE, {contextPath, collapseChildren, childrenContextPaths, childrenCollapsedByDefault});
 const startLoading = () => createAction(actionTypes.START_LOADING);
 const stopLoading = () => createAction(actionTypes.STOP_LOADING);
 const reloadTree = () => createAction(actionTypes.RELOAD_TREE);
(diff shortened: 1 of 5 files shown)


crydotsnake:
Hello @Devclaim 

Thanks for your pull request!

Since this is BUGFIX PR it needs to be opened against the `7.3` branch. Because the `7.3` LTS gets bugfix updates until december 2023.

Checkout our [roadmap](https://www.neos.io/features/release-process.html) for more informations.

Devclaim:
Hello, sorry I did not have that in mind. I opened a PR for 7.3 here: https://github.com/neos/neos-ui/pull/3635  . Luckily it's almost the exact same PR, only the name of the css file is different in the 7.3 branch. Do I close this PR now? I am not sure what happens next and how the changes -if accepted- make it into the 8.3 branch.

markusguenther:
Guess we can close this as a duplicate of https://github.com/neos/neos-ui/pull/3635
Thanks for your work 💙

PR #3634, opened 2023-10-06
BUGFIX: Show the right error message if only child is published 

### LINKED ISSUE #3633: BUG: Inconsistent error message when publishing childpage after create and copy
### Current Behavior

When I copy/create a page and a childpage into a different dimension, and publish only the child page, an error message is shown:
Bildschirmfoto 2023-10-06 um 14 49 43

### Expected Behavior

Usually the error message should be consistent with other error messages that help the user find the problem with the publishing: 

Bildschirmfoto 2023-10-06 um 14 59 22

### Steps To Reproduce

Discovered on Neos Demo: 

1. Go to childpage of Features, for example "Multiple Columns"
2. Copy page to different dimension (german) using the Dimension dropdown and 'create and copy' option
3. Publish only the child page 'Mulitple Columns'

### Environment

```markdown
- Flow:
- Neos: 9.0
- PHP: 8.2
```

pKallert:
closes #3633 

**What I did**
I adjusted the message that is output to the user if the parentnode does not exist in the current dimension.


ahaeslich (approved):

Classes/Controller/BackendServiceController.php
```diff
@@ -243,6 +244,11 @@ public function publishAction(array $nodeContextPaths, string $targetWorkspaceNa
                     'Node could not be published, probably because of a missing parentNode. Please check that the parentNode has been published.',
                     1682762156
                 );
+            } catch (NodeAggregateDoesCurrentlyNotCoverDimensionSpacePoint $e) {
+                throw new NodeAggregateDoesCurrentlyNotCoverDimensionSpacePoint(
```
mhsdesign: Thanks for improving the error ;)

I find the pattern of rethrowing the same error class a bit odd. I see that this pattern was used also in the code above and introduced via https://github.com/neos/neos-ui/pull/3468

But I like to suggest to use a generic RuntimeException and additionally pass the original exception as third parameter for both cases.

pKallert: I had a look and all of the exceptions are domain exceptions, so it would make more sense to rethrow that to me. 
Looks like this now: 

Bildschirmfoto 2023-12-08 um 09 08 32

dlubitz (approved):
:+1:

mhsdesign (approved):
Can be merged after solving the conflicts ;)

PR #3640, opened 2023-10-12
TASK: cleanup node service

mhsdesign:
i leveraged 
```
$this->contentRepositoryRegistry->subgraphForNode($node)
            ->findClosestNode($node->nodeAggregateId, FindClosestNodeFilter::create(nodeTypeConstraints: NodeTypeNameFactory::NAME_DOCUMENT));
```

over manually iterating over the parents. Helper methods were inlined in the few places used.
The "NodeService" was renamed to something which seems more internal and should not be used from outside ;)

**What I did**

**How I did it**

**How to verify it**

requires https://github.com/neos/neos-development-collection/pull/4601 to be merged first.


nezaniel (approved):
Fine by me

PR #3644, opened 2023-10-19
TASK: Add phpstan

### LINKED ISSUE #3631: Static code analysis
we need phpstan

mhsdesign:
**What I did**

**How I did it**

**How to verify it**


Classes/Controller/BackendServiceController.php
```diff
@@ -208,9 +199,8 @@ public function publishAllAction(): void
     /**
      * Publish nodes
      *
-     * @param array $nodeContextPaths
+     * @param list<string> $nodeContextPaths
```
dlubitz, line 202: Why `list` here instead of array? AFAIS it is not necessary and would simply set stronger constraints than needed.

Nothing blocking, I'm just currious,

dlubitz (approved):

Classes/Controller/BackendServiceController.php
mhsdesign: phpstan wants us to specify the subtype of the array. And yes indeed we could also use `array<string>` or `string[]` - i think an integer list conveys best the intention.

Classes/Fusion/Helper/NodeInfoHelper.php
```diff
@@ -78,12 +63,6 @@ class NodeInfoHelper implements ProtectedContextAwareInterface
      */
     protected $baseNodeType;
 
-    /**
-     * @Flow\InjectConfiguration(path="userInterface.navigateComponent.nodeTree.loadingDepth", package="Neos.Neos")
-     * @var string
-     */
-    protected $loadingDepth;
-
```
nezaniel, line 86: where is this used instead?

nezaniel (approved):

Classes/Fusion/Helper/NodeInfoHelper.php
mhsdesign: The property is also unused in this class in Neos 7.3

However the feature still works

PR #3649, opened 2023-10-22
BUGFIX: Respect also min value in width of input field on RangeEditor

### LINKED ISSUE #3648: Negative values in RangeEditor get cut off
### Description

A negative value in the RangeEditor get's cut off. When you have a minimal value of 0 and high negative number as min, it is even worse, as only the `max` option is taken into considering on how long the input field should be

### Steps to Reproduce

Create a property like this

```yaml
propertyName:
  type: integer
  ui:
    label: Possible negative Value
    inspector:
      editor: 'Neos.Neos/Inspector/Editors/RangeEditor'
      editorOptions:
        min: -80
        max: 80
        step: 1
        unit: px
```

Try to set a negative value

#### Expected behavior

The value is still fully visible

#### Actual behavior

image

### Affected Versions

Neos: 8.3.6  
UI: 8.3.4

jonnitto:
Closes #3648

## Before

image

## After

image


mhsdesign (approved):
;) Thx

PR #3651, opened 2023-10-23
BUGFIX: Allow float numbers in RangeEditor

### LINKED ISSUE #3650: Range editor doesn't support steps below 1
### Description

If you have a property set with the RangeEditor and set the step below 1 (e.g. 0.1) the range editor will still use step 1. Also it isn't possible the use floats as min or max values

```yaml
property:
      type: float
      ui:
        inspector:
          editor: 'Neos.Neos/Inspector/Editors/RangeEditor'
          editorOptions:
            min: -50
            max: 50
            step: 0.1
```

### Affected Versions

Neos: 8.3.6  
UI: 8.3.4

jonnitto:
Fix #3650

## Before

https://github.com/neos/neos-ui/assets/4510166/b2fcc2fe-8c37-468b-856e-5fe23570d33d

## After

https://github.com/neos/neos-ui/assets/4510166/15a41c18-405a-448b-994c-6d67be33fc14

### DIFF
--- a/packages/neos-ui-editors/src/Editors/Range/index.js
+++ b/packages/neos-ui-editors/src/Editors/Range/index.js
@@ -65,7 +65,7 @@ class RangeEditor extends PureComponent {
         const {value, highlight} = this.props;
         const valueAsString = value === 0 ? '0' : (value || '');
         // Calculate the width of the input field based on the length of the min, max and step values
-        const numLength = (value) => value.toString().length;
+        const numLength = value => value.toString().length;
         const additionalStepLength = numLength(options.step) - 1;
         const styleWidth = Math.max(numLength(options.min), numLength(options.max)) + additionalStepLength + 'ch';
 
--- a/packages/neos-ui-editors/src/Editors/Range/index.js
+++ b/packages/neos-ui-editors/src/Editors/Range/index.js
@@ -41,7 +41,7 @@ class RangeEditor extends PureComponent {
     handleChange = event => {
         const {options} = this.props;
         const {target} = event;
-        const useParseInt = options.step % 1 === 0;
+        const useParseInt = (options.step || 1) % 1 === 0;
 
         let value = useParseInt ? parseInt(target.value, 10) : parseFloat(target.value, 10);
         if (isNaN(value)) {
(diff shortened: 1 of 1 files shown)


packages/neos-ui-editors/src/Editors/Range/index.js
```diff
@@ -42,7 +42,7 @@ class RangeEditor extends PureComponent {
         const {options} = this.props;
         const {target} = event;
 
-        let value = parseInt(target.value, 10);
+        let value = parseFloat(target.value, 10);
```
mhsdesign: i guess we should use `parseFloat` based on the currently selected type for the property?

The `Editor` props currently dont contain the type, but id be fine with adding it to the props.

Alternatively i have fetched the type in a custom plugin like this

(pseudo code)

```ts
const nodeType = useNodeType(
    useCurrentlyFocusedNode()!.nodeType
)!;
const propertyType = nodeType.properties?.[props.identifier]?.type;
```

jonnitto: I'm currently working on a easier solution: Something like `const useParseInt = options.step % 1 === 0;`

jonnitto:
Basically, we should also add a debounce function to the `<input type="text" />` for the direct input. But this is another topic

packages/neos-ui-editors/src/Editors/Range/index.js
```diff
@@ -41,8 +41,9 @@ class RangeEditor extends PureComponent {
     handleChange = event => {
         const {options} = this.props;
         const {target} = event;
+        const useParseInt = options.step % 1 === 0;
```
mhsdesign: clever idea with the steps ^^

```suggestion
        const useParseInt = (options.step || 1) % 1 === 0;
```

otherwise this would be false (if `step` is not defined) 

```
({}).step % 1 === 0
```

so but that means for full float support you have to set the step to stm like `step: 0.1`

mhsdesign (approved):
At some point we should probably also consider the `type` (`int` vs `float`) here and depending on this limit `step` to a natural number, but i think for now this is a good fix

PR #3656, opened 2023-10-30
BUGFIX: Apply multiple properties from inspector

mhsdesign:
Resolves #4656 

This inverts the `UpdateNodeInfo` feedback by applying the respective last similar feedback instead of the first, since nodes don't update their internal state any more.

**How to verify it**

Change two inspector properties at the same time


nezaniel (approved):
Looks plausible and works

PR #3657, opened 2023-10-31
TASK: Remove Shortcut Override

mhsdesign:
This logic was moved back to Neos https://github.com/neos/neos-development-collection/pull/3372

**What I did**

**How I did it**

**How to verify it**


mficzel (approved):
Can confirm that this works together with https://github.com/neos/neos-development-collection/pull/4495

PR #3662, opened 2023-11-08
BUGFIX: Make long nodetype names readable in the node infoview

Sebobo:
**What I did**

Prevent nodetype name from overflowing the node info view.

**How I did it**

Insert soft hyphens after dots and colons in the nodetype name for the NodeInfoView.
And force word break via CSS and hide browser hyphen character.

**How to verify it**

Before:

Bildschirmfoto 2023-11-08 um 14 33 54

After:

Bildschirmfoto 2023-11-08 um 14 33 30

### DIFF
--- a/packages/neos-ui-views/src/NodeInfoView/index.js
+++ b/packages/neos-ui-views/src/NodeInfoView/index.js
@@ -36,6 +36,8 @@ export default class NodeInfoView extends PureComponent {
         };
 
         const nodeType = $get('nodeType', node);
+        // Insert soft hyphens after dots and colon to make the node type more readable
+        const hyphenatedNodeTypeName = nodeType.replace(/([.:])/g, '$1\u00AD');
 
         return (
             <ul className={style.nodeInfoView}>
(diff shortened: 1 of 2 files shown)


Sebobo:
Targeting 7.3 caused a conflict mess.

ahaeslich (approved):

jonnitto (approved):
Thank you for this fix ❤️

Sebobo:
This actually caused a regression when copying the nodetype name as it now contains the visible hyphens...
Will need to find a better solution.

jonnitto:
We could add a function to copy it on clipboard on click, wdyt?

bwaidelich:
> This actually caused a regression when copying the nodetype name as it now contains the visible hyphens

I stumble upon this a lot, especially when copying the node type for a db query because those hyphens are not directly visible always but lead to the db queries to return wrong results :)

As @Sebobo [mentioned](https://github.com/neos/neos-ui/pull/3700#issuecomment-2047661503), this could probably be fixed by using `<wbr>` instead of `\u00AD`.
A "copy to clipboard" button would be nice on top :)

Sebobo:
I added a copy to clipboard in my customers project as they have an extended variant of the node info view with improved readability. I can probably transfer both the fix and the button from there (which was anyway my plan)

PR #3665, opened 2023-11-14
FEATURE: Add workspace Sync button

pKallert:
This will add a rebase button to the UI for rebasing the personal workspace of a user

**What I did**
I added a button to the Neos UI that shows up if the personal workspace of a user is outdated or in error state.

**How I did it**
Neos UI when personal workspace is outdated: 
Bildschirmfoto 2023-11-24 um 12 04 32

Clicking on the button shows popup: 
Bildschirmfoto 2023-11-24 um 11 57 55

Clicking on "Synchronize now" will rebase the workspace

Neos UI when personal workspace is outdated and in error mode: 
Bildschirmfoto 2023-11-24 um 12 03 03

Clicking on the button shows popup: 
Bildschirmfoto 2023-11-24 um 12 03 14

Related: https://github.com/neos/neos-ui/issues/3803


markusguenther:
I replaced the custom Icons with default `fontawesome` icons. As the inner circle is too small for additional signs, I added a badge as a layer. The result looks now like that:

**Up-to-date:**
Screenshot 2023-12-11 at 13 28 27

**Outdated but can be rebased:**
Screenshot 2023-12-11 at 13 28 06

**Conflict:**
Screenshot 2023-12-11 at 13 17 43

Also, the dialog looks a bit different now.

**Outdated but can be rebased:**
Screenshot 2023-12-11 at 15 54 05

**Conflict:**
Screenshot 2023-12-11 at 15 54 34

markusguenther:
I also taking care of the end to end tests.

markusguenther:
Is it OK that I additionally mark the workspace as read only, so that users cannot edit content when the workspace is falling in a conflicting state? It makes no sense to be able to edit content when it cannot be used anyway.

packages/neos-ui-redux-store/src/CR/Workspaces/index.ts
```diff
@@ -75,6 +78,11 @@ const confirmDiscard = () => createAction(actionTypes.DISCARD_CONFIRMED);
  */
 const changeBaseWorkspace = (name: string) => createAction(actionTypes.CHANGE_BASE_WORKSPACE, name);
 
+/**
+ * Rebase the user workspace
+ */
+const rebaseWorkspace = (name: string) => createAction(actionTypes.REBASE_WORKSPACE, name);
```
mhsdesign, line 84: is there no reducer needed for REBASE_WORKSPACE? i guess because its handled via saga?

mhsdesign (commented):
The code looks good so far and i also like the icon and gui :heart:

But i noticed some things while testing:

- if changes were made to the currently viewed page and a rebase is excecuted the document must be reloaded (and maybe also the page tree generally?)
- when rebasing there are two spinners which seems odd?
image
- one can only rebase the user workspace and detect conflicts there. But if the base workspace is a custom workspace and not live and it is outdated or in conflict the ui doesnt tell

Classes/Controller/BackendServiceController.php
```diff
@@ -618,4 +619,34 @@ public function generateUriPathSegmentAction(string $contextNode, string $text):
...
+            $error = new Error();
+            $error->setMessage($error->getMessage());
+
+            $this->feedbackCollection->add($error);
+            $this->view->assign('value', $this->feedbackCollection);
+            return;
+        }
+
+        $success = new Success();
+        $success->setMessage(sprintf('Successfully synced User workspace'));
```
bwaidelich: Is that something that needs to be localized?

bwaidelich (commented):
I can't really judge on the ReactJS parts though and I didn't get around testing this yet but everything that I understand looks really neat and great! :)

Classes/Controller/BackendServiceController.php
markusguenther: Indeed will take care later!

markusguenther:
@mhsdesign the spinner is a known issue but as it reacts on the same states we would need a new state then for the rebase in process 🤷‍♂️

Classes/Controller/BackendServiceController.php
mhsdesign: It might be that those messages are not a flashmessage but just show up in the console as debug info in which case translation is irrelevant

markusguenther:
@bwaidelich @mhsdesign Hope that I could fulfil your wishes :), we have now translator labels and an extra remote state and therefore also a reducer, so now the indicator is just for the rebase.

Classes/Controller/BackendServiceController.php
pKallert: We should probably then also localize all of the other $success->setMessage calls in the Controller to be consistent

dlubitz (approved):
Looks good to me by reading! Thank you @pKallert for taking care!

PR #3670, opened 2023-12-05
FEATURE: Allow grouping in DimensionSwitcher

JamesAlias:
**What I did**
If (and only if) a `Neos.ContentRepository.contentDimensions.[...].presets` preset contains a `group` then the `DimensionSwitcher` dropdown is grouped.

Example:
```yaml
Neos:
  ContentRepository:
    contentDimensions:
      language:
        label: 'Language'
        icon: icon-language
        default: en_UK
        defaultPreset: en_UK
        presets:
          en_UK:
            label: 'English (UK)'
            values:
              - en_UK
            uriSegment: uk
            group: NON-EU
          de:
            label: German
            values:
              - de
            uriSegment: de
            group: EU
```

**How I did it**
I additionally map the `group` field when transforming language presets to select box options.

**How to verify it**
Wrote a test for it.

Add groups to language presets and open the DimensionSwitcher.

Screenshot 2023-12-05 at 1 29 04 PM

### DIFF
--- a/Tests/IntegrationTests/Fixtures/1Dimension/SitePackage/Configuration/Settings.yaml
+++ b/Tests/IntegrationTests/Fixtures/1Dimension/SitePackage/Configuration/Settings.yaml
@@ -12,35 +12,42 @@ Neos:
             values:
               - en_US
             uriSegment: en
+            group: NON-EU
           en_UK:
             label: 'English (UK)'
             values:
               - en_UK
               - en_US
             uriSegment: uk
+            group: NON-EU
           de:
             label: German
             values:
               - de
             uriSegment: de
+            group: EU
           fr:
             label: French
             values:
               - fr
             uriSegment: fr
+            group: EU
           nl:
             label: Dutch
             values:
               - nl
               - de
             uriSegment: nl
+            group: EU
           da:
             label: Danish
             values:
               - da
             uriSegment: da
+            group: EU
           lv:
             label: Latvian
             values:
               - lv
             uriSegment: lv
+            group: EU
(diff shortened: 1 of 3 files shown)


markusguenther:
As it is a feature we need to target Neos 9.0 here, or we need to discuss a potential 8.4 version. Did talk about that already at the hamburg sprint.

But ATM it would target 9.0. Like the feature, btw. ;)

JamesAlias:
> As it is a feature we need to target Neos 9.0 here, or we need to discuss a potential 8.4 version. Did talk about that already at the hamburg sprint.
> 
> But ATM it would target 9.0. Like the feature, btw. ;)

I wasn't sure which version to target, but we need it in Neos 8.x for a customer 🙈 
It _is_ a feature, but I could _mark_ it as patch 😅 

8.4 would make sense as well as upgrading to 9.x would be a big step for most afaik.

skurfuerst:
@markusguenther hey, we need this for a customer on 8.3 which we won't update too soon. So it would be really good to have this landing on an 8.x branch, as otherwise we need to patch forever... and I think the feature has little risk :) 

All the best, Sebastian

markusguenther:
By the way, I wrote a short proposal on Discuss to getting the feature out the door without tricking SemVer.
https://discuss.neos.io/t/feedback-wanted-for-neos-ui-features/6466

markusguenther (approved):
Thanks for the nice improvement. The failing end-to-end test is also broken in the 8.3 and 8.4 branches. For some reason in 8.2 and 9.0 it works and locally it works also.

markusguenther:
Ok, acceptance tests now also work :)

Sebobo (approved):
Fine by reading

Issue #3673, opened 2023-12-18
MultiSelectBox renders SelectBox Header when displaySearchBox is false


markusguenther:
### Description

While testing the PR #3536 I tried to enable and disable the search box via the feature list.
For the `MultiSelectBox` the disabled `SearchBox` was still visible.

#### Expected behavior

The space of the search field is gone when the search is disabled.

#### Actual behavior

The bottom space is the empty select box header.
<img width="321" alt="Screenshot 2023-12-18 at 20 25 47" src="https://github.com/neos/neos-ui/assets/1014126/60fd8fa5-79ed-49ab-b7ec-38d907283a19">

### Affected Versions

Neos: 7.3

UI: 7.3

PR #3674, opened 2023-12-18
BUGFIX: Prevent rendering of SearchBox

markusguenther:
When the `displaySearchBox` property is false, we should not render the `SelectBox` that is responsible for the search.

Fixes: #3673


markusguenther:
Ok need to check the failing test cases

Sebobo:
Shouldn't a default value be set for the `displaySearchbox`?
Also the test needs to be adjusted.

markusguenther:
@Sebobo It is disabled but with a multi selectbox, you see the disabled box that takes a lot of space and leads to a bad UX.
Screenshot 2023-12-18 at 20 25 47

markusguenther:
The whole concept of the components regarding select boxes is leading to issues, and it seems not that trivial to remove/hide the search box... need to think about it, and therefore it will not make it into the pre-Christmas release.

PR #3678, opened 2023-12-20
BUGFIX: Adds group for dimension configuration

markusguenther:
In version 8.4 of the neos-ui we added the group property for dimensions, that is, for grouping dimensions visually. Therefore, we need the configuration information in the UI.

The original change: https://github.com/neos/neos-ui/pull/3670


markusguenther:
Not sure if we need to add the group for the real CR as well. Currently it is just a visual grouping.

markusguenther:
With this PR the feature will also work in 9.0. The up merge was not enough...

ahaeslich (approved):

PR #3597, opened 2023-08-16
DRAFT FEATURE: add information to dimensions menu

MiauzGenau:
Related to: https://github.com/neos/neos-ui/issues/3413

What I did
Show in the dimension menu, if a dimension already exists.
Add links to existing dimensions, to be able to open them in new browser tabs.

How I did it
getNodeByContextPath gives us a Node with otherNodeVariants attached.
With this information we can find out, which content dimensions already exist and display it in the menu.

For the URL building, I use the current contextPath and append the respective dimensions.

How to verify it
Look at dimension switcher on pages, where not all dimensions exist.

How it looks
Before:
(grey = disallowed)
Screenshot 2023-07-27 at 16 25 28

After:
(grey = does not exist, strike-through = disallowed)
Screenshot 2023-07-27 at 16 22 44

PR #3680, opened 2023-12-28
FEATURE: Display document status of other dimensions in dimension dropdown

### LINKED ISSUE #3413: FEATURE: Usability Improvements for Dimensions Menu

pKallert:
Solves partially https://github.com/neos/neos-ui/issues/3413

**What I did**
In the single-dimensions dropdown, the Languages with existing documents are now a linked and can be opened in a new tab. 
The dropdown now shows which dimensions the current already document exists in. "Not allowed" dimensions have a new styling 

**How I did it**

For the dimensions menu there are two scenarios. This here describes the current implementation, but everything is still up for discussion

***Single dimension***
"Shine through" for en_UK and en_US  
Bildschirmfoto 2024-01-02 um 11 00 58
=> English(US) and English(UK) can be opened in new tab 

For "normal" languages: only one dimensions highlighted
Bildschirmfoto 2024-01-02 um 11 01 54

***Multiple dimensions***
(Currently) no shine through for dimensions. 
No linking of dimensions, this is what the apply button is for, I think? 
Disallowed dimensions are marked with strike-through (for now)
Bildschirmfoto 2024-01-02 um 13 50 49

Available dimensions are shown top-down. 
For example, if in the first dropdown "Deutsch" is chosen as the language, in the second dropdown only the variants with language deutsch and country * will be chosen. 
Bildschirmfoto 2024-01-02 um 13 57 05

**ToDos**
* full width of links in dropdown
* write e2e tests

**To be discussed**
* Do we even want the "available" Elements shown in multiple dimensions? Is it understandable?
* Visualize "Shine through" of languages in dropdown menu?
* Are shine-through dimensions counted as "available"?
* Better display of "disallowed" vs "available"?


markusguenther:
This is a successor PR of https://github.com/neos/neos-ui/pull/3597

markusguenther:
Hello, and thanks to @pKallert for highlighting that fantastic feature. I've reviewed the code and experimented with it, and it functions as you outlined. However, I'm not particularly fond of the strikeout method for disallowed dimensions. With the introduction of the new grouping feature in UI version 8.4, I thought about utilizing groups for this purpose.

This approach could potentially enhance the user experience, as users would better comprehend the implications. We could incorporate a helpful text within the groups to explain the meaning of terms like "untranslated" or "disallowed." Additionally, it might be beneficial to consider hiding disallowed dimensions, as they occupy space without serving any practical purpose.

What do you and the others think about that?
Maybe a help message would be a nice feature extension for the groups anyway.

Screenshot 2024-01-02 at 19 54 09

markusguenther:
Addendum: Linking to other dimensions in a new tab doesn't work for me in Firefox, Chrome and Safari

pKallert:
I thought a little bit more about the styling of the elements and came up with a new solution: 

Element does not exist in dimension: use grey background color 
Bildschirmfoto 2024-01-30 um 08 58 50

Element is disallowed: Text color is grey 
Bildschirmfoto 2024-01-30 um 15 00 40

Example: 
Bildschirmfoto 2024-01-30 um 15 09 15
Francais-> France exists, 
Francais-> Germany is not allowed, and does also not exist 
Francais-> US is allowed, but does not exist 

This would also maintain the current logic where disallowed Elements are grey/have opacity.

markusguenther:
That's funny, I also had this topic yesterday on my desk to continue with that.
I came to this idea.

1. **Colored Bars:**
    * Using a small colored bar for each entry is a concise and visually efficient way to communicate the status of dimensions.
    * The traffic light metaphor (red, orange, etc.) is intuitive and widely understood, making it user-friendly.
2. Not Permitted Dimensions:
    * Greyed-out entries with a red bar make it clear that these dimensions are not permitted. This combination of visual cues effectively signals to users that these dimensions are restricted.
3. Not Translated Dimensions:
    * Introducing an orange bar and a zebra background for dimensions that are not translated adds a distinct visual element to highlight this state.
    * The use of a 45-degree angle for the zebra background creates a visually interesting effect and sets it apart from the regular and greyed-out dimensions.
    
I am also half the way of the implementation. What do you think about that?

pKallert:
I tried it out locally and now have the following design: 
Not permitted: 
Bildschirmfoto 2024-03-15 um 08 19 37

Not translated: 
Bildschirmfoto 2024-03-15 um 08 20 33
Bildschirmfoto 2024-03-12 um 11 46 46
For not tranlated we could also not do the striped background: 

Bildschirmfoto 2024-03-12 um 12 41 01

As a follow-up we could also visualize the shine-through with fallback dimensions.

dlubitz (approved):
Thanks for this one. ❤️ This will make my life much easier. 😄 

Looks good by reading

ahaeslich:
@pKallert nice visualization. As addition and because I'm not sure if I'm missing it: are we explaining the color codes to the user, e.g. via using the title attribute?

dlubitz:
And pipeline is complaining and needs to get fixed before merge!

pKallert:
> @pKallert nice visualization. As addition and because I'm not sure if I'm missing it: are we explaining the color codes to the user, e.g. via using the title attribute?

Not right now, but I will get on that :)

Sebobo:
The title says draft, but we are ready for review correct?

packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/index.js
```diff
@@ -147,17 +161,28 @@ export default class DimensionSwitcher extends PureComponent {
 
     handleToggle = () => {
         this.setState({isOpen: !this.state.isOpen});
+
```
Sebobo: This newline looks like an accident

packages/react-ui-components/src/SelectBox_Option_SingleLine/style.module.css
```diff
@@ -0,0 +1,14 @@
+.dropdownLink{
```
Sebobo:
```suggestion
.dropdownLink {
```

packages/react-ui-components/src/SelectBox_Option_SingleLine/style.module.css
```diff
@@ -0,0 +1,14 @@
...
+    padding: 5px 14px;
+    display: inline-block;
+    width: 100%;
+
+}
+.hasIcon {
+    width: calc(100% - 2em);
+    padding: 5px 0px;
+}
+.linkedItem{
```
Sebobo:
```suggestion
.linkedItem {
```

packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/DimensionSelectorOption.js
```diff
@@ -3,24 +3,47 @@ import PropTypes from 'prop-types';
...
 import SelectBox_Option_SingleLine from '@neos-project/react-ui-components/src/SelectBox_Option_SingleLine/index';
+import mergeClassNames from 'classnames';
 
 export default class DimensionSelectorOption extends PureComponent {
     static propTypes = {
         option: PropTypes.shape({
             label: PropTypes.string.isRequired,
-            disallowed: PropTypes.bool
+            disallowed: PropTypes.bool,
+            existing: PropTypes.bool,
```
Sebobo: @nezaniel told us that the wording whether a node exists in the desired dimension is "coverage" or "covered"

```suggestion
            covered: PropTypes.bool,
```

Sebobo (changes requested):
Thx, looks great, left some suggestions.

packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/index.js
pKallert: Removed 👍

packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/DimensionSelectorOption.js
pKallert: I adjusted all the variables to use "covered" as naming

pKallert:
Quick opinion poll: 

Do you prefer with or without striped background to display that a document does not exist in this dimension? 

313102123-64ce50f0-dad0-4be4-b970-35245acef569313102293-f78c2116-a8f4-4578-9540-933f2203c930

Striped background: 🚀  reaction 
No striped background: 👍  reaction

crydotsnake:
> Quick opinion poll:
> 
> Do you prefer with or without striped background to display that a document does not exist in this dimension?
> 
> 313102123-64ce50f0-dad0-4be4-b970-35245acef569313102293-f78c2116-a8f4-4578-9540-933f2203c930
> 
> Striped background: 🚀 reaction No striped background: 👍 reaction

If the document doesent exist in the dimension its also not clickable, right? 🤔

ahaeslich:
The striped background is more different to the 'edited' marking we have on other places in the CMS. Therefore, it might highlight the difference a bit better.

dlubitz:
@ahaeslich is right here. You could think, that there are unpublished nodes in this dimension.

Sebobo (changes requested):
I just tested the feature again. 
It looks great but there is one problem, the state is not properly updated after creating the non-existing variant. 

1. So in the demo I'm in English
2. then German has the stripes
3. then I create the German variant via the modal
4. dropdown shows all variants normally
5. I go back to English
6. German is still shown as non-existing (but it won't show the create variant dialog when I click again)

Resources/Private/Translations/en/Main.xlf
```diff
@@ -380,6 +380,12 @@
             <trans-unit id="copyNodeTypeNameToClipboard" xml:space="preserve">
                 <source>Copy node type to clipboard</source>
             </trans-unit>
+            <trans-unit id="dimensions.doesNotExistsInDimension" xml:space="preserve">
+                <source>The current document does not exist in the dimension</source>
```
bwaidelich, line 384: I might miss the context, but I was wondering: What is "the dimension"? Would something like "...does not exist in the currently active content dimension" be clearer?

bwaidelich (commented):
I can't really judge the JS code, but it looks quite impressive to me :)

So I just left my 2ct on the xlf file – I do understand _these_ :)

dlubitz:
@Sebobo Can you please check again and remove your request for change to get this done?

Sebobo (approved):
Thx still works well. Though the issue I mentioned is still not solved, but that is related to the backend and not your implementation.

The problem is that the state of the node in the original language is outdated and therefore the dropdown shows the old state (meaning German does not exist). 
The cause of this is, that the backend returns only nodes for the current context and not the variants that have actually also been updated in the meantime.

CleanShot 2025-07-28 at 09 50 32@2x

When switching between variants via the dropdown, the information about the current document is only partially reloaded and the request for `get-additional-node-metadata` is skipped, which would update the dimension values.

So I'm fine with merging this change, but we need another issue to fix the loading of the additional metadata to prevent going out of sync while working on variants.

PR #3681, opened 2024-01-02
FEATURE Add timeable node visibility

dlubitz:
Adopt UI to use the new properties for the Timeable Node Visibility.

Rename `_hiddenBeforeDateTime` to `enableAfterDateTime` and `_hiddenAfterDateTime` to `disableAfterDateTime`.


mhsdesign (approved):
Will have to test this probably as i had the feeling it worked already without this change?

mhsdesign (approved):
Checked again an now it makes sense after the adjustments

Issue #3119, opened 2022-05-14
Turn Neos UI into a scaffold for Plugin/BackendModule integration


grebaldi:

Sebobo:
So is our goal to make it work similar to the content module, that we have an outer module application and the module itself lives in an iframe and can optionally communicate with the "Neos backend app"?

mhsdesign:
So i always wanted to know how the ui archives its superior behavior as content module. Because indeed its not just another backend module.

### Questions i had in mind.

- ### how does the content module achieve its autogenerated "submodule" points, the sites?

this logic is deeply embedded to the module integration/rendering in Neos.Neos
https://github.com/neos/neos-development-collection/blob/720e3a080339b3ab1379ff1c1774b5d9aa6ee80a/Neos.Neos/Classes/Controller/Backend/MenuHelper.php#L78

-  ### why does the module drawer at the right feel in the ui a little different?
because it _is_ different / completely rebuild

this for example is the `<Drawer />` all the menu items are inlined in the page as json and used when booting the application.
https://github.com/neos/neos-ui/blob/e53f8e20ee70ca832828f033f5a69c8223a2deab/packages/neos-ui/src/Containers/Drawer/index.js#L25 

even the neos icon and the user select box are rebuild - everything is rebuild (see next question).

-  ### is the neos ui influenced by the backend css styles of Neos.Neos?
nope - fun fact it doesnt use the module api at all - we only declare it to be a module, so it pops up in the menu

```yaml
Neos:
  Neos:
    modules:
      content:
        label: 'Neos.Neos:Main:content'
        controller: 'Neos\Neos\Ui\Controller\BackendController'
```

one can still navigate  to `/neos/content` without this config in place
https://github.com/neos/neos-ui/blob/e53f8e20ee70ca832828f033f5a69c8223a2deab/Configuration/Settings.yaml#L5-L12
and this policy? it seems useless too (just tested without it): https://github.com/neos/neos-ui/blob/e53f8e20ee70ca832828f033f5a69c8223a2deab/Configuration/Policy.yaml#L18-L21

#### So why does the ui still work?

It works because `neos/content` is a subroute declared in:
https://github.com/neos/neos-ui/blob/e53f8e20ee70ca832828f033f5a69c8223a2deab/Configuration/Routes.Backend.yaml#L1-L3

which uses the `Neos\Neos\Ui\Controller\BackendController` and is now completely independent.

btw - the above linked controller in the module definition, will never be triggered that way.
[...]

Issue #3098, opened 2022-04-08
`globalRegistry.get('frontendConfiguration')` doesnt work in plugin bootstrap


mhsdesign:
### Description

when one registers a new plugin in the `manifest` and tries to call `globalRegistry.get('frontendConfiguration')` null is returned.

### Steps to Reproduce

1. Write a manifest like:
```js
import manifest from '@neos-project/neos-ui-extensibility';
manifest('Fooo', {}, (globalRegistry) => {
    console.log(globalRegistry.get('frontendConfiguration'))
}
```
#### Expected behavior

to get the actual `FrontendConfigurationSynchronousRegistry`

#### Actual behavior

we log null

### What actually happened

the `FrontendConfigurationSynchronousRegistry` is not initialized, when the bootstraps are called.

the manifests are booted here in line 76 https://github.com/neos/neos-ui/blob/6ce277470258599c1d29f9ec2f5983e7e44d4338/packages/neos-ui/src/index.js#L76

but only in line 121 https://github.com/neos/neos-ui/blob/6ce277470258599c1d29f9ec2f5983e7e44d4338/packages/neos-ui/src/index.js#L121-L127
the raw "yaml/php-array" frontendConfiguration array is transferred to a `SynchronousRegistry`

### ... Sure we can get hold of the `frontendConfiguration` in another way:
1. Write a manifest like:
```js
manifest('Fooo', {}, (globalRegistry, { frontendConfiguration }) => {
    console.log(frontendConfiguration)
}
```
now it works, but note that this is the raw "yaml/php-array" and not a `SynchronousRegistry` so `frontendConfiguration.get('My.Package')`
doesnt work - but it does other places (when you *actually* have the `FrontendConfigurationSynchronousRegistry`)

### Outcome

is there a reason, why the FrontendConfigurationSynchronousRegistry is not initialized before boot and passed?
im now used to this but i find this behavior odd - and would like to have at all places a `FrontendConfigurationSynchronousRegistry` available.

### Affected Versions
UI: Since ever

mhsdesign:
Same applies for `i18n` - the translations are not available at first. `setTranslations` will only be called, after the manifests are bootet.

A workaround is to register a saga which waits until the system initializes (when the frontent config and translations are loaded.)

```js
import manifest from "@neos-project/neos-ui-extensibility";
import { take } from 'redux-saga/effects';
import { actionTypes } from '@neos-project/neos-ui-redux-store';

[...]

PR #3682, opened 2024-01-10
!!!FEATURE: Reform foundation and boot process of the UI

grebaldi:
refs:  #3119
next: https://github.com/neos/neos-ui/pull/3773

This is the first in a series of PRs that will enable the creation of multiple applications based on a Neos UI core.

This PR turns the `Backend/Index.html` Fluid template into a dedicated `ApplicationView`.  This `ApplicationView` has the sole purpose of building the foundational HTML document (via string concatenation) for bootstrapping the Neos UI.

Formerly, Fusion was used to retrieve and render the initial data for the UI. This task has been delegated to a set of dedicated classes, namely:

- `ConfigurationProvider` - retrieves the `nodeTree` and `structureTree` segments from `Neos.Neos.userInterface.navigateComponent` settings, `allowedTargetWorkspaces` from the ContentRepository's `WorkspaceService` as well as the `nodeTypeSchema` and `translations` endpoints.
- `RoutesProviderInterface` - retrieves all other routes/endpoints required for communication with the Neos server application
- `FrontendConfigurationProvider` - reads and preprocesses the `Neos.Neos.Ui.frontendConfiguration` settings that are mostly used by third-party plugins to share data between server and client
- `NodeTypeGroupsAndRolesProviderInterface` - retrieves information about node type roles and node type groups. Roles are a dedicated UI-concept that is meant to distinguish between document, content and collection nodes. Groups refer the grouping of node types in the creation dialog.
- `MenuProvider` - retrieves all data needed to render the main burger menu located in the top left corner of the UI.
- `InitialStateProvider` - reads and preprocesses the `Neos.Neos.Ui.initialState` settings that are used to hydrate the UI's redux store

The responsibilities of these classes is entirely derived from their former Fusion counterparts. All of them have been marked `@internal` to allow for later removal. Nonetheless, it remains possible to implement their accompanying interfaces and replace their implementations via `Objects.yaml`. This enables rare edge cases and unplanned extensibility.

The splash screen, formerly a Fluid template, is now hard-coded into the `ApplicationView` class, which renders the `Neos.Neos.Ui.splashScreen` setting obsolete. It has therefore been removed.

Furthermore, this PR reforms the client-side boot process of the UI. This includes the following changes:

* Scripts (The UI main script, as well as plugins) are now loaded in the `<head/>` with `defer`
  * This way, the UI does not need to listen to DOMContentLoaded in order to discover the #appContainer element. This allows a lot of the UI boot process to become synchronous.
* The main boot function is no longer a Redux Saga. It is instead an asynchronous function.
  * Also, the various boot-related sub-processes (initializing plugins, loading translations, etc...) have been wrapped in dedicated functions.
* The initial data for the boot process is now `JSON.parse`d instead of `window['_NEOS_UI_*']`ed
  * `JSON.parse` is somewhat more performant, because for JSON only a subset of ECMAScript needs to be parsed. (Though, this is not all that significant given the amount of data, but still :slightly_smiling_face:)
* The `data-first-tab` attribute has been removed from #appContainer
  * I was unable to find any references to it within the code base, so I removed it, because I didn't know what exactly to put in there. I took it for a historical leftover. Please let me know if I'm mistaken :)

**Remaining TODOs**

- [x] Replace Fusion with a dedicated application view class
- [x] Use `<head>`-scripts + `defer` instead of `<body>`-scripts for script inclusion
- [-] <s>Turn the remaining Fluid templates into Fusion prototypes as well (optional)</s>
  - [-] <s>`Backend/ConfigurationVersion.html`</s>
  - [-] <s>`Backend/Guest.html`</s>
  - [-] <s>`Backend/GuestNotificationScript.html`</s>
  - [-] <s>`Error/ErrorMessage.html`</s>
  - **EDIT:** Scope is large enough as it is - these can be done later
- [x] Remove the `actions.System.boot()` dispatch?
  - Noone seems to listen to it...
- [x] Remove `Neos.Neos.Ui.splashScreen` configuration
  - Relies on fluid partial that no longer exists
  - Replacement? -> **EDIT:** None. It's hard-coded now. We can open this up later if there's demand for it.
- [-] <s>Remove fusion objects that are no longer relevant</s>
  - [-] <s>`Neos.Neos.Ui:RenderConfiguration`</s>
  - **EDIT:** Nope, this is still being used for the guest frame
- [-] <s>Deprecate EEL-Helpers that are no longer relevant</s>
  - [-] <s>`Neos.Ui.Workspace`</s>
  - **EDIT:** Nope, this is still being used in settings
- [x] Add doc comment that explains node type roles and node type groups concepts
- [x] Rename `NodeTypesProviderInterface` -> `NodeTypeGroupsAndRolesProviderInterface`
- [x] Use an interface to className mapping in `Objects.yaml` instead of explicitly configuring `BackendController` properties
- [x] Prevent duplication of `defer` attribute on `<script/>`-includes through `Neos.Neos.Ui.resources` settings

**Note to fellow maintainers**

In order to preserve the usefulness of the commits in this PR, please do not merge any of the version branches into this branch. To update the code with the latest changes in `9.0`, please do an interactive rebase instead:

```
git fetch upstream
git checkout -b feature/multi-app/01-foundation upstream/feature/multi-app/01-foundation
git rebase -i upstream/9.0 feature/multi-app/01-foundation
```

This will require
```
git push -f upstream feature/multi-app/01-foundation
```
afterwards.

This way, every commit will be replayed on top of `9.0`, and can be adjusted seperately.


mhsdesign:
Thank you so much!!! This really cleans up many things that are needlessly messy and hard to understand. I added already a dec comments to your text and will look over the code later ;)

> Remove the actions.System.boot() dispatch?

I only know that I have used the init see:
https://github.com/neos/neos-ui/issues/3098#issuecomment-1303492943
(And just noticed the issue might be generally interesting for you;))

About:
> Remove Neos.Neos.Ui.splashScreen configuration

and
> The data-first-tab attribute has been removed from #appContainer

I have never heard of those things so I assume they can go;)


> are now loaded in <body/> instead of <head/>

Today I actually prefer the use of the defer attribute … feels cleaner to have them in the head … but it’s really not important 

--------

Note to myself:
the commit is probably the one which will get outdated the easiest and has to be taken care of carefully.
> TASK: Group all initial data variables into a single view variable

Also i think we should incorporate https://github.com/neos/neos-ui/pull/2920 into here.

-------

Thanks again!!!! :D

Resources/Private/Fusion/Backend/Component/HeadIcons.fusion
```diff
@@ -0,0 +1,57 @@
...
+ * This file is part of the Neos.Neos.Ui package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+include: resource://Neos.Fusion/Private/Fusion/Root.fusion
```
mhsdesign: this is sadly not a good idea to use fusion imo. The fusion parser will / cannot optimize this and the whole neos.fusion stuff is reparsed. That has to be done unfortunately because fusion allows global paths that can be overridden.

Resources/Private/Fusion/Backend/Component/Scripts.fusion
```diff
@@ -0,0 +1,16 @@
...
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+include: resource://Neos.Fusion/Private/Fusion/Root.fusion
+
+/** @internal */
+prototype(Neos.Neos.Ui:Component.Scripts) < prototype(Neos.Fusion:Value) {
+    value = ${scripts}
```
mhsdesign: i know this is just a draft, but i wanted to point out that this context variable `scripts` is super sneaky and unexpected here ^^

Resources/Private/Fusion/Backend/Component/HeadIcons.fusion
grebaldi: You're confirming my worst fears there :sweat_smile:

grebaldi:
Hi @mhsdesign,

thanks a lot for your very valuable feedback :+1: 

> Today I actually prefer the use of the defer attribute … feels cleaner to have them in the head … but it’s really not important

I'm always afraid of introducing race conditions when it comes to this stuff :sweat_smile:. Well, MDN states:

> This Boolean attribute is set to indicate to a browser that the script is meant to be executed after the document has been parsed, but before firing [DOMContentLoaded](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event).
(see: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#defer)

So, I guess this means we have a guarantee that the script can see the whole document? I mean `document has been parsed` sounds like that, but `before firing DOMContentLoaded` strikes me the wrong way - so as if we're at a point of execution at which the state of the DOM isn't exactly determined.

I'm with you, that we should use `defer`, especially if that means that the browser's rendering thread doesn't have to block for parsing our UI bundle before displaying anything. So, if you can take my above stated fear from me, I'll gladly change the script integration to `defer` :D

I also have some ideas bouncing around in my head with regards to the code issues you've pointed out. I similarly feel that the fusion foundation as of right now is less than ideal. I see two possible way out of here:

**A) Build dedicated Fusion `@class` implementations for resource collection** 

So, basically, this would mean to back the following prototypes with classes:
- `Neos.Neos.Ui:Component.InitialData`
- `Neos.Neos.Ui:Component.HeadStylesheets`
- `Neos.Neos.Ui:Component.HeadIcons`
- `Neos.Neos.Ui:Component.Scripts`

This would move all the resource-collecting logic out of the controller and out of the fusion code. It'd also eliminate the need for `include: resource://Neos.Fusion/Private/Fusion/Root.fusion` (at least I think it would :)).

**B) Replace Fusion with a dedicated application view class**

Because: what's the point of using Fusion to initialize the UI anyway?

A dedicated view could take care of hiding the resource-collecting logic from the controller. The `initialData` contents would map nicely onto view variables. The view could just concatenate the HTML document - nice, short and fast :slightly_smiling_face: 

wdyt?

grebaldi:
Oh, and regarding #3098:

I don't think that this issue would be fixed as of right now. But I'm going to try to move around some bits of the boot process (should be possible), so plugins have an easier time reading this stuff.

mhsdesign:
I actually dont know that much about `defer` - just used it - so lets stick at first with js body scripts ^^

> B) Replace Fusion with a dedicated application view class
> Because: what's the point of using Fusion to initialize the UI anyway?

And yeah i thought so too... well lets just use b? :D
The only thing would be of course unplanned extensibility, but i havent used that and everything useful is already available as api via settings. 

especially things like
```
# TODO: dirty hack to not have to re-implement neos:backend.configurationCacheVersion VH
version = Neos.Fusion:Template {
    templatePath = 'resource://Neos.Neos.Ui/Private/Templates/Backend/ConfigurationVersion.html'
    @process.trim = ${String.trim(value)}
}
```
make me rethink if fusion is the right choice 😅 

Aaand we might finally remove this weirdly namespaced `Neos.Ui.PositionalArraySorter.sort` and `Neos.Ui.Api` helpers 😂

grebaldi:
@mhsdesign 

> I actually dont know that much about `defer` - just used it - so lets stick at first with js body scripts

I looked into the spec, but it is just as imprecise about the exact timing as MDN is:
> If the [async](https://html.spec.whatwg.org/multipage/scripting.html#attr-script-async) attribute is not present but the [defer](https://html.spec.whatwg.org/multipage/scripting.html#attr-script-defer) attribute is present, then the classic script will be fetched [in parallel](https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel) and evaluated when the page has finished parsing.
>
(see: https://html.spec.whatwg.org/multipage/scripting.html#attr-script-defer)

Then I found this article that is very confident in saying:
> Whereas DEFER scripts don’t execute until the HTML document is done being parsed (AKA, DOM Interactive or `performance.timing.domInteractive`).
>
(see: https://calendar.perfplanet.com/2016/prefer-defer-over-async/#attributes)

Sadly no sources to back that up...

I guess it wouldn't make sense for any `defer` implementation to not have the DOM available after document parsing. So, I think we can relatively safely assume that `<head>`-scripts + `defer` will work fine. I'll add that to the list :)

> The only thing would be of course unplanned extensibility

Actually, you'd have to do quite some acrobatics to hook into the backend fusion view. It cannot be extended via regular ol' `autoInclude` - which was by design, because this view is not supposed to be interfered with.

I think the reasoning for using fusion back then was that it provided a unified high-level API for both MVC/Routing and ContentRepository, which are both major concerns for initial hydration of the UI. But this was always supposed to be a strictly hidden implementation detail.

> [...] everything useful is already available as api via settings.

https://github.com/neos/neos-ui/blob/2204bb11070d8fc43bb28cdbc1d208cefa255b91/Configuration/Settings.yaml#L94-L96

:point_up: Well... Actually those settings are not supposed to be API. However, this is certainly not the best way to hide stuff :sweat_smile: 

I'd like to get rid of this mechanism as well, but that exceeds the scope of this PR.

> well lets just use b? 

I'll add that to the list as well :)

mhsdesign:
Thanks for the housekeeping and implementing the `defer` :D

> Actually, you'd have to do quite some acrobatics to hook into the backend fusion view. It cannot be extended via regular ol' autoInclude - which was by design, because this view is not supposed to be interfered with.

yes i fully agree and also talked to @Sebobo about this. There is absolutely no need to have it in fusion and php would be a better fit.

But to keep this pr rather simpler, i would say that we could keep the fusion code in one messy file for now and refactor it to php with another followup pr?

I especially dont want to merge the `include: resource://Neos.Fusion/Private/Fusion/Root.fusion` part :D

But if you think its doable in scope of this pr i wouldnt mind moving to php quicker.

grebaldi:
> But if you think its doable in scope of this pr i wouldnt mind moving to php quicker.

Already on it :D

Classes/Domain/NodeTypesProviderInterface.php
```diff
@@ -0,0 +1,26 @@
...
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+declare(strict_types=1);
+
+namespace Neos\Neos\Ui\Domain;
+
+/**
+ * @internal
```
mhsdesign: your commit message expresses really well the definition of the roles and id very much like to have that as a doc comment here ^^

> Retrieves information about node type roles and
node type groups.
>
> Roles are a dedicated UI-concept that is meant to
distinguish between document, content and collection nodes. Groups refer
the grouping of node types in the creation dialog.

Additionally i would like to propose a different more obvious and specific naming like `NodeTypeGroupsAndRolesProviderInterface`

Configuration/Objects.yaml
```diff
@@ -0,0 +1,24 @@
...
+    routesProvider:
+      object: Neos\Neos\Ui\Infrastructure\MVC\RoutesProvider
+    frontendConfigurationProvider:
+      object: Neos\Neos\Ui\Infrastructure\Configuration\FrontendConfigurationProvider
+    nodeTypesProvider:
+      object: Neos\Neos\Ui\Infrastructure\ContentRepository\NodeTypesProvider
+    menuProvider:
+      object: Neos\Neos\Ui\Infrastructure\Neos\MenuProvider
+    initialStateProvider:
+      object: Neos\Neos\Ui\Infrastructure\Configuration\InitialStateProvider
```
mhsdesign: I would propose to go for a interface to className mapping instead of declaring just the properties for the backend-controller. Because the ConfigurationProvider also uses the `CacheConfigurationVersionProviderInterface` and via a global interface override it would become obvious how the dependencies are resolved.


```yaml
'Neos\Neos\Ui\Domain\NodeTypesProviderInterface':
  className: Neos\Neos\Ui\Infrastructure\ContentRepository\NodeTypesProvider
# ...
```

mhsdesign (commented):
WOW ❤️ 🎉 thank you so much for so cleanly moving the neos ui integration to php.
The code looks way more obvious and is easier to follow! Thanks a lot!

Classes/Domain/Service/StyleAndJavascriptInclusionService.php
```diff
@@ -61,7 +61,7 @@ class StyleAndJavascriptInclusionService
     public function getHeadScripts(): string
     {
         return $this->build($this->javascriptResources, function ($uri, $additionalAttributes) {
-            return '<script src="' . $uri . '" ' . $additionalAttributes . '></script>';
+            return '<script src="' . $uri . '" ' . $additionalAttributes . ' defer></script>';
```
mhsdesign, line 64: if one configures 

```yaml
resource: fooo
defer: true
```

or
```yaml
resource: fooo
attributes:
  defer: true
```

there will be actually a second `defer` attribute in $additionalAttributes already 🤯 shouldn't be a real problem though ^^

i could imagine omitting the `defer` below in the `htmlAttributesArrayToString`.

And it seems that if someone uses an esm plugin with 

```yaml
resource: fooo
attributes:
  type: "module"
```

we can omit this defer tag, as its implicitly set and doppeltgemoppelt? https://wiki.selfhtml.org/wiki/HTML/Attribute/defer

mhsdesign:
Fyi this will be a bit nasty to rebase to https://github.com/neos/neos-ui/pull/3689 but i think doable, so i would propose to merge that first?

grebaldi:
@mhsdesign 
> Fyi this will be a bit nasty to rebase to https://github.com/neos/neos-ui/pull/3689 but i think doable, so i would propose to merge that first?

#3689 should definitely go first. I don't think rebase will be that big a deal :)

Classes/Domain/Service/StyleAndJavascriptInclusionService.php
mhsdesign: But i could confirm that `<script src="plugin" type="module" defer=""></script>` works ;)

mhsdesign: And i see you fixed my initial concerns via https://github.com/neos/neos-ui/pull/3682/commits/4325fb0d5c239954e37491cf56c683ad4a85dbfd ;)

mhsdesign:
Id say this is nearly ready to be merged. Because im a little paranoid i actually diffed the initial bootstrap html before and after and found a few differences worth noting.

First it seems that the `User Settings` module link was not included previously, but now shows up and will be rendered in the menu:

```json
        {
            "label": "Neos.Neos:Modules:user.label",
            "icon": "fas fa-users",
            "children": [
                {
                    "icon": "fas fa-user",
                    "label": "Neos.Neos:Modules:userSettings.label",
                    "uri": "http:\/\/localhost:8081\/neos\/user\/usersettings",
                }
            ]
        }
```

There seems to be a `module.hideInMenu` or `submodule.hideInMenu` flag to toggle this.

image

And second, it seems that we boot the ui with more data: I found that everything mapped to the legacy globals but there is a new entry `settings` with seems to be an complete dump of the whole Neos.Neos.Ui configuration, at least by looking at its keys:

- 'resources'
- 'contentCanvas'
- 'frontendConfiguration'
- 'frontendDevelopmentMode'
- 'nodeTypeRoles'
- 'configurationDefaultEelContext'
- 'documentNodeInformation'
- 'initialState'
- 'changes'
- 'outOfBandRendering'

Edit. I found the cause. We json_encode all the $this->variables, but it was once thought be a good feature to have the settings automagically available see https://github.com/neos/flow-development-collection/commit/b944521345d1801ce146a3aa6be80b5871b3224e.
So i guess we need to namespace all the variables in something like `booting` or `value` like the json view idk.

mhsdesign (approved):
I adjusted the previously mentioned parts a little and now everything is like expected.
I would prefer another or two reviews though ^^
Thanks again for the hard work!

Classes/Domain/CacheConfigurationVersionProviderInterface.php
```diff
@@ -0,0 +1,27 @@
...
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
+ * This package is Open Source Software. For the full copyright and license
+ * information, please view the LICENSE file which was distributed with this
+ * source code.
+ */
+
+declare(strict_types=1);
+
+namespace Neos\Neos\Ui\Domain;
```
mhsdesign: shouldnt we group all these `ProviderInterface`s into one sub namespace like `Neos\Neos\Ui\Domain\BackendProviders`?

That way `Neos\Neos\Ui\Domain` is not that cluttered and there is also "space" for other interfaces not directly related to the backend, like the NodeCreationHandlerInterface for example?

I could take care of that refactoring, but i wanted to ask for your blessing @grebaldi ^^

mhsdesign: Or maybe `Neos\Neos\Ui\Domain\InitialData` or `Neos\Neos\Ui\Domain\Booting` or `Neos\Neos\Ui\Domain\InitialDataProvider`?

mhsdesign: Discussed this with Wilhelm and he was okay with moving it to `InitialData`.

Configuration/Views.yaml
```diff
@@ -1,7 +1,6 @@
 -
   requestFilter: 'isPackage("Neos.Neos.Ui") && isController("Backend")'
-  viewObjectName: 'Neos\Fusion\View\FusionView'
+  viewObjectName: 'Neos\Neos\Ui\Presentation\ApplicationView'
```
Sebobo: Do we still need the Views.yaml as the ApplicationView is already assigned in the Controller?

Sebobo (approved):
LGTM, just left one question regarding the necessity of the Views.yaml

Configuration/Views.yaml
mhsdesign: yes we should probably remove that.

mhsdesign: Fixed via https://github.com/neos/neos-ui/pull/3682/commits/ba8cf91b2df323fd112945229362173f5ccadd46 now using the $defaultViewObjectName as that is more predictable and views yaml is rather considered for extensibility nowadays

PR #3693, opened 2024-01-22
BUGFIX: Fix dimension dropdown on browser zoom

### LINKED ISSUE #3639: BUG: DimensionSwitcher dropdown not visible at specific zoom levels
### Description

At certain browser zoom levels, the language selector dropdown disappears. This only affects the DimensionSelector component and none of the other dropdowns in the primary toolbar.

### Steps to Reproduce

Set your browser zoom to something like 50% (we have seen this vary from computer to computer depending on the available screen-size). 
Click to toggle the DimensionSwitcher

#### Expected behavior

The dimension-selection dropdown should appear

#### Actual behavior

The dimension-selection dropdown does not appear

### Affected Versions

Neos: 8.3.5

UI: 8.3.4

pKallert:
resolves #3639 

**What I did**
I had a look at why the language dropdown disappears when the browser is zoomed out. 

**How I did it**
The SelectBox is loading DropDown.Contents with scrollable="true" which triggers a check if the element is out of bound. 
The check for const elementBoundingBox = el.getBoundingClientRect(); is returning -0.05 for top when I zoom out 80% and thus is "out of bounds" of the viewport. There is apparently some css that makes the dimensions dropdown slightly higher than the right topbar. 
I could not find a good reason why the Element is out of bounds by 0.05, perhaps there is a better solution? 
To resolve the issue, I placed the dimensions menu at the top of the bar, thus it cannot go over the top of the bar. 

**How to verify it**
When zooming out, the dimensions menu is visible again: 
Bildschirmfoto 2024-01-22 um 14 23 53
Before that, Menu was not opening: 
Bildschirmfoto 2024-01-22 um 14 25 31

Also, I checked on desktop without browser zoom and there was no visible change in styling with the additional CSS

### DIFF
--- a/packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/style.module.css
+++ b/packages/neos-ui/src/Containers/PrimaryToolbar/DimensionSwitcher/style.module.css
@@ -71,6 +71,7 @@
 .singleDimensionDropdown {
     display: flex;
     align-items: center;
+    align-self: start;
 
     .dimensionSwitcherDropDown {
         background: unset;


grebaldi (approved):
Hi @pKallert,

thanks for taking care of this issue :)

I was able to reproduce that (indeed very weird) behavior and can verify that your change fixes it :+1:

markusguenther:
Thanks @pKallert, but I guess we should target 8.3 here :)

pKallert:
> Thanks @pKallert, but I guess we should target 8.3 here :)

I switeched it out 👍

markusguenther (approved):

markusguenther:
Thanks for taking care 💙

Issue #3616, opened 2023-09-15
BUG: Nodecreation position icons not shown


mhsdesign:
### Description

<img width="436" alt="image" src="https://github.com/neos/neos-ui/assets/85400359/a55e9711-9b1e-4856-8604-0dcd76495d3b">

expected:

<img width="276" alt="image" src="https://github.com/neos/neos-ui/assets/85400359/a1ef9ddc-b667-4113-ab5c-06ce6e96db55">

Regression since using custom icons: https://github.com/neos/neos-ui/pull/3423

We currently just concatenate the uri on client side which doesnt respect cases where one uses resources in the cloud.
The bigger scope of the issue is described here: https://github.com/neos/neos-ui/issues/2092

[Description of the bug or feature]

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

#### Expected behavior

[What you expected to happen]

#### Actual behavior

[What actually happened]

### Affected Versions

Neos:

UI:

You can get this information by running `composer show` or using the package management module within Neos.
The UI version can also be found inside the left sidebar after version 1.0.3.

Sebobo:
Why do we use separate icons with their own uris for this and not the iconset?

Sebobo:
Just for this issue we could also embed the SVGs to skip the requests.

ahaeslich:
Okay now we have two use cases:

- cloud resources
- a Neos installation residing in a "subfolder": #3712

PR #3695, opened 2024-01-23
bugfix/2092 resolve resource uris on server

mhsdesign:
Resolves: https://github.com/neos/neos-ui/issues/2092
Resolves: https://github.com/neos/neos-ui/issues/3616

- BUGFIX: Extract resource resolving magic out of `ResourceIcon`
- BUGFIX: Move resource logic to server

### DIFF
--- a/Classes/Controller/BackendServiceController.php
+++ b/Classes/Controller/BackendServiceController.php
@@ -31,6 +31,7 @@ use Neos\Flow\Mvc\View\JsonView;
 use Neos\Flow\Persistence\Exception\IllegalObjectTypeException;
 use Neos\Flow\Persistence\PersistenceManagerInterface;
 use Neos\Flow\Property\PropertyMapper;
+use Neos\Flow\ResourceManagement\ResourceManager;
 use Neos\Neos\Domain\Service\ContentContextFactory;
 use Neos\Neos\Domain\Service\ContentDimensionPresetSourceInterface;
 use Neos\Neos\Service\PublishingService;
@@ -156,6 +157,12 @@ class BackendServiceController extends ActionController
      * @var NodeUriPathSegmentGenerator
      */
     protected $nodeUriPathSegmentGenerator;
+
+    /**
+     * @Flow\Inject
+     * @var ResourceManager
+     */
+    protected $resourceManager;
 
     /**
      * Set the controller context on the feedback collection after the controller
@@ -569,5 +576,11 @@ class BackendServiceController extends ActionController
     {
         $slug = $this->nodeUriPathSegmentGenerator->generateUriPathSegment($contextNode, $text);
         $this->view->assign('value', $slug);
+    }
+
+    public function redirectToResourceUriAction(string $resourcePath): void
+    {
+        $target = $this->resourceManager->getPublicPackageResourceUriByPath($resourcePath);
+        $this->redirectToUri($target, 0, 301);
     }
 }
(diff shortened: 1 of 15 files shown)


mhsdesign:
I decided against using a react context to pass the neos ui service redirect uri route along (`/neos/ui-services/redirect-resource-uri`) and instead its now hardcoded. 

With Neos 9 we will establish service routes as part of the dom like so (see https://github.com/neos/neos-ui/pull/3804)

```html
 <link
   id="neos-ui-uri:/neos/xliff.json"
   href="https://mysite.example/neos/xliff.json?locale=de-DE"
   data-locale="de-DE"
   data-locale-plural-rules="one,other"
 >
...
 <link
   id="neos-ui-uri:/neos/redirect-resource-uri"
   href="https://mysite.example/neos/ui-services/redirect-resource-uri"
 >
 ```

to make it more dynamic.

But there is a catch. In local dev mode at least, it seems to take some time until the icons are visible on the first visit: (and somehow my browser didnt cache it after a reload :()

https://github.com/neos/neos-ui/assets/85400359/451f860f-7649-4021-a7bf-d0dc0e53c8c3

Sebobo:
I think we shouldn't introduce additional requests to resolve icons for the default UI.
Rather we should use something like https://github.com/KrakenTyio/esbuild-svg-symbol-plugin to generate svg symbols for the stuff we need and reference them. This way they can be preloaded by the document with one request.
Or keep the SVG itself in React.

Is there a benefit besides resolving some issues with a few icons and the dummy fallback in the image editor?
3rd party icons and images are already resolved in the configuration or not?

mhsdesign:
Option 1 of https://github.com/neos/neos-ui/issues/2092#issuecomment-1906983074 is currently not resolved by the server, but could be. On the otherhand resolving the resource in the server already via a node type preprocessor is not a good idea and will crash in CLI context (which basically forbids doing anything node type related once we use this functionality via cli)

see also https://github.com/neos/neos-development-collection/issues/4888

soo as alternative solution approach i could imagine using indeed inline svgs (like for the neos ui logo) and move the `resource://` replacement logic into the node type schema endpoint instead of a node type preprocessor, as we are in the web context at that time and know the correct host

mhsdesign:
So i experimented a bit and tried out the `file` loader for the Neos logo and the custom icons for the node creation dialog.

The problem with the file loader is that we dont know the correct public base path.

For example if the svg is copied next to the javascripts we should be able to link it via

`http://127.0.0.1:8081/_Resources/Static/Packages/Neos.Neos.Ui/Build/logo-AUKGDP6U.svg`

buuut by default it will link to `./logo-AUKGDP6U.svg` which leads to the icon not being found. Configuring the [`publicPath`](https://esbuild.github.io/api/#public-path) is also not really possible as we dont know if the files are placed in `Neos.Neos.Ui` `Neos.Neos.Ui.Compiled` or located in the cloud (the problem we are initially trying to solve)

Additionally having each svg separate as file is not the best option and we should check into the svg-sprite approach. Though im not sure if we wont face the same `publicPath` dilemma there as well?
Additionally with svg-sprite `xlink:href` cannot point to a different domain, wont that already be a problem for cloud hosted resources without proxy?

On the other-hand we dont need to optimise loading of 4 icons when font-awesome inlines thousands and completely bloats up our bundle:

image
 
So we could decide to fully inline these few icons into the build (either as data-url or as plain text) and for any real optimisations look into how we can make font-awesome awesome? (see https://docs.fontawesome.com/v5/web/advanced/svg-sprites)

mhsdesign:
Seb agreed with my reasoning and well just include our 3 additional svgs in the bundle
See https://github.com/neos/neos-ui/pull/3836

PR #3696, opened 2024-01-24
BUGFIX: Enforce closing dialog

### LINKED ISSUE #3679: BUG: Creation dialog cannot be closed after validation threw an error
### Description

In a creation dialog with properties that have a validator, such as a `NotEmpty` validator, the dialog cannot be closed as soon as validation threw any kind of error.

### Steps to Reproduce

1. Configure a node type to have at least two fields in the creation dialog, at least one of them with a `NotEmpty` validator.
2. Open the creation dialog. Fill in the non-mandatory field and press tab to trigger validation on the mandatory one. You now see the red validation error.
3. Try using the "Back" button to close the creation dialog.

image

#### Expected behavior

The creation dialog closes.

#### Actual behavior

The creation dialog stays open and cannot be escaped unless you reload the whole UI.

### Affected Versions

Neos: 7.3+
UI: 7.3+

markusguenther:
Resolve creation dialog closure issue with property validation errors by introducing a 'force' parameter. The 'close' function now accepts an optional 'force' parameter, defaulting to false, allowing the dialog to be closed even when validation errors occur.

Fixes: #3679

**What I did**
Added a force parameter to the close function of the dialog that is false by default and optional, to be not too breaky.
So we can change the behavior for the KeyboardEvent or click outside.

**How to verify it**
Create a page, for instance in the Neos.Demo and don't enter a title and click create. So you will get a validation error, before the adjustment you cannot close the dialog with ESC or click outside the dialog.
With the PR it will work!

### DIFF
--- a/packages/react-ui-components/src/Dialog/DialogManager.ts
+++ b/packages/react-ui-components/src/Dialog/DialogManager.ts
@@ -4,7 +4,7 @@ export interface EventRoot {
 }
 
 export interface Dialog {
-    close: () => boolean;
+    close: (force?: boolean) => boolean;
 }
 
 export class DialogManager {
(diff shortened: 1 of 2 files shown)


markusguenther:
The only question is if the shaking makes sense because I see no scenario where it will occur.

markusguenther:
But when we change it for Escape, it should be the same for the click outside behavior.

mhsdesign (changes requested):
Sorry no :D 

That basically reverts my efforts in preventing any accidental exit. Thats why the back button should be used? I dont see the bug in https://github.com/neos/neos-ui/issues/3679?


See https://github.com/neos/neos-ui/pull/3579

PR #3697, opened 2024-01-25
TASK: Translate backend messages

pKallert:
**What I did**
In https://github.com/neos/neos-ui/pull/3665 we added the Translation trait to the Backend Service Controller. 
Currently, only one message is translated in the backend, but we should translate all messages.


Resources/Private/Translations/en/Main.xlf
```diff
@@ -326,6 +326,14 @@
                     <source>Discarded {0} changes.</source>
                 </trans-unit>
             </group>
+            <trans-unit id="switchedBaseWorkspace" xml:space="preserve">
+                <source>Switched base workspace to "{workspace}".</source>
```
bwaidelich, line 330: Didn't check and I assume you _did_, but just to make sure: `{workspace}` as placeholder works here? (I only knew that we support numeric placeholders)

bwaidelich (approved):
+1 by reading

dlubitz (approved):
👍 Looks also good to me by reading.
(if `{workspace}` really works as placeholder 🤞 )

crydotsnake (approved):
👁️

Resources/Private/Translations/en/Main.xlf
pKallert: Tried it out again to confirm and it looks good: 
Bildschirmfoto 2024-06-14 um 15 26 38

Issue #3664, opened 2023-11-14
Non-scrollable Neos.Neos/Inspector/Editors/RichTextEditor


auwaerter:
### Description

Within my Neos instance the [Neos.Neos/Inspector/Editors/RichTextEditor](https://neos.readthedocs.io/en/stable/References/PropertyEditorReference.html#property-type-string-richtexteditor-full-screen-rich-text-editor) seems to be non-scrollable when i'm having words in an amount that overlaps the browser window. 

I'm using the following workaround(s) to apply changes:
- Click outside of the RichTextEditor, e.g. to another item within the Content Tree. This gives you the "You still have changes. What do you want to do with them?" dialog, where you can "Throw away" or "Apply" your modifications.
- Move down with the arrow down key, but for that you barley can click at the Discard/Apply button, as only 1/4 of it is shown.

I'm using the following NodeType definition (same as the link above):

```yaml
'Super.Vendor:Content.Text':
  superTypes:
    'Neos.Neos:Content': true
    'Super.Vendor:Constraint.Content.Carousel': true
    'Super.Vendor:Constraint.Content.Column': true
    'Super.Vendor:Constraint.Content.Main': true
    'Super.Vendor:Constraint.Content.Teaser': true
    'Super.Vendor:Constraint.Content.Footer': true
  ui:
    label: i18n
    icon: file-text
    position: 20
    inspector:
      groups:
        settings:
          label: 'Settings'
  properties:
    'source':
      type: 'string'
      ui:
        label: 'Toggle the editor'
        reloadIfChanged: true
        inspector:
          group: 'settings'
          editor: 'Neos.Neos/Inspector/Editors/RichTextEditor'
          editorOptions:
            placeholder: '<p>Text</p>'
            autoparagraph: true
            linking:
              anchor: true
              title: true
              relNofollow: true
              targetBlank: true
              download: true
            formatting:
              strong: true
              em: true
              u: true
              sub: true
              sup: true
              del: true
              p: true
              h1: true
              h2: true
              h3: true
              h4: true
              h5: true
              h6: true
              pre: true
              underline: true
              strikethrough: true
              removeFormat: true
              left: true
              right: true
              center: true
              justify: true
              table: true
              ol: true
              ul: true
              a: true
```

Question: Is this a bug or a mistake (in e.g. the editorOptions, where it configures the non-scrolling part) i've did somewhere?

### Steps to Reproduce

1. Use Google Chrome Version 119.0.6045.124
[...]

PR #3699, opened 2024-01-26
BUGFIX: Prevent overlaying of upper toolbars in RichTextEditor

markusguenther:
The `RichTextEditor`, utilizing the `CKEditor` component, experienced an issue where the wrapping component failed to scroll appropriately with increased content, causing it to grow and overlay the upper toolbar. This commit addresses the problem, ensuring proper container resizing and preventing toolbar overlay.

Fixes: #3664

https://github.com/neos/neos-ui/assets/1014126/9d394351-3a94-4098-8c23-8f4b3989ec0a


Sebobo (approved):
Works as expected, thx!

PR #3700, opened 2024-01-26
BUGFIX: Add ellipsis to nodeInfoView content

### LINKED ISSUE #3637: NodeTypeName is cut off in info panel
### Description

image

while this is a silly example, there are legit cases for long nodeType names (for example if the vendor prefix is not small and its a sub component of something)

We should make this field scrollable

image

markusguenther:
The NodeTypeName in the info panel was getting cut off, so I implemented a solution to enhance the display. Now, an ellipsis is added to prevent overflow within the container, and the full NodeType name is accessible through a tooltip that appears when hovered over.

Fixes: #3637

Screenshot 2024-01-26 at 12 38 07

### DIFF
--- a/packages/neos-ui-views/src/NodeInfoView/style.css
+++ b/packages/neos-ui-views/src/NodeInfoView/style.css
@@ -29,5 +29,8 @@
     &__content {
         font-size: 13px;
         color: var(--colors-ContrastBright);
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
     }
 }


markusguenther:
We can also think about an improvement to be able to get the NodeType name to the clipboard by an extra button or just clicking it. But that's more a feature then.

mhsdesign (dismissed):
Thanks, but seboo fixed it differently for 8.3 https://github.com/neos/neos-ui/pull/3662

its still not perfect, but i wouldnt change legacy behavour.

markusguenther:
Thank you for your feedback. I appreciate your concern about the consistency of the solution across different branches. However, I would like to clarify that my PR does not introduce a behavior change, but rather a visual improvement. The text in the nodetype name info view was floating out of the container and looking broken, which was not a good user experience. My PR adds an ellipsis to indicate that there is more text available, and the popup with the full name was always there on hover. This is a common UI pattern that is used in many applications. I hope this explains my reasoning behind the PR. Please let me know if you have any questions or suggestions. The change from @sebobo was not linked in the issue btw ;)

Sebobo:
We just have to adjust the change during upmerge as this one will break the intention of my change

mhsdesign:
Thanks for your in depth elaboration. As it seems that seb is fine change as well I won’t be blocking this any further.
And I appreciate it a lot that you put so much love into the final 7.3 release. Please forgive my hard criticism for any last minute changes 😬^^

(and yes unfortunately sebs fix was not linked in my mentioned issue above. I’m sorry for that, as it caused additional effort on your end)

Sebobo (approved):
It's fine for 7.3 and works as expected. But please make sure to verify the behaviour of my change in 8.3 when upmerging.

bwaidelich:
Hi all,
I have noticed that node type names contain weird whitespace characters between the segments when copied from the `NodeInfoView`, e.g. `Some.SHYPackage:SHYDocument.SHYFoo`. Could this be related to this change, @markusguenther ?

Sebobo:
@bwaidelich a fix of mine causes this, I know of a better solution now and will create a new PR soonish

bwaidelich:
@Sebobo ah, thx for the info. Could you share a link to the change with the "regression"?
I stumble upon this a lot, e.g. when filtering the db for nodes of a type because those whitespaces are not directly visible

Sebobo:
@bwaidelich https://github.com/neos/neos-ui/pull/3662
The improved solution just needs a `<wbr>` instead of the shy character, from my tests this then fulfils both readability and copy&paste requirements.

bwaidelich:
@Sebobo great, thanks for the info

PR #3701, opened 2024-01-26
BUGFIX: Resolve issue with custom data source displaying additional l…

### LINKED ISSUE #3675: BUG: Custom Data Source preview with label and secondaryLabel displays wrong
Hello
I noticed something with custom data source, that it displays the secondary label with use of label and preview not correctly
Bildschirm_foto 2023-12-15 um 13 44 27

See discussion on Neos Slack: https://neos-project.slack.com/archives/C0U0KEGDQ/p1702645369911609

markusguenther:
Previously, there was an issue with the custom data source where the secondary label, utilized with label and preview, was not being displayed correctly. This commit addresses the problem, ensuring proper alignment and display of the secondary label in conjunction with the label and preview elements. The implementation includes the use of a CSS grid layout to achieve consistent and accurate rendering of the custom data source information.

Fixes: #3675

**Before:**
Screenshot 2024-01-26 at 13 28 55

**After with preview:**
Screenshot 2024-01-26 at 14 13 09

**After without preview and 3rd label:**
Screenshot 2024-01-26 at 14 17 11

### DIFF
--- a/packages/react-ui-components/src/SelectBox_Option_MultiLineWithThumbnail/style.css
+++ b/packages/react-ui-components/src/SelectBox_Option_MultiLineWithThumbnail/style.css
@@ -1,6 +1,23 @@
 .multiLineWithThumbnail__item {
+    display: grid;
+    grid-template-columns: auto 1fr;
+    gap: -var(--spacing-Half);
+    align-items: start;
     box-sizing: content-box;
     background-color: var(--colors-ContrastDarkest);
+}
+
+.multiLineWithThumbnail__item span {
+    grid-column: 2;
+    grid-row: 1;
+
+    &.multiLineWithThumbnail__secondaryLabel {
+        grid-row: 2;
+    }
+
+    &.multiLineWithThumbnail__tertiaryLabel {
+        grid-row: 3;
+    }
 }
 
 .multiLineWithThumbnail__item--multiLine {
@@ -26,6 +43,8 @@
 .multiLineWithThumbnail__image {
     width: calc(1.33 * (var(--spacing-GoldenUnit) + var(--spacing-Full)));
     height: calc(var(--spacing-GoldenUnit) + var(--spacing-Full));
+    grid-column: 1;
+    grid-row: 1 / 4;
     object-fit: contain;
     background-color: #fff;
     background-size: 10px 10px;
--- a/packages/react-ui-components/src/SelectBox_Option_MultiLineWithThumbnail/style.css
+++ b/packages/react-ui-components/src/SelectBox_Option_MultiLineWithThumbnail/style.css
@@ -1,7 +1,7 @@
 .multiLineWithThumbnail__item {
     display: grid;
     grid-template-columns: auto 1fr;
-    gap: -var(--spacing-Half);
+    column-gap: var(--spacing-Half);
     align-items: start;
     box-sizing: content-box;
     background-color: var(--colors-ContrastDarkest);
@@ -10,6 +10,9 @@
 .multiLineWithThumbnail__item span {
     grid-column: 2;
     grid-row: 1;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
 
     &.multiLineWithThumbnail__secondaryLabel {
         grid-row: 2;
@@ -50,12 +53,10 @@
     background-size: 10px 10px;
     background-position: 0 0, 25px 25px;
     background-image: linear-gradient(45deg, #cccccc 25%, transparent 25%, transparent 75%, #cccccc 75%, #cccccc), linear-gradient(45deg, #cccccc 25%, transparent 25%, transparent 75%, #cccccc 75%, #cccccc);
-    margin-right: .75em;
     display: inline-block;
     vertical-align: middle;
-    margin-left: -var(--spacing-Full);
-    margin-top: -var(--spacing-Half);
-    margin-bottom: -var(--spacing-Half);
+    margin-left: -var(--spacing-Half);
+    align-self: center;
 }
 
 .multiLineWithThumbnail__item:hover .multiLineWithThumbnail__secondaryLabel,


markusguenther:
For easier testing here my DataSource:

```php
<?php

namespace Neos\Demo\Service\DataSource;


use Neos\Flow\Annotations as Flow;
use Neos\Flow\Persistence\PersistenceManagerInterface;
use Neos\Neos\Domain\Service\UserService;
use Neos\Neos\Service\DataSource\AbstractDataSource;

class EditorsDataSource extends AbstractDataSource
{

    /**
     * @var string
     */
    static protected $identifier = 'neos-demo-editors';

    /**
     * @Flow\Inject
     * @var UserService
     */
    protected $userService;

    /**
     * @Flow\Inject
     * @var PersistenceManagerInterface
     */
    protected $persistenceManager;

    /**
     * @param NodeInterface $node The node that is currently edited (optional)
     * @param array $arguments Additional arguments (key / value)
     * @return array
     */
    public function getData(NodeInterface $node = null, array $arguments = [])
    {
        $options = [];
        foreach ($this->userService->getUsers() as $user) {
			$options[] = [
                'value' => $this->persistenceManager->getIdentifierByObject($user),
                'label' => $user->getLabel(),

                // additional optional parameters:
                'secondaryLabel' => "works here",
                'tertiaryLabel' => "is a nice person",
                'preview' => 'data:image/svg+xml;base64,PHN2ZyB4bWxucz0naHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmcnIHZpZXdCb3g9JzAgMCA0NDggNTEyJz48IS0tIUZvbnQgQXdlc29tZSBGcmVlIDYuNS4xIGJ5IEBmb250YXdlc29tZSAtIGh0dHBzOi8vZm9udGF3ZXNvbWUuY29tIExpY2Vuc2UgLSBodHRwczovL2ZvbnRhd2Vzb21lLmNvbS9saWNlbnNlL2ZyZWUgQ29weXJpZ2h0IDIwMjQgRm9udGljb25zLCBJbmMuLS0+PHBhdGggZD0nTTIyNCAxNmMtNi43IDAtMTAuOC0yLjgtMTUuNS02LjFDMjAxLjkgNS40IDE5NCAwIDE3NiAwYy0zMC41IDAtNTIgNDMuNy02NiA4OS40QzYyLjcgOTguMSAzMiAxMTIuMiAzMiAxMjhjMCAxNC4zIDI1IDI3LjEgNjQuNiAzNS45Yy0uNCA0LS42IDgtLjYgMTIuMWMwIDE3IDMuMyAzMy4yIDkuMyA0OEg0NS40QzM4IDIyNCAzMiAyMzAgMzIgMjM3LjRjMCAxLjcgLjMgMy40IDEgNWwzOC44IDk2LjlDMjguMiAzNzEuOCAwIDQyMy44IDAgNDgyLjNDMCA0OTguNyAxMy4zIDUxMiAyOS43IDUxMkg0MTguM2MxNi40IDAgMjkuNy0xMy4zIDI5LjctMjkuN2MwLTU4LjUtMjguMi0xMTAuNC03MS43LTE0M0w0MTUgMjQyLjRjLjYtMS42IDEtMy4zIDEtNWMwLTcuNC02LTEzLjQtMTMuNC0xMy40SDM0Mi43YzYtMTQuOCA5LjMtMzEgOS4zLTQ4YzAtNC4xLS4yLTguMS0uNi0xMi4xQzM5MSAxNTUuMSA0MTYgMTQyLjMgNDE2IDEyOGMwLTE1LjgtMzAuNy0yOS45LTc4LTM4LjZDMzI0IDQzLjcgMzAyLjUgMCAyNzIgMGMtMTggMC0yNS45IDUuNC0zMi41IDkuOWMtNC44IDMuMy04LjggNi4xLTE1LjUgNi4xem01NiAyMDhIMjY3LjZjLTE2LjUgMC0zMS4xLTEwLjYtMzYuMy0yNi4yYy0yLjMtNy0xMi4yLTctMTQuNSAwYy01LjIgMTUuNi0xOS45IDI2LjItMzYuMyAyNi4ySDE2OGMtMjIuMSAwLTQwLTE3LjktNDAtNDBWMTY5LjZjMjguMiA0LjEgNjEgNi40IDk2IDYuNHM2Ny44LTIuMyA5Ni02LjRWMTg0YzAgMjIuMS0xNy45IDQwLTQwIDQwem0tODggOTZsMTYgMzJMMTc2IDQ4MCAxMjggMjg4bDY0IDMyem0xMjgtMzJMMjcyIDQ4MCAyNDAgMzUybDE2LTMyIDY0LTMyeicvPjwvc3ZnPg=='
            ];
        }
        return $options;
    }
}

```

And the nodetype stuff:
```yaml
properties:
    authors:
      type: array
      ui:
        label: 'Author(s)'
        reloadIfChanged: true
        inspector:
          group: 'document'
          position: '200'
          editorOptions:
            placeholder: Choose
            dataSourceIdentifier: neos-demo-editors
```

packages/react-ui-components/src/SelectBox_Option_MultiLineWithThumbnail/style.css
```diff
@@ -1,8 +1,25 @@
 .multiLineWithThumbnail__item {
+    display: grid;
+    grid-template-columns: auto 1fr;
+    gap: -var(--spacing-Half);
+    align-items: start;
     box-sizing: content-box;
     background-color: var(--colors-ContrastDarkest);
 }
 
+.multiLineWithThumbnail__item span {
```
Sebobo, line 10: A looong tittle makes the whole element wider and the ellipsis of the secondary and tertiary labels disappear.

packages/react-ui-components/src/SelectBox_Option_MultiLineWithThumbnail/style.css
```diff
@@ -1,8 +1,25 @@
 .multiLineWithThumbnail__item {
+    display: grid;
+    grid-template-columns: auto 1fr;
+    gap: -var(--spacing-Half);
```
Sebobo: A negative gap is marked as invalid value and also doesn't seem to have any effect.

The styling would also be simpler when the gap is actually used and the padding of the element disabled and therefore the image correctly positioned (also the image margin needs to be disabled.

The result is looking the same but without all the CSS pushing around.

Sebobo (changes requested):
Thx, looks better (though I don't like the image hanging loose there, which is not a problem of this PR), but this change needs a cleaner CSS implementation.

markusguenther:
Thanks for the valuable feedback. With very long labels, it looked like that:
Screenshot 2024-02-02 at 13 35 23
Screenshot 2024-02-02 at 13 35 12

So the idea was to break the text, but it leads to huge heights.
Screenshot 2024-02-02 at 13 31 58

Now we add an ellipsis and, like in the `NodeInfoView` we have a tooltip by default.

**Safari:**
Screenshot 2024-02-02 at 14 18 28
**Chrome:**
Screenshot 2024-02-02 at 14 17 30
**Firefox:**
Screenshot 2024-02-02 at 14 17 18

packages/react-ui-components/src/SelectBox_Option_MultiLineWithThumbnail/style.css
```diff
@@ -26,17 +46,17 @@
...
     background-size: 10px 10px;
     background-position: 0 0, 25px 25px;
     background-image: linear-gradient(45deg, #cccccc 25%, transparent 25%, transparent 75%, #cccccc 75%, #cccccc), linear-gradient(45deg, #cccccc 25%, transparent 25%, transparent 75%, #cccccc 75%, #cccccc);
-    margin-right: .75em;
     display: inline-block;
     vertical-align: middle;
-    margin-left: -var(--spacing-Full);
-    margin-top: -var(--spacing-Half);
-    margin-bottom: -var(--spacing-Half);
+    margin-left: -var(--spacing-Half);
```
Sebobo, line 58: Not so happy about still having a negative margin, but I think its fine for now.

Sebobo (approved):
Great much better!

packages/react-ui-components/src/SelectBox_Option_MultiLineWithThumbnail/style.css
markusguenther: Yeah, but in the select boxes I am always afraid that I break something else 🙈

PR #3703, opened 2024-01-30
TASK: Avoid heavily use of controller context and make `Neos.Ui.NodeInfo` internal

mhsdesign:
_If_ for some reason `Neos.Ui.NodeInfo` was used in consumer fusion land it must be migrated.
For migration please copy the parts of the `NodeInfo` helper to your own helper.
Note that the helper is sorely an internal detail of the user interface.

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/Classes/Fusion/Helper/NodeInfoHelper.php
+++ b/Classes/Fusion/Helper/NodeInfoHelper.php
@@ -31,6 +31,7 @@ use Neos\Neos\Ui\Domain\Service\UserLocaleService;
 use Neos\Neos\Utility\NodeTypeWithFallbackProvider;
 
 /**
+ * @internal
  * @Flow\Scope("singleton")
  * @todo EEL helpers are still to be declared as internal
  */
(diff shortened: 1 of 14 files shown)


Classes/Fusion/Helper/NodeInfoHelper.php
```diff
@@ -30,6 +30,7 @@
 use Neos\Neos\Utility\NodeTypeWithFallbackProvider;
 
 /**
+ * @internal
  * @Flow\Scope("singleton")
  * @todo EEL helpers are still to be declared as internal
```
grebaldi, line 35:
```suggestion
```

This should go, since `NodeInfoHelper` is marked `@internal` now in L33.

grebaldi (approved):
Hi @mhsdesign,

nice work :) A left a tiny nitpicky comment, but the rest LGTM :+1:

Classes/Fusion/Helper/NodeInfoHelper.php
grebaldi: **UPDATE:** Just saw that #3711 contains all changes of #3703 and removes the `@todo` annotation. Maybe #3711 can just supersede #3703, so #3703 can be closed? wdyt?

mhsdesign: a i forgot how i did it, but yes this pr is basically step one and later we followup with https://github.com/neos/neos-ui/pull/3711 so ill merge it first ;)
Thanks for looking so deep into it :D

PR #3704, opened 2024-01-30
TASK: Declare public api in php classes

mhsdesign:
Depends on https://github.com/neos/neos-ui/pull/3682/ which must be merged first.

I recently had someone ask me how to use this `Property` thing in Neos9. At first i was confused. And then i understood he was trying to use the Neos.Ui "change" classes.
To avoid that in the future and make it also easier to clean up our php mess in the Neos UI ^^ i placed a few internal annotations around.

and the changes now explicitly warn from being used :D (which is spoiler really hard anyway ^^ - otherwise we would have tests *g)

```php
/**
 * Changes a property on a node
 * @internal These objects internally reflect possible operations made by the Neos.Ui.
 *           They are sorely an implementation detail. You should not use them!
 *           Please look into the php command API of the Neos CR instead.
 */
class Property extends AbstractChange
```

_If_ you happen to use any class from the Neos Ui previously (which should not be the case, as the ui has nothing to offer for php users ^^), which is not declared as api anymore, you should please prepare to refactor it.

**What I did**
 

**How I did it**

**How to verify it**

### DIFF
--- a/Classes/Fusion/Helper/WorkspaceHelper.php
+++ b/Classes/Fusion/Helper/WorkspaceHelper.php
@@ -22,6 +22,7 @@ use Neos\Neos\Ui\ContentRepository\Service\WorkspaceService;
 
 /**
  * The Workspace helper for EEL contexts
+ * @internal
  */
 class WorkspaceHelper implements ProtectedContextAwareInterface
 {
--- a/Classes/Fusion/Helper/WorkspaceHelper.php
+++ b/Classes/Fusion/Helper/WorkspaceHelper.php
@@ -22,7 +22,7 @@ use Neos\Neos\Ui\ContentRepository\Service\WorkspaceService;
 
 /**
  * The Workspace helper for EEL contexts
- * @internal
+ * @todo EEL helpers are still to be declared as internal
  */
 class WorkspaceHelper implements ProtectedContextAwareInterface
 {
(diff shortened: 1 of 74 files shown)


Classes/NodeCreationHandler/NodeCreationHandlerInterface.php
```diff
@@ -17,6 +17,7 @@
 /**
  * Contract for Node Creation handler that allow to hook into the process just before a node is being added
  * via the Neos UI
+ * @api
```
mhsdesign, line 20: to be transparent with you ;) that is the only php class/interface in the neos.ui which deserves an `@api` notation. Any objections? Or do you need another thing to be declared `@api` as well?

Classes/Fusion/Helper/WorkspaceHelper.php
```diff
@@ -22,6 +22,7 @@
 
 /**
  * The Workspace helper for EEL contexts
+ * @todo EEL helpers are still to be declared as internal
```
markusguenther, line 25: And why you did not declare it ;)

markusguenther (approved):
Looks good to me, and thanks for going through all the files :)

Classes/Fusion/Helper/WorkspaceHelper.php
mhsdesign: because this would be to controversy to declare that in this pr. Will be discussed in a followup if the helper was actually useful in some obscure case.

PR #3705, opened 2024-01-31
TASK: Upgrade Node.js to version 20 LTS

markusguenther:
Node.js 16 has reached its end of life, and even version 18 is now in maintenance phase. This PR updates the Node.js version to 20, ensuring we benefit from the latest updates.


mhsdesign:
but as _that_ only dev tooling i would like to target 8.3 so we dont have to switch locally that often :O

Sorry i must confuse you with my principals of when to target what i hope they make sense? 😂

markusguenther:
> but as _that_ only dev tooling i would like to target 8.3 so we dont have to switch locally that often :O
> 
> Sorry i must confuse you with my principals of when to target what i hope they make sense? 😂

I never switch manually :P
May Terminal is looking for the version and switches via NVM :)

But can target 8.3.

mhsdesign (approved):
Thanks for taking care ^^

PR #3707, opened 2024-01-31
TASK: Upgrade testcafe to the latest version

markusguenther:
Upgrade the end to end dependencies :
- testcafe to version 3.5
- testcafe-browser-provider-saucelabs to version 3.0
- testcafe-react-selectors to 5.0

As we increase the Node.js version to the new LTS (20.9) we should also upgrade testcafe, as version 2.3.1 was the last version for node 16 and is end of life in April 2024. (we used version 2.1.0)

- [x]  Upgrade dependencies
- [ ]  Need to change the SauceLabs integration
- [ ] Check current Tests


markusguenther:
This PR is based upon the Node.js upgrade. #3705

mhsdesign:
_if_ possible i'd like to target here 8.3 as well. This would make our lives much easier in the end :D

mhsdesign:
Partly obsolete with https://github.com/neos/neos-ui/pull/3850

PR #3710, opened 2024-02-03
TASK: Fix phpstan error and add `composer run lint:phpstan`

mhsdesign:
**What I did**

**How I did it**

**How to verify it**


markusguenther (approved):
Looks good to me :)

PR #3713, opened 2024-02-06
BUGFIX: Neos redirection issue in subfolder configuration

markusguenther:
Addressing an issue where the redirection to `/neos` fails when Neos runs in a subfolder. This commit introduces a new route to the `defaultModule`. The index action of the backend controller is responsible for ensuring the correct backend module. Additionally, consideration for the `preferredStartModules` configuration has been incorporated here.

Fixes: https://github.com/neos/neos-development-collection/issues/4797

### DIFF
--- a/packages/neos-ui-backend-connector/src/index.ts
+++ b/packages/neos-ui-backend-connector/src/index.ts
@@ -32,6 +32,7 @@ export const initializeJsAPI = (parent: {[propName: string]: any}, {alias = 'neo
     addLibrary('use', initializeUse(addLibrary, neos));
     addLibrary('q', initializeFlowQuery(routes));
     addLibrary('endpoints', initializeEndpoints(routes));
+    addLibrary('routes', routes);
 
     //
     // Attach Neos API to the parent object
--- a/packages/neos-ui-backend-connector/src/index.ts
+++ b/packages/neos-ui-backend-connector/src/index.ts
@@ -32,7 +32,6 @@ export const initializeJsAPI = (parent: {[propName: string]: any}, {alias = 'neo
     addLibrary('use', initializeUse(addLibrary, neos));
     addLibrary('q', initializeFlowQuery(routes));
     addLibrary('endpoints', initializeEndpoints(routes));
-    addLibrary('routes', routes);
 
     //
     // Attach Neos API to the parent object
--- a/packages/neos-ui-sagas/src/UI/Impersonate/index.js
+++ b/packages/neos-ui-sagas/src/UI/Impersonate/index.js
@@ -38,7 +38,8 @@ export function * impersonateRestore({globalRegistry}) {
             } else {
                 yield put(actions.UI.FlashMessages.add('restoreUserImpersonateUser', errorMessage, 'error'));
             }
-            window.location.pathname = '/neos';
+
+            window.location.href = $get('core.modules.defaultModule', window.neos.routes);
         } catch (error) {
             yield put(actions.UI.FlashMessages.add('restoreUserImpersonateUser', errorMessage, 'error'));
         }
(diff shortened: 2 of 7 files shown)


markusguenther:
If this is ok, we can maybe also use it here in the neos-ui manifest
```
// This is an extreme case when even the top node does not exist in the given dimension
                // TODO: still find a nicer way to break out of this situation
                if (!redirectContextPath) {
                    window.location = '/neos';
                    break;
                }
```

ahaeslich (approved):
Tested in project with subfolder configuration. Works like a charm. Thx for taking care 💖

ahaeslich:
We will have to make sure this is not lost during upmerges as of the changes in https://github.com/neos/neos-ui/pull/3682/.

packages/neos-ui-backend-connector/src/index.ts
```diff
@@ -32,6 +32,7 @@ export const initializeJsAPI = (parent: {[propName: string]: any}, {alias = 'neo
     addLibrary('use', initializeUse(addLibrary, neos));
     addLibrary('q', initializeFlowQuery(routes));
     addLibrary('endpoints', initializeEndpoints(routes));
+    addLibrary('routes', routes);
```
mhsdesign: i dont think exposing this and possibly increasing our surface of the experimental extensibility api makes sense.
I can only assume that the routes were explicitly no part of `window` and that we should always `yield select()` the state.

mhsdesign (dismissed):
yes and using this in the other `/neos` spot also makes absolutely sense.

packages/neos-ui-backend-connector/src/index.ts
markusguenther: Hmm, ok but declaring the `defaultModule` as endpoint also sounds bad and something like `$get('routes.core.modules.defaultModule', neos)` is not working in the saga. So I need to check if I can use it somehow in the saga.

mhsdesign: hmm i see the `routes` are indeed directly passed to `<Root />` which creates a react context but we cant use those in redux.

In that case i would say we either have to store the value in redux which makes no sense right? Or we store it in a value in a module. And refactor the code later. That will make the code smell obvious without using `window` and exposing too much.

markusguenther: Stupid question: Would you save all routes to the state, or just the single one? Think about bringing the routes to the system state. So we would have next to the `system.authenticationTimeout` also `system.routes`.

Just would ask for the intention before building it and then reverting it again ;)

mhsdesign: Il have to ask wilhelm, to hard the question is

mhsdesign: he solved the puzzle :D We can add the `routes` via sage middleware run as initial parameter to each saga.

SEE https://github.com/neos/neos-ui/pull/3713/commits/bf48d911d5076c58199b8865a93ef713e6faecf1

markusguenther:
Hmm, now the end-to-end tests are not working anymore 🙈

mhsdesign:
sorry :D

packages/neos-ui-sagas/src/UI/Impersonate/index.js
```diff
@@ -4,7 +4,7 @@ import {actionTypes, actions} from '@neos-project/neos-ui-redux-store';
 import backend from '@neos-project/neos-ui-backend-connector';
 import {$get} from 'plow-js';
 
-export function * impersonateRestore({globalRegistry}) {
+export function * impersonateRestore({globalRegistry, routes}) {
```
markusguenther, line 7: At this point, routes is undefined and therefore the whole UI is not working. For the impersonation Saga it works here we need another approach.

markusguenther:
```js
bootstrap(globalRegistry, {store, frontendConfiguration, configuration, routes}))
```

That's the reason why it fails ;)

mhsdesign: oh sorry 😅

PR #3714, opened 2024-02-07
Bugfix/3477 infinite loading indicator

### LINKED ISSUE #3477: BUG: Occasions of the loading indicator running in an endless loop
### Description

There are a couple of occasions in which the loading indicator doesn't notice that some async process has failed and just ends up looping endlessly.

Several older issues mention this phenomenon:
- https://github.com/neos/neos-ui/issues/2590#issuecomment-1531490615
- https://github.com/neos/neos-ui/issues/1718
- https://github.com/neos/neos-ui/issues/1664
- https://github.com/neos/neos-ui/issues/2358
- https://github.com/neos/neos-development-collection/issues/4121

#### Repro 1

1. Open a page in the backend
2. Run `./flow session:destroyall`
3. Navigate to a different page in the backend, so the login dialog shows up
4. See that the loading indicator runs in an endless loop

#### Repro 2

1. Render a page with just one `Neos.Fusion:Debug` output like this:
```
prototype(Vendor.Site:Document.Page) < prototype(Neos.Fusion:Component) {
    renderer = Neos.Fusion:Debug {
        foo = 'bar'
    }
}
```
2. Open that page in the backend
3. See that the loading indicator runs in an endless loop

#### Repro 3

1. Provoke an error by rendering a page like this:
```
prototype(Vendor.Site:Document.Page) < prototype(Neos.Fusion:Component) {
    ren/derer = 'foo'
}
```
2. Open that page in the backend
3. See an error message an the loading indicator running in an endless loop

### Expected behavior

The loading indicator should not run in an endless loop, but stop when an error occurs.

### Actual behavior

Peek 2023-05-02 15-44

### Affected Versions

Neos: >= 7.3 (and older)

UI: >= 7.3 (and older)

mhsdesign:
Resolves: #3477

Alternate fix replaces https://github.com/neos/neos-ui/pull/2566 which is good and we need to make the page handler non neos ui specific so it can go "back" to neos

The initial idea to spin until the ui informs to be ready was that if possibly a lot of ckeditors have to be initialised and that the content manager should wait until the indicator stops spinning before clicking around.
IN my problems i could not verify that and as this is just an indicator it does no harm stopping a few milliseconds earlier. It never prevented anything or is used for anything critical.
In fact the infinite spinning indicator is much worse when developing and debugging 

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/packages/neos-ui-redux-store/src/UI/ContentCanvas/index.ts
+++ b/packages/neos-ui-redux-store/src/UI/ContentCanvas/index.ts
@@ -17,7 +17,6 @@ export interface State extends Readonly<{
     formattingUnderCursor: Formatting;
     currentlyEditedPropertyName: string | null;
     isLoading: boolean;
-    loadingFailed: boolean;
     focusedProperty: string | null;
     backgroundColor: string;
     shouldScrollIntoView: boolean;
@@ -30,7 +29,6 @@ export const defaultState: State = {
     formattingUnderCursor: {},
     currentlyEditedPropertyName: null,
     isLoading: true,
-    loadingFailed: false,
     focusedProperty: null,
     backgroundColor: '#ffffff',
     shouldScrollIntoView: false,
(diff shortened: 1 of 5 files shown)


mhsdesign:
@73nici

PR #3716, opened 2024-02-10
BUGFIX: Explicit value object support for node properties

mhsdesign:
Previously this logic was triggered implicitly though the property mapper, but as this is such a primary use case and requires special behavior in the serializer: https://github.com/neos/neos-development-collection/pull/4638 there should be an obvious code flow.

This change does not modify behavior but equals having the `DenormalizingObjectConverter` priority set to infinity :D


markusguenther (approved):
Looks good by reading

Sebobo (commented):
How can I test or verify this?

PR #3718, opened 2024-02-15
TASK: Use `docker-compose` rather than `docker compose`

grebaldi:
I've been using `start-neos-dev-instance.sh` for local E2E tests as of late and stumbled upon an issue.

The `Tests/IntegrationTests/start-neos-dev-instance.sh` script refers to docker-compose thusly:
https://github.com/neos/neos-ui/blob/fe77db35c48e886b7a0421ccf3e2f0da6d098e52/Tests/IntegrationTests/start-neos-dev-instance.sh#L5-L8

Now, the comment `# use the docker composer plugin` sounds to me like a misconception, but apart from that there's the rather unusual `docker compose` call (without dash).

I suppose that in some environments docker-compose works like a docker plugin, so that it is available as a subcommand of the docker cli. But usually it is invoked via its own CLI binary `docker-compose`.

Since `docker compose` stopped working on my machine (though I'm not sure whether it ever did work), I'd like to suggest to change this call to `docker-compose`, which *should* work everywhere.

### DIFF
--- a/Tests/IntegrationTests/start-neos-dev-instance.sh
+++ b/Tests/IntegrationTests/start-neos-dev-instance.sh
@@ -4,7 +4,7 @@ set -e
 
 function dc() {
     # use the docker composer plugin
-    docker compose -f ./Tests/IntegrationTests/docker-compose.neos-dev-instance.yaml $@
+    docker-compose -f ./Tests/IntegrationTests/docker-compose.neos-dev-instance.yaml $@
 }
 
 echo "#############################################################################"


mhsdesign:
The comment has to be my foult, though it was brought out of context :D https://github.com/neos/neos-ui/pull/3369/commits/02c788669641c58228b1f4d8cc658851c847b486

I remember changing this like you suggested, but i think this would be the legacy syntax (with dash) ... and robert and me figured that it works without, thus it was removed again ...

all part of this squashed pr: https://github.com/neos/neos-ui/pull/3369

kdambekalns (changes requested):
`docker compose` is the proper way to do it. See https://docs.docker.com/compose/reference/ and https://docs.docker.com/compose/migrate/

grebaldi:
Yet it doesn't work as expected:

```sh
> docker --version
Docker version 24.0.5, build 24.0.5-0ubuntu1~22.04.1

> docker compose
docker: 'compose' is not a docker command.
See 'docker --help'
```

I'm pretty sure this came with an update. Maybe a problem with `docker` itself?

`docker-compose` works without a problem :shrug:

grebaldi:
**UPDATE:**

I found the issue. On Ubuntu, you need to install an additional package for this to work:

```sh
sudo apt install docker-compose-v2
```

source: https://github.com/docker/compose/issues/8630#issuecomment-1825685695

kdambekalns:
With Homebrew I once did this, when I dumped Docker Desktop:

```
mkdir -p ~/.docker/cli-plugins
ln -sfn /opt/homebrew/opt/docker-compose/bin/docker-compose ~/.docker/cli-plugins/docker-compose
```

PR #2596, opened 2019-11-28
FEATURE: Node Creation Handler for the "CreationDialogProperties"

bwaidelich:
Since the interface for creation handlers is defined in the Neos.Neos.Ui
package this change is required for neos/neos-development-collection#2785
to work properly.

Related: neos/neos-development-collection#2173

PR #3720, opened 2024-02-16
BUGFIX Fix conversion in CreationDialogPropertiesCreationHandler

### LINKED ISSUE #3719: Inconsistent conversion in `CreationDialogPropertiesCreationHandler`
### Description

The `CreationDialogPropertiesCreationHandler` does property value conversion from "raw" to "typed" on it's own, and in a different way compared to the `NodeConverter` from the Content Repository or even the `NodePropertyConversionService` from the UI itself.

### Steps to Reproduce

1. Create a nodetype with a datetime property
2. Create a node using  the creation dialog
3. Create such a node without the creation dialog
4. Compare the stored dattime values in the database

#### Expected behavior

The same timezone handling is applied, dates are stored in the server's default timezone.

#### Actual behavior

Dates are stored in the users' timezone, as sent by the browser.

### Affected Versions

UI: 8.3 (and earlier)

kdambekalns:
This makes the `CreationDialogPropertiesCreationHandler` use the `convert()` method from the `NodePropertyConversionService` to provide consistent behaviour.

**How to verify it**

See #3719 for steps to reproduce

### DIFF
--- a/Classes/NodeCreationHandler/CreationDialogPropertiesCreationHandler.php
+++ b/Classes/NodeCreationHandler/CreationDialogPropertiesCreationHandler.php
@@ -37,7 +37,7 @@ class CreationDialogPropertiesCreationHandler implements NodeCreationHandlerInte
                 continue;
             }
             $propertyType = TypeHandling::normalizeType($propertyConfiguration['type'] ?? 'string');
-            if ($propertyValue === '' && !TypeHandling::isSimpleType($propertyType)) {
+            if ($propertyValue === null || ($propertyValue === '' && !TypeHandling::isSimpleType($propertyType))) {
                 continue;
             }
             $propertyValue = $this->nodePropertyConversionService->convert($node->getNodeType(), $propertyName, $propertyValue, $node->getContext());
--- a/Classes/NodeCreationHandler/CreationDialogPropertiesCreationHandler.php
+++ b/Classes/NodeCreationHandler/CreationDialogPropertiesCreationHandler.php
@@ -13,10 +13,8 @@ namespace Neos\Neos\Ui\NodeCreationHandler;
 
 use Neos\ContentRepository\Domain\Model\NodeInterface;
 use Neos\Flow\Annotations as Flow;
-use Neos\Flow\Property\Exception as PropertyException;
-use Neos\Flow\Property\PropertyMapper;
-use Neos\Flow\Property\TypeConverter\PersistentObjectConverter;
-use Neos\Flow\Security\Exception as SecurityException;
+use Neos\Neos\Ui\Domain\Service\NodePropertyConversionService;
+use Neos\Neos\Ui\NodeCreationHandler\NodeCreationHandlerInterface;
 use Neos\Utility\ObjectAccess;
 use Neos\Utility\TypeHandling;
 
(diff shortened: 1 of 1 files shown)


bwaidelich (approved):
Oh nice, we have a service for this. Neos is truly service-oriented :)

+1 by reading

kitsunet:
Yep, this place would be the _third_ implementation of how to convert incoming stuff for node properties 😆 and it does things differently than the other two.

grebaldi (approved):
Hi @kdambekalns,

I ran the scenario you've described in the issue and can confirm that it works. :+1: for having consistency :)

Though I actually do believe that the property value should stick to the client's timezone, because server time may run counter to user's intent :/

kdambekalns:
But clients may be all over the world, so storing in **one** timezone and converting on output/display/use seems better. Plus, within the UI we rely on UTC…

Classes/NodeCreationHandler/CreationDialogPropertiesCreationHandler.php
```diff
@@ -27,37 +25,22 @@ class CreationDialogPropertiesCreationHandler implements NodeCreationHandlerInte
...
-
-        foreach ($node->getNodeType()->getConfiguration('properties') as $propertyName => $propertyConfiguration) {
+        foreach ($data as $propertyName => $propertyValue) {
+            $propertyConfiguration = $node->getNodeType()->getConfiguration('properties')[$propertyName];
             if (!isset($propertyConfiguration['ui']['showInCreationDialog']) || $propertyConfiguration['ui']['showInCreationDialog'] !== true) {
                 continue;
             }
             $propertyType = TypeHandling::normalizeType($propertyConfiguration['type'] ?? 'string');
-            if (!isset($data[$propertyName])) {
+            if ($propertyValue === null || ($propertyValue === '' && !TypeHandling::isSimpleType($propertyType))) {
```
mhsdesign, line 40: You introduced this `null` skipping here but i makes sense i guess.

I am more curios about the  `$propertyValue === '' && !TypeHandling::isSimpleType($propertyType)` part because it seems to have gone missing in 9.0 after the ESCR [adjustments](https://github.com/neos/neos-ui/commit/1498c8b4c6d84faf993dac1dd3756ad20993e4dd#diff-773a367aeb44c8b1dc0134079eaa8f3483e76936c98682d720234ce87a6f44a8L55-L57).

Originally it was introduced to fix https://github.com/bwaidelich/Wwwision.Neos.CreationDialogProperties/issues/10

But when using the `nodePropertyConversionService` we seem to get this `$rawValue !== ''` check for free in:

https://github.com/neos/neos-ui/blob/01ac51c52e66a2ad6fb9d9dcbff9623739e5da14/Classes/Domain/Service/NodePropertyConversionService.php#L87

So not skipping this property would write an empty string to it but that is fine and seems to be expected because of this check 😅 

https://github.com/neos/neos-development-collection/blob/eada1dbfb569de6e3932b26b4b3c630401020e05/Neos.ContentRepository/Classes/Domain/Model/Node.php#L924-L931

we might as well keep it for 8.3 like this bu my brain is exploding when i think about 9.0 :D

mhsdesign (commented):
Thanks for cleaning that up :D

It seems the whole thing was introduced with https://github.com/neos/neos-ui/pull/2596 and never really changed since. I stumbled upon this in Neos 9 and assumed it was "special" legacy handling code but was confused that it also exists in 8.3.

I plan with Neos9 to extract this whole part with https://github.com/neos/neos-ui/pull/3519 deeper into the Ui php so that each `NodeCreationHandlerInteface` just gets a value object with deserialized elements. That way no creation handler has to come to the idea of doing it theirselves.

Classes/NodeCreationHandler/CreationDialogPropertiesCreationHandler.php
```diff
@@ -27,37 +25,22 @@ class CreationDialogPropertiesCreationHandler implements NodeCreationHandlerInte
...
-     * @throws PropertyException | SecurityException
-     */
     public function handle(NodeInterface $node, array $data): void
     {
-        $propertyMappingConfiguration = $this->propertyMapper->buildPropertyMappingConfiguration();
-        $propertyMappingConfiguration->forProperty('*')->allowAllProperties();
-        $propertyMappingConfiguration->setTypeConverterOption(PersistentObjectConverter::class, PersistentObjectConverter::CONFIGURATION_OVERRIDE_TARGET_TYPE_ALLOWED, true);
-
-        foreach ($node->getNodeType()->getConfiguration('properties') as $propertyName => $propertyConfiguration) {
+        foreach ($data as $propertyName => $propertyValue) {
```
mhsdesign, line 34: i think this is dangerous to now loop of $data, as this might $somehow contain entries not from the `creationDialog.elements` configuration, like in [Psmb.SplitAdd](https://github.com/psmb/Psmb.SplitAdd/tree/master) were we do some magic like https://github.com/psmb/Psmb.SplitAdd/blob/ceb3d9c3d98755cce1177355e0585e684faa9512/Classes/PropertiesNodeCreationHandler.php#L17

so it should be safer to loop over the configured properties like before and check if its promoted via `showInCreationDialog`

kitsunet: This is prevented in 47/37 by continuing if it#s not allowed.

mhsdesign: ahh i see but actually not really yet. Line 35 expects the property to exist which is unsafe *AND WILL CRASH* even for legal other creation dialog elements so this access as to be `??` first.

i would argue not to change much of the codeflow in this bugfix and keep looping over `$node->getNodeType()->getConfiguration('properties')` so the actual part of this fix becomes more clear.

For neos 9 i prepared a little refactoring that will cleanup this mess completely by deserialising the elements beforehand

Benjamin-K:
We have stumpled upon this several times now. If we select 10:00 in the creation dialog, 11:00 will be stored (because of UTC + 1).

Classes/NodeCreationHandler/CreationDialogPropertiesCreationHandler.php
kdambekalns: Well, it is what the `isset()` checks does, but not "as strict", i.e. anything that is `null` or not a non-empty string representing some non-literal, is decteted.

Classes/NodeCreationHandler/CreationDialogPropertiesCreationHandler.php
kdambekalns: > i think this is dangerous to now loop of $data, as this might $somehow contain entries not from the creationDialog.elements configuration

No, now it will *only* loop over what was submitted, because there is no need to loop over all properties instead. Right? Not sure I understand your reference to Psmb.SplitAdd…

But you are right, *if* the configuration is not set, this would break, so `??` makes sense.

mhsdesign: yeah no fine with me lets loop over $data WITH the null check in place ;)
`Psmb.SplitAdd` is just a random case where we will have 100% a key in $data that is not part of the schema.

Classes/NodeCreationHandler/CreationDialogPropertiesCreationHandler.php
mhsdesign: Okay and regarding 9.0 i created a followup https://github.com/neos/neos-ui/issues/3755

Classes/NodeCreationHandler/CreationDialogPropertiesCreationHandler.php
mhsdesign: adjusted ;)

mhsdesign (approved):

PR #3723, opened 2024-02-17
BUGFIX: Neos Ui JSON serializable property values

mhsdesign:
Ports https://github.com/neos/neos-development-collection/pull/4638 to the Neos Ui, as the class was moved in Neos 9.
Required for https://github.com/neos/neos-ui/pull/3996

Unfortunately the property serialisation for the NeosUi does NOT use the expected `\JsonSerializable::jsonSerialize` but the property mapper and the `ArrayFromObjectConverter` for object types to get an array that will be json encoded.

This works mostly fine but in some cases it fails:
- when your `fromArray` fields and property names values dont match
- when a "object" subtypes the object mapper is no convertable like the `GuzzleHttp\Psr7\Uri`:

```
Too few arguments to function GuzzleHttp\Psr7\Uri::isAbsolute(), 0 passed in /core/neos-manufacture-highest/Packages/Framework/Neos.Utility.ObjectHandling/Classes/ObjectAccess.php on line 151 and exactly 1 expected
```

Current workarounds are AOP:

https://github.com/sitegeist/Sitegeist.InspectorGadget/blob/78f5f4a206287b1c4bedf5cb88582ed51cb4a311/Classes/Infrastructure/NodeInfo/NodeInfoPostProcessingAspect.php#L17

Or to use a dumb property mapper:
  https://github.com/sitegeist/Sitegeist.Archaeopteryx/blob/9322b9cb8e4824bcaf7aaa247c23b1244a2f1167/Classes/LinkToArrayForNeosUiConverter.php#L12C16-L12C78

------

We fix this by early returning any \JsonSerializable so the native `json_encode` is used in a later step. Additionally we also check if the class of the value object itself would know how to be deserialised from a plain scalar value. This is done by checking for `fromArray` and friends which is necessary to be implemented by a value object in the Flow and Neos universe.

---------

During the upmerge to 9.0 i noticed that we obviously dont use the `DenormalizingObjectConverter` of Flows property mapper anymore. Instead we use custom written denormalisers for symfony. But they again dont allow for a simple check like `DenormalizingObjectConverter::isDenormalizable()`. And in order to avoid adding this verbose "hacky" code this upmerge keeps using this Flow helper as its essentially what the CR would do as well:

```php
(new ValueObjectArrayDenormalizer())->supportsDenormalization([], $propertyValue::class)
&& (new ValueObjectBoolDenormalizer())->supportsDenormalization(true, $propertyValue::class)
&& (new ValueObjectFloatDenormalizer())->supportsDenormalization(1.1, $propertyValue::class)
&& (new ValueObjectIntDenormalizer())->supportsDenormalization(1, $propertyValue::class)
&& (new ValueObjectStringDenormalizer())->supportsDenormalization('', $propertyValue::class)
```

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/Classes/Domain/Service/NodePropertyConverterService.php
+++ b/Classes/Domain/Service/NodePropertyConverterService.php
@@ -27,6 +27,7 @@ use Neos\Flow\Property\Exception as PropertyException;
 use Neos\Flow\Property\PropertyMapper;
 use Neos\Flow\Property\PropertyMappingConfiguration;
 use Neos\Flow\Property\PropertyMappingConfigurationInterface;
+use Neos\Flow\Property\TypeConverter\DenormalizingObjectConverter;
 use Neos\Flow\Property\TypeConverterInterface;
 use Neos\Neos\Utility\NodeTypeWithFallbackProvider;
 use Neos\Utility\ObjectAccess;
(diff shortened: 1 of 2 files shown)


Classes/Domain/Service/NodePropertyConverterService.php
```diff
@@ -209,15 +214,34 @@ public function getPropertiesArray(Node $node)
...
+     * @param string $dataType the property type from the node type schema
+     * @return \JsonSerializable|int|float|string|bool|null|array<mixed>
      * @throws PropertyException
      */
     protected function convertValue($propertyValue, $dataType)
     {
+        if (
+            $propertyValue instanceof \JsonSerializable
+            // todo maybe combine them for performance with a static `isDenormalizable` utility.
+            && (new ValueObjectArrayDenormalizer())->supportsDenormalization([], $propertyValue::class)
```
mhsdesign: `fromBoolean` and `fromInteger` (the long form) which was allowed in flow https://github.com/neos/flow-development-collection/pull/2766 seems to be missing as a feature for the new cr...

mhsdesign:
Merging as this is an up-merge and i have already all the reviews in https://github.com/neos/neos-development-collection/pull/4638 ;)

PR #3734, opened 2024-03-06
BUGFIX: document-node is redirected to the nearest parent on workspace change

c4ll-m3-j4ck:
Previously, on changing from a workspace to another one that didn't contain the current document-node, an exception would be triggered if no new document-node could be found a single level up in the node-tree.

This PR changes this behaviour to walk up the node-tree until a document-node can be found (going up to the site-node eventually) and sets the new document node to the first shared one that can be found in the rootline.

### DIFF
--- a/Classes/Controller/BackendServiceController.php
+++ b/Classes/Controller/BackendServiceController.php
@@ -359,11 +359,6 @@ class BackendServiceController extends ActionController
             $updateWorkspaceInfo->setWorkspace($userWorkspace);
             $this->feedbackCollection->add($updateWorkspaceInfo);
 
-            // Construct base workspace context
-            $contextProperties = $documentNode->getContext()->getProperties();
-            $contextProperties['workspaceName'] = $targetWorkspaceName;
-            $contentContext = $this->contextFactory->create($contextProperties);
-
             // If current document node doesn't exist in the base workspace, traverse its parents to find the one that exists
             $nodesOnPath = $documentNode->getContext()->getNodesOnPath($sitePath, $originalNodePath);
             $redirectNode = $nodesOnPath[count($nodesOnPath) - 1];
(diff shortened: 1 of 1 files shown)


Classes/Controller/BackendServiceController.php
```diff
@@ -357,24 +360,13 @@ public function changeBaseWorkspaceAction(string $targetWorkspaceName, NodeInter
...
-                    break;
-                } else {
-                    $redirectNode = $redirectNode->getParent();
-                    if (!$redirectNode) {
-                        throw new Exception(sprintf('Wasn\'t able to locate any valid node in rootline of node %s in the workspace %s.', $documentNode->getContextPath(), $targetWorkspaceName), 1458814469);
-                    }
-                }
-            }
+            $nodesOnPath = $documentNode->getContext()->getNodesOnPath($sitePath, $originalNodePath);
+            $redirectNode = $nodesOnPath[count($nodesOnPath) - 1];
```
Sebobo:
```suggestion
            $redirectNode = array_pop($nodesOnPath);
```
Then you might get `null` as result and then we could instead use the site node of the context as redirectNode

c4ll-m3-j4ck: Like this?
`$redirectNode = array_pop($nodesOnPath) ?? $documentNode->getContext()->getRootNode;`

Sebobo: `$documentNode->getContext()->getCurrentSiteNode()` should work.
The root node is not loadable in the ui.

Sebobo (approved):
Thx

mhsdesign:
I could reproduce the bug

> Wasn't able to locate any valid node in rootline of node /sites/neosdemo/features/navigation-elements/third-item/node-nh1g56sefvfkf/node-jwe1o0ggldihe@user-admin;language=en_US in the workspace live.


Scenario 1 is working currently:

- create workspace
- create new subpage in workspace and publish and highlight it
- change workspace to live again



But scenario 2 fails currently:

- create workspace
- create new subpage in workspace
- create new subsubpage in workspace
- publish and highlight the new subsubpage
- change workspace to live again

mhsdesign (approved):
I cant say if this covers or breaks every edgecase but both scenario 1 and 2 are now working correctly (im redirected to the closest page that exists in the base ws)

PR #3736, opened 2024-03-11
BUGFIX: Remove duplicated apply button in DateInput & handle onChange on date selection

### LINKED ISSUE #3421: FEATURE: Remove separate "apply"-Button from DateTime Editor
### Description

The "Apply" Button of the DateTime Editor is often not visible when the datetime editor ist in the lower half of the screen. This leads to confusion when editors choose a date but the Inspector apply button stays inactive. 

Also it it a questionable ui-solution to have to push to different apply buttons in a row to actually select a date. 

This could be nicely combined with #3394 

#### Expected behavior

It would be great if the apply button at the datepicker would be removed entirely and the date would be set once the user sets a date (date-mode) or a (time).

Main Problem tody is that apply does two things: 1. Apply value, 2. Close DatePicker. If we find a different rule for closing the datepicker after editing apllying to values once they are slected should feel natural to editors.

[What actually happened]

### Affected Versions

UI: since ever

Alvadda:
The current date picker required a separate click to a blue "apply" button to actually set the field value. Afterwards the editors still had to click the inspector apply which leads to confusion. This is especially problematic in the CreationDialog where the apply button often is below the visible area which causes additional confusion.

This change improved DateTime Editor UX by auto-applying date/time selections which allowed to remove the DatePicker's "Apply" button entirely. The DateTime Editor still closes once the editor clicks outside.

For testings: Define a DateTime property and verify that you can set the date value without second apply.

Fixes: #3421

### DIFF
--- a/packages/neos-ui-editors/src/Editors/DateTime/index.js
+++ b/packages/neos-ui-editors/src/Editors/DateTime/index.js
@@ -62,7 +62,6 @@ class DateTime extends PureComponent {
                 timeOnly={!hasDateFormat(options.format)}
                 placeholder={i18nRegistry.translate($get('placeholder', options) || 'Neos.Neos:Main:content.inspector.editors.dateTimeEditor.noDateSet')}
                 todayLabel={i18nRegistry.translate('content.inspector.editors.dateTimeEditor.today', 'Today', {}, 'Neos.Neos', 'Main')}
-                applyLabel={i18nRegistry.translate('content.inspector.editors.dateTimeEditor.apply', 'Apply', {}, 'Neos.Neos', 'Main')}
                 locale={interfaceLanguage}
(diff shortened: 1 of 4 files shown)


grebaldi (approved):
Nice one :+1: 

I was able to verify that this change works as expected. The good ol' apply button of the DateTimeEditor has been a pain for many, so thanks a lot for finally removing it @Alvadda!

**EDIT:** Follow-up for Neos.Neos, because the translation label for the button will be obsolete once this is merged: https://github.com/neos/neos-development-collection/issues/4928

markusguenther (approved):
Thanks for your first contribution 💙
That is a change editors will love.

grebaldi:
This needs to target `8.3`, right? @markusguenther

Issue #3412, opened 2023-03-07
FEATURE: Usability Improvements for Date Time Editor


markusguenther:
### Description

Currently, in Neos, the date time editor is somewhat cumbersome to use. Users must click through several dropdown menus to select the desired time and date, which can be time-consuming and inefficient.

I propose the implementation of usability improvements for the date time editor in Neos 8.3, including the addition of a 24-hour format and the ability to input time directly via keyboard. These improvements would make it easier and more efficient for users to enter dates and times in Neos, improving the overall user experience.

The 24-hour format option would allow users to select the format they prefer for displaying time, providing greater flexibility and customization. Additionally, the ability to input time directly via keyboard would save time and reduce the number of clicks required to select the desired time.

This feature would be particularly useful for users who frequently work with dates and times in Neos, such as editors and content managers. It would make the process of entering dates and times more efficient and user-friendly.

Additional specifications for this feature will need to be discussed and defined, but I believe it would be a valuable addition to Neos 8.3, enhancing the usability of the date time editor and improving the overall user experience.

This is a feature idea from the community brainstorming thread on discuss:
https://discuss.neos.io/t/attention-neos-community/6206/10

crydotsnake:
> and the ability to input time directly via keyboard

#3394 👍🏽

markusguenther:
Also related https://github.com/neos/neos-ui/issues/3421

PR #3737, opened 2024-03-11
BUGFIX: Use 12/24 hour format for DateTimePicker depending on configured `format`

Alvadda:
Currently the DateTimePicker always uses the 12h time format even if the time format specified is in 24h format.

Currently supported hour formats from docs:
> g: hour without leading zeros - 12-hour format - 1 through 12
> G: hour without leading zeros - 24-hour format - 0 through 23
> h: 12-hour format of an hour with leading zeros - 01 through 12
> H: 24-hour format of an hour with leading zeros - 00 through 23

This change ensures that the 24h format is used for entering times when a 24h format was configured.

For testing:
```
dateTime24h:
  type: DateTime
    inspector:
      editorOptions:
        format: d-m-Y H:i
dateTime12h:
  type: DateTime
    inspector:
      editorOptions:
        format: d-m-Y h:i a
```

Fixes partly : [#3412](https://github.com/neos/neos-ui/issues/3412)

### DIFF
--- a/packages/react-ui-components/src/DateInput/dateInput.tsx
+++ b/packages/react-ui-components/src/DateInput/dateInput.tsx
@@ -60,6 +60,11 @@ export interface DateInputProps {
     readonly timeOnly?: boolean;
 
     /**
+     * Show the Time in 24 hours or 12 hours format
+     */
+    readonly is24Hour?: boolean;
+
+    /**
      * Add some constraints to the timepicker.
      * It accepts an object with the format { hours: { min: 9, max: 15, step: 2 }},
      * this example means the hours can't be lower than 9 and higher than 15,
@@ -104,7 +109,7 @@ interface DateInputTheme {
 }
 
 const defaultProps: PickDefaultProps<DateInputProps, 'labelFormat' | 'timeConstraints'> = {
-    labelFormat: 'DD-MM-YYYY hh:mm',
+    labelFormat: 'DD-MM-YYYY hh:mm A',
     timeConstraints: {
         minutes: {
             min: 0,
@@ -141,6 +146,7 @@ export class DateInput extends PureComponent<DateInputProps, DateInputState> {
             labelFormat,
             dateOnly,
             timeOnly,
+            is24Hour,
             locale,
             disabled
         } = this.props;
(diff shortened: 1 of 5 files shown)


packages/react-ui-components/src/DateInput/dateInput.tsx
```diff
@@ -176,6 +182,15 @@ export class DateInput extends PureComponent<DateInputProps, DateInputState> {
             },
         );
 
+        const getTimeFormat = () => {
+            if (dateOnly) {
+                return false
+            }
+
+            console.log('getTimeFormat', is24Hour);
```
grebaldi:
```suggestion
```

gotcha :)

packages/react-ui-components/src/DateInput/dateInput.tsx
```diff
@@ -176,6 +182,15 @@ export class DateInput extends PureComponent<DateInputProps, DateInputState> {
             },
         );
 
+        const getTimeFormat = () => {
```
grebaldi: This should not be a function imho. The result from `getTimeFormat` is used exactly once per render - so it might as well be calculated in place, e.g.:

```typescript
const timeFormat = dateOnly ? false : (is24Hour ? 'HH:m' : 'hh:m A');
```

But, since nested ternaries are awkward, it would also be okay to turn `getTimeFormat` into a private calculated property of the component, e.g.:

```typescript
private get timeFormat(): false | string {
    if (this.props.dateOnly) {
        return false;
    }

    return this.props.is24Hour ? 'HH:m' : 'hh:m A';
}
```

grebaldi (changes requested):
Hi @Alvadda,

nice work! Deriving this from the configured format makes complete sense and I was able to test your change locally to verify that it works like a charm :+1: 

Just one minor nitpick on the code, that's all :)

Since this is being treated as `BUGFIX`, the PR should target `8.3` rather than `8.4`.

packages/react-ui-components/src/DateInput/dateInput.tsx
Alvadda: Good Catch

packages/react-ui-components/src/DateInput/dateInput.tsx
Alvadda: I like your suggestion with the private function in the class component. 

I don't rly like nested ternaries and think a function is way more readable than nesting ternaries.

Thanks for your suggestion I will change this.

grebaldi (approved):
Thanks for the adjustments, @Alvadda :+1:

mhsdesign (approved):

Issue #3671, opened 2023-12-11
ImageCropper AspectRatioDropDown can always deselect preset


PRGfx:
### Description

Sometimes the user may select one of the configured aspect-ratios, e.g. choose between landscape and portrait. However, as soon as I provide the option to select an aspect-ratio preset, I can clear the selection and freely draw my cropping area.

### Steps to Reproduce

1. Configure a property with an image, forcing the editor to crop the image to one of the provided aspect-ratios
  ```yaml
  properties:
    image:
      type: Neos\Media\Domain\Model\ImageInterface
      ui:
        inspector:
          editorOptions:
            features:
              crop: true
             crop:
               aspectRatio:
                 # we want to force an aspect-ratio when selecting the image
                 forceCrop: true
                 # we explicitely don't want custom aspect-ratios
                 allowCustom: false
                 # we set a defaultOption so we don't end up without selected aspect-ratio
                 defaultOption: portrait
                 options:
                   portrait:
                     label: Portrait
                     width: 2
                     height: 3
                   landscape:
                     label: Landscape
                     width: 3
                     height: 2
  ```
2. Select an image, clear the selected aspect ratio, modify the cropping area at will

#### Expected behavior

The user should not be able to clear the selected preset.
Arguably `defaultOption` should be optional as well (and default to the first valid option) if the cropping-configuration is not locked and custom configuration is not permitted.

#### Actual behavior

There is a button to clear the selected aspect ratio in the AspectRatioDropDown as soon as something is selected and there is no way to configure the editor to not provide this option.

After deselecting the configured preset the user can select any aspect-ratio they want.

### Affected Versions

Neos: 5.3 - today
UI: This seems to never having changed (also see #2909)

NicoleNitschke:
I would like to fix this issue in the ccurent sprint. :)

ahaeslich:
Fixed with https://github.com/neos/neos-ui/pull/3739

PR #3738, opened 2024-03-12
#3671 - add check if custom ratios for images are allowed

NicoleNitschke:
**What I did**
I prevent the clear button in the image cropper view to appear, if the yaml-configuration says `allowCostum: false`. This way, we show only allowed aspect ratios.

**How I did it**
I changed the code, so that the yaml settings 'allowCustom' for the image editor is respected.

**How to verify it**
As described in the issue:
Configure a property with an image, forcing the editor to crop the image to one of the provided aspect-ratios.
```
  properties:
    image:
      ui:
        inspector:
          group: 'media'
          position: 50
          editorOptions:
            crop:
              aspectRatio:
                # we want to force an aspect-ratio when selecting the image
                forceCrop: true
                # we explicitely don't want custom aspect-ratios
                allowCustom: false
                # we set a defaultOption so we don't end up without selected aspect-ratio
                defaultOption: portrait
                options:
                  portrait:
                    label: Portrait
                    width: 2
                    height: 3
                  landscape:
                    label: Landscape
                    width: 3
                    height: 2
```

### Before:
grafik

### With bugfix:
grafik

### DIFF
--- a/packages/neos-ui-editors/src/SecondaryEditors/ImageCropper/AspectRatioDropDown/index.js
+++ b/packages/neos-ui-editors/src/SecondaryEditors/ImageCropper/AspectRatioDropDown/index.js
@@ -38,13 +38,14 @@ export default class AspectRatioDropDown extends PureComponent {
             PropTypes.instanceOf(AspectRatioOption)
         ),
         placeholder: PropTypes.string,
+        allowCustomRatios: PropTypes.boolean,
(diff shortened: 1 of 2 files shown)


NicoleNitschke:
This PR will be closed because Neos 7.3 is no longer maintained. See #3739 instead.

PR #3739, opened 2024-03-12
BUGFIX #3671: add check if custom ratios for images are allowed

### LINKED ISSUE #3671: ImageCropper AspectRatioDropDown can always deselect preset

NicoleNitschke:
**What I did**
I prevent the clear button in the image cropper view to appear, if the yaml-configuration says `allowCostum: false`. This way, we show only allowed aspect ratios.

**How I did it**
I changed the code, so that the yaml settings 'allowCustom' for the image editor is respected.

**How to verify it**
As described in the issue:
Configure a property with an image, forcing the editor to crop the image to one of the provided aspect-ratios.
```
  properties:
    image:
      ui:
        inspector:
          group: 'media'
          position: 50
          editorOptions:
            crop:
              aspectRatio:
                # we want to force an aspect-ratio when selecting the image
                forceCrop: true
                # we explicitely don't want custom aspect-ratios
                allowCustom: false
                # we set a defaultOption so we don't end up without selected aspect-ratio
                defaultOption: portrait
                options:
                  portrait:
                    label: Portrait
                    width: 2
                    height: 3
                  landscape:
                    label: Landscape
                    width: 3
                    height: 2
```

### Before:
grafik

### With bugfix:
grafik

### DIFF
--- a/packages/neos-ui-editors/src/SecondaryEditors/ImageCropper/AspectRatioDropDown/index.js
+++ b/packages/neos-ui-editors/src/SecondaryEditors/ImageCropper/AspectRatioDropDown/index.js
@@ -37,13 +37,14 @@ export default class AspectRatioDropDown extends PureComponent {
             PropTypes.instanceOf(AspectRatioOption)
         ),
         placeholder: PropTypes.string,
+        allowCustomRatios: PropTypes.boolean,
 
         onSelect: PropTypes.func.isRequired,
         onClear: PropTypes.func.isRequired
     };
 
     render() {
-        const {options, current, placeholder, onSelect, onClear} = this.props;
+        const {options, current, placeholder, allowCustomRatios, onSelect, onClear} = this.props;
 
         const dropDownHeaderClasses = mergeClassNames({
             [style.dropDown__btn]: true,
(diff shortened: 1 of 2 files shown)


grebaldi (approved):
Hi @NicoleNitschke.

this makes sense indeed. I was able to verify that your change works like a charm! :+1: Thanks a lot for your work! :)

mhsdesign (approved):

PR #3740, opened 2024-03-12
FEATURE: move user menu to left drawer

### LINKED ISSUE #3685: Relocate User Menu to Left Drawer for Improved Navigation
### Description:
The current positioning of the user menu in the primary bar has been identified as a potential area for improvement. Considering user behavior, where the user menu is primarily utilized for actions like logging out, we propose relocating it to the left drawer. This adjustment aligns with established UX patterns observed in other systems, where the user menu is de-emphasized and accessed when needed. 

### Proposed Change:
Move the user menu from the primary bar to the left drawer to minimize its prominence in the interface. This approach reflects the user behavior, where interactions with the user menu are infrequent and often limited to specific actions such as logging out (which is not the most used action). Backend modules should be updated to maintain a consistent and streamlined user interface.

### Expected Benefits:

* Reduced prominence of the user menu, allowing the primary focus on essential application features.
* Simplified user interface, aligning with established UX patterns observed in other systems.
* Enhanced user experience by reducing visual clutter and emphasizing core functionalities.
* The drawer has way more space, and therefore we are able to add additional information, like a user avatar.

### Additional Information:
We've identified an opportunity to enhance the Neos User Interface (UI) by making it more consistent and user-friendly. The proposed changes include restructuring the primary and secondary toolbars, moving the workspace selector to the top level, and incorporating feedback from users, such as Daniel Kestler. This adjustment is part of our ongoing efforts to optimize the Neos User Interface based on user behavior and industry-standard UX patterns. 

https://github.com/neos/neos-ui/issues/3247
https://github.com/neos/neos-ui/issues/3803

laurahaenel:
**What I did**

I moved the user menu to the left drawer to free space in the primary toolbar.

**How I did it**

- Move the UserMenu Component to the left drawer
- Rearrange the order of the dropdown items
- Add a new UserImage Component that currently renders the users initials
- Styling

**How to verify it**

With your eyes.

Old:

image

New:

image

Related issues: https://github.com/neos/neos-development-collection/issues/4825

Neos.Neos PR: https://github.com/neos/neos-development-collection/pull/4934


laurahaenel:
This change is affected by the open bug https://github.com/neos/neos-development-collection/issues/4590

packages/neos-ui/src/Containers/Drawer/UserDropDown/UserImage.js
```diff
@@ -0,0 +1,26 @@
...
+
+import {neos} from '@neos-project/neos-ui-decorators';
+
+import style from './style.module.css';
+
+@connect(state => ({
+    userFirstName: state?.user?.name?.firstName,
+    userLastName: state?.user?.name?.lastName
+}))
+@neos()
```
grebaldi:
```suggestion
```

`@neos` is not needed here, because this component does not interact with any global object promoted by the decorator (like the `i18nRegistry` or `containerRegistry`).

packages/neos-ui/src/Containers/Drawer/UserDropDown/UserImage.js
```diff
@@ -0,0 +1,26 @@
...
+export default class UserImage extends PureComponent {
+    static propTypes = {
+        userFirstName: PropTypes.string.isRequired,
+        userLastName: PropTypes.string.isRequired
```
grebaldi:
```suggestion
        userFirstName: PropTypes.string,
        userLastName: PropTypes.string
```

In L21 both props are treated as optional (which makes sense).

grebaldi (changes requested):
Hi @laurahaenel,

thanks so much for taking on this feature :+1: I like how you solved it - especially the `UserImage` component - well done! :slightly_smiling_face: 

I left some code comments (nothing big). Aside from that, I was able to verify that your change works (including user impersonation).

On a meta note: I noticed that introducing this feature will potentially lead to confusion in combination with user impersonation. With the user menu hidden in the drawer, users will no longer have any visual indication that they're currently impersonating someone else (I think this is also what @Sebobo is referring to in https://github.com/neos/neos-development-collection/issues/4825#issuecomment-1991720648). 

Maybe we can find a way to indicate this somehow? I think it's not necessary to see exactly which account is being impersonated, but we need some visual clue, that the user is currently not themselves :) (I think as much as a border around the entire UI would suffice - but I welcome more sophisticated suggestions :sweat_smile:)

packages/neos-ui/src/Containers/Drawer/UserDropDown/UserImage.js
laurahaenel: Applied :)

packages/neos-ui/src/Containers/Drawer/UserDropDown/UserImage.js
laurahaenel: Applied :)

grebaldi (approved):
Thanks for the adjustments @laurahaenel :)

PR #3745, opened 2024-03-13
BUGFIX: Select box shows invalid value

### LINKED ISSUE #3520: BUG: Select Box show as empty when current value doesnt match anymore the values

73nici:
replaces the initial solution https://github.com/neos/neos-ui/pull/3526 with a refactored approach applied from outside the select-box to reduce the logic there and add translation for the label.

Resolves: #3520

**What I did**

For the first and second case (mentioned in #3520) we introduced a warning icon, a proper label and additionally allow to reset the value like:

image

And for the multi select box it will look like this, ensuring the sorting will still work

https://github.com/neos/neos-ui/assets/85400359/55a92ddc-a591-4fe0-9ee4-39712b2e6098

For a (slow) datasource it looks like:

https://github.com/neos/neos-ui/assets/85400359/890717cc-986e-42c4-a1e4-d74202fd6b2f

and for multiple: true it looks like:

https://github.com/neos/neos-ui/assets/85400359/ef8c27a1-1693-481a-8b28-a7f1623165ac

Related discussion https://github.com/neos/neos-ui/pull/3526#issuecomment-1629025114


mhsdesign:
<details><summary>Old description from Niclas</summary>
<p>

Currently the selectbox shows the "choose" placeholder, if an invalid option is selected. This is misleading, cause a value is still selected and therefore can be used.

If the current selected value is not a valid option, the placeholder changes to "invalid value" without  the value being updated.
To check this, remove a selected value from a select box. After reloading, the placeholder should read "invalid value".

Fixes #3520 

New behaviour:
image

Old behaviour: 
image


</p>
</details> 




Hi Niclas ;)

i adjusted your pr a bit to what we discussed at the sprint i guess... following todos are still todo:

- [x] transform the helper to ts
- [x] add unit tests for it
- [ ] console error when used standalone as component to indicate there are missing values -> not needed for now
- [x] update description
- [x] fix index when dragging and dropping
- [x] adjust label wording?

mhsdesign:
I added some test for `processSelectBoxOptions` and also run them against the legacy implementation to make sure the behaviour matches. The result ist promising, only the tests for the new features fail (`creates missing option`). So the behaviour is correct ;)

``` 
FAIL  packages/neos-ui-editors/src/Editors/SelectBox/selectBoxHelpers.spec.ts
  processSelectBoxOptions
    ✓ transforms an associative array with labels to list of objects (1 ms)
    ✓ keeps valid shape of list of objects intact
    ✓ overrules the array key with the explicit value
    ✓ uses numeric string array key for list of objects
    ✓ omits entries that are invalid and empty (1 ms)
    ✕ creates missing option for unmatched string value (2 ms)
    ✕ creates missing options for unmatched additional array value (1 ms)
    ✕ creates missing options for unmatched additional multiple array values (1 ms)
    ✓ ignored current value being null or undefined and doesnt create missing option
```

mhsdesign:
e2e failure is currently related:


```
 ✓ Create a text node in a new container element at the correct position

 Create new nodes
 ✖ Check ClientEval for dependencies between properties of NodeTypes in Creation Dialog

   1) AssertionError: expected [ Array(6) ] to deeply equal [ Array(5) ]

      + expected - actual

         {
           label: 'label_9'
           value: 9
         }
      -  {
      -    icon: 'exclamation-triangle'
      -    label: 'Invalid value: ""'
      -    value: ''
      -  }
       ]


      Browser: Firefox 127.0 / Windows 10 (https://app.saucelabs.com/tests/8ce1a59a62ca48fcb7c99bae4d08ac28)

         41 |            {'label': 'even', value: 'even'}
         42 |        ])
         43 |
         44 |    await t
         45 |        .expect(dependingPropertySelectBox.props.value).eql('')
       > 46 |        .expect(dependingPropertySelectBox.props.options).eql([
         47 |            {label: 'label_1', value: 1},
         48 |            {label: 'label_3', value: 3},
         49 |            {label: 'label_5', value: 5},
         50 |            {label: 'label_7', value: 7},
         51 |            {label: 'label_9', value: 9}

         at <anonymous> (/home/circleci/app/Packages/Application/Neos.Neos.Ui/Tests/IntegrationTests/Fixtures/1Dimension/createNewNodes.e2e.js:46:59)
         at asyncGeneratorStep (/home/circleci/app/Packages/Application/Neos.Neos.Ui/Tests/IntegrationTests/Fixtures/1Dimension/createNewNodes.e2e.js:4:38)
         at _next (/home/circleci/app/Packages/Application/Neos.Neos.Ui/Tests/IntegrationTests/Fixtures/1Dimension/createNewNodes.e2e.js:4:38)
```

mhsdesign:
I dont understand the test completely, how can there be an empty string in an integer nodetype property???

https://github.com/neos/neos-ui/blob/7c9fc845981d13c7a413efb33094949026071dfa/Tests/IntegrationTests/Fixtures/1Dimension/createNewNodes.e2e.js#L45

https://github.com/neos/neos-ui/blob/2be77a4967f0f755bd76cff5cd7959d907f1d694/Tests/IntegrationTests/SharedNodeTypesPackage/NodeTypes/Content/NodeWithDependingProperties.yaml#L31

i guess empty string means unset??? we should check on that but ill adjust the "create magic select box option" logic to ignore empty string

Sebobo (approved):
Works on my machine, thx!
Even the createNew thing which is nowhere documented I think.
I also have to add those options to my NodeTypes schema.

73nici:
Hey @mhsdesign I think fixed the CI :). Only a snapshot update for selectbox was needed.

Sebobo (approved):

grebaldi (approved):

PR #3747, opened 2024-03-13
BUGFIX: Disable node text selection in document tree

Alvadda:
It was possible to select the HTML Text from the node elements in the documentation tree via `shift + left click`. 
This bug was found in the PR [3625](https://github.com/neos/neos-ui/pull/3635) via the [commit](https://github.com/neos/neos-ui/pull/3635#issuecomment-1898625946)

The issue arrived premier in the `safari` browser.

The issue was fixed be applying `user-select: none` to the whole tree element to prevent all children to be HTML text selectable. 

To verify this, try to select multiple elements with `shift + left click`  and there should not have the default blue selection highlight.

Wrong:

Correct:
image

### DIFF
--- a/esbuild.js
+++ b/esbuild.js
@@ -3,6 +3,8 @@ const {compileWithCssVariables} = require('./cssVariables');
 const {cssModules} = require('./cssModules');
 const esbuild = require('esbuild');
 const { version } = require('./package.json')
+const { browserslistToTargets } = require('lightningcss');
+const browserslist = require('browserslist');
 
 const isProduction = process.argv.includes('--production');
 const isE2ETesting = process.argv.includes('--e2e-testing');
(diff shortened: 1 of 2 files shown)


esbuild.js
```diff
@@ -75,8 +75,12 @@ const options = {
         cssModules(
             {
                 visitor: compileWithCssVariables(),
-                targets: {
-                    chrome: 80 // aligns somewhat to es2020
+                targets: { // only support es2020 browser
```
mhsdesign, line 78: please add these links to the explanation

https://lightningcss.dev/transpilation.html

https://github.com/evanw/esbuild/issues/121#issuecomment-646956379

ahaeslich (approved):

ahaeslich:
➕ by 👀

mhsdesign (approved):
Thanks for fixing this. it works on my mashine as well

PR #3748, opened 2024-03-13
BUGFIX: 9.0 infinite (real) reloads after exception

73nici:
**What I did**
Update the NodeInfoHelper to build an initial absolute URL, instead of a relative one.

**How to verify it**

Throw an exception (or return an abnormal response) in any places and expect the iframe to be only loaded once.
The orange loading bar will keep spinning but that is part of  #3477.

```php
throw new \Exception(rand());
``` 

But currently this results in an endless reload, because in the condition of the frame `win.location.href` will not equal `this.props.src` causing `win.location.replace(this.props.src)` being retriggered infinitely:

https://github.com/neos/neos-ui/blob/5e1675897032a0dba73b3a1bb445ffad55162680/packages/react-ui-components/src/Frame/frame.tsx#L96

### DIFF
--- a/Classes/Fusion/Helper/NodeInfoHelper.php
+++ b/Classes/Fusion/Helper/NodeInfoHelper.php
@@ -344,7 +344,10 @@ class NodeInfoHelper implements ProtectedContextAwareInterface
             $nodeAddressFactory = NodeAddressFactory::create($contentRepository);
             $nodeAddress = $nodeAddressFactory->createFromNode($nodeAddress);
         }
-        return (string)NodeUriBuilder::fromRequest($actionRequest)->uriFor($nodeAddress);
+        $uriBuilder = new UriBuilder();
+        $uriBuilder->setRequest($actionRequest);
+        $uriBuilder->setCreateAbsoluteUri(true);
+        return (string)NodeUriBuilder::fromUriBuilder($uriBuilder)->uriFor($nodeAddress);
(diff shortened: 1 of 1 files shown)


mhsdesign (approved):
I have to look at this again, it seems after navigating via the site tree the infinite reloads continue.

Basically there are two ways to trigger this bug, one via navigating via the iframe to a broken page or via the document tree.
Both should stop loading after encountering the first error. (The orange indicator might continue spinning but there should not be any real reloads)

In 8.3 always absolute src are used for `ui.contentCanvas.src`: `"http://127.0.0.1:8081/neos/preview?node%5B__contextNodePath%5D=%2Fsites%2Fneosdemo%2Ffeatures%2Fnavigation-elements%2Ffirst-item%40user-admin%3Blanguage%3Den_US"`

mhsdesign (changes requested):
.

73nici:
Hey @mhsdesign,
can i still help with something to get this PR merged :)?

mhsdesign (approved):
Thanks now it works. Neither the inline navigation (via the iframe) nor the navigation via the document tree leads to endless exception flodding like currently:

https://github.com/neos/neos-ui/assets/85400359/b02ecfca-8c94-4d85-9aed-9eb1a570e7b2

That way we restored the same behaviour as in Neos 8.3

PR #3750, opened 2024-03-14
WIP Bugfix/3477 infinite loading indicator remove dom event

laurahaenel:
see #3477

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/Classes/Fusion/ExceptionHandler/PageExceptionHandler.php
+++ b/Classes/Fusion/ExceptionHandler/PageExceptionHandler.php
@@ -102,10 +102,6 @@ class PageExceptionHandler extends AbstractRenderingExceptionHandler
         $fluidView->setFormat('html');
         $fluidView->setTemplatePathAndFilename('resource://Neos.Neos.Ui/Private/Templates/Error/ErrorMessage.html');
 
-        $guestNotificationScript = new StandaloneView();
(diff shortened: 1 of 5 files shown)


Sebobo:
Hi @laurahaenel, what is the state of this PR?

I recently helped two people again who had problems with the loading indicator due to Fusion issues and I thought about fixing this and showing more helpful errors if there is an actual loading error.

laurahaenel:
Hi @Sebobo ,
If I remember correctly, there was a discussion at the Hamburg Sprint '24 about the condition from the two events Neos.Neos.Ui.ContentReady and load (or something). Maybe @grebaldi or @mhsdesign can remember :D 
(see https://github.com/neos/neos-ui/issues/3477#issuecomment-2002400111) 
AFAIK there was no outcome on how to proceed. Maybe we can just close this.

mhsdesign:
Hi so the the problem is that relying on the custom even is quite cumbersome and breaks easily if the response is not conforming. Instead we wanted to use the native events and possibilities to detect if the iframe is loaded to afterwards initialise everything. It was kind of a rabbit hole too and we feared that we might break something unintentionally though i still believe this is the correct direction, we just have to make it bullet proof.

Sebobo:
Ok. Either way, I think we also need a timeout to help with debugging. So if the iframe doesn't load in a certain time and doesn't trigger our event, I would like to show some information to the user and print some debug information to the console. F.e. if the guest frame is missing crucial metadata and such things.

PR #3752, opened 2024-03-14
FEATURE: Refactor publication and discarding endpoints

nezaniel:
This implements explicit endpoints for publishing and discarding sites or documents, adjusting Neos' behavior to the ESCR.

(see also: https://github.com/neos/neos-development-collection/pull/4943)


grebaldi:
* PHP linting check will be fixed once https://github.com/neos/neos-development-collection/pull/4943 has been merged
* JS unit tests will be green once https://github.com/neos/neos-ui/pull/3757 has been upmerged to `9.0`
* E2E tests are green, because of https://github.com/neos/neos-ui/pull/3752/commits/3a06560087b6c7736e5175dcbc4b6d312c653e8c - **We need to revert this commit before merging!**

Classes/Domain/Model/Changes/Remove.php
```diff
@@ -104,4 +104,17 @@ public function apply(): void
             $this->feedbackCollection->add($updateParentNodeInfo);
         }
     }
+
+    private function getRemovalAttachmentPoint(): ?NodeAggregateId
+    {
+        $subgraph = $this->contentRepositoryRegistry->subgraphForNode($this->subject);
+
+        if ($this->subject->nodeType->isOfType(NodeTypeNameFactory::NAME_DOCUMENT)) {
```
mhsdesign:
```suggestion
        if ($this->subject?->nodeType->isOfType(NodeTypeNameFactory::NAME_DOCUMENT)) {
```

otherwise one wouldnt be able to remove nodes without nodeType, or in Neos.Neos we use the `getNodeType` trait that provides a fallback. (NodeTypeWithFallbackProviderTrait or sth)

grebaldi: Is `subject` nullable here or is it rather `subject->nodeType`? So:

```php
if ($this->subject->nodeType?->isOfType(NodeTypeNameFactory::NAME_DOCUMENT)) {
```

mhsdesign: no sorry of course the latter :)

Classes/Domain/Model/Changes/Remove.php
```diff
@@ -104,4 +104,17 @@ public function apply(): void
             $this->feedbackCollection->add($updateParentNodeInfo);
         }
     }
+
+    private function getRemovalAttachmentPoint(): ?NodeAggregateId
```
mhsdesign, line 108: as pointed out here: https://github.com/neos/neos-development-collection/pull/4943#pullrequestreview-1949957107

this logic should imo reside in Neos.Neos so that there is an API way of deleting nodes so they will still be deletable via the workspace publisher.

Classes/ContentRepository/Service/WorkspaceService.php
```diff
@@ -35,6 +34,11 @@
  */
 class WorkspaceService
 {
+    private const NODE_HAS_BEEN_CREATED = 0b0001;
+    private const NODE_HAS_BEEN_CHANGED = 0b0010;
+    private const NODE_HAS_BEEN_MOVED = 0b0100;
+    private const NODE_HAS_BEEN_DELETED = 0b1000;
```
mhsdesign, line 40: that seems clever and cost effective but isnt this a bit hard to grasp as the information is interpreted in the js codbase?

mhsdesign:
i added my comments over here https://github.com/neos/neos-development-collection/pull/4943#pullrequestreview-1949957107

also we should probably keep in mind that at some point we want to publish the node removal more transparently see
https://github.com/neos/neos-ui/pull/2974

i guess the whole `removalAttachmentPoint` is hacky and we will need to change it to support this case ...

as noted above i think all the change projection logic should reside in Neos.Neos so we can refactor this anytime.
The change projection is market as experimental and unstable, so i would like to avoid it to leak into the NeosUi and to all places where people are interested in the current changes.

Maybe this is doable in the scope of this pr maybe not. Just an idea. Thanks for the refactoring so far.

Classes/ContentRepository/Service/WorkspaceService.php
grebaldi: In the other PR #3759, I've already refactored the client side part of this: https://github.com/neos/neos-ui/pull/3759/commits/3ce8499159180c8a6ed3e3a39944fe6a5e908024

Bitmasks work in JS just as they do in PHP, so this can simply be mirrored.

Classes/Domain/Model/Changes/Remove.php
grebaldi: Indeed. Neos.Neos should provide a high level API for all change operations imho. But I would argue that this is out of scope for this PR.

We definitely need to straighten out the boundaries between Neos.Neos and Neos.Neos.Ui. I believe however that by now everything in Neos.Neos.Ui is treated as internal (mostly thanks to your effort :)), so it would be permissible for a deeper refactoring to take place post-release.

wdyt?

Classes/Domain/Model/Changes/Remove.php
grebaldi: Adjusted :)

mhsdesign:
Thanks for all the adjustments and discussions so far. It seems the mutations are now handled by the Neos `Workspace` but we dont use the new `Workspace::countChangesInSite` api. Should we do this in a followup or later :D?

grebaldi:
@mhsdesign:
> Thanks for all the adjustments and discussions so far. It seems the mutations are now handled by the Neos `Workspace` but we dont use the new `Workspace::countChangesInSite` api. Should we do this in a followup or later :D?

Definitely later. Currently, we need the actual changes in the UI to calculate dirty-state in various places. The new API doesn't solve this for us yet, but that's an implementation detail and can be adjusted later down the line. We need to take care of conflict resolution first.

Classes/Controller/BackendServiceController.php
```diff
@@ -182,56 +170,70 @@ public function changeAction(array $changes): void
...
     /**
-     * Publish all nodes
+     * Publish all changes in the current site
+     *
+     * @phpstan-param array<string,string> $command
      */
-    public function publishAllAction(): void
+    public function publishChangesInSiteAction(array $command): void
     {
+        /** @todo send from UI */
```
mhsdesign, line 179: todo remove todo, or later

mhsdesign (approved):
Thanks for the change i will run the Neos Ui e2e locally and then i suppose we can merge this.

grebaldi:
Just a reminder:

- https://github.com/neos/neos-development-collection/pull/4943 should be merged first
- https://github.com/neos/neos-ui/commit/3a06560087b6c7736e5175dcbc4b6d312c653e8c must be reverted before this one gets merged

mhsdesign:
Locally all tests pass, except an unrelated unit test failure that should be fixed after an upmerge. Ill push also a commit to fix the silly ci (sorry i meant style ci)

mhsdesign:
reopen to trigger ci

mhsdesign:
Huh e2e still fail

> Fatal error: Cannot redeclare Neos\Neos\Domain\Workspace\Workspace::getCurrentContentStreamId() in /home/circleci/app/Packages/Neos/Neos.Neos/Classes/Domain/Workspace/Workspace.php on line 131

also on 9.0 after merge ...

Classes/Controller/BackendServiceController.php
```diff
@@ -182,56 +170,70 @@ public function changeAction(array $changes): void
...
-            $currentAccount = $this->securityContext->getAccount();
-            $workspaceName = WorkspaceNameBuilder::fromAccountIdentifier($currentAccount->getAccountIdentifier());
-
-            $nodeIdentifiersToPublish = [];
-            foreach ($nodeContextPaths as $contextPath) {
-                $nodeAddress = $nodeAddressFactory->createFromUriString($contextPath);
-                $nodeIdentifiersToPublish[] = new NodeIdToPublishOrDiscard(
-                    $nodeAddress->nodeAggregateId,
-                    $nodeAddress->dimensionSpacePoint
-                );
```
mhsdesign, line 226: FYI, as we dont use the node address from the Neos Ui here anymore and their dimension we accidentally trashed partial publishing with respect to dimensions.

But this might be okay as discussed :D

PR #3753, opened 2024-03-15
BUGFIX: Correctly append cache buster value to existing query parameters for included CSS & JS

Sebobo:
Prevent invalid urls when static Uris already have query parameters and Neos UI tries to add its own cache buster value.

### DIFF
--- a/Classes/Domain/Service/StyleAndJavascriptInclusionService.php
+++ b/Classes/Domain/Service/StyleAndJavascriptInclusionService.php
@@ -95,7 +95,7 @@ class StyleAndJavascriptInclusionService
                 $hash = substr(md5_file($resourceExpression), 0, 8);
                 $resourceExpression = $this->resourceManager->getPublicPackageResourceUriByPath($resourceExpression);
             }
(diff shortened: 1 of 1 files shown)


Classes/Domain/Service/StyleAndJavascriptInclusionService.php
```diff
@@ -95,7 +95,7 @@ protected function build(array $resourceArrayToSort, \Closure $builderForLine)
                 $hash = substr(md5_file($resourceExpression), 0, 8);
                 $resourceExpression = $this->resourceManager->getPublicPackageResourceUriByPath($resourceExpression);
             }
-            $finalUri = $hash ? $resourceExpression . '?' . $hash : $resourceExpression;
+            $finalUri = $hash ? $resourceExpression . (str_contains($resourceExpression, '?') ? '&' : '?') . $hash : $resourceExpression;
```
mhsdesign, line 98: is this really _the_ flow way of doing that? With Flow 9 we could use https://github.com/neos/flow-development-collection/pull/3316

but for 8.3 this might be all weve got :/

mhsdesign (approved):

PR #3663, opened 2023-11-08
BUGFIX: Show full labels in nodetrees if possible

Sebobo:
**What I did**

Previously labels were cut even if there would have been enough space due to the constraint to 264px.
Also since 8.3 the trees can be resized and the labels didn’t adjust to the increased width.

**How I did it**

Now they will use the available width via CSS.

**How to verify it**

Before:

Bildschirmfoto 2023-11-08 um 15 31 58
Bildschirmfoto 2023-11-08 um 15 31 51

After:

Bildschirmfoto 2023-11-08 um 15 32 17
Bildschirmfoto 2023-11-08 um 15 32 10

PR #3763, opened 2024-04-10
BUGFIX: Copy nodetype name from nodeinfoview without hyphens

Sebobo:
Fixes a regression introduced in https://github.com/neos/neos-ui/pull/3663 by using the wbr tag instead of the shy character which is part of the copied text. The wbr tag is ignored.

Also adds a copy button for easier interaction.
This change keep the feature that the nodetype name wraps when it’s longer than the inspector is wide.

CleanShot 2024-04-10 at 22 16 09@2x

### DIFF
--- a/Resources/Private/Translations/en/Main.xlf
+++ b/Resources/Private/Translations/en/Main.xlf
@@ -374,6 +374,9 @@
             <trans-unit id="errorBoundary.footer" xml:space="preserve">
                 <source>For more information about the error please refer to the JavaScript console.</source>
             </trans-unit>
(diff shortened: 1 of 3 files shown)


Sebobo:
The alignment of the copy button is a bit off, as I aligned it based on its hover state:
 
CleanShot 2024-04-10 at 22 19 57@2x

When I align it with the `Type` text then the hover state looks off.

bwaidelich (approved):
Awesome, thanks for taking care!
+1 by reading

ahaeslich (approved):
by 👀

PR #3764, opened 2024-04-16
Delegate selection of relation distribution strategy

### LINKED ISSUE #3587: BUG 9.0 NodeMove should respect `aggregate: true`
otherwise with gather all its not possible to move content around if they dont have the same parent in all dimensions: https://github.com/neos/neos-development-collection/issues/5368

Also for content it might just not be desired to be in sync.

see https://github.com/neos/neos-ui/blob/5d69c44b337b77ed344bc100f605a13722e6ecb0/Classes/Domain/Model/Changes/MoveInto.php#L90

```php
$nodeType->isAggregate()
    ? RelationDistributionStrategy::STRATEGY_SCATTER
    : RelationDistributionStrategy::STRATEGY_GATHER_ALL
```

Related to https://github.com/neos/neos-development-collection/issues/3340

basically currenlty everything is aggregate ^^

### Description

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

#### Expected behavior

[What you expected to happen]

#### Actual behavior

[What actually happened]

### Affected Versions

Neos:

UI:

nezaniel:
... to the command handler for now.

As long as the UI doesn't support a dialog for this, we fall back to the node type declaration.

based on [#4994](https://github.com/neos/neos-development-collection/pull/4994)
resolves #3587

### DIFF
--- a/Classes/Domain/Model/Changes/MoveAfter.php
+++ b/Classes/Domain/Model/Changes/MoveAfter.php
@@ -91,7 +91,7 @@ class MoveAfter extends AbstractStructuralChange
                 $workspace->workspaceName,
                 $subject->subgraphIdentity->dimensionSpacePoint,
                 $subject->nodeAggregateId,
-                RelationDistributionStrategy::STRATEGY_GATHER_ALL,
+                null,
                 $hasEqualParentNode ? null : $parentNodeOfPreviousSibling->nodeAggregateId,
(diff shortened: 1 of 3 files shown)


mhsdesign:
duplicate of https://github.com/neos/neos-ui/pull/3876 which was merged and thus now obsolete ;)

PR #3770, opened 2024-04-24
BUGFIX: Improve UI loading performance by removing UI nodedata script tag and using less data

Sebobo:
**What I did**

With this change the minimal required nodedata for each node in the rendered
content is inserted as data attribute and not as inline script anymore.
This improves performance as no extra function call is executed for each node.

Additonally the reduction in rendered node attributes reduces the output filesize and again improves loading time.

To prevent just-in-time loading of nodes all incomplete nodedata is requested after the guest frame has finished loading.

One big benefit of that is that the scripts tags won't interfere with JS plugins anymore and using the Fusion Augmenter will not wrap a single object with a ContentWrapping with another tag due to the script tag.

Fixes https://github.com/neos/neos-development-collection/issues/2988

**How to verify it**

To test this change one can use the `bugfix/remove-ui-script-tag-compiled-artifacts` branch in composer:

```
"neos/neos-ui": "dev-bugfix/remove-ui-script-tag-compiled-artifacts as 8.3.8",
```

* Check whether content is still properly editable
* Inspector shouldn't show a loading spinner when content is selected
* `preview` document should be smaller (10-30%) than before and load quicker (10-40%)
* No script tags with node data in the raw preview html document

- [x] Load nodedata async after guest frame has loaded
- [ ] Visualise for user that backend still loads data before page is fully editable

### DIFF
--- a/Classes/ContentRepository/Service/NodeService.php
+++ b/Classes/ContentRepository/Service/NodeService.php
@@ -13,6 +13,7 @@ namespace Neos\Neos\Ui\ContentRepository\Service;
 
 use Neos\ContentRepository\Domain\Model\NodeInterface;
 use Neos\ContentRepository\Domain\Model\Workspace;
+use Neos\ContentRepository\Domain\Service\Context;
 use Neos\ContentRepository\Domain\Service\ContextFactoryInterface;
 use Neos\ContentRepository\Domain\Utility\NodePaths;
 use Neos\Eel\FlowQuery\FlowQuery;
@@ -45,6 +46,11 @@ class NodeService
      * @var DomainRepository
      */
     protected $domainRepository;
+
+    /**
+     * @var array<string, Context>
+     */
+    protected array $contextCache = [];
 
     /**
      * Helper method to retrieve the closest document for a node
@@ -86,36 +92,43 @@ class NodeService
         $nodePath = $nodePathAndContext['nodePath'];
         $workspaceName = $nodePathAndContext['workspaceName'];
         $dimensions = $nodePathAndContext['dimensions'];
+        $siteNodeName = $site ? $site->getNodeName() : explode('/', $nodePath)[2];
 
-        $contextProperties = $this->prepareContextProperties($workspaceName, $dimensions);
+        // Prevent reloading the same context multiple times
+        $contextHash = md5(implode('|', [$siteNodeName, $workspaceName, json_encode($dimensions), $includeAll]));
+        if (isset($this->contextCache[$contextHash])) {
+            $context = $this->contextCache[$contextHash];
+        } else {
+            $contextProperties = $this->prepareContextProperties($workspaceName, $dimensions);
 
-        if ($site === null) {
-            list(, , $siteNodeName) = explode('/', $nodePath);
-            $site = $this->siteRepository->findOneByNodeName($siteNodeName);
-        }
+            if ($site === null) {
+                $site = $this->siteRepository->findOneByNodeName($siteNodeName);
+            }
 
-        if ($domain === null) {
-            $domain = $this->domainRepository->findOneBySite($site);
-        }
+            if ($domain === null) {
+                $domain = $this->domainRepository->findOneBySite($site);
+            }
 
-        $contextProperties['currentSite'] = $site;
-        $contextProperties['currentDomain'] = $domain;
-        if ($includeAll === true) {
-            $contextProperties['invisibleContentShown'] = true;
-            $contextProperties['removedContentShown'] = true;
-            $contextProperties['inaccessibleContentShown'] = true;
-        }
+            $contextProperties['currentSite'] = $site;
+            $contextProperties['currentDomain'] = $domain;
+            if ($includeAll === true) {
+                $contextProperties['invisibleContentShown'] = true;
+                $contextProperties['removedContentShown'] = true;
+                $contextProperties['inaccessibleContentShown'] = true;
+            }
 
-        $context = $this->contextFactory->create(
-            $contextProperties
-        );
-
-        $workspace = $context->getWorkspace(false);
-        if (!$workspace) {
-            return new Error(
-                sprintf('Could not convert the given source to Node object because the workspace "%s" as specified in the context node path does not exist.', $workspaceName),
-                1451392329
+            $context = $this->contextFactory->create(
+                $contextProperties
             );
+
+            $workspace = $context->getWorkspace(false);
+            if (!$workspace) {
+                return new Error(
+                    sprintf('Could not convert the given source to Node object because the workspace "%s" as specified in the context node path does not exist.', $workspaceName),
+                    1451392329
+                );
+            }
+            $this->contextCache[$contextHash] = $context;
         }
 
         return $context->getNode($nodePath);
--- a/Classes/Controller/BackendServiceController.php
+++ b/Classes/Controller/BackendServiceController.php
@@ -532,6 +532,7 @@ class BackendServiceController extends ActionController
         $createContext = array_shift($chain);
         $finisher = array_pop($chain);
 
+        // we deduplicate passed nodes here
         $nodeContextPaths = array_unique(array_column($createContext['payload'], '$node'));
 
         $flowQuery = new FlowQuery(array_map(
--- a/Classes/Controller/BackendServiceController.php
+++ b/Classes/Controller/BackendServiceController.php
@@ -567,7 +567,7 @@ class BackendServiceController extends ActionController
                 $nodeTypeFilter = $finisher['payload']['nodeTypeFilter'] ?? null;
                 $result = $nodeInfoHelper->renderNodesWithParents(
                     array_filter($flowQuery->get()),
-                    $this->getControllerContext(), 
+                    $this->getControllerContext(),
                     $nodeTypeFilter
                 );
                 break;
--- a/Classes/Controller/BackendServiceController.php
+++ b/Classes/Controller/BackendServiceController.php
@@ -532,11 +532,13 @@ class BackendServiceController extends ActionController
         $createContext = array_shift($chain);
         $finisher = array_pop($chain);
 
+        $nodeContextPaths = array_unique(array_column($createContext['payload'], '$node'));
+
         $flowQuery = new FlowQuery(array_map(
-            function ($envelope) {
-                return $this->nodeService->getNodeFromContextPath($envelope['$node']);
+            function ($contextPath) {
+                return $this->nodeService->getNodeFromContextPath($contextPath);
             },
-            $createContext['payload']
+            $nodeContextPaths
         ));
 
         foreach ($chain as $operation) {
@@ -548,8 +550,8 @@ class BackendServiceController extends ActionController
 
         switch ($finisher['type']) {
             case 'get':
-                $result = $nodeInfoHelper->renderNodes(array_filter(
-                    $flowQuery->get()),
+                $result = $nodeInfoHelper->renderNodes(
+                    array_filter($flowQuery->get()),
                     $this->getControllerContext()
                 );
                 break;
--- a/Classes/Controller/BackendServiceController.php
+++ b/Classes/Controller/BackendServiceController.php
@@ -548,14 +548,25 @@ class BackendServiceController extends ActionController
 
         switch ($finisher['type']) {
             case 'get':
-                $result = $nodeInfoHelper->renderNodes(array_filter($flowQuery->get()), $this->getControllerContext());
+                $result = $nodeInfoHelper->renderNodes(array_filter(
+                    $flowQuery->get()),
+                    $this->getControllerContext()
+                );
                 break;
             case 'getForTree':
-                $result = $nodeInfoHelper->renderNodes(array_filter($flowQuery->get()), $this->getControllerContext(), true);
+                $result = $nodeInfoHelper->renderNodes(
+                    array_filter($flowQuery->get()),
+                    $this->getControllerContext(),
+                    true
+                );
                 break;
             case 'getForTreeWithParents':
                 $nodeTypeFilter = $finisher['payload']['nodeTypeFilter'] ?? null;
-                $result = $nodeInfoHelper->renderNodesWithParents(array_filter($flowQuery->get()), $this->getControllerContext(), $nodeTypeFilter);
+                $result = $nodeInfoHelper->renderNodesWithParents(
+                    array_filter($flowQuery->get()),
+                    $this->getControllerContext(), 
+                    $nodeTypeFilter
+                );
                 break;
         }
 
--- a/Classes/Domain/Service/UserLocaleService.php
+++ b/Classes/Domain/Service/UserLocaleService.php
@@ -38,6 +38,13 @@ class UserLocaleService
     protected $rememberedContentLocale;
 
     /**
+     * Remebered content locale for locale switching
+     *
+     * @var Locale
+     */
+    protected $rememberedUserLocale;
+
+    /**
      * For serialization, we need to respect the UI locale, rather than the content locale
      *
      * @param boolean $reset Reset to remebered locale
@@ -47,11 +54,15 @@ class UserLocaleService
         if ($reset === true) {
             // Reset the locale
             $this->i18nService->getConfiguration()->setCurrentLocale($this->rememberedContentLocale);
+        } elseif ($this->rememberedUserLocale) {
+            // Restore the local
+            $this->i18nService->getConfiguration()->setCurrentLocale($this->rememberedUserLocale);
         } else {
             $this->rememberedContentLocale = $this->i18nService->getConfiguration()->getCurrentLocale();
             $userLocalePreference = ($this->userService->getCurrentUser() ? $this->userService->getCurrentUser()->getPreferences()->getInterfaceLanguage() : null);
             $defaultLocale = $this->i18nService->getConfiguration()->getDefaultLocale();
             $userLocale = $userLocalePreference ? new Locale($userLocalePreference) : $defaultLocale;
+            $this->rememberedUserLocale = $userLocale;
             $this->i18nService->getConfiguration()->setCurrentLocale($userLocale);
         }
     }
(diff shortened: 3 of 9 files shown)


Sebobo:
I'm not sure how breaking this could be right now, therefore draft for now ;) 

@grebaldi I remember you started working on a replacement for the script tag, so your insights here would be very welcome.

This doesn't need to be included in 8.3 but maybe we can :)

Sebobo:
Some test results with 82 nodes (haven't found a good way to measure time to interactive with the iframe):

## With Script Tag + with full Properties

Document size: 245KB / 14.6KB
Document rendering time: ~220ms
Transferred: 3.3MB
Finish: ~1.7s
Load: ~620ms
Requests: 28

## Without Script Tag + with MinimalProperties + ondemand node loading

Document size: 198KB / 13.3KB (~80%)
Document rendering time: ~140ms (~65%)
Finish: ~1.6s
Load: ~620ms
Requests: 28

## Without Script Tag + with MinimalProperties + Node batch loading

Document size: 198KB / 13.3KB (~80%)
Document rendering time: ~140ms (~65%)
Finish: ~2.1s
Load: ~660ms
Requests: 29

Classes/Domain/Service/UserLocaleService.php
```diff
@@ -37,6 +37,13 @@ class UserLocaleService
      */
     protected $rememberedContentLocale;
 
+    /**
+     * Remebered content locale for locale switching
+     *
+     * @var Locale
+     */
+    protected $rememberedUserLocale;
```
bwaidelich: Are those changes related?

Sebobo: Not a 100%, just some optimisation as the locale is switched twice for every node triggered by the augmentation

Sebobo:
I tested a variant of this change in my customers project and the document size was reduced from 1.8mb to 1.2mb.
Without any nodedata it's 700kb. So I think we should invest further time in the future to really only include what we need.

Also there are two child nodes queries for each node to get document and content child nodes. Maybe we also don't need those for the initial rendering with some adjustments. Right now it would break some parts of the ui.

Sebobo:
OK optimised a bit more and completely removed nodedata from the guest frame:

## Without NodeData in augmenter

Document size: 108KB / 7.5KB (~44%)
Document rendering time: ~123ms (~55%)
Finish: ~2.1s
Load: ~610ms
Requests: 30


When clicking through the demo site each page is about 50% quicker to load which actually feels much nicer.

bwaidelich:
Great initiative, thanks for taking care.
Just one consideration: additional HTTP requests are fast locally, but they _can_ be a bottleneck on remote servers. Not suggesting, that it is in this case but we should keep it in the back of our heads and test with throttled bandwidth

Sebobo:
Yes of course. At the end we need the same amount of nodedata. But we kill the overhead of the function calls and document parsing.
Even if the combined loading time is the same as we load the nodedata, it feels faster as the page is already displayed much earlier while the data is fetched and the JS doesn't block the rendering. We send a bunch of queries at the initial load, so while the user starts moving the mouse to an element ideally the data is already there and simply be injected into the store instead of being parsed.
This makes especially a difference when you quickly navigate between pages.
The inspector also shows a loading spinner if the node is not fully loaded yet. So we have to check whether other places might need some visual feedback that we are still loading something if the connection is slow.

Sebobo:
@bwaidelich I tested the loading performance a while yesterday with "Fast 3G" connection + Caching in my local Dev Context.

Result: The page loads much faster, but if you try to immediately edit text in a CKEditor, the first click only selects the element instead of showing the caret to start editing. The second click then works.

I want to test this on an actual remote server in Production Context to get a feeling for it.

I also checked whether I can adjust the LoadingBar to wait for the GuestFrame to be initialised, but this is not a simple change right now. I would make sense to have a loading indicator of some kind for the guest initialisation too. This is already currently problem, when `initializeGuestFrame` takes a bit longer. Just now with my change it takes a bit longer depending on the connection.
But the request runs in parallel with other requests like fetching node policies etc. So it's not simply adding to the loading time. But of course this depends a bit on the setup too.

We currently also load some nodes twice as the structure tree needs them too. I want to check whether I can then skip some of them.

Benjamin-K:
Thanks for taking care of this. As a side effect, this will also help with some CSS selectors, as with the current approach selectors like `.element:last-child` don't work as expected in the backend.

Sebobo:
Realised today this will also fix the issue https://github.com/neos/neos-development-collection/issues/2988 with the Fusion Augmenter when augmenting a single content element. Previously it added another tag as it counted the element itself and the script tag instead of modifying the outside tag of the content element.

Sebobo:
The two main performance hogs which makes the async request for the nodedata slow is in `\Neos\Neos\Ui\Fusion\Helper\NodeInfoHelper::renderChildrenInformation` which queries each nodes children twice. 
In my tests this takes at least 50-60% of the duration.
The other one is the very inefficient conversion of each node context path from the query to an actual node. 

It seems this was less of a problem in the previous version as during rendering the same first-level-node-cache was automatically used where possible.
I think this could be optimised quite a bit in a separate PR.

Right now I'm getting great results by optimising \Neos\Neos\Ui\ContentRepository\Service\NodeService::getNodeFromContextPath in skipping the creation of a completely new context with site and domain for each requested node.
For the child node query I don't have a great idea right now. This could be easier with Neos 9 maybe.

grebaldi (approved):
@Sebobo 
> @grebaldi I remember you started working on a replacement for the script tag, so your insights here would be very welcome.

My idea was to replace the content element wrapping technique we're using right now (HTML Augmenter) with an alternative approach using HTML comments instead. This would prevent collisions with userland markup and I also expect it to perform better, since HTML augmentation is quite slow.

That idea does not at all interfere with your change here. I actually thought we had rid ourselves of those script tags already :see_no_evil: 

Thanks a lot for taking care of it! :+1: 

I've tested it locally (unthrottled), and I did notice some lag with the guest frame initialization (especially noticeable with lots of greyed-out not-inline-editable elements). But I think the trade-off is well worth it and we can optimize the initialization later still.

I've also re-run the CircleCI e2e tests, which were red due to connectivity issues. Everything's green now :)

Sebobo:
Hi @grebaldi thx for the review!

This morning while walking the 🦮 I thought about batch loading the nodes in the UI as a future optimisation. Meaning instead of loading all of them at once, We could load them in batches of 20(?) nodes or so and this way would make the nodes "above the fold" quicker accessible than the rest.
But we definitely also need to adjust the behaviour of the progress bar to actually show the state of the guest frame application instead of just the DomContentLoaded event. But would also do this in a separate PR

Sebobo:
@bwaidelich what do you think, should we merge this change and open a follow up to rework the ui loading state for 8.4 f.e.?
Or combine both into 8.4 to be safe

bwaidelich:
> what do you think, should we merge this change

I can't really give a qualified opinion (just wanted to chime in regarding potential pitfalls with n+1 requests) but my tendency would be to go the safer route (merge it into 8.4)

packages/neos-ui-guest-frame/src/initializeGuestFrame.js
```diff
@@ -60,13 +61,29 @@ export default ({globalRegistry, store}) => function * initializeGuestFrame() {
         return;
     }
 
-    const nodes = Object.assign({}, guestFrameWindow['@Neos.Neos.Ui:Nodes'], {
-        [documentInformation.metaData.documentNode]: documentInformation.metaData.documentNodeSerialization
-    });
+    // Load legacy node data scripts from guest frame - remove with Neos 9.0
+    const legacyNodeData = guestFrameWindow['@Neos.Neos.Ui:NodeData'] || {};
```
mhsdesign:
```suggestion
    const legacyNodeData = guestFrameWindow['@Neos.Neos.Ui:Nodes'] || {};
```

mhsdesign:
Fyi for up merging it has to be adjusted in Neos 9 in Neos.Neos https://github.com/neos/neos-development-collection/blob/bf4df1ea7445e482df9c2e6a9c647915f6a33a53/Neos.Neos/Classes/Service/ContentElementWrappingService.php#L108

mhsdesign:
As we dont profit anymore from the fusion cache the ajax request might actually be slower.
For example here https://github.com/neos/neos-ui/blob/cb5bccf7f76b2f068b65f50763e6a168babe5cb4/Resources/Private/Fusion/Prototypes/Page.fusion#L11-L47 we explicitly cache the script tags hard for performance.

Sebobo:
Sadly some additional tests in a large page in a real page with almost 300 nodes make the usability worse, as the async request takes 3-4s. I don't unterstand right now, why the flow query from the UI is sooo much slower than simply rendering the same nodes with their data in the html.

The page itself got reduced from 1.8mb to 700kb, and loads in 2.8s instead of 4.2 uncached which is nice, but the overall feeling is worse with this change.

Therefore I wouldn't merge this without improving the flow query response time.

mhsdesign (changes requested):
see above, we have to investigate into caching

Sebobo:
I did some further modifications by removing duplicate node context paths (which the tested page had a lot of) and I created a separate branch with further optimisations which include loading all the required nodedata from the db with custom queries.

I now have the following results:

Flow query performance for loading a large page in a large project with the page tree, content tree and nodes required for the displayed content. The page tree requests 175 nodes (the 332KB request), the content tree requests 58 nodes ( the 129kb request) and the guest frame requests 191 nodes (the 615kb request) which haven't been loaded by the content tree yet and with the duplicates removed.
The preview document shrinks from 1.8mb to 784kb with this branch compared to 8.3.8 as all the script tags are gone.

Performance UI 8.3.8:
neosui-83-838

Performance with this  branch:
neosui-83-remove-ui-script-tag

Performance with branch #3817:
neosui-83-optimise-ui-flowqueries

The project I'm testing with has a slightly worse usability with this branch than with 8.3.8 as the additional flowquery that fetches the missing content nodedata takes a few seconds. But #3817 fully fixes that improving the time until inline text can be edited from 7.5s to 6.5s.

packages/neos-ui-guest-frame/src/initializeGuestFrame.js
```diff
@@ -60,13 +61,45 @@ export default ({globalRegistry, store}) => function * initializeGuestFrame() {
...
+    // Load remaining list of not fully loaded nodes from the backend if there are any
+    const fullyLoadedNodesFromContent = notFullyLoadedNodeContextPaths.length > 0 ? (yield q(notFullyLoadedNodeContextPaths).get()).reduce((nodes, node) => {
+        nodes[node.contextPath] = node;
+        return nodes;
+    }, {}) : [];
+
+    const nodes = Object.assign(
+        {},
+        legacyNodeData, // Merge legacy node data from the guest frame - remove with Neos 9.0
+        nodesAlreadyPresentInStore,
```
mhsdesign: nodesAlreadyPresentInStore are already present in the store so we dont need to  `actions.CR.Nodes.merge(nodes)` them? Its not a replace...

Sebobo: Correct, I still need to merge them into one list for initialisation, but not merge them into the store. Thx!

Classes/Domain/Service/UserLocaleService.php
```diff
@@ -37,6 +37,13 @@ class UserLocaleService
      */
     protected $rememberedContentLocale;
 
+    /**
+     * Remebered content locale for locale switching
```
mhsdesign:
```suggestion
     * Remebered user locale for locale switching (cached for performance)
```

packages/neos-ui-guest-frame/src/initializeGuestFrame.js
```diff
@@ -60,13 +61,48 @@ export default ({globalRegistry, store}) => function * initializeGuestFrame() {
...
+    const nodeContextPathsInGuestFrame = findAllNodesInGuestFrame().map(node => node.getAttribute('data-__neos-node-contextpath'));
+
+    // Filter nodes that are already present in the redux store and duplicates
+    const nodesByContextPath = store.getState().cr.nodes.byContextPath;
+    const nodesAlreadyPresentInStore = {};
+    const notFullyLoadedNodeContextPaths = [...new Set(nodeContextPathsInGuestFrame)].filter((contextPath) => {
+        const node = nodesByContextPath[contextPath];
+        const nodeIsLoaded = node !== undefined && node.isFullyLoaded;
+        if (nodeIsLoaded) {
+            nodesAlreadyPresentInStore[contextPath] = node;
```
mhsdesign: this side effect here from the otherwise pure filter function looks odd and only seems to server the usecase that only the nodes on the page found should be passed later to `initializeContentDomNode`.

Instead id just pass the whole nodes to `initializeContentDomNode` instead:

```
store.getState().cr.nodes.byContextPath
```

that will also get rid of

```
    // Combine new and existing nodes into a list for initialisation in the guest frame
     nodes = Object.assign(nodes, nodesAlreadyPresentInStore);
```

and using `let`

Classes/ContentRepository/Service/NodeService.php
```diff
@@ -86,36 +92,43 @@ public function getNodeFromContextPath($contextPath, Site $site = null, Domain $
         $nodePath = $nodePathAndContext['nodePath'];
         $workspaceName = $nodePathAndContext['workspaceName'];
         $dimensions = $nodePathAndContext['dimensions'];
+        $siteNodeName = $site ? $site->getNodeName() : explode('/', $nodePath)[2];
```
mhsdesign, line 95: this `explode('/'` will be impossible to upmerge to neos 9 ... but we wont need it there i guess as `currentSite` will not be needed so fine with me ;)

Classes/Controller/BackendServiceController.php
```diff
@@ -532,11 +532,13 @@ public function flowQueryAction(array $chain): string
         $createContext = array_shift($chain);
         $finisher = array_pop($chain);
 
+        $nodeContextPaths = array_unique(array_column($createContext['payload'], '$node'));
```
mhsdesign, line 536: i dont think we need the `array_unique` here as flowquery does permit passing twice the same node, even if stupid. And on the calling site we already use a `Set`.

So in the end i guess we dont need this whole `array_column` change at all?

packages/neos-ui-guest-frame/src/initializeGuestFrame.js
```diff
@@ -60,13 +61,48 @@ export default ({globalRegistry, store}) => function * initializeGuestFrame() {
...
+            return false;
+        }
+        return true;
     });
 
+    // Load remaining list of not fully loaded nodes from the backend if there are any
+    const fullyLoadedNodesFromContent = notFullyLoadedNodeContextPaths.length > 0 ? (yield q(notFullyLoadedNodeContextPaths).get()).reduce((nodes, node) => {
+        nodes[node.contextPath] = node;
+        return nodes;
+    }, {}) : [];
```
mhsdesign: otherwise we modify the argument by reference and that is not the way a `reduce` should be used in js i fear? IDK

```suggestion
    const fullyLoadedNodesFromContent = notFullyLoadedNodeContextPaths.length > 0 ? (yield q(notFullyLoadedNodeContextPaths).get()).reduce((nodes, node) => ({
        ...nodes,
        [node.contextPath]: node
    }), {}) : [];
```

mhsdesign (commented):
Had one more look at the code

packages/neos-ui-guest-frame/src/initializeGuestFrame.js
Sebobo: Removed it

packages/neos-ui-guest-frame/src/initializeGuestFrame.js
Sebobo: My variant is faster and less memory intensive as it reuses an existing object and doesn't create new ones for each iteration. It probably make little difference here, but I try to stick with the way that's most efficient.

Benjamin-K: Yeah, you shouldn't, as this would update the object passed to the function. But in this case a new object is created and therefore I would prefer the original implementation if we look at performance.

Benjamin-K: But: why do we return an array instead of an object if `notFullyLoadedNodeContextPaths.length === 0`? This will result in different types...

Sebobo:
@mhsdesign are we finished here now? I merge the current 8.3 and the node updates on click disappeared again thx to Wilhelms fix.

packages/neos-ui-guest-frame/src/initializeGuestFrame.js
Sebobo: The return type was a mistake that Marc already fixed I think while we had a pairing session. Will check again

Sebobo:
@grebaldi @mhsdesign one thing this PR changes is that rendering the page content will not update the stored nodes anymore with the state from the rendered nodedata. 

It relies on the store which is mostly prefilled by the content tree before the content is rendered.
Afterwards switching between pages will not initiate new requests.

There it could be that changes of another user will be rendered in the content, but will not update the stored nodes.
Only refreshing the content tree will update those nodes, which is not intuitive. This could result in outdated information in the inspector if I'm not mistaken.

Option 1: We could let the content tree always load nodes from the server when switching pages.
Option 2: We introduce something like a timestamp or hash that tells us if our data is outdated for a page.

Wdyt?

grebaldi:
@Sebobo 

I think Option 1 sounds quite fair :)

Option 2 may save some HTTP requests, but I don't think it's worth the effort. 

To track content changes per document (via hash or timestamp), we would have to traverse each document's content tree anyway. In the best case we save an HTTP request, in the worst case, we do the traversal twice.

Sebobo:
@grebaldi @mhsdesign I removed some code in the content tree document change saga and this way the content tree always reloads the nodes and this way we have up-to-date content.
The UX is also still good, as the document node shows the spinner when switching between pages back and forth, but besides that, the content stays displayed and can be interacted with while the data is loading.

IMO we can merge then.

mhsdesign (approved):
Thanks for all the effort

PR #3773, opened 2024-04-30
WIP: Isolate Drawer menu from the rest of the UI

grebaldi:
previous: https://github.com/neos/neos-ui/pull/3682
depends on: #3792, #3804 

WIP.


mhsdesign:
~Okay after merging https://github.com/neos/neos-ui/pull/3792 and https://github.com/neos/neos-ui/pull/3804 i attempted to get this pr resolved of its conflicts. Now a rebase is a little cruel as there are lots of conflicts for the mentioned feature prs because this pr contains them too but in different commits. A merge worked just find but leaves a particular odd history obviously now. So im thinking of either squashing this pr when finished or work with some git magic to remove the other feature branch commits here ... or maybe dont take it too seriously like i do :D~

Edit because the commits were split cleanly i was able to reset hard to 9.0 and just cherry pick all the changes you did after the original translation changes were made. That effectively drops the double commits. To confirm that the changes are correct i did a diff to my merge approach and even found an error in my merge resolution: (i had kept the `loadImpersonateStatus` function of f65f3cb1b9e8facd59e2b00a0d1124d76513ade5) So now i pushed the rewrite and ist just the actual ca.20 commits instead of 60+ !!!

Aside from that i checked it out and fixed the linter and it compiles and from the looks of it works already. As this is wip there obviously still some todos in the code. Thanks again for your work as always!

PR #3776, opened 2024-05-11
TASK: Dont add useless node names on copy

mhsdesign:
Should have been adjusted via https://github.com/neos/neos-ui/pull/3515 but better late than sorry

**What I did**

**How I did it**

**How to verify it**


nezaniel (approved):
lgtm

PR #3786, opened 2024-05-18
FEATURE: Level up to PHPStan 8

mhsdesign:
This gives us proper type errors when doing things on possible null values which is super important.

Following (non behaviour changing) adjustments were made:
- Handle nullability correctly
- Make `ChangeInterface::getSubject` not nullable (as we always call `setSubject`, and simplify code not having to deal with the possible nullability here (which was also not done previously correctly for all cases)
  - (Until commit https://github.com/neos/neos-ui/pull/3786/commits/716966726afc1d131890a2c52055ee46d0847848 i tried to go down the route of always handling the possible null case, which was started from sebastianK with the ESCR adjustments, but its just unnecessary complex and gives no advantage to have a `Change` without context node. Its useless)
- Make some internal helpers `final` for sanity
- wrap `$view->setControllerContext` (for out of band rendering) into a `method_exists` as the method is with FLow9 not part of the interface anymore
- Declare workspaceName of `UpdateWorkspaceInfo` non nullable as its set always in the constructor
- Declare node of `UpdateNodeInfo ` non nullable as its set always via setter

### DIFF
--- a/Classes/Controller/BackendServiceController.php
+++ b/Classes/Controller/BackendServiceController.php
@@ -418,6 +418,7 @@ class BackendServiceController extends ActionController
         $nodeAddressFactory = NodeAddressFactory::create($contentRepository);
 
         $currentAccount = $this->securityContext->getAccount();
+        assert($currentAccount !== null);
         $userWorkspaceName = WorkspaceNameBuilder::fromAccountIdentifier(
             $currentAccount->getAccountIdentifier()
         );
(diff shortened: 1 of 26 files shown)


Classes/Controller/BackendServiceController.php
```diff
@@ -476,24 +478,28 @@ public function changeBaseWorkspaceAction(string $targetWorkspaceName, string $d
         // todo ensure that https://github.com/neos/neos-ui/pull/3734 doesnt need to be refixed in Neos 9.0
         $redirectNode = $documentNode;
         while (true) {
```
mhsdesign, line 480: this code is probably broken but i didnt touch it here ... this deserves a follwup -> https://github.com/neos/neos-ui/issues/3815

grebaldi (approved):
Thanks so much for sustaining the type-safety efforts, @mhsdesign! :+1: :+1: :+1:

Issue #3771, opened 2024-04-25
TASK: Replace `NeosUiDefaultNodesOperation` with `ReloadNodesQueryHandler`


grebaldi:
_Originally posted by @mhsdesign in https://github.com/neos/neos-ui/pull/3759#discussion_r1579424028_:
>              ... but we dont seem to remove the NeosUiDefaultNodesOperation yet?

PR #3789, opened 2024-05-25
FEATURE: Find nodes by their id in Neos Backend

bwaidelich:
Extends the custom `SearchOperation` such that it respects node aggregate ids

Related: #3771


bwaidelich:
I have no clue why the e2e tests fail..

Classes/FlowQueryOperations/SearchOperation.php
```diff
@@ -76,17 +73,24 @@ public function evaluate(FlowQuery $flowQuery, array $arguments): void
         /** @var Node $contextNode */
         $contextNode = $context[0];
         $subgraph = $this->contentRepositoryRegistry->subgraphForNode($contextNode);
+        $matchingNodeByAggregateId = [];
```
grebaldi:
```suggestion
        $matchingNodeByAggregateId = null;
```

dlubitz (approved):

mhsdesign:
im still confused what the reason for the `search` operation is in general but it seems to be used via javascript:

https://github.com/neos/neos-ui/blob/9e47fdfd4d8ef345f75cf88ce34c178c0b161637/packages/neos-ui-sagas/src/UI/PageTree/index.js#L163

bwaidelich:
Oh my, repeating history.. #4093 is nicer IMO

PR #3759, opened 2024-03-19
!!!FEATURE: Lock the UI and present a modal during publish/discard

grebaldi:
<div align=center>

<table>
<tbody>
<tr>
<td></td>
<td></td>
</tr>
</tbody>
</table>
</div>

solves: #3743
builds on:  #3752

The PublishDiscardDialog is a modal that is supposed to pop up whenever an editor starts publishing or discarding their workspace.

Editors need to confirm their intent, after which they are presented with a process indicator that blocks the screen as long as the remote operation is not finished yet.

Finally, Editors see a success message or an error message - depending on the outcome. In both cases, they need to acknowledge the info. If they see an error message, they are also given an opportunity to retry the action.

**Publish Workflow**

<details>
<summary>Show Video</summary>

https://github.com/neos/neos-ui/assets/2522299/e7e7f54a-406f-4400-b59b-96796abee64a

</details>

For publishing, users are now first presented with a confirmation dialog. Formerly, publishing didn't show this dialog. This PR makes both `Publish` and `Discard` consistent in this regard (as you wouldn't want to accidentally publish anything either). 

If the user confirms the action, they see an uncancellable modal that illustrates the ongoing operation. Formerly, the operation was only indicated by a spinning icon alongside the `PublishingDropdown` head. The UI remained usable during the operation, which could potentially lead to invalid actions. Now, the UI is locked, including a `beforeunload` handler that warns the user if they're about to reload or leave the page.

The progress indicator remains open for as long as any async operation is still running. This includes the publishing operation itself, but also the adjacent reload of nodes and the reload of the guest frame. Only after the UI has reached a stable and consistent state, the lock is released.

Finally, the user is presented with a dialog that lets them acknowledge the operation. Formerly, a successful publishing process would have been indicated by a fleeting flash message. The acknowledgement dialog remains open until the user has taken notice. This way users can safely leave the tab, do something else, and return without losing the context of what their last action has been.

**Discard Workflow**

<details>
<summary>Show Video</summary>

https://github.com/neos/neos-ui/assets/2522299/2d5e0453-923d-496d-a5f1-ad4e3f27b4ea

</details>

The `Discard` workflow behaves exactly like the `Publish` workflow. The only difference is in the coloring and the progress indicator illustration.

**Error Handling**

<details>
<summary>Show Video</summary>

https://github.com/neos/neos-ui/assets/2522299/1cccee8f-0382-488f-b2e1-10f9e20e1f0f

</details>

[...]

PR #3792, opened 2024-05-30
!!!FEATURE: Centralize error handling

grebaldi:
prelude to: https://github.com/neos/neos-ui/pull/3773

The `@neos-project/neos-ui-error` package had been introduced in https://github.com/neos/neos-ui/pull/3759. During the discussion of that PR, @mhsdesign suggested it would be a good idea to centralize other error-handling mechanisms in that package (namely the high-level `ErrorBoundary`, see: https://github.com/neos/neos-ui/pull/3759#discussion_r1575109688).

As I was working on https://github.com/neos/neos-ui/pull/3773, I realized that I would need to decouple FlashMessages from the redux store and expose a different mechanism to use them. So I decided to take the centralization step first.

Bottom-line, this PR moves the following concepts into `@neos-project/neos-ui-error`:

- The high-level `<ErrorBoundary/>`-component (introduced in https://github.com/neos/neos-ui/pull/3485)
  - Catches errors that may occur *during* React rendering (anywhere in the component tree)
- The `terminateDueToFatalInitializationError`-function (introduced in https://github.com/neos/neos-ui/pull/3682)
  - Catches errors that may occur *before* any React component can be rendered
- FlashMessages (have been there since ever)
  - Are used for handling anticipated, user-facing errors in various places

The `@neos-project/neos-ui-error` package now exposes a `showFlashMessage` function that can be used to display FlashMessages from anywhere in the system. The state of FlashMessages is now decoupled from the redux store. It is handled in a global, singleton `Observable` instead.

To achieve this, I introduced a new state handling mechanism via the `@neos-project/framework-observable` package using the approach I had outlined in https://github.com/neos/neos-ui/pull/3331.

Thus far, I have implemented the `createObservable` and `createState` mechnisms with additional tests and documentation. Later on, this package will allow us to replace redux opportunistically where needed.

This change is breaking, because I have removed the now obsolete parts of the redux store that were related to flash messages. It is unlikely that third parties are relying on the parts I've removed, since they are implementation details related to the read-side (So unless a plugin needs to know what flash messages are currently displayed - which I doubt -, this should be safe).

The only thing still intact is the `UI.FlashMessages.add` action, because plugins may actually use it and currently have no other way to display flash messages. I have implemented a saga to redirect those action calls to the new mechanism.

My plan is to expose `showFlashMessage` publicly via a global JS API, that I will introduce in https://github.com/neos/neos-ui/pull/3773. In that PR I will then deprecate the intermediate saga and the `UI.FlashMessages.add` action.


packages/neos-ui-error/src/container/FlashMessages/FlashMessage.tsx
```diff
@@ -0,0 +1,82 @@
...
+    const isSuccess = severity === 'success';
+    const isError = severity === 'error';
+    const isInfo = severity === 'info';
+    const isClosing = React.useRef(false);
+    const handleClose = React.useCallback(() => {
+        const {onClose, id} = props;
+
+        if (!isClosing.current) {
+            isClosing.current = true;
+            setTimeout(() => onClose(id), 100);
```
mhsdesign, line 37: i do not quite grasp this timeout here and also was wondering if its ideal that we delegate this timeout logic to this nested component instead of handling it in a higher level ... but please dont see this as critique but rather me trying to learn the tricks ;)

packages/framework-observable/README.md
```diff
@@ -0,0 +1,152 @@
+# @neos-project/framework-observable
```
mhsdesign, line 1: thank you for providing extensive documentation for these framework parts.

I think we already partially discussed the idea of the "The Client-side framework" from https://github.com/neos/neos-ui/pull/3331:

> ### `@neos-project/framework-observable`
> 
> The `@neos-project/framework-observable` package provides a very simple implementation of the `Observable`-pattern (see: https://github.com/tc39/proposal-observable).
> 
> Furthermore, based on `Observable`s, it establishes two key primitives for use in the UI:
> 
> #### State
> 
> A `State` observable represents a value that changes over time. It can be asked about its current value and any subscriber will receive the current value immediately.
> 
> It also provides an `update` method, that allows to push the next value calculated from the current value.
> 
> #### Channel
> 
> A `Channel` observable represents a sort of message bus. A subscriber will receive all messages that have been published to the `Channel` since the subscription was established.
> 
> It provides a `publish` method to push messages onto the channel.
> 
> ### `@neos-project/framework-react-hooks`
> 
> The `@neos-project/framework-react-hooks` package contains a bunch of essential React hooks to allow binding state to React components.
> 
> Most notably it contains the `useLatestValueFrom` hook, that allows to bind `Observable`s to components.

And for the DTO serialisation and de-serialisation (`@neos-project/framework-schema`) you elaborated why it makes sense to provide a custom implementation instead of using a npm library: https://github.com/neos/neos-ui/pull/3331#issuecomment-1518032438

Regarding observers and state handling i remember having talked to you regarding using rx-js vs a custom implementation like `@neos-project/framework-observable` and i think the reasons for a custom implementation were customisability (limited feature scope we only need to implement what we need), smaller bundle size, and that  `Observable` will hopefully come to native ECMA script which we definitely want to profit from if available and this can be ensured by having it all under control.
But im a little fuzzy on the details and dont quite remember it all. So it would be great to discuss this to fill in the gaps. In the end this adds complexity to our codebase and thus the benefits and downsides should be carefully considered. When we decide for this library an outlined discussion would also help future maintainers to understand our decision ;) 

EDIT:
I just skimmed over the implementation and its a super slim layer so +1 from my side ;) And we could consider this dicussion closed ^^ The documentation and tests are more than the code :D

mhsdesign (commented):
Well done ❤️ especially thanks for keeping `actionTypes.UI.FlashMessages.ADD` in tact as we promote this as part of a guide "how to use neos ui redux" ^^  https://docs.neos.io/guide/advanced/extending-the-user-interface/react-extensibility-api#sagas

packages/neos-ui-error/src/container/FlashMessages/FlashMessage.tsx
grebaldi: I've abstained from refactoring this stuff. This is only copied over from the original implementation and I agree that it's a bit smelly, but I didn't want to do more than strictly necessary to achieve the architectural goals.

packages/framework-observable/README.md
grebaldi: You've summarized it perfectly :)

packages/neos-ui-error/src/container/FlashMessages/FlashMessage.tsx
mhsdesign: Perfect i oversaw that. In that case im even more grateful that its just moved as this should simplify the review :)

markusguenther:
This PR is now in sync with the latest 9.0 changes. Thanks for all the work @grebaldi ❤️

kitsunet (approved):
LGTM

mhsdesign (approved):
As discussed with christian and markus we like this change very much thanks :)

We have to keep in mind that moving back from redux of course means that the store is not easily debuggable via browser extension and that plugins cannot hook into the process. This means we likely end up with more explicit defined extension points instead of one big store. But for the ui is this exactly what wee need. Small components like flashmessage especially profit from that as there is absolutely no need for extensibiliity.

PR #3799, opened 2024-06-01
BUGFIX: `LinkEditor` placeholder now uses the custom placeholder correctly

### LINKED ISSUE #3797: BUG: placeholder for the `LinkEditor` cannot be set
### Description

If you try to use the `LinkEditor` you cannot set a custom placeholder. If you try to set one, the default placeholder gets replaced with another different fallback placeholder.

### Steps to Reproduce

1. create a new property in your `.yaml` file for any node-type, setting its `editor:` to `Neos.Neos/Inspector/Editors/LinkEditor`
2. set a placeholder
3. look at your configured node-type in the neos frontend

#### Expected behavior

When I set a custom placeholder through my `.yaml` configuration, I expect to see my custom placeholder in the frontend.

#### Actual behavior

If I set a custom placeholder, it is ignored and a static fallback placeholder is used instead.

### Affected Versions

Neos: -

UI: every Version since 11.09.2018, currently in 8.3

dr-format-d:
Fixes #3797

**What I did**
If the user now sets a placeholder for the `LinkEditor` the placeholder will now be correctly shown in the neos frontend.

**How I did it**
I removed a (imo) unnecessary condition, which caused the custom placeholder to be always overridden with a static placeholder.

**How to verify it**
Try to set a placeholder for the `LinkEditor` like you would/expect with any other editor component

**Before:**
Video before

**After:**
Video after

### DIFF
--- a/packages/neos-ui-editors/src/Library/LinkInput.js
+++ b/packages/neos-ui-editors/src/Library/LinkInput.js
@@ -297,7 +297,7 @@ export default class LinkInput extends PureComponent {
                     value={''}
                     plainInputMode={isUri(this.state.searchTerm)}
                     onValueChange={this.handleValueChange}
-                    placeholder={this.props.i18nRegistry.translate($get('options.placeholder', this.props))}
+                    placeholder={this.props.i18nRegistry.translate($get('options.placeholder', this.props) || 'Neos.Neos:Main:content.inspector.editors.linkEditor.search', $get('options.placeholder', this.props) || 'Paste a link, or type to search')}
                     displayLoadingIndicator={this.state.isLoading}
                     displaySearchBox={true}
                     setFocus={this.props.setFocus}
--- a/packages/neos-ui-editors/src/Library/LinkInput.js
+++ b/packages/neos-ui-editors/src/Library/LinkInput.js
@@ -297,7 +297,7 @@ export default class LinkInput extends PureComponent {
                     value={''}
                     plainInputMode={isUri(this.state.searchTerm)}
                     onValueChange={this.handleValueChange}
-                    placeholder={this.props.i18nRegistry.translate($get('options.placeholder', this.props) || 'Neos.Neos.Ui:Main:ckeditor__toolbar__link__placeholder', 'Paste a link, or search')}
+                    placeholder={this.props.i18nRegistry.translate($get('options.placeholder', this.props))}
                     displayLoadingIndicator={this.state.isLoading}
                     displaySearchBox={true}
                     setFocus={this.props.setFocus}


packages/neos-ui-editors/src/Library/LinkInput.js
```diff
@@ -297,7 +297,7 @@ export default class LinkInput extends PureComponent {
                     value={''}
                     plainInputMode={isUri(this.state.searchTerm)}
                     onValueChange={this.handleValueChange}
-                    placeholder={this.props.i18nRegistry.translate($get('options.placeholder', this.props) || 'Neos.Neos.Ui:Main:ckeditor__toolbar__link__placeholder', 'Paste a link, or search')}
+                    placeholder={this.props.i18nRegistry.translate($get('options.placeholder', this.props))}
```
grebaldi:
```suggestion
                    placeholder={this.props.i18nRegistry.translate($get('options.placeholder', this.props) || 'Neos.Neos.Ui:Main:ckeditor__toolbar__link__placeholder', $get('options.placeholder', this.props) || 'Paste a link, or search')}
```

Otherwise, if no manual placeholder has been configured, the placeholder ends up empty:
Screenshot_2024-06-03_13-13-22 LinkEditor empty

grebaldi (changes requested):
Hi @dr-format-d,

thanks a lot for your first PR to the Neos UI! :tada: 

I tested your change and it makes custom placeholders work correctly. However, I'll need a little adjustment (see comment below), so that it still works, when no custom placeholder is configured.

packages/neos-ui-editors/src/Library/LinkInput.js
dr-format-d: Thank you for reviewing! I applied the code you provided, except for the chosen label. I don't see why the label should change from: `Neos.Neos:Main:content.inspector.editors.linkEditor.search` to `Neos.Neos.Ui:Main:ckeditor__toolbar__link__placeholder` only in this case.

Consequently, i adjusted `Paste a link, or search` to `Paste a link, or type to search` to better fit the default placeholder configured here:
image

grebaldi: Well, it's not obvious, sadly, but the `<LinkInput/>` is used in two places. One is the LinkEditor (for which the `Neos.Neos:Main:content.inspector.editors.linkEditor.search` label is a fitting choice), the other is the RichTextEditorToolbar, which is why the fallback to `Neos.Neos.Ui:Main:ckeditor__toolbar__link__placeholder` exists.

But, I guess it's okay to unify this and put LinkEditor in the lead. We just gotta remember to clean up the other one :sweat_smile:

grebaldi (approved):

Sebobo (approved):
Thx for this fix! 🙂

PR #3802, opened 2024-06-05
TASK: adjust to overhauled node uri building

mhsdesign:
Ui adjustments for https://github.com/neos/neos-development-collection/pull/4892

- we use `Neos.Ui.NodeInfo.previewUri` in eel instead of `uri` as the `documentNode` will always be in a non live workspace either way and to better symbolise that we want to render a ABSOLUTE preview uri. The `uri` method could be removed.
- the `redirectTo` action of the `Neos.Neos.Ui` `BackendController` will already work with the new node address
  - `Neos.Ui.NodeInfo.createRedirectToNode` redirected in 8.3 to the node in the base workspace, this behaviour was restored as currently the user workspace is targeted which would be new accidental? behaviour change
- replaced last reference of `LinkingService::createNodeUri` with new uri builder

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/Classes/Controller/BackendController.php
+++ b/Classes/Controller/BackendController.php
@@ -228,7 +228,7 @@ class BackendController extends ActionController
         $nodeAddress = NodeAddress::fromJsonString($node);
 
         $this->redirectToUri(
-            $this->nodeUriBuilderFactory->forRequest($this->request->getHttpRequest())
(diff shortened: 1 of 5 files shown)


mhsdesign:
Blocked by https://github.com/neos/neos-development-collection/issues/5150 as this crashes the E2E tests

PR #3804, opened 2024-06-14
!!!FEATURE: Reform i18n mechanism

grebaldi:
prelude to: https://github.com/neos/neos-ui/pull/3773

This PR reforms the UI's i18n mechanism by exposing a new function `translate` directly from the `@neos-project/neos-ui-i18n` package.

The function looks as follows:
```typescript
function translate(
    fullyQualifiedTranslationAddressAsString: string,
    fallback: string | [string, string],
    parameters: Parameters = [],
    quantity: number = 0
): string;
```

And can be used like this:
```typescript
translate(
  'Neos.Neos:Main:workspaces.allChangesInWorkspaceHaveBeenDiscarded',
  'All changes from workspace "{0}" have been discarded.',
  ['user-admin']
);

// output (en): All changes from workspace "user-admin" have been discarded.
```

The new mechanism is completely independent from registries. There's no longer a need for injecting the `I18nRegistry` to use translations, instead `translate` can be used directly.

`translate` will only work after the translations have been loaded during the bootstrapping process of `@neos-project/neos-ui`. If it is used too early, it throws an exception.

The function now also properly supports plural rules (via [`Intl` Web API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl)). Tha `I18nRegistry` had only supported English plurals, but with broken rules (quantity=0 yielded singular instead of plural).

Other changes included in this PR are:

- Everything in `@neos-project/neos-ui-i18n` is now TypeScript (save `manifest.js`)
- The `<I18n/>`-component was deprecated
- The `I18nRegistry` is now properly documented, but was also deprecated
- The `xliff.json` endpoint is now discovered via `<link>`-tag, instead of initial data
  - The `<link>`-tag is marked with `prefetch` to slightly speed up the fetch request
- The `@neos-project/neos-ui-i18n` now has a comprehensive README.md

This change is breaking, because translations may have relied on the broken plural rules. It is also breaking, because the signature for substitution parameters has changed slightly - although this should be non-breaking at runtime.


kitsunet:
I like this, but wow it's also A LOT. Thanks for digging in!

markusguenther:
Rebased to latest 9.0 state.

markusguenther:
Locally the e2e tests run fine:
```
 Create new nodes
 - Create content collection node
 - Create text node in container
 ✓ Create a text node in a new container element at the correct position

 Create new nodes
 ✓ Check ClientEval for dependencies between properties of NodeTypes in Creation Dialog
 - Open create dialog node
 - Open context help and check for Markdown rendering
 ✓ Check the nodetype help in create dialog
 ✓ Check that nodetype without help has no help button
 - Create Image node
 - Select image from media library
 ✓ Create an Image node from ContentTree
 ✓ Can create a new page
 - Create a headline node
 - Type something inside of it
 - Inline validation
 - Skipped: Create a link to node
 ✓ Can create content node from inside InlineUI
 - Create an inline headline node
 - Insert text into the inline text and press enter
 ✓ Inline CKEditor mode `paragraph: false` works as expected
 ✓ Supports secondary inspector view for element editors

 Discarding
 - Create a document node
 - Create another node inside it
 - Discard all nodes and hope to be redirected to root
 ✓ Discarding: create multiple nodes nested within each other and then discard them
 - Create a document node
 - Discard that node
 ✓ Discarding: create a document node and then discard it
 - Navigate via the page tree
 - Delete that page
 - Discard page deletion
 ✓ Discarding: delete a document node and then discard deletion
 - Create a content node
 - Discard that node
 ✓ Discarding: create a content node and then discard it
 - Delete this headline
 - Discard page deletion
 ✓ Discarding: delete a content node and then discard deletion

 Image editor
 ✓ Can crop an image

 Inspector
 - Rename Discarding page via inspector
 - Test unapplied changes dialog - resume
 - Test unapplied changes dialog - discard
 - Test unapplied changes dialog - apply
 ✓ Can edit the page title via inspector
 ✓ Check ClientEval for dependencies between properties of NodeTypes in Inspector

 InspectorValidation
 - Remove homepage title
 - Check error badge for one error
 ✓ Remove homepage title to get one error
 - Clean title and uri path segment field
 - Check error badge for two errors
 ✓ Remove Page title and URI segment to get two errors
 - Remove homepage title
 - Check error badge for one error
 - Enter new title Home
 ✓ Remove homepage title to get one error and resolve error by new title

 FIX #3184: Discarded node move changes are reflected correctly in the document tree
 ✓ Scenario #1: Moving nodes and then discarding that change does not lead to an error
 ✓ Scenario #2: Moved nodes do not just disappear after discarding the move change
 ✓ Scenario #3: Discarding a move change while being on a moved node does not lead to an error in the guest frame

 TODO Reimplement / Fix Node Tree Presets for 9.0 https://github.com/neos/neos-ui/issues/3702
 - Node tree preset "default" removes all blog related nodes and only loads nodes with depth <= 2
 - Node tree preset "blog" shows nothing but page [🗋 Blog]
 - In node tree preset "blog", page [🗋 Blog] has no toggle handle
 - Reloading the node tree while in preset "blog" results in nothing but page [🗋 Blog]
 - Node tree preset "blog-articles" shows page [🗋 Blog] and all articles beneath it
 - BUG #3816: Switching back from node tree preset "blog" does not affect loading children in node tree preset "default"
 - BUG #2583: Searching the document tree does not break expansion in node tree preset "default"
 - BUG #2800 1/2: Moving pages before/after in a filtered view does not lead to the disappearance of nodes
 - BUG #2800 2/2: Moving pages into each other in a filtered view does not break expansion

 Refresh Document Tree
 - Refresh the Document Tree and wait
 - Rename home page via inspector
 ✓ Refresh Document tree

 Rich text editor
 ✓ Can crop an image

 Select Boxes
 - SelectBox contents open below the SelectBox.
 ✓ SelectBox opens below and breaks out of the creation dialog if there's enough space below.
 - SelectBox contents open above if the SelectBox is just above the screen bottom.
 - SelectBox contents disappear when SelectBox is scrolled out of sight.
 ✓ SelectBox opens above in creation dialog if there's not enough space below.
 - SelectBox contents open above if the SelectBox is just above the screen bottom.
 - When the inspector tab panel is scrolled just enough, so that there's enough space, SelectBox contents jump below the SelectBox.
 ✓ SelectBox opens above in inspector if there's not enough space below.

 Sidebar toggle
 - LeftSideBar
 - RightSideBar
 ✓ Can toggle sidebars

 Switching dimensions
 - Navigate to some inner page and switch dimension
 - Switch back to original dimension
 ✓ Switching dimensions
 ✓ Grouping of dimensions

 Syncing
 - Create a headline node in the document
 - Type something inside of it
 ✓ Syncing: Create a conflict state between two editors and choose "Discard all" as a resolution strategy during rebase
 - Create a headline node in the document
 - Type something inside of it
 ✓ Syncing: Create a conflict state between two editors and choose "Drop conflicting changes" as a resolution strategy during rebase
 - Create a headline node in the document
 - Type something inside of it
 ✓ Syncing: Create a conflict state between two editors, start and cancel resolution, then restart and choose "Drop conflicting changes" as a resolution strategy
 during rebase
 - Create a headline node in the document
 - Type something inside of it
 ✓ Syncing: Create a conflict state between two editors and choose "Drop conflicting changes" as a resolution strategy, then cancel and choose "Discard all" as a
 resolution strategy during rebase
 - Create a headline node in the document
 - Type something inside of it
 ✓ Publish + Syncing: Create a conflict state between two editors, then try to publish and choose "Drop conflicting changes" as a resolution strategy during automatic
 rebase
 - Create a headline node in the document
 - Type something inside of it
 ✓ Publish + Syncing: Create a conflict state between two editors, then try to publish the document only and choose "Drop conflicting changes" as a resolution
 strategy during automatic rebase

 Tree multiselect
 ✓ Move multiple nodes via toolbar
 ✓ Move multiple nodes via DND, CMD-click
 ✓ Move multiple nodes via DND, SHIFT-click

 Tree search
 - Search the page tree
 - Set the Shortcut filter
 - Clear search
 - Clear filter
 ✓ PageTree search and filter
 - Search is initially hidden and can be opened
 - Close the search input again
 ✓ Pagetree search field can be toggled
 - Search is initially hidden and we open it
 - We reload the page
 - We expect to have a visible search field
 ✓ Pagetree search field state is stored
 - Search is initially hidden and we open it with "t s"
 ✓ Pagetree search field toggles on hotkey


 44 passed (8m 58s)
 9 skipped
```

packages/neos-ui-i18n/README.md
mhsdesign, line 1: lovely inline documentation thanks ❤️ !11!!!

packages/neos-ui-i18n/src/global/initializeI18n.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+        };
+        jest.spyOn(global, 'fetch' as any).mockImplementation(server as any);
+    });
+
+    afterEach(() => {
+        jest.resetAllMocks();
+        teardownI18n();
+    });
+
+    it('loads the translation from the location specified in the current HTML document', async () => {
```
mhsdesign, line 47: tests 💯 :D

packages/neos-ui-i18n/src/model/Locale.ts
```diff
@@ -0,0 +1,66 @@
...
+    public getPluralFormIndexForQuantity(quantity: number): number {
+        return this.pluralRules.getIndexOf(
+            PluralRule.fromString(
+                this.intlPluralRules.select(quantity)
+            )
+        );
+    }
+}
+
+export class InvalidLocale extends Error {
```
mhsdesign, line 53: i like that you adapted the php code style regarding exceptions with factory methods like `InvalidLocale.becauseOfInvalidIdentifier()` as proposed and used in the ESCR https://discuss.neos.io/t/rfc-updated-default-code-style/5836!!

packages/neos-ui-i18n/src/model/Locale.ts
```diff
@@ -0,0 +1,66 @@
...
+    private readonly intlPluralRules: Intl.PluralRules;
+
+    private constructor(
+        private readonly intlLocale: Intl.Locale,
+        private readonly pluralRules: PluralRules
+    ) {
+        this.intlPluralRules = new Intl.PluralRules(this.intlLocale.toString());
+    }
+
+    public static create = (identifier: string, pluralRulesAsString: string): Locale => {
```
mhsdesign, line 23: this is basically how we write value objects in php too !!! this is a great direction :D

packages/neos-ui-i18n/src/model/TranslationRepository.ts
```diff
@@ -0,0 +1,44 @@
...
+ * source code.
+ */
+import {Locale} from './Locale';
+import {Translation, type TranslationDTO} from './Translation';
+import type {TranslationAddress} from './TranslationAddress';
+
+export type TranslationsDTO = Record<string, Record<string, Record<string, TranslationDTO>>>;
+
+export class TranslationRepository {
+    private _translationsByAddress: Record<string, null | Translation> = {};
```
mhsdesign: a little nitpick: i  think the underscore is obsolete in combination with the `private` visibility. Even its js i like to refer to psr12 which denotes that underscores should not be used to denote visibility.

mhsdesign (approved):
Thank you so much for refactoring the translation functionality.

When reading the code i really enjoyed finding common modern php patterns now reflected here in the code base like immutable value objects, collection of vo's, enums, explicit exceptions, etc (and also repositories) while on the other hand embracing functions when applicable like the main `translation` function which is just that - a function - not a service, registry or anything which is quite how javascript was indented. I think this unique mix is a really good pattern for the future and this example i18n implementation gives a good boilerplate for future similar endpoints or functionality.

For a moment i was wondering if we had to respect the `Neos.Neos.userInterface.scrambleTranslatedLabels` option but funnily this is sorely handled by the `XliffService` and thus the rendered json will already be containing the `#######` scrambles.

Soo much work and a well designed model :D Thanks!

packages/neos-ui-i18n/src/model/TranslationRepository.ts
markusguenther: Yes, the convention of using an underscore (_) prefix for private variables, like _translationsByAddress, is considered relatively outdated in modern TypeScript. But with a huge PR like that, it is kind a harsh to outline this :)

markusguenther (approved):

packages/neos-ui-i18n/src/model/TranslationRepository.ts
mhsdesign: > But with a huge PR like that, it is kind a harsh to outline this :)

jup 😂 evil me 😈

mhsdesign:
In https://github.com/neos/neos-ui/pull/3907 i made some followup up adjustments to ensure the old i18n with its forgiving behaviour is kept in the old places. Im not really sure how we want to replace that but i think we need a  `tryTranslate` method in cases its a user label as the ui should never crash for those cases.

PR #3809, opened 2024-06-21
BUGFIX: Let document tree searchbox scale with sidebar width

Sebobo:
**What I did**

Let the search box use the space available when the left sidebar is resized

**How I did it**

Replace fixed width with flex value.

**How to verify it**

Before:

CleanShot 2024-06-21 at 13 37 40@2x

After

CleanShot 2024-06-21 at 13 37 27@2x


mhsdesign (approved):
👀

PR #3810, opened 2024-06-21
FEATURE: Separated references and properties in Neos UI

pKallert:
**What I did**
Currently, we are merging the references into the properties in the Neos UI. Since they are now separated into properties: and references: in the config, we should also handle them as such in the UI. 

The logic for the core is changed here: https://github.com/neos/neos-development-collection/pull/5156

### DIFF
--- a/packages/neos-ts-interfaces/src/index.ts
+++ b/packages/neos-ts-interfaces/src/index.ts
@@ -147,16 +147,8 @@ export interface ReferencesConfiguration {
     ui?: {
         label?: string;
         reloadIfChanged?: boolean;
-        inline?: {
-            editor?: string;
-            editorOptions?: {
-                [propName: string]: any;
-            };
-        }
-        inlineEditable?: boolean;
         inspector?: {
             hidden?: boolean;
-            defaultValue?: string;
             editor?: string;
             editorOptions?: {
                 [propName: string]: any;
@@ -168,9 +160,6 @@ export interface ReferencesConfiguration {
             message?: string;
             thumbnail?: string;
         };
-    };
-    validation?: {
-        [propName: string]: ValidatorConfiguration | undefined;
     };
     properties?: {
         [propName: string]: PropertyConfiguration | undefined;
--- a/packages/neos-ts-interfaces/src/index.ts
+++ b/packages/neos-ts-interfaces/src/index.ts
@@ -161,9 +161,6 @@ export interface ReferencesConfiguration {
             thumbnail?: string;
         };
     };
-    properties?: {
-        [propName: string]: PropertyConfiguration | undefined;
-    };
 }
 export interface NodeType {
     name?: string;
@@ -238,7 +235,7 @@ export interface NodeType {
         [propName: string]: PropertyConfiguration | undefined;
     };
     references?: {
-        [propName: string]: ReferencesConfiguration | undefined;
+        [referenceName: string]: ReferencesConfiguration | undefined;
     };
 }
 
--- a/packages/neos-ts-interfaces/src/index.ts
+++ b/packages/neos-ts-interfaces/src/index.ts
@@ -142,7 +142,40 @@ export interface PropertyConfiguration {
         [propName: string]: ValidatorConfiguration | undefined;
     };
 }
-
+export interface ReferencesConfiguration {
+    type?: string;
+    ui?: {
+        label?: string;
+        reloadIfChanged?: boolean;
+        inline?: {
+            editor?: string;
+            editorOptions?: {
+                [propName: string]: any;
+            };
+        }
+        inlineEditable?: boolean;
+        inspector?: {
+            hidden?: boolean;
+            defaultValue?: string;
+            editor?: string;
+            editorOptions?: {
+                [propName: string]: any;
+            }
+            group?: string;
+            position?: number | string;
+        };
+        help?: {
+            message?: string;
+            thumbnail?: string;
+        };
+    };
+    validation?: {
+        [propName: string]: ValidatorConfiguration | undefined;
+    };
+    properties?: {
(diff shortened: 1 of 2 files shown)


dlubitz (approved):
LGTM!

packages/neos-ts-interfaces/src/index.ts
```diff
@@ -142,7 +142,40 @@ export interface PropertyConfiguration {
...
+        inline?: {
+            editor?: string;
+            editorOptions?: {
+                [propName: string]: any;
+            };
+        }
+        inlineEditable?: boolean;
+        inspector?: {
+            hidden?: boolean;
+            defaultValue?: string;
```
mhsdesign:
```suggestion
```

not legit options for references i think

packages/neos-ts-interfaces/src/index.ts
```diff
@@ -142,7 +142,40 @@ export interface PropertyConfiguration {
...
+            position?: number | string;
+        };
+        help?: {
+            message?: string;
+            thumbnail?: string;
+        };
+    };
+    validation?: {
+        [propName: string]: ValidatorConfiguration | undefined;
+    };
```
mhsdesign:
```suggestion
```

not legit options for references i think

packages/neos-ts-interfaces/src/index.ts
```diff
@@ -142,7 +142,40 @@ export interface PropertyConfiguration {
...
+    properties?: {
```
mhsdesign: is this a new feature ✨Native support in the Neos ui?? :D Hurray

packages/neos-ts-interfaces/src/index.ts
```diff
@@ -142,7 +142,40 @@ export interface PropertyConfiguration {
...
+    ui?: {
+        label?: string;
+        reloadIfChanged?: boolean;
+        inline?: {
+            editor?: string;
+            editorOptions?: {
+                [propName: string]: any;
+            };
+        }
+        inlineEditable?: boolean;
```
mhsdesign:
```suggestion
```

not legit options for references i think

packages/neos-ui-contentrepository/src/registry/NodeTypesRegistry.ts
```diff
@@ -215,6 +218,20 @@ export default class NodeTypesRegistry extends SynchronousRegistry<NodeType> {
                                     helpThumbnail: property.ui?.help?.thumbnail
                                 })
                             ),
+                        ...references.filter(p => p.ui?.inspector?.group === group.id)
+                                .map(property => ({
```
mhsdesign:
```suggestion
                                .map(reference => ({
```

... and the rest :D

packages/neos-ui-contentrepository/src/registry/NodeTypesRegistry.ts
```diff
@@ -215,6 +218,20 @@ export default class NodeTypesRegistry extends SynchronousRegistry<NodeType> {
...
+                                    type: 'editor',
+                                    id: property.id,
+                                    label: property.ui?.label,
+                                    editor: property.ui?.inspector?.editor,
+                                    editorOptions: property.ui?.inspector?.editorOptions,
+                                    position: property.ui?.inspector?.position,
+                                    hidden: property.ui?.inspector?.hidden,
+                                    helpMessage: property.ui?.help?.message,
+                                    helpThumbnail: property.ui?.help?.thumbnail,
+                                    properties: property?.properties
```
mhsdesign: ah we just expose it here but its not interpreted anywhere? I think we should use a better term here ...

mhsdesign (commented):
Ah i see its indeed doable :D

packages/neos-ui-contentrepository/src/registry/NodeTypesRegistry.ts
pKallert: Yes, would be a follow-up to use it in the editor, for now it is not used anywhere. 

Would perhaps edgeProperties be better?

packages/neos-ts-interfaces/src/index.ts
pKallert: Well, rather the preparation for using edge properties 😄

packages/neos-ui-contentrepository/src/registry/NodeTypesRegistry.ts
mhsdesign: or `referenceProperties`? Maybe introduce this thing first with the mentioned followup and not now? ... on the other hand its just the yaml mapped 1to1 so we should probably not mess with it too much? ... but its not 1to1 really so maybe `referenceProperties`? :D

pKallert: `referenceProperties` works for me, it is also used fairly often in the documentation: https://docs.neos.io/guide/content-repository/nodetype/references-vs-properties 
I can also remove it for now, it does not really make a difference if we add it now or later.

mhsdesign: > it does not really make a difference if we add it now or later.

true in this case its simple ... but id do it from time to time to delay the discussion and let the reviewers focus more on the core parts :D ... Id vote for removing it now as without using this thing we dont know what the editor actually needs.
For example this pr would just pass the properties 1to1 but maybe they should also be mapped and get some view configuration? For people already wanting to experiment, there are alternative ways to fetch the reference properties by getting the js nodetype so i think that would be fine :D

pKallert: I removed it for now 👍

mhsdesign (approved):
Thanks looks perfect ;) ... but we should wait with merging for the neos change ofc

PR #3813, opened 2024-06-24
!!! FEATURE: Separate properties and references in clientEval

pKallert:
**What I did**
As discussed in https://github.com/neos/neos-development-collection/pull/5156 references should be available separately in ClientEval 

**How I did it**

I adjusted the logic for loading data for the right sidebar to be separated into properties and references. 

depends on https://github.com/neos/neos-ui/pull/3810  

**How to verify it**

Use ClientEval:console.log(node) in the right to see that references are no longer in the properties array but now their separate item: 
Bildschirmfoto 2024-06-24 um 18 01 14


grebaldi (approved):
Nice one, thx @pKallert! :+1: by :eyes:

mhsdesign:
Looks good sofar but ill have to try this out ;) It seems that with the transientValues we still mix properties and references but that seems sensible ^^

packages/neos-ui/src/Containers/RightSideBar/Inspector/InspectorEditorEnvelope/index.js
```diff
@@ -67,8 +67,7 @@ export default class InspectorEditorEnvelope extends PureComponent {
         //
         // nodeType needs to be read directly from node
         //
-        const sourceValueRaw = id === '_nodeType' ? node?.nodeType : node?.properties?.[id];
-        const sourceValue = sourceValueRaw;
+        const sourceValue = id === '_nodeType' ? node?.nodeType : (node?.references?.[id] ? node?.references?.[id] : node?.properties?.[id]);
```
mhsdesign, line 70: its a bit random that the source value is fetched here inside the component and not from the outside ... are there other places like this?

mhsdesign (changes requested):
image

i think i found a bug: the neos ui doesnt notice the deletion of a reference property. It will  show up again after hitting apply.

this could be related to logic like this https://github.com/neos/neos-ui/blob/60262de6e4cded95ddc1269e356cfd020bd39402/packages/neos-ui-sagas/src/Changes/index.js#L28-L38 or similar that relies on `node.properties`

PR #3817, opened 2024-06-26
FEATURE: Speed up UI flow queries with custom queries

Sebobo:
**What I did**

This is a follow up for #3770 in which we load all nodes requested by the UI directly from the database instead of querying each node by itself and its document and content children separately.

**How I did it**

Extract some methods from the NodeDataRepository and added them to the NodeService in the UI to resolve all requested nodes with one db query and prefetch all child nodes of the resulting nodes to fill them into the 1st-level cache.

**How to verify it**

Flow query performance for loading a large page in a large project with the page tree, content tree and nodes required for the displayed content. The page tree requests 175 nodes (the 332KB request), the content tree requests 58 nodes ( the 129kb request) and the guest frame requests 191 nodes (the 615kb request) which haven't been loaded by the content tree yet and with the duplicates removed.

Performance UI 8.3.8:
neosui-83-838

Performance with branch #3770:
neosui-83-remove-ui-script-tag

Performance with this branch:
neosui-83-optimise-ui-flowqueries

### DIFF
--- a/Classes/Aspects/AugmentationAspect.php
+++ b/Classes/Aspects/AugmentationAspect.php
@@ -40,21 +40,9 @@ class AugmentationAspect
 
     /**
      * @Flow\Inject
-     * @var UserLocaleService
-     */
-    protected $userLocaleService;
-
-    /**
-     * @Flow\Inject
      * @var HtmlAugmenter
      */
     protected $htmlAugmenter;
-
-    /**
-     * @Flow\Inject
-     * @var NodeInfoHelper
-     */
-    protected $nodeInfoHelper;
 
     /**
      * @Flow\Inject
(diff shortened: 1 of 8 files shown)


Sebobo:
Have to rebase, something is off with all those changes

Sebobo:
Also have to check whether I broke the "presets" feature of the document tree

Sebobo:
After discussing this a bit with @mhsdesign, it can make sense to add a "findNodesByIdentifiers" to the NodeDataRepository for 8.4 and use the new method instead of adding queries in this PR. This could also be useful for Neos 9. I will open an issue and link it.

Sebobo:
Needs a rebase when #3770 was unmerged into 8.4

Sebobo:
I'll close this as #3961 already fixes most of the problems without adding more complexity in comparison to this change.

PR #3821, opened 2024-07-02
FEATURE: Allow to set maximum rows to restrict height of the textarea editor in inspector

dlubitz:
The textarea editor is not restricted in height. If you need to add a lot of content it grows as long as the content is. This PR allows to set a maximum rows and defines a default of max 24 rows.


dlubitz:
Related documentation: https://github.com/neos/neos-development-collection/pull/5170
Related schema: https://github.com/Sebobo/Shel.Neos.Schema/pull/23

Sebobo (approved):
LGBR

IMO a default of 10 would also be fine.

pKallert (approved):
Tried it out locally and it works as expected 😃 

We could also think about adjusting Neos Demo to use this in the contact form message

Issue #3702, opened 2024-01-30
Neos 9.0 Either reimplement or fully remove `presetBaseNodeType`


mhsdesign:
currently this feature is broken see https://neos-project.slack.com/archives/C0L41E268/p1698743660587269

basti wrote on slack

> Well, partly.. If a different preset is selected, the tree doesn't work anymore – but that can be fixed in UI I assume.

```
Neos:
  Neos:
    userInterface:
      navigateComponent:
        nodeTree:
          loadingDepth: 2
          presets:
            default:
              baseNodeType: 'Neos.Demo:Document.Blog,Neos.Demo:Document.BlogPosting'
            'not blog':
              ui:
                label: 'Not Blog'
                icon: 'icon-book'
              baseNodeType: '!Neos.Demo:Document.Blog,!Neos.Demo:Document.BlogPosting'
```

> Well, partly.. If a different preset is selected, the tree doesn't work anymore – but that can be fixed in UI I assume.

![image](https://github.com/user-attachments/assets/70173d8b-1a0a-4f81-9568-49075ace822d)

mhsdesign:
With https://github.com/neos/neos-ui/pull/3822 we introduced new e2e tests (nodeTreePresets.e2e.js) which have to be disabled at first as the feature has to be reimplemented first.

PR #3822, opened 2024-07-02
BUGFIX: Fix inconsistencies in document tree with node tree presets

### LINKED ISSUE #2583: Searching for children of already expanded nodes breaks expansion after removing search term
### Description

When searching for a node that is on the third level inside of the node tree, upon removing the search term, the parent node on the first level cannot be expanded again.

### Steps to Reproduce

1. Configure nodeTree loadingDepth to be 1
```
Neos:
  Neos:
    userInterface:
      navigateComponent:
        nodeTree:
          loadingDepth: 1
```
2. Expand a node that has two levels of children
3. Open the search input
4. Search for a node that is two levels below the collapsed node
5. Remove search term
6. Expand the same node again

#### Expected behavior

The node should expand again, showing revealing all of its child nodes.

#### Actual behavior

Only the icon next to the node switches to the collapsed state, but no child nodes are visible.

See the following GIF for a demonstration using the demo:
neos-pagetree-search

### Affected Versions

Neos: 4.3.8 and 5.0.4

UI: 4.0.1 and 5.0.1

### LINKED ISSUE #2800: Moving nodes with Preset Nodetype filter
Screen-Recording-2020-10-12-at-1

## Steps to reproduce:

1. Toggle the tree preset
2. Move a Chapter node
3. Refresh the tree

Observe the node disappear. It won't re-appear even after reload.
It's a backend issue, a bug in getNodeDataForParentAndNodeType as far as I can tell: it doesn't find nodes that have been moved but not yet published, but it's really hard for me to dig through that CR code... Maybe some backend people could lend a hand? @bwaidelich, could you?

## Affected Versions
Neos: 4.x and up

UI: 4.x and up

Related: #1691

### LINKED ISSUE #3816: BUG: NodeTree presets, break expanding collapsed subtree
### Description

After using NodeTree filter presets, the default NodeTree does not reset properly and does not allow the expanding of sub-trees, 
which where expanded before the filter usage.

### Steps to Reproduce

1. Add config:
```
Neos:
  Neos:
    userInterface:
      navigateComponent:
        nodeTree:
          presets:
            blog:
              ui:
                label: 'Blog'
                icon: 'icon-folder'
              baseNodeType: 'Neos.Demo:Document.Blog'
```
2. Expand sub-tree in default NodeTree
3. Use filter preset
4. Reset filter
5. Try to expand the sub-tree => No sub-tree items get loaded

#### Expected behavior

Sub-tree items should get loaded.

#### Actual behavior

Screen Recording 2024-06-25 at 14 17 17(1)

### Affected Versions

Neos: 8.3

UI: 8.3

grebaldi:
fixes: #3816, #2583, #2800

## Add settings during E2E test run

This PR indroduces a mechanism to add additional settings during E2E test run.

This is necessary, because the newly introduced `nodeTreePresets.e2e.js` test fixture requires node tree presets to be configured. Node tree presets are not active by default, so we need to set up additional settings for this test fixture specifically.

<details>
<summary>Read details on how this is achieved</summary>

The mechanism is achieved by providing two public endpoints in the test distribution:

### `POST /test/write-additional-settings`

> [!WARNING]
> This endpoint is only available in the test distribution and should never be replicated for production use. **DO NOT DO THIS IN PRODUCTION!**

This endpoint accepts a JSON-encoded object as a body. It's top-most key must be `settings`. Any value passed to `settings` will be YAML-serialised and written to `FLOW_PATH_CONFIGURATION/Settings.Additional.yaml`.

This endpoint then also registers a signal handler for `bootstrapShuttingDown`, in which it will flush all caches, so that the additional configuration will be used upon the next request.

#### EXAMPLE

```sh
$ curl -X POST -d '{"settings":{"foo": "bar"}}' http://127.0.0.1:8081/test/write-additional-settings --header "Content-Type:application/json"
{"success":true}
```

This will write the following `Settings.Additional.yaml`:
```yaml
foo: bar
```

### `POST /test/remove-additional-settings`

> [!WARNING]
> This endpoint is only available in the test distribution and should never be replicated for production use. **DO NOT DO THIS IN PRODUCTION!**

This endpoint does not require any data and will remove the `FLOW_PATH_CONFIGURATION/Settings.Additional.yaml` file, if it exists.

If the file was removed, this endpoint then also registers a signal handler for `bootstrapShuttingDown`, in which it will flush all caches, so that the removed configuration will no longer be used upon the next request.

#### EXAMPLE

```sh
$ curl -X POST -d '' http://127.0.0.1:8081/test/remove-additional-settings --header "Content-Type:application/json"
{"success":true}
```

This will remove `Settings.Additional.yaml`, if it exists.

</details>

### The Problem

The `PageTree` redux state is tracking `hidden` nodes, presumably for the purpose to hide all nodes that are not part of the current search result set.

This is indicated by how `hiddenContextPaths` for page tree search results are calculated:
https://github.com/neos/neos-ui/blob/fd5c917e70b1ba3f1abe2d4681fd5f4dad5229bc/packages/neos-ui-sagas/src/UI/PageTree/index.js#L191-L193

When node type filter and search term are cleared, the nodes that have been flagged as  `hidden` during the last search, remain `hidden` aftwards, because the page tree state is not properly cleaned up.

### The Solution

For clarity, I decided to flip the `hidden` state into a `visible` state. By default `visible` is set to `null`, indicating that every node should be shown in the tree.

When a search term or node type filter is set, `visible` will at first contain all context paths in the search result set (including intermediaries that are only ancestors of matching nodes).

While `visible` is not `null`, every node that is added to the in-memory store will be added to the `visible` list as well.

`visible` here only means "This node is invited to the tree party". A node that is invited, may still not be shown, because its parent is collapsed. Every node that is not invited, will never be shown, no matter what other toggle state may be set.

### The Problem

Since https://github.com/neos/neos-ui/pull/2794 all structural change operations handle an optional `baseNodeType` to account for currently set tree filters.

#2794 had been introduced in October 2020. About a year before that, in December 2019, batch operations for tree nodes had been introduced via https://github.com/neos/neos-ui/pull/2568. #2794 did not consider the changes made in #2568, so the `baseNodeType` isn't being sent when tree nodes are moved. This leads to the disappearance of all nodes beneath the reference parent.

The logic for this would have needed to be handled in the `moveDroppedNodes.js` saga. Oddly enough, the 4 years old merge commit of #2794 shows that it wasn't present then:
https://github.com/neos/neos-ui/tree/ed9e04cdcaf5774914a60d866a8589709b4704f8/packages/neos-ui-sagas/src/CR/NodeOperations

In the 5 years old merge commit for #2568, `moveDroppedNodes.js` does indeed exist:
https://github.com/neos/neos-ui/tree/48233fee50c106334a949647ed3deef6d66dfa3c/packages/neos-ui-sagas/src/CR/NodeOperations

I have no explanation for this.

### The Solution

I added the missing `baseNodeType` to `moveDroppedNodes.js`.

### The problem

Similar to #3816 & #2583, the PageTree state did not recognize when a context path changed (which is what happens, if a node gets moved "into" another one).

### The solution

I modified the PageTree reducer, so it tracks the `CR.Nodes.UPDATE_PATH` action as well.

## Additional Changes

- One more bugfix: Reloading the tree when in filtered state was broken -> this is fixed now
- Tree nodes no longer show superfluous toggle handles
  - For this I had to add a nodeTypeFilter option to the `getForTreeWithParents` finisher of the FlowQuery API
  - The result is still not technically correct (as the entire API is quite broken), but it'll at least *look* correct :sweat_smile:
- I have added deprecation markers to all *singular* node operations (stuff like `copy` as opposed to `copyMultiple`, `move` as opposed to `moveMultiple` and so on)
  - While I was attempting to fix #2800, I stumbled upon this confusion, and since all singular operations can just as well be expressed as *multiple* operations, we should drop the singular ones entirely
  - Follow up to this PR: The operations should be removed in 9.0 (this will require additional work)

## Remaining TODOs

- [x] Write E2E test cases for node tree presets
  - [x] Create Mechanism to add additional settings during E2E test run
  - [x] Write some basic assertions that verify that node tree presets are basically working
  - [x] Write a failing test for #3816
  - [x] Write a failing test for #2583
  - [x] Write a failing test for #2800
- [x] Fix #3816
- [x] Fix #2583
- [x] Fix #2800
  - [x] Part 1: Nodes of an entire level disappear when any of them is moved
  - [x] Part 2: Parent cannot be uncollapsed after a node has been moved "into" it
- [x] Fix: Reloading the node tree while filters are set leads to wrong or empty result
  - [x] Happens on Discard as well
- [x] Fix: "Toggle children"-handle shows up, even though the node should have no children as per node type filter

### DIFF
--- a/Classes/Fusion/Helper/NodeInfoHelper.php
+++ b/Classes/Fusion/Helper/NodeInfoHelper.php
@@ -275,9 +275,10 @@ class NodeInfoHelper implements ProtectedContextAwareInterface
     /**
      * @param array $nodes
      * @param ControllerContext $controllerContext
+     * @param null|string $nodeTypeFilter
      * @return array
      */
-    public function renderNodesWithParents(array $nodes, ControllerContext $controllerContext): array
+    public function renderNodesWithParents(array $nodes, ControllerContext $controllerContext, ?string $nodeTypeFilter = null): array
     {
         // For search operation we want to include all nodes, not respecting the "baseNodeType" setting
         $baseNodeTypeOverride = $this->documentNodeTypeRole;
@@ -287,7 +288,7 @@ class NodeInfoHelper implements ProtectedContextAwareInterface
         foreach ($nodes as $node) {
             if (array_key_exists($node->getPath(), $renderedNodes)) {
                 $renderedNodes[$node->getPath()]['matched'] = true;
-            } elseif ($renderedNode = $this->renderNodeWithMinimalPropertiesAndChildrenInformation($node, $controllerContext, $baseNodeTypeOverride)) {
+            } elseif ($renderedNode = $this->renderNodeWithMinimalPropertiesAndChildrenInformation($node, $controllerContext, $nodeTypeFilter ?? $baseNodeTypeOverride)) {
                 $renderedNode['matched'] = true;
                 $renderedNodes[$node->getPath()] = $renderedNode;
             } else {
--- a/packages/neos-ui-redux-store/src/UI/PageTree/index.ts
+++ b/packages/neos-ui-redux-store/src/UI/PageTree/index.ts
@@ -2,6 +2,7 @@ import produce from 'immer';
 import {action as createAction, ActionType} from 'typesafe-actions';
 
 import {actionTypes as system, InitAction, GlobalState} from '../../System';
+import {actionTypes as nodes, Action as NodesAction} from '../../CR/Nodes';
 import {NodeContextPath, SelectionModeTypes} from '@neos-project/neos-ts-interfaces';
 
 import * as selectors from './selectors';
@@ -10,7 +11,7 @@ import {calculateNewFocusedNodes} from '../../CR/Nodes/helpers';
 export interface State extends Readonly<{
     focused: NodeContextPath[];
     toggled: NodeContextPath[];
-    hidden: NodeContextPath[];
+    visible: null | NodeContextPath[];
     intermediate: NodeContextPath[];
     loading: NodeContextPath[];
     errors: NodeContextPath[];
@@ -21,7 +22,7 @@ export interface State extends Readonly<{
 export const defaultState: State = {
     focused: [],
     toggled: [],
-    hidden: [],
+    visible: null,
     intermediate: [],
     loading: [],
     errors: [],
@@ -55,7 +56,7 @@ interface CommenceSearchOptions extends Readonly<{
 }> {}
 const commenceSearch = (contextPath: NodeContextPath, {query, filterNodeType}: CommenceSearchOptions) => createAction(actionTypes.COMMENCE_SEARCH, {contextPath, query, filterNodeType});
 interface SearchResult extends Readonly<{
-    hiddenContextPaths: NodeContextPath[];
+    visibleContextPaths: null | NodeContextPath[];
     toggledContextPaths: NodeContextPath[];
     intermediateContextPaths: NodeContextPath[];
 }> {}
@@ -80,11 +81,34 @@ export type Action = ActionType<typeof actions>;
 //
 // Export the reducer
 //
-export const reducer = (state: State = defaultState, action: InitAction | Action, globalState: GlobalState) => produce(state, draft => {
+export const reducer = (state: State = defaultState, action: InitAction | NodesAction | Action, globalState: GlobalState) => produce(state, draft => {
     switch (action.type) {
         case system.INIT: {
             const contextPath = action.payload.cr.nodes.documentNode || action.payload.cr.nodes.siteNode;
             draft.focused = contextPath ? [contextPath] : [];
+            break;
+        }
+        case nodes.ADD:
+        case nodes.MERGE: {
+            if (state.visible !== null) {
+                const visible = new Set([
+                    ...state.visible,
+                    ...Object.keys(action.payload.nodeMap)
+                ]);
+
+                draft.visible = [...visible];
+            }
+            break;
+        }
+        case nodes.SET_STATE: {
+            if (state.visible !== null) {
+                const visible = new Set([
+                    ...state.visible,
+                    ...Object.keys(action.payload.nodes)
+                ]);
+
+                draft.visible = [...visible];
+            }
             break;
         }
         case actionTypes.FOCUS: {
@@ -123,7 +147,7 @@ export const reducer = (state: State = defaultState, action: InitAction | Action
             break;
         }
         case actionTypes.SET_SEARCH_RESULT: {
-            draft.hidden = action.payload.hiddenContextPaths;
+            draft.visible = action.payload.visibleContextPaths;
             draft.toggled = action.payload.toggledContextPaths;
             draft.intermediate = action.payload.intermediateContextPaths;
             break;
--- a/packages/neos-ui-redux-store/src/UI/PageTree/index.ts
+++ b/packages/neos-ui-redux-store/src/UI/PageTree/index.ts
@@ -111,6 +111,16 @@ export const reducer = (state: State = defaultState, action: InitAction | NodesA
             }
             break;
         }
+        case nodes.UPDATE_PATH: {
+            if (state.visible !== null) {
+                const visible = new Set(state.visible);
+                visible.delete(action.payload.oldContextPath);
+                visible.add(action.payload.newContextPath);
+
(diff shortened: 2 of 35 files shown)


Sebobo:
You are awesome! Also love the part with the deprecation of the singular operations!
Will look for some time asap to review.

mhsdesign:
Wow ❤️ What is going on :D Thanks i guess :D The tests will probably allow us to reimplement this feature for Neos 9 then as it seems useful after all: https://github.com/neos/neos-ui/issues/3702 (currently its broken in 9.0 - as its an Neos Ui attached concept where were intercept link clicks ant attach query params)

Also nice idea with the post routes for adjusting settings. I run into the same problem and decided to use different Flow sub context that i dynamically set via header and special magic in `index.php` ... but a route seems much more flexible.

packages/react-ui-components/src/SelectBox/selectBox.js
```diff
@@ -43,6 +43,11 @@ export default class SelectBox extends PureComponent {
          */
         disabled: PropTypes.bool,
 
+        /**
+         * Optional id for the element
+         */
+        id: PropTypes.string,
```
pKallert, line 49: The id can be set here, but is only used for the reset button, should it not also be used for the dropdown itself?

pKallert (commented):
I tried it out locally and works like a charm 👍

packages/react-ui-components/src/SelectBox/selectBox.js
grebaldi: Excellent point!

Based on your suggestion, I have added a commit that will distribute the given SelectBox id across the DropDown* components: <s>https://github.com/neos/neos-ui/pull/3822/commits/81e8d237a557dc689fe30f273c0b4c92882bfa7a</s> https://github.com/neos/neos-ui/pull/3822/commits/9e830313351645f0d1eebac47714a7623eb88c34 (**EDIT:** First attempt wasn't null-safe :sweat_smile:)

The DropDownWrapper then carries the SelectBox id, the DrownDownHeader gets an id that is prefixed with the SelectBox id and suffixed with `-header`. Same for DropDownContents.

pKallert (approved):
Looks good to me! Thank you so much for these fixes 🌷

bwaidelich (approved):
Great work as usual, thanks!
I can't really judge the JS parts, but I can confirm that this fixes some nasty bugs with a Neos 8.3 installation of mine and I could not (yet *g) observe any unwanted side effects

Classes/Fusion/Helper/NodeInfoHelper.php
```diff
@@ -275,9 +275,10 @@ public function renderNodes(array $nodes, ControllerContext $controllerContext,
     /**
      * @param array $nodes
      * @param ControllerContext $controllerContext
+     * @param null|string $nodeTypeFilter
      * @return array
      */
-    public function renderNodesWithParents(array $nodes, ControllerContext $controllerContext): array
+    public function renderNodesWithParents(array $nodes, ControllerContext $controllerContext, ?string $nodeTypeFilter = null): array
     {
         // For search operation we want to include all nodes, not respecting the "baseNodeType" setting
```
mhsdesign, line 283: that statement is not correct anymore i guess ;)

packages/neos-ui-redux-store/src/UI/PageTree/index.ts
```diff
@@ -80,13 +81,46 @@ export type Action = ActionType<typeof actions>;
...
 //
-export const reducer = (state: State = defaultState, action: InitAction | Action, globalState: GlobalState) => produce(state, draft => {
+export const reducer = (state: State = defaultState, action: InitAction | NodesAction | Action, globalState: GlobalState) => produce(state, draft => {
     switch (action.type) {
         case system.INIT: {
             const contextPath = action.payload.cr.nodes.documentNode || action.payload.cr.nodes.siteNode;
             draft.focused = contextPath ? [contextPath] : [];
             break;
         }
+        case nodes.ADD:
```
mhsdesign, line 91: thanks for providing tests for this new logic ❤️

Sebobo (approved):
Works as advertised and change looks sensible, awesome ❤️

mhsdesign (commented):
forgot to publish my review ;) But looks all good by reading though ^^

PR #3823, opened 2024-07-03
BUGFIX: Add float case handling to NodePropertyConversionService

c4ll-m3-j4ck:
This extends the case handling in NodePropertyConversionService with a simple check for float and adds a conversion from string to float/double

**What I did**

**How I did it**

**How to verify it**


kitsunet (approved):

mhsdesign:
I just wonder shouldn't this go into 8.3 and what was the exact bug?

PR #3824, opened 2024-07-04
BUGFIX: Allow float-point numbers without decimal places in FloatValidator

c4ll-m3-j4ck:
The current implementation of the floatValidator uses parseFloat, which strips decimal point and places if a whole number (eg. .0) is provided. This then causes the regex validation to fail because it always assumes there to be a decimal point and following decimal places. 

This change makes the decimal places in the regex-validation optional.

### DIFF
--- a/packages/neos-ui-validators/src/Float/index.tsx
+++ b/packages/neos-ui-validators/src/Float/index.tsx
@@ -8,7 +8,7 @@ interface FloatOptions {
 }
 const Float = (value: any, validatorOptions: FloatOptions) => {
     const number = parseFloat(value);
-    if (value !== null && value !== undefined && value.length !== 0 && ((isNaN(number) || value.toString().match(/^[-+]?[0-9]*\.[0-9]+([eE][-+]?[0-9]+)?$/) === null))) {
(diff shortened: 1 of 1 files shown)


Sebobo (approved):

jonnitto:
Why is this against the Neos 9 branch? As this is a bugfix, shouldn't this go into Neos 8.3?

Sebobo:
Rebased to 8.3

c4ll-m3-j4ck:
@jonnitto Honestly it was based on 9.0 because thats what I was working on at the time. I don't mind this being targeted against 8.3/8.4

Sebobo:
@c4ll-m3-j4ck that's what I thought 😉

jonnitto (approved):

markusguenther:
The Unit Tests were not adjusted and failed at this point. (https://github.com/neos/neos-ui/actions/runs/14224592204)

Adjusted now just the test in the following commit:
https://github.com/neos/neos-ui/commit/c551f07beb3e1fde59fbe814e0a240083d431e2b

PR #3827, opened 2024-07-12
FEATURE: Make MultiSelectBox drag&drop sorting feature optional

### LINKED ISSUE #3715: FEATURE: Make MultiSelectBox drag&drop sorting feature optional
The MultiSelectBox in `react-ui-components` currently automatically makes all items draggable to allow changing the sorting.
But this should be optional as not every use case needs this and then we have a feature that only confuses users.

Example: Sorting Tags and Collections for an Asset in the Media.UI has no benefit, see https://github.com/Flowpack/media-ui/issues/57

Example from the Demo:

SCR-20240208-kftb

laurahaenel:
Solves #3715 .

**What I did**

- MultiSelectBox will default to non-dragging behavior.

**How I did it**

- Added new prop for the MultiSelectBox to ALLOW dragging.

**How to verify it**

- E.g. AssetEditor add prop `allowDragging={true}` (`Packages/Application/Neos.Neos.Ui/packages/neos-ui-editors/src/Editors/AssetEditor/index.js`)

With `allowDragging={false}`:
image

Default:
image


packages/react-ui-components/src/MultiSelectBox/multiSelectBox.js
```diff
@@ -31,6 +31,11 @@ class MultiSelectBox extends PureComponent {
          */
         disabled: PropTypes.bool,
 
+        /**
+         * Allows dragging.
+         */
+        allowDragging: PropTypes.bool,
```
mhsdesign, line 37: im not totally found of the `allow*` naming schema, but it seems were following the pattern of `allowEmpty` so it makes sense :D

mhsdesign (commented):
Thanks for this additional option. This really makes sense for things that are unordered either way! :D

Though im not sure about the breakiness here ... the pr is not declared as breaking (no `!!!` in the title) ... and technically it is not allowed to be breaking for a minor like 8.4 ... at least not too breaky but that is discussable.

But as you said
> MultiSelectBox will default to non-dragging behavior.

It is very much a new behaviour that might be unexpected at first and our users will have to look how to revert it.

I really like that we introduce a "positive" flag for the `MultiSelectBox` rather than negating stuff over and over but maybe it will make sense to introduce an `allowDragging=true` default for certain - if not all - editors that use the `MultiSelectBox` implicitly like the `Neos.Neos/Inspector/Editors/SelectBoxEditor`

A default for each editor can be set via this config:

https://github.com/neos/neos-development-collection/blob/fbc6ea95f05315dfaf2f7b58c6a398955a40177c/Neos.Neos/Configuration/Settings.yaml#L283-L284

Wdyt?


Edit: @Sebobo wrote

> But this should be optional as not every use case needs this and then we have a feature that only confuses users.


so maybe this is one of these slighly breaking changes that is allowed in a minor :D

Sebobo:
Thx so much for the feature!

If you disable it by default this will break at least one of my plugins. Though in the Media UI I will finally be able to turn it off where it had no effect.

So I'm pro default "on".

packages/react-ui-components/src/MultiSelectBox/multiSelectBox.js
laurahaenel: Could also be `enableDragging` if you like that more?

mhsdesign: no thanks I was just reflecting

mhsdesign (approved):
Thänks

laurahaenel:
Mh so I assume every native Neos.Editor which makes use of the MultiSelectBox should then be defaulting to `allowDragging={true}`, did I get this right?

Sebobo:
> Mh so I assume every native Neos.Editor which makes use of the MultiSelectBox should then be defaulting to `allowDragging={true}`, did I get this right?

The components default should be `true`. Then we don't need to set it on the editors. 
As this component is used in 3rd party plugins, we cannot change the default right now.

Sebobo (changes requested):
Please add `allowDragging: true` to the defaultProps of the multiselectbox

Sebobo (approved):

Sebobo (changes requested):
Thx, the tests still need some adjustments though.

laurahaenel:
What about the circleci test? I guess I dont get why it fails

Sebobo:
It complains that it cannot see an editor field, as the select box covers it 

```
Warnings (1):
 --
  TestCafe cannot interact with the <label for="__neos__editor__property---uriPathSegment" class="Z9_R_G_label _49YZqG_reset hN5w1W_envelope__label">...</label> element because another element obstructs it.
  When something overlaps the action target, TestCafe performs the action with the topmost element at the original target's location.
  The following element with a greater z-order replaced the original action target: <div class="jYGwrW_listPreviewElement _49YZqG_reset jYGwrW_listPreviewElement--isHighlighted" role="button">...</div>.
  Review your code to prevent this behavior.
```

Could be also unrelated. Did you rebase recently?

crydotsnake:
Ping @laurahaenel

laurahaenel:
Mh I cant rerun the circle ci tests (no permissions). I tried to run them locally but the setup is not correct on my machine. I will have a deeper look into it to get the same (failing) test results. Or should I just rebase?

laurahaenel:
I think I will make a new branch, this one is outdated and neither rebase nor merge is feasible 😅 Cherry-picking it is
image
I also broke the history when I forced it 🙄 Not using the outdated branch.

This is the new MR https://github.com/neos/neos-ui/pull/3874

PR #3833, opened 2024-07-17
BUGFIX: Prevent unneccessary server calls when inline editor is unchanged

### LINKED ISSUE #3832: BUG: Simply clicking into a ckeditor text field and afterwards selecting something else triggers a change
>  I fear this change introduced a regression which was not found during testing nor > is covered by e2e.
> I noticed that simply clicking into a ckeditor text field and afterwards selecting > something else triggered a change.
> This happens for every text element:
> 
>  image

_Originally posted by @mhsdesign in https://github.com/neos/neos-ui/issues/3751#issuecomment-2233110498_

To illustrate that a little more:

https://github.com/user-attachments/assets/9050d860-9cd9-4c50-a3ae-30a3b0642929

**Affected Versions:**

Neos UI: 8.3

grebaldi:
fixes: #3832

Straight-forward fix: This PR adds a check to prevent calls to the change API when the inline editor is unchanged.

### DIFF
--- a/packages/neos-ui-guest-frame/src/initializePropertyDomNode.js
+++ b/packages/neos-ui-guest-frame/src/initializePropertyDomNode.js
@@ -1,6 +1,6 @@
 import {$get, $contains} from 'plow-js';
 
-import {actions} from '@neos-project/neos-ui-redux-store';
+import {actions, selectors} from '@neos-project/neos-ui-redux-store';
 import {validateElement} from '@neos-project/neos-ui-validators';
 
 import {getGuestFrameWindow, closestContextPathInGuestFrame} from './dom';
@@ -74,6 +74,14 @@ export default ({store, globalRegistry, nodeTypesRegistry, inlineEditorRegistry,
                         actions.Changes.persistChanges([change])
                     ),
                     onChange: value => {
+                        const node = selectors.CR.Nodes.byContextPathSelector(contextPath)(store.getState());
+                        if (node) {
+                            const oldValue = node.properties[propertyName];
+                            if (oldValue === value) {
+                                return;
+                            }
+                        }
+
                         const validationResult = validateElement(value, $get(['properties', propertyName], nodeType), globalRegistry.get('validators'));
                         // Update inline validation errors
                         store.dispatch(


packages/neos-ui-guest-frame/src/initializePropertyDomNode.js
```diff
@@ -74,6 +74,14 @@ export default ({store, globalRegistry, nodeTypesRegistry, inlineEditorRegistry,
                         actions.Changes.persistChanges([change])
                     ),
                     onChange: value => {
+                        const node = selectors.CR.Nodes.byContextPathSelector(contextPath)(store.getState());
```
mhsdesign, line 77: i though hey why not use the transientValues ... but we only calculate this dirty state for the inspector ... so i guess this is the way ;) ...

a possibly 'cleaner'? in terms of not - depending - on the node - here - and doing it before `onChange`  called - alternative could be to track what the ckeditor changed currently via `change:data` and depending on this invoke the callback?

mhsdesign (commented):
Thank you so much for investigating so quickly ;) It makes totally sense that it fails currently because of this new line: https://github.com/neos/neos-ui/pull/3751/files#diff-f2abb91e96ac7035f92ab489c0b6393a31683dadd9f5d3946e4cecab5b9e5373R62

Sebobo (approved):
LGBR

I assume the strict comparison should work fine.

packages/neos-ui-guest-frame/src/initializePropertyDomNode.js
grebaldi: While probably possible, I wouldn't regard it as cleaner. Tracking CKEditor state is a whole can of worms in its own right, especially since we are only interested in changes that happened since the last time the property was focused.

Checking for actual changes before doing the remote call on the other hand *always* makes sense - and renders all tracking on CKEditor level moot. I was actually surprised that we didn't do that already :sweat_smile:

mhsdesign: okay you got me mostly convinced ... i just dislike callback - in this case the `onChange` - being fired when they simply shouldnt.
Nothing changed if we defocus a not - dirty - ckeditor (maybe ckeditor even has a flag for this?) so thats what i was referring to being "cleaner".

But still seb approved as well so feel free to merge ;)

Issue #3712, opened 2024-02-06
Creation dialog icons are not shown in Neos installations with subfolder setting


ahaeslich:
![image](https://github.com/neos/neos-ui/assets/1756367/efb3c6a4-498f-4660-ae51-b5746f66e8fc)

Steps To Reproduce

1. Use a Neos installation in a subfolder (e.g. [www.myproject.com/subfolder/neos](http://www.myproject.com/subfolder/neos))
2. Trigger the creation dialog by creating a new node in the document or content tree

This is essentially caused in the `ResourceIcon` component:

https://github.com/neos/neos-ui/blob/fe77db35c48e886b7a0421ccf3e2f0da6d098e52/packages/react-ui-components/src/Icon/resourceIcon.tsx#L26

Relates: #3423

mhsdesign:
Closing as dub of https://github.com/neos/neos-ui/issues/3616 and its about to be fixed 🎉

PR #3836, opened 2024-08-08
BUGFIX: Always bundle custom svgs instead of using `resource://` paths

mhsdesign:
Resolves: https://github.com/neos/neos-ui/issues/3616
Resolves: https://github.com/neos/neos-ui/issues/3712

See https://github.com/neos/neos-ui/issues/2092#issuecomment-2283775887 for an explanation

**What I did**

The custom icons are now included into the bundle instead of requested by path.

image

Also to unify things and simplify the build-stack, the Neos icon is now also bundled as plain svg instead of data-url:

image

Also the fallback image of the image editor will be fixed for cloud hosted resources now:

| before | after |
|--------|--------|
| image | image | 

Surprisingly i could not observer any pixel shifting. So this should be 100% the same visual look.

**How I did it**

**How to verify it**


packages/react-ui-components/src/Icon/resourceIcon.tsx
```diff
@@ -13,6 +13,11 @@ export interface ResourceIconProps extends Omit<IconProps, 'theme'> {
     readonly theme?: ResourceIconTheme;
 }
 
+/**
+ * @deprecated please refrain from using resource paths
```
Sebobo, line 17: The `Icon` component is still using the ResourceIcon when defining custom icons in yaml. What is the intended replacement?

Sebobo (changes requested):
Works and looks great, but I don't see yet how we can deprecate the now old `ResourceIcon` if the new one does something else and from what I see doesn't support custom icons.

mhsdesign:
As written in https://github.com/neos/neos-ui/issues/2092

> Case (3.) is more complicated, I'm afraid. The `<ResourceIcon/>` is a specialization of the (unfortunately public API) `<Icon/>`-component. We will need to deprecate its behavior, and (prior to that) find a way to allow for custom Icon URIs that can be resolved on the server side (we'll some sort of configurable mapping thing here).

My idea is to deprecate the behaviour but thats just for us and without real effect obviously.
To get rid of all usages we will - and i will plan do so - make the `NodeTypeSchemaBuilder` transform all `resource://` paths on the server. 

That will leave us with no more usages of client `resource://` paths. Wdyt?

Sebobo (approved):
Works on my machine <3

packages/neos-ui-editors/src/Editors/Image/resource/dummy-image.dataurl.svg
mhsdesign, line 1: This is now duplicated here and in `Neos.Neos` but yeah there is no way around this ... and it somehow belongs here imo

grebaldi (approved):
Looks & works very well :)

PR #3837, opened 2024-08-09
BUGFIX: Hide preview central group if empty

Sebobo:
**What I did**

Only show group "Preview central" in edit/preview modes dropdown if there are entries.

**How to verify it**

Disable the preview modes in the Neos.Demo and the `desktop` mode of Neos.Neos.
The dropdown shouldn't not contain the header anymore without any entries.


mhsdesign:
looks good. would have approved by reading but only if there is a screenshot attached now ill see to try it out :)

Sebobo:
Sorry, you are right:

Before:
CleanShot 2024-08-10 at 09 21 29@2x

After:
CleanShot 2024-08-10 at 09 21 00@2x

mhsdesign (approved):
thx

PR #3838, opened 2024-08-09
TASK: Adjust to extract workspace metadata and user-assignment to Neos

bwaidelich:
Counter-part to https://github.com/neos/neos-development-collection/pull/5146

Related: https://github.com/neos/neos-development-collection/issues/4726


Tests/IntegrationTests/TestDistribution/composer.json
```diff
@@ -14,7 +14,7 @@
     "require": {
       "neos/flow-development-collection": "9.0.x-dev as 9.0",
 
-      "neos/neos-development-collection": "9.0.x-dev as 9.0",
+      "neos/neos-development-collection": "feature/4726-extract-workspace-metadata-and-user-assignment-to-neos-dev as 9.0",
```
mhsdesign: revert me :)

Classes/Controller/BackendServiceController.php
```diff
@@ -335,18 +321,16 @@ public function discardChangesInSiteAction(array $command): void
         try {
             /** @todo send from UI */
             $contentRepositoryId = SiteDetectionResult::fromRequest($this->request->getHttpRequest())->contentRepositoryId;
-            $command['contentRepositoryId'] = $contentRepositoryId->value;
-            $command['siteId'] = $this->nodeService->deserializeNodeAddress(
+            $siteNodeAddress = $this->nodeService->deserializeNodeAddress(
                 $command['siteId'],
                 $contentRepositoryId
-            )->nodeAggregateId->value;
-            $command = DiscardChangesInSite::fromArray($command);
```
mhsdesign: i know this currently looks weird but this was just introduced by Bernhard and Wilhelm and we plan - as the comment says `@todo send from UI` - to send the array shape directly from the neos ui and then call `DiscardChangesInSite::fromArray` just to have type safety. Eventually each command will have its own proper command handler etc but right now its in an in between state. So i would like to keep the usages to the commands.

mhsdesign (dismissed):

Classes/Controller/BackendServiceController.php
mhsdesign: reintroduced the usages. Otherwise we would also have ugly conflicts with https://github.com/neos/neos-ui/pull/3858

mhsdesign (approved):
see https://github.com/neos/neos-development-collection/pull/5146#pullrequestreview-2356423998

Issue #3839, opened 2024-08-19
BUG: CKEditor writes "unexpected" markup after "clicking around" - not even typing


mhsdesign:
The logic introduced with https://github.com/neos/neos-ui/pull/3833 has a downside ... if a ckeditor plugin is ill-programmed the logic - comparing the transformed html after defocus - rather than if a user really changed stuff will lead to unexpected markup being written even-though nothing was might have changed in the eyes of the editor if they were just clicking around.

A more prominent example will probably be that if you uninstall a ckeditor plugin or change their config.
For example if you adjust https://www.neos.io/download-and-extend/packages/techdivision/techdivision-ckstyles.html to use `text-green` instead of `text-red` as css class on spans, ckeditor will cease to understand the previous without an migration and thus just removes the markup.

This is not a new problem, but previously editors could influence when something was saved and that was only when typing.

NeosUi: 8.3.9

mhsdesign:
Another unexpected usecase is the following:

Imagine you have an inline editable property which - dont ask me - is also shown in the inspector.

Now if you insert text in the inspector like "Me & Myself" (containing special chars) after reloading and clicking into the text via the ckeditor it will save a converted version "Me &amp; Myself" on defocusing the text.
Any editor will not understand that subtle new change now which appears in the publishing dropdown. And also the workspace-diff will show that looks the same.

mhsdesign:
Another case is when the `autoparagraph` setting was changed for the property. It might be unexpected that the conversation is triggered here implicitly as well?

Or basically anything like disallowing lists or links afterwards.

mhsdesign:
So after talking to @dlubitz  and @kitsunet  about this we might need to consider making silent changes like this transparent.
One way would be to show an dialog. A mini poc can be seen here:

https://github.com/user-attachments/assets/c88b0bb0-081d-4442-b9b8-610a8814e881

Following can be seen or not seen in the clip:
- The headline is inline editable directly because its not broken
- The paragraph was changed to `autoparagraph: false` and thus we have a potential problem here
- The editor can decide to actually take the new ckeditor version, or leave everything untouched which will leave the editor in an readonly state
- If confirmed the new markup is saved immediately and the editor may proceed with their actual changes

It could be great to also show in the readonly state the actual content of the live site instead of the upcasted data but for that we would have to lazily initialise ckeditor everywhere.
Regarding implementation we must be careful that calling `editor.getData` on initialisation of every inline element is not too expensive...

[...]

PR #3842, opened 2024-08-26
BUGFIX: 3839 show warn dialog before ckeditor formatting gets lost

mhsdesign:
Resolves: #3839

https://github.com/user-attachments/assets/02c42bdc-4452-493e-bf74-9cf07408d0b2

**What I did**

**How I did it**

**How to verify it**


grebaldi (changes requested):
Looks good after a first quick skim, but I noticed a naming issue:

`derivation, *Derivation` should be `deviation, *Deviation`, because

|English|German|
|-|-|
|derivation|Ableitung|
|deviation|Abweichung|

PR #3846, opened 2024-09-05
BUGFIX: 3839 ckeditor only save changes if (really) dirty

mhsdesign:
Resolves: https://github.com/neos/neos-ui/issues/3839
Alternative fix for: https://github.com/neos/neos-ui/issues/3832

**What I did**

The `onChange` callback is since also fired after defocusing an editor (https://github.com/neos/neos-ui/pull/3751). A followup introduced a check to compare the contents if the change is indeed new (https://github.com/neos/neos-ui/pull/3833).

The comparison is fragile and not always in the users intention. (https://github.com/neos/neos-ui/issues/3839)

This pr reverts the check and introduces a more stable isDirty handling.

**How I did it**

While i researched the isDirty tracking i stumbled upon https://github.com/ckeditor/ckeditor5/issues/996#issuecomment-407360730, naively i did a little tunnel through space and had to frizzle a lot with state handling.
But when trying to solve the problem of deduplicating the debounce _after_ the focus was lost i stumbled upon `.flush` which is EXACTLY what we need here. No further logic needed. And it guarantees a stable, simple to read and best possible behaviour.

**How to verify it**

### DIFF
--- a/packages/neos-ui-ckeditor5-bindings/src/ckEditorApi.js
+++ b/packages/neos-ui-ckeditor5-bindings/src/ckEditorApi.js
@@ -57,19 +57,12 @@ export const createEditor = store => async options => {
     return NeosEditor
         .create(propertyDomNode, ckEditorConfig)
         .then(editor => {
-            let isDirty = false;
-            editor.model.document.on('change:data', () => {
-                isDirty = true;
-            });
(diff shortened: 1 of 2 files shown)


Sebobo (approved):
Makes sense, works great and reduces save operations even further. Love it.

PR #3847, opened 2024-09-09
BUGFIX: Remove duplicate focus ckeditor border

mhsdesign:
With https://github.com/neos/neos-development-collection/pull/3618 the legacy ember css class `.neos-inline-editable` was removed (previously set here: https://github.com/neos/neos-development-collection/blob/881675b91185cbd8c41cb5e64f7467bf78710e04/Neos.Neos/Classes/Service/ContentElementEditableService.php#L69)

This leads to the default ckeditor focus outline styling not being removable.

Instead, we use `[data-__neos-property][contenteditable]` to target ckeditor inline elements now.

| before | after |
|--------|--------|
| image | image |

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/packages/neos-ui-guest-frame/src/style.module.css
+++ b/packages/neos-ui-guest-frame/src/style.module.css
@@ -22,11 +22,11 @@
     outline-color: var(--colors-Warn);
 }
 
-:global(.neos-inline-editable:focus) {
+:global([data-__neos-property][contenteditable]) {
     outline: none;
 }
 
-:global([data-neos-inline-editor-is-initialized]:hover) {
+:global([data-__neos-property][contenteditable]:hover) {
     outline-offset: 5px;
     outline: 2px dashed var(--colors-PrimaryBlue);
 }


Sebobo (approved):
Looks and works great. Thx

mhsdesign:
Apparently this also fixes a similar glitch on windows as reported on [slack](https://neos-project.slack.com/archives/C0U0KEGDQ/p1725974034350719) where there is a white outline:


image


image


which is set by the user-agent 


image

> I'm using Chrome 128.0.6613.138 (Official Build) (64-Bit) on Windows 11 and it is also visible in Firefox. It's coming from the user agent stylesheets. The [Neos.Neos:Editable](https://neos.readthedocs.io/en/stable/References/NeosFusionReference.html#neos-neos-editable) always seems to have the white outline, but it uses a span instead of a div when block=false and then the white outline is visible for each line.

mhsdesign:
To be totally sure i just retested this on a Neos UI version: v7.3.19 and can confirm that this fully restores the original behaviour, that includes NO outline being shown in focus mode and only on hover.

PR #3848, opened 2024-09-09
BUGFIX: Remove required flag from showResetButton prop on selectbox

Sebobo:
The prop is actually not required in its implementation and the UI tends to throw warnings because of the required flag.


mhsdesign (approved):
```
const {showResetButton} = this.props;
        if (showResetButton) {
```

is safe indeed.

Thanks ❤️

PR #3851, opened 2024-09-17
TASK: Use saucectl with configuration

markusguenther:
#### Overview:
We are using sauce labs as a remote browser service. The versions we were using are on an API version that has been deprecated. So this Pull Request brings the sauce labs usage to the latest versions. For that, we are now using the CLI tool of sauce labs `saucectl`. To be able to access the remote browsers, `saucectl` needs on CircleCI a tunnel. So we also need to use the `Sauce Connect Proxy 5 CLI`. 

#### Changes:
1. **Use `saucectl` CLI Tool**:
   - We have transitioned to using the official `saucectl` CLI for managing and executing Sauce Labs tests.
   - `saucectl` runs the `testcafe` tests
   - To enable access to remote browsers on CircleCI, we now utilize the **Sauce Connect Proxy 5** CLI tool, which creates the necessary tunnel for Sauce Labs.

3. **Removal of Sauce Labs Test Links in PR Comments**:
   - The current approach of adding links to Sauce Labs test results in PR comments was unreliable and has been removed.
   - In the future, we will look into leveraging the new features from the latest tools, such as webhooks, to automatically add test-related information back to the PR.

4. **New Features Enabled**:
   - The updated tools introduce several improvements:
     - **Enhanced test management**: Test cases now appear directly in the Sauce Labs dashboard.
     - **Improved test failure analysis**: You can now jump directly to failed test videos.
     - **Simplified multi-browser testing**: It is now easier to add and manage test suites across multiple browsers.

     
**New view of tests:**
Screenshot 2024-09-19 at 09 56 01

**Old view of tests:**
Screenshot 2024-09-19 at 09 55 35

#### Future Improvements:
With these updates, we have laid the groundwork for:
- Automating test information reporting in PRs through [webhooks](https://docs.saucelabs.com/web-apps/automated-testing/testcafe/yaml/#json).
- Extending our testing to more browsers and devices with minimal configuration.


markusguenther:
Added another little improvement - at sauce labs we now are able to filter for the target branch of the PR.
That makes it faster and easier to find the right recording.

Screenshot 2024-09-20 at 08 48 47

markusguenther:
We had a ticket in the past for more browsers in the acceptance tests. I could not find the issue, but which browsers we should test and is it ok for us that the tests take much longer?

Currently, I am using just Firefox, but we can use everything from the table below.
Screenshot 2024-09-20 at 09 26 11

markusguenther:
As some tests are flaky and just need a rerun occasionally, the new setup does that.
Here, the Windows suite failed in one test case. The Mac test suite started right after the finish of the Windows test suite, but as these failed, the Windows suite runs again. And in the end, everything was green.

Screenshot 2024-09-20 at 09 50 00

skurfuerst (approved):
Thanks so much <3 <3

Makefile
```diff
@@ -108,7 +108,7 @@ test:
 
 ## Executes integration tests on saucelabs.
 test-e2e-saucelabs:
-	bash Tests/IntegrationTests/e2e.sh "saucelabs:Firefox@latest:Windows 10"
+	bash Tests/IntegrationTests/e2e-saucelabs.sh
```
mhsdesign, line 111: why do we need another script that is almost `e2e.sh`? that will it make hard to maintain, especially since in neos 9 the script is partly rewritten. So we def need adjustments for Neos 9 ... and it probably better to have a shared script for local and sauce e2e tests?

PR #3850, opened 2024-09-13
8.4: fix e2e tests

skurfuerst:
This change fixes the TestCafe Setup

PR #3855, opened 2024-09-24
TASK: Backport the end to end tests from 8.4

markusguenther:
Related: https://github.com/neos/neos-ui/pull/3850
Related: https://github.com/neos/neos-ui/pull/3851

In the 8.4 branch, we worked on improving the acceptance test setup. Since we are supporting both Neos 8.3 and 8.4 simultaneously, the new test setup includes updated versions of TestCafe and SauceLabs.

This PR just backports the functionality.

The e2e.sh got some new parameters to be able to run local tests and saucelabs. Additionally, we introduce a help parameter to get more details on the usage.

```e2e.sh --help```
Screenshot 2024-09-24 at 20 58 25

For the saucelabs usage we need the environment variables `SAUCE_USERNAME` and `SAUCE_ACCESS_KEY`
```e2e.sh --saucelabs```

Local usage with e.g. the chrome browser works like that
```e2e.sh --browser chrome```


markusguenther:
As I got the feedback that we have too many shell scripts for the tasks, I would try to unify the scripts somehow.
So that we have on script that can be run saucelabs by a parameter and local if it is not set.

mhsdesign (approved):
Thank youuuuuuu so muuuuch :) <3

mhsdesign:
With testcafe 3 we can use native cdp automation for chromium. So to ensure that our tests also work on chromium based browsers locally we should add chrome to the test matrix. Thats why i change the test to run on chrome on mac os.

Additionally i added exit code handling to the bash script to signal if a test failed.

Issue #3859, opened 2024-09-26
BUG: Remove leading and tailing whitespace in LinkInput


markusguenther:
### Description

Editors often had the case that they copied links from an Excel list and then pasted them into e.g. LinkEditor fields.

If the people do not notice that a space at the end of the URL was copied or a tab at the beginning, the link immediately leads to a 404. 

### Steps to Reproduce

https://github.com/user-attachments/assets/06052606-d2d8-4383-bd24-1b0c6d57e989

#### Expected behavior

The links is trimmed so that we do not save the leading or tailing whitespace.

### Affected Versions

Neos: 8.3 (but guess it happens also in earlier versions)

UI: 8.3

Jai-Marothiya:
Hello @markusguenther,

I am interested in solving this bug and I believed I can fix it. Please assign this issue to me.

markusguenther:
Thank you for taking on this issue. However, we already have an open pull request for this. You are welcome to review it.

Jai-Marothiya:
Cool, can you assign me 1-2 relevant issues similar to this or may be different on which I can work?

Jai-Marothiya:
I reviewed your PR and it looks good.

PR #3860, opened 2024-09-26
BUGFIX: Trim leading and tailing whitespaces

markusguenther:
Whitespaces like a tab can lead to a 404 error if a URL has been pasted to a `LinkEditor` or `LinkInput`.

As data sources, assets, URLs and slugs should not start with a whitespace, we can remove them.

Fixes: #3859


markusguenther:
If I missed something, please let me know. Also thought about an editor option to trim that, but that would be more a feature than a bug fix then.


https://github.com/user-attachments/assets/9c247eeb-85ab-4918-b26f-1870df594bd4

packages/neos-ui-editors/src/Library/LinkInput.js
```diff
@@ -149,6 +149,10 @@ export default class LinkInput extends PureComponent {
     }
 
     handleSearchTermChange = searchTerm => {
+        // trim leading and trailing whitespace as it can cause issues
+        // with the further processing
+        searchTerm.trim()
```
Sebobo: Could there be a case in which `searchTerm` is null or undefined?

markusguenther: Right, that should be checked!

markusguenther:
The acceptance tests will run when https://github.com/neos/neos-ui/pull/3855 has been merged. At the moment they are broken.

packages/neos-ui-editors/src/Library/LinkInput.js
markusguenther: Normally, it should be always a string, but it is better to prevent a potential error

Sebobo (approved):
Looks good to me by reading, thx!

pKallert:
Thanks for having a look! 
I tried it out, but it only works when entering something by hand. Copy/pasting a URL with tab/space still saves with the space.

markusguenther:
> Thanks for having a look! I tried it out, but it only works when entering something by hand. Copy/pasting a URL with tab/space still saves with the space.

Will test it again, but also used the Clipboard.

markusguenther:
@pKallert Don't know why it worked before for me 🙈. Thanks for the test and review.

markusguenther:
Ok , as the `handleSearchTermChange` is called on each keystroke we trim also spaces while entering text, so I now just trim the beginning at this point and when the Link should be applied manually we know that nothing should change anymore, so I can also trim the tailing whitespaces.

Hope that we now have a solution that works for all cases 🙈

markusguenther:
Finally, tests are also passing 🥳

pKallert (approved):
Works like a charm, thank you!

PR #3861, opened 2024-09-27
FIX: Sizing of icons in alerts 

pKallert:
**What I did**
With the FontAwesome icons update the icons in the alerts now are too wide. 

Bildschirmfoto 2024-09-27 um 13 10 40

**How I did it**
FontAwesome sets the box-sizing to content box now:

Bildschirmfoto 2024-09-27 um 13 11 07

**How to verify it**
Icons are small again: 

Bildschirmfoto 2024-09-27 um 13 10 27

Do not know if we should change box-sizing for icons globally or not, but I could not find any other place where it does not work.


mhsdesign:
perfect thanks lets go with that and in case we notice something else we come back to it :)

mhsdesign (approved):

mhsdesign:
im not totally sure but i think there are also other cases around in the ui:

image

PR #3862, opened 2024-09-27
TASK: Run acceptance tests also on chrome

markusguenther:


.sauce/config1Dimension.yml
```diff
@@ -18,15 +18,20 @@ testcafe:
...
-  - name: "One dimensional Tests in Firefox on MacOS"
-    # todo use chrome here and fix ci https://github.com/neos/neos-ui/issues/3591
+  - name: "One dimensional Tests on Firefox on Windows"
     browserName: "firefox"
+    src:
+      - "Tests/IntegrationTests/Fixtures/1Dimension/*.e2e.js"
+    platformName: "Windows 10"
+    screenResolution: "1280x1024"
+  - name: "One dimensional Tests in Edge on MacOS"
+    browserName: "microsoftedge"
```
mhsdesign, line 34: why not chrome? ^^

.sauce/config2Dimension.yml
```diff
@@ -3,7 +3,7 @@ kind: testcafe
 sauce:
   region: us-west-1
   concurrency: 1 # Controls how many suites are executed at the same time.
-  retries: 0
+  retries: 3
```
mhsdesign, line 6: retries / smart retries dont work as far as i tried.

Tests/IntegrationTests/Fixtures/1Dimension/nodeTreePresets.e2e.js
```diff
@@ -40,20 +40,6 @@ const SETTINGS_WITH_NODE_TREE_PRESETS = {
                         presets: {
                             'default': {
                                 baseNodeType: 'Neos.Neos:Document,!Neos.TestNodeTypes:Document.Blog,!Neos.TestNodeTypes:Document.BlogArticle'
-                            },
-                            'blog': {
```
mhsdesign, line 44: i think we should try to keep it and make it work on chrome. I dont know why it doesnt work in ci. Locally it works splendidly. We have to debug the http request

.sauce/config1Dimension.yml
markusguenther: It was just a try, in safari we also have issues in the tests ;)

Tests/IntegrationTests/Fixtures/1Dimension/nodeTreePresets.e2e.js
markusguenther: I had the issue also locally, and therefore I changed this and also did not see why we need to change it. But I was not involved in the change back then.

.sauce/config2Dimension.yml
markusguenther: While testing the PR for 8.4 I had the case that one run failed and the retry succeeded. But did not enforce it after that.

PR #3863, opened 2024-10-07
WIP: Ckeditor5 upgrade

mhsdesign:
**What I did**

**How I did it**

**How to verify it**

PR #3864, opened 2024-10-09
BUGFIX: Restore old Placeholder behavior to keep shown when focused

mhsdesign:
In Neos 7.3.17 and before the placeholder in ckeditor would not disappear when focused. And only when typing. By switching to the native placeholder feature of ckeditor in https://github.com/neos/neos-ui/pull/3558 we introduced a regression as that implementation would hide the placeholder while just clicking into. This is especially noticeable for centered texts as the cursor needs one millisecond to adjust and its glitchy.

 This will eventually be fixed in Ckeditor 24 something. As a hotfix for our version 16 i introduce the patch of https://github.com/ckeditor/ckeditor5/pull/8474 (without the css absolute change as that would make centered text uneditable). The mentioned fix was not merged directly into ckeditor but made way more complicated: https://github.com/ckeditor/ckeditor5/pull/8867
 We dont need the more complicated fix as we dont use the mentioned features of the placeholder in a image description and such.

before (placeholder might disappear fully and text is not editable unless clicking on another element)

https://github.com/user-attachments/assets/4c1db00a-8d61-4c05-9307-a42c8e1e733f

after

https://github.com/user-attachments/assets/81cdbf88-b5b7-43e9-9356-eb716b3b2e54

**What I did**

**How I did it**

**How to verify it**


markusguenther (approved):
Looks good by reading. Will test later!

Sebobo:
Works as expected, though my cursor (In Vivaldi / Chrome) is at the end of the placeholder which is confusing:

CleanShot 2024-10-09 at 11 15 17@2x

In Safari it works as in your video.

markusguenther:
I looked bit into the latest CSS changes of the Placeholder view, but there was nothing special. So I adjusted the position. Now we have the cursor at the first position.

**Safari:**
Screenshot 2024-10-11 at 14 28 40


**Chrome:**
Screenshot 2024-10-11 at 14 28 47

markusguenther (approved):
For me, it works like I would expect it.
Thanks for taking care @mhsdesign 💙

mhsdesign:
Thanks for the adjustments ... i dared to try in firefox but well:

image

The placeholder dropped out.

Also on safari it doesnt blink anymore but the styling looks good at least.

This obviously needs a bit more of love. Ill see how it behaves in CK-Editor 40 or something here https://github.com/neos/neos-ui/pull/3863 just for testing ;)

markusguenther:
Hmm, ok the Firefox renders the `br` a bit different, and it seems the after position overlays the span, and so we don't see the cursor blinking. Hope to find some time tomorrow or Tuesday 🙈

markusguenther:
Tested the behavior in Firefox (132), Chrome (129) and Safari (18.0.1) and it behaves the same. 

https://github.com/user-attachments/assets/066d4dd4-8331-4c70-b944-85a3002320cf

Sebobo (approved):
For fine for me now in latest Firefox (after updating), Vivaldi (Chrome) and Safari.

mhsdesign:
Thank you so much for carrying this one @markusguenther <3

mhsdesign:
Btw for headlines, something is still off but that was before as well i think:

image

PR #3867, opened 2024-10-14
BUGFIX: Fix hidden state evaluation

mhsdesign:
The Neos 8.3 LinkingService contained this logic
```
$action = $node->getContext()->getWorkspace()->isPublicWorkspace() && !$node->isHidden() ? 'show' : 'preview';
```
ensuring that disabled nodes can still be previewed. In Neos 9 a no route matched error will be thrown.

To restore the old behaviour we evaluate the hidden state

**What I did**

**How I did it**

**How to verify it**


Classes/Controller/BackendController.php
```diff
@@ -226,9 +228,29 @@ public function redirectToAction(string $node): void
...
+            $nodeAddress->dimensionSpacePoint,
+            VisibilityConstraints::withoutRestrictions()
+        )->findNodeById($nodeAddress->aggregateId);
+
+        // we always want to redirect to the node in the base workspace.
+        $workspace = $contentRepository->getWorkspaceFinder()->findOneByName($nodeAddress->workspaceName);
+
+        $nodeAddressInBaseWorkspace = NodeAddress::create(
+            $nodeAddress->contentRepositoryId,
+            $workspace->baseWorkspaceName ?? WorkspaceName::forLive(),
```
bwaidelich: If the selected workspace has no base workspace – shouldn't that be an error?

bwaidelich (commented):
Thanks for fixing!
The amount of domain logic in the controller makes me a little unsure, but in general it makes sense to me..
Will we be able to cover this with a test to prevent this from breaking again in the future?

Classes/Controller/BackendController.php
mhsdesign: i this is like one of these `if ($siteNode) {}` calls that are just there to prevent an error than to run into a fatal error.
If some logic is changed that we for example work on the live workspace directly we should also still be able to preview it ... but i think youre right ... for that case $workspace->baseWorkspaceName ?? $workspace->workspaceName should be enough;)

mhsdesign: Adjusted:)

Classes/Controller/BackendController.php
```diff
@@ -225,9 +226,29 @@ public function redirectToAction(string $node): void
...
+        $contentRepository = $this->contentRepositoryRegistry->get($nodeAddress->contentRepositoryId);
+
+        $nodeInstance = $contentRepository->getContentGraph($nodeAddress->workspaceName)->getSubgraph(
+            $nodeAddress->dimensionSpacePoint,
+            VisibilityConstraints::withoutRestrictions()
+        )->findNodeById($nodeAddress->aggregateId);
+
+        $workspace = $contentRepository->findWorkspaceByName($nodeAddress->workspaceName);
+
+        // we always want to redirect to the node in the base workspace.
```
bwaidelich:
```suggestion
        // we always want to redirect to the node in the base workspace unless we are on a root workspace in which case we stay on that (currently that will not happen)
```

bwaidelich (approved):

PR #3868, opened 2024-10-15
TASK: Adjust to Add workspace content stream mapping to content graph projection

bwaidelich:
Neos.Ui counter part for https://github.com/neos/neos-development-collection/pull/5096


mhsdesign (approved):
<3 thanks ;) youre reading my mind! Can be merged because of the b/c layer already

PR #3872, opened 2024-10-21
TASK: further fix e2e test cases for Neos 9

skurfuerst:


Tests/IntegrationTests/pageModel.js
```diff
@@ -59,7 +59,7 @@ export class PublishDropDown {
 
     static publishDropdownDiscardAll = ReactSelector('PublishDropDown ContextDropDownContents').find('button').withText('Discard all');
 
-    static publishDropdownPublishAll = ReactSelector('PublishDropDown ShallowDropDownContents').find('button').withText('Publish all');
+    static publishDropdownPublishAll = ReactSelector('PublishDropDown ContextDropDownContents').find('button').withText('Publish all');
```
mhsdesign, line 62: i dont get that one but whatever :D

mhsdesign (approved):
i dont use the local tests in Tests/IntegrationTests/docker-compose.neos-dev-instance.yaml and you and @JamesAlias are the only ones i think? Wilhelm runs them differently as well so i think this is good to go :D

PR #3873, opened 2024-10-21
fix copy paste across dimensions

skurfuerst:
related https://github.com/neos/neos-development-collection/issues/4614

After thoroughly thinking this through, I decided
that moving nodes across dimensions is effectively a copy followed by
a delete, thus it CHANGES IDENTITIES of nodes then.

This means it will work in all cases, even if the moved node already
existed with the same ID in the target dimension.

Lateron, we additionally need to expose CreateNodeVariant (which creates
connected variants) in the UI as well.

Resolves: neos/neos-development-collection#4614

### DIFF
--- a/Classes/Domain/Model/Changes/MoveAfter.php
+++ b/Classes/Domain/Model/Changes/MoveAfter.php
@@ -13,9 +13,13 @@ namespace Neos\Neos\Ui\Domain\Model\Changes;
  */
 
 use InvalidArgumentException;
+use Neos\ContentRepository\Core\DimensionSpace\OriginDimensionSpacePoint;
+use Neos\ContentRepository\Core\Feature\NodeDuplication\Command\CopyNodesRecursively;
 use Neos\ContentRepository\Core\Feature\NodeMove\Command\MoveNodeAggregate;
 use Neos\ContentRepository\Core\Feature\NodeMove\Dto\RelationDistributionStrategy;
-use Neos\Neos\Ui\Domain\Model\Feedback\Operations\RemoveNode;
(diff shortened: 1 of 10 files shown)


Classes/Domain/Model/Changes/MoveAfter.php
```diff
@@ -71,19 +75,49 @@ public function apply(): void
...
+                    $subject->workspaceName,
+                    $subject->dimensionSpacePoint,
+                    $subject->aggregateId,
+                    RelationDistributionStrategy::STRATEGY_GATHER_ALL,
+                    $hasEqualParentNode ? null : $parentNodeOfPreviousSibling->aggregateId,
+                    $precedingSibling->aggregateId,
+                    $succeedingSibling?->aggregateId,
+                );
+                $contentRepository->handle($command);
+            }
```
mhsdesign, line 120: im not sure i understand the full implications of this ... as far as i know is move node super well tested and supports a variety of cases to distribute node aggregates and i dont know if we want to trade that in for the use of this `CopyNodesRecursively` plus removing the original node?

isnt this supported already by `MoveNodeAggregate` or do we need to extend the core?

mhsdesign:
Found an alternative fix in https://github.com/neos/neos-development-collection/issues/4614#issuecomment-2470222193

mhsdesign:
As elaborated in https://github.com/neos/neos-development-collection/issues/5475 i could reproduce one super edge case where moving nodes does not work because they have been varied already, but mostly the current behaviour is stable. I think we dont need this change for now and as i discussed with you we want to go into the direction first to prevent impossible behaviour of cut instead of magically making every edgecase work.

PR #3875, opened 2024-10-21
BUGFIX: Skip `null` when publishing nodes

kdambekalns:
The `getNodeFromContextPath(…)` in the Neos.Ui `NodeService` may return `null`, passing that to `publishNodde(…)` fails:

```
Exception in line 240 of /…/Flow_Object_Classes/Neos_Neos_Ui_Controller_BackendServiceController.php:
Neos\Neos\Service\PublishingService_Original::publishNode(): Argument #1 ($node) must be of type
Neos\ContentRepository\Domain\Model\NodeInterface, null given, called in
/…/Neos_Neos_Ui_Controller_BackendServiceController.php on line 240
```

This change adds a check and skips those nodes.

**How to verify it**

I see this error in a client project quite often, but sadly do not have a reproducer. Looking at the code shows this can happen, though.

### DIFF
--- a/Classes/Controller/BackendServiceController.php
+++ b/Classes/Controller/BackendServiceController.php
@@ -237,6 +237,14 @@ class BackendServiceController extends ActionController
 
             foreach ($nodeContextPaths as $contextPath) {
                 $node = $this->nodeService->getNodeFromContextPath($contextPath, null, null, true);
+                if ($node === null) {
+                    $error = new Info();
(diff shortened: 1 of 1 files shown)


kdambekalns:
I found this on my machine today (after quite some time trying to find a difference in installed packages between local and production setups). PhpStorm's local history says the code was added on October 16th, but I don't remember writing it. 🙃 So maybe it's from somewhere else, but I could not find anything. So I created this PR.

skurfuerst:
makes sense, thanks :)

PR #3876, opened 2024-10-23
FEATURE: Add `options.moveNodeStrategy` configuration to NodeTypes

### LINKED ISSUE #3587: BUG 9.0 NodeMove should respect `aggregate: true`

pKallert:
Resolves: https://github.com/neos/neos-ui/issues/3587
Resolves partially: https://github.com/neos/neos-development-collection/issues/5368

In order to mimic the behavior of 8.3 we decided to introduce a setting for move strategies on NodeType `options`. 

In 8.3 when moving a document node, we (try to) move it in all dimensions. When moving a content node in 8.3 we do not try to move it in all dimensions. 

Currently the move strategy is hard-coded in the Neos UI. With this PR we load the move strategy from the NodyType to specify it in the command. 

This gives us the flexibility in the future to implement some custom move strategies in the UI but lets us keep the 8.3 behavior for now. 

Further this neos pr removes the old `aggregate` flag fully as its totally unused but was in 8.3 indicator for this behaviour https://github.com/neos/neos-development-collection/pull/5314 

**What I did**

**How I did it**

**How to verify it**


Configuration/NodeTypes.yaml
```diff
@@ -34,6 +35,7 @@
     nodeCreationHandlers:
       promotedElements:
         factoryClassName: 'Neos\Neos\Ui\Infrastructure\ContentRepository\CreationDialog\PromotedElementsCreationHandlerFactory'
+    moveNodeStrategy: scatter
```
mhsdesign, line 38: as discussed today for content, especially as scatter also moves virtual variants the behaviour is correct

> Still, specializations pointing to the same node using the fallback mechanism will be kept gathered.

Configuration/NodeTypes.yaml
```diff
@@ -24,6 +24,7 @@
         factoryClassName: 'Neos\Neos\Ui\Infrastructure\Neos\UriPathSegmentNodeCreationHandlerFactory'
       promotedElements:
         factoryClassName: 'Neos\Neos\Ui\Infrastructure\ContentRepository\CreationDialog\PromotedElementsCreationHandlerFactory'
+    moveNodeStrategy: gatherAll
```
mhsdesign, line 27: We also discussed that it might make sense to use `gatherSpecializations` or `scatter` on document level from time to time (to create different structures for each dimension). Maybe we need to introduce some ui dialog with a little "magnet" slider which dimensions to also gather

But we should also keep in mind that non gatherAll behaviour might be broken for the Neos Ui as it relies on the information here: https://github.com/neos/neos-development-collection/blob/a14bbe701335c013df85579d1703f9dd4b018cc5/Neos.Neos/Classes/Controller/Service/NodesController.php#L334-L335

mhsdesign (approved):

Classes/Domain/Model/Changes/MoveAfter.php
```diff
@@ -71,14 +71,20 @@ public function apply(): void
             $hasEqualParentNode = $parentNode->aggregateId
                 ->equals($parentNodeOfPreviousSibling->aggregateId);
 
-
             $contentRepository = $this->contentRepositoryRegistry->get($subject->contentRepositoryId);
-
+            $rawMoveNodeStrategy = $this->getNodeType($this->subject)?->getConfiguration('options.moveNodeStrategy');
```
kitsunet, line 75: this btw. should still be part of the schema  @mhsdesign in Neos.Neos as otherwise you cannot validate it....

mhsdesign: irigh zrk who actually uses the validator 😅 but if you think so ... are the validation things merged? then it can live here in the ui ... other wise id say no its not worth it. its not something anyone is gonna use.

kitsunet: 😂 I think we cannot merge schemas? Lets skip it for now, but IMHO would be good to have somewhere...

PR #3880, opened 2024-10-24
BUGIFX: Show hidden state only for directly disabled nodes without inheritance

dlubitz:
Before this fix the display of the hidden state in the content and document tree was broken. Initially only directly disabled nodes where marked as hidden. After opening a document node below a directly disabled node the icon of the selected node has changed also to "hidden".

In Neos <9 we only show disabled nodes with the "hidden" indicator and not the children. So this change ensures the same behavior as before.

In future we should think about a new indicator, which shows a "nodes hidden/disabled by inheritance" state.

Before:
image

Now: 
image


nezaniel (approved):
Nice one, already ran into this

bwaidelich (approved):
Perfect

PR #1282, opened 2017-11-16
FEATURE: Respect policy configuration in UI

grebaldi:
This PR adds an aspect around the NodeInfoHelper to either prevent the rendering of nodes
or to add relevant policy information.

Based on said information, several Components in the UI are disabled or prevented from
initialization to ensure, that the policy is visibly applied.

TODOS:

- [x] Add policy info for NodeTreePrivilege
- [x] Add policy info for CreateNodePrivilege
- [x] Add policy info for RemoveNodePrivilege
- [x] Add policy info for EditNodePrivilege
- [x] Add policy info for EditNodePropertyPrivilege

- [x] Enforce NodeTreePrivilege
- [x] Enforce CreateNodePrivilege in NodeTypes dialog
- [x] Enforce RemoveNodePrivilege in Page Tree
- [x] Enforce RemoveNodePrivilege in Content Tree
- [x] Enforce RemoveNodePrivilege in Inline node toolbar
- [x] <s>Enforce EditNodePrivilege for inline editing</s> (<- this is already managed during content element wrapping)
- [x] Enforce EditNodePrivilege for inspector
- [x] Enforce EditNodePropertyPrivilege for inline editing
- [x] Enforce EditNodePropertyPrivilege for inspector

- [x] Enforce EditNodePrivilege in Page Tree
- [x] Enforce EditNodePrivilege in Content Tree
- [x] Enforce EditNodePrivilege in Inline node toolbar

- [x] Enforce EditNodePropertyPrivilege in Page Tree in case _hidden is amongst the disallowed properties
- [x] Enforce EditNodePropertyPrivilege in Content Tree in case _hidden is amongst the disallowed properties
- [x] Enforce EditNodePropertyPrivilege in Inline node toolbar in case _hidden is amongst the disallowed properties

Resolves: #1154

PR #3886, opened 2024-11-13
BUGFIX: Allow to insert node inside not editable content collection (via policy)

mhsdesign:
The `canEdit` check is evaluated by another privilege - the `EditNodePrivilege` - which is separate from the `CreateNodePrivilege` to allow or disallow node creation.

If `EditNodePrivilege` restricts almost all nodes except a few - say headline (by `nodeIsOfType`) - and `CreateNodePrivilege` does the same: only allow to create the headline (by `createdNodeIsOfType`), this policy is correctly enforced via the cr. The Neos ui though is a little to restrictive and allows LESS than what is allowed by the core. The `EditNodePrivilege` would be required to be granted for content collections to be able to insert content. To avoid having to grant this privilege - which comes with the burden of otherwise having to allow content collections AND its subtypes which could be sliders, and to be technically correct, we just dont evaluate the can edit here.

The button is still grey-d out if ALL nodes are disallowed by policy see: https://github.com/neos/neos-ui/blob/2597ffef66f7901569b865e35ea634a941b50a7b/packages/neos-ui-redux-store/src/CR/Nodes/selectors.ts#L292-L299

This behaviour has been since the introduction of Neos Ui policy evaluation: https://github.com/neos/neos-ui/pull/1282

Though inside a commit shows that during the development the behavior was changed:

https://github.com/neos/neos-ui/commit/72b2614fc2eb5b3dbb1b88a2a60aad5abe03e201#diff-8b6869be808b6aad3e3ac19f6312c029d54112eb72aa09ffd757e4a7a686bcefR200

I can only assume that in "Enforce EditNodePrivilege for inline editing" the use of the `canEdit` flag was added to restrictive.

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/packages/neos-ui-redux-store/src/CR/Nodes/selectors.ts
+++ b/packages/neos-ui-redux-store/src/CR/Nodes/selectors.ts
@@ -273,7 +273,7 @@ export const makeGetAllowedChildNodeTypesSelector = (nodeTypesRegistry: NodeType
         (_: GlobalState, {role}: {reference: NodeContextPath | null, role: string, subject: NodeContextPath | null}) => role
     ],
     (referenceNode, referenceParentNode, role) => {
-        if (referenceNode === null || (referenceNode.policy && referenceNode.policy.canEdit === false)) {
+        if (referenceNode === null) {
             return [];
         }
         const isSubjectNodeAutocreated = referenceNode.isAutoCreated;


mhsdesign:
For now ill be patching this like so:

```patch
Patch to allow writing policies allowing to create content inside of a content collection

The Neos Ui requires for that that the content collection is editable. But that is wrong and too restrictive than the core actually is.
As its cumbersome to patch the bundled Neos Ui javascript, this patch allows for a new method `isExactlyContentCollection` that can be
used instead of `nodeIsOfType(["Neos.Neos:ContentCollection"])` as this would do inheritance checks in the edit node privilege.

See also BUGFIX: Allow to insert node inside not editable content collection (via policy) https://github.com/neos/neos-ui/pull/3886

diff --git a/Neos.ContentRepository/Classes/Security/Authorization/Privilege/Node/NodePrivilegeContext.php b/Neos.ContentRepository/Classes/Security/Authorization/Privilege/Node/NodePrivilegeContext.php
index aba4e98892..333e71bc7e 100644
--- a/Neos.ContentRepository/Classes/Security/Authorization/Privilege/Node/NodePrivilegeContext.php
+++ b/Neos.ContentRepository/Classes/Security/Authorization/Privilege/Node/NodePrivilegeContext.php
@@ -142,6 +142,14 @@ class NodePrivilegeContext
         return false;
     }

+    public function isExactlyContentCollection(): bool
+    {
+        if ($this->node === null) {
+            return true;
+        }
+        return $this->node->getNodeType()->getName() === "Neos.Neos:ContentCollection";
+    }
+
     /**
      * Matches if the selected node belongs to one of the given $workspaceNames
      *
```

Sebobo (approved):

markusguenther (approved):
Looks good by reading, thanks :)

PR #3890, opened 2024-12-01
BUGFIX: Hide paste button in inline ui if no nodes are in clipboard

mhsdesign:
before:

button is always visible:

image

image

after:

button is hidden, todo maybe just grey out to prevent moving of buttons?

image

though after copying from inline, the buttons jump now:

image

**What I did**

**How I did it**

**How to verify it**


markusguenther (approved):
Thanks for the fix :)

PR #3892, opened 2024-12-07
BUGFIX: Prevent copy nodes across dimensions

mhsdesign:
see https://github.com/neos/neos-development-collection/issues/5054#issuecomment-2523060805
    
Previously the `$targetDimensionSpacePoint` was also wrongly chosen, and it was attempted to paste the node into the subjects home dimension

Also we hide the paste button in the Neos Ui if its across dimensions (im not sure if this is the best idea but yeah) - maybe an error is better?

**What I did**

**How I did it**

**How to verify it**


kitsunet (approved):
works nicely

Issue #3894, opened 2024-12-10
TASK: React warnings regarding unrecognized properties rendered to the DOM 


markusguenther:
### Description

When interacting with the CKEditor component, specifically clicking on the editor and attempting to type, the following React warnings are triggered:

    Warning: React does not recognize the executeCommand prop on a DOM element.
    Warning: React does not recognize the formattingUnderCursor prop on a DOM element.

**Root Cause**

These warnings occur because the executeCommand and formattingUnderCursor props are passed directly to DOM elements within the CKEditor component. React does not recognize these as valid DOM attributes, leading to unnecessary warnings in the console.

### Steps to Reproduce

1. Take care that  ```editorOptions: formatting:code``` is true (for instance, test with Neos.Demo)
2. Open a page and create a text content node
3. Click inside an inline editable

#### Expected behavior

No, react errors in the console.

#### Actual behavior

<img width="1258" alt="Screenshot 2024-12-10 at 15 07 39" src="https://github.com/user-attachments/assets/58c259d5-1f8d-4364-889f-1eda2aa6af9c">

### Affected Versions

Neos: 8.3

UI: 8.3

PR #3895, opened 2024-12-10
BUGFIX: Prevent warnings for unrecognized DOM props

markusguenther:
This change addresses React warnings related to invalid DOM attributes:
- Removed `executeCommand` and `formattingUnderCursor` from being passed to DOM elements.
- Ensured these props are only handled within components where they are relevant, preventing them from being incorrectly rendered on DOM elements.

These update prevent unnecessary warnings in the console.

Resolves: #3894

### DIFF
--- a/packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/Helpers/renderToolbarComponents.js
+++ b/packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/Helpers/renderToolbarComponents.js
@@ -31,7 +31,7 @@ export default richtextToolbarRegistry => {
                 const restProps = omit(props, ['isVisible', 'isActive']);
                 const isActiveProp = isToolbarItemActive(formattingUnderCursor, inlineEditorOptions)(componentDefinition);
 
-                const finalProps = {
+                let finalProps = {
                     ...restProps,
                     key: index,
                     isActive: isActiveProp,
@@ -40,6 +40,11 @@ export default richtextToolbarRegistry => {
                     formattingUnderCursor,
                     [callbackPropName]: e => e.stopPropagation() || executeCommand(commandName, ...commandArgs)
                 };
+
+                // filter out props that are not needed by the component and lead to a render warning
+                if (commandName === 'code') {
+                    finalProps = omit(finalProps, ['executeCommand', 'formattingUnderCursor']);
+                }
 
                 const Component = component;
 
--- a/packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/Helpers/renderToolbarComponents.js
+++ b/packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/Helpers/renderToolbarComponents.js
@@ -41,7 +41,11 @@ export default richtextToolbarRegistry => {
                     [callbackPropName]: e => e.stopPropagation() || executeCommand(commandName, ...commandArgs)
                 };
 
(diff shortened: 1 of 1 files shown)


packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/Helpers/renderToolbarComponents.js
```diff
@@ -41,6 +41,11 @@ export default richtextToolbarRegistry => {
                     [callbackPropName]: e => e.stopPropagation() || executeCommand(commandName, ...commandArgs)
                 };
 
+                // filter out props that are not needed by the component and lead to a render warning
+                if (commandName === 'code') {
```
mhsdesign, line 49: why does this have to be conditionally wrapped? And what does `code` mean?

markusguenther: The component that is used for the code command (class U) is not using the two props, and they end up in the DOM and that leads to the issue. In other places, we also omit properties to prevent that they end up in the IconButtons of the toolbar.

Sebobo: Does it mean the "code" Dom element specifically or some CK internal thing?
If it's a CK internal, a reference in the comment would be great. Like this it just looks like a random magic string.

markusguenther: The formatting type code comes from CKEditor and we just pass thru the props to the components.

markusguenther: We actually use just code. But the newer CKEditor also has `codeBlock`, but we just don`t have it yet.

markusguenther: I have updated the code and added a more detailed comment to describe this.

Sebobo (approved):

mhsdesign (changes requested):
Thanks for elaborating on the comment, i had a more deep look into it myself and now im pretty sure this is the wrong place.

The `code` plugin is actually shipped with the Neos.Demo and registered there: https://github.com/neos/Neos.Demo/blob/6ac18378d6d54a87663c8e20fe4eef1e30a50a3f/Resources/Private/CodeTagPlugin/src/manifest.js#L28

And here in the rendering we can also see the mistake, as we just pass the `finalProps` directly to the button without omitting anything:

https://github.com/neos/Neos.Demo/blob/6ac18378d6d54a87663c8e20fe4eef1e30a50a3f/Resources/Private/CodeTagPlugin/src/IconButtonComponent.jsx#L7

Omitting the props in the Ui based on the arbitrary name is too high-level.

Sebobo (changes requested):
I agree, this is too specific. I encountered the same issues with the toolbar when working on the toolbar UX changes in #3972. 

The problem is that f.e. the toolbar renders the registered components with all props and our Button components forwards all props to their innermost button element.
That's something we would have to fix there I think, as the plugin developer has no influence over that.

Sebobo:
Don't think we need this anymore. If it is still relevant please reopen.

Issue #3897, opened 2024-12-10
BUG: CKEditor table button icons missing


markusguenther:
### Description

The buttons of the CKEditor table actually don't show the icon. The fallback labels take a lot of space and are not well aligned.
This happens since Neos UI version 8.3.10.

### Steps to Reproduce

1. Create a text node
2. Add a table
3. Check the row and col buttons

#### Expected behavior

The buttons have just an icon and don't render the fallback label.
<img width="1325" alt="Screenshot 2024-12-10 at 18 52 03" src="https://github.com/user-attachments/assets/c74f18bf-1d1a-4e2e-bfd6-5cdbd2eeef42">

#### Actual behavior

<img width="1324" alt="Screenshot 2024-12-10 at 18 38 57" src="https://github.com/user-attachments/assets/52e5d1b6-2d44-4bc1-8eff-c19a6f70fce3">

The fallback labels are rendered.

### Affected Versions

Neos: 8.3

UI: 8.3.10

PR #3883, opened 2024-11-10
!!! FEATURE: Upgrade CKEditor 5 from version 16.0.0 to 44.0.0

skurfuerst:
This update represents a major leap forward, spanning several years of development. Due to the extensive changes and numerous breaking changes, it’s challenging to provide a comprehensive list of all API alterations. As such, this PR introduces a **breaking change**.

The CKEditor API has evolved significantly, and any plugins relying on outdated APIs may no longer function as expected. We cannot provide a detailed upgrade guide for your custom or third-party plugins. Instead, we recommend consulting the release notes of CKEditor 5 for the relevant versions. Note that during testing, some packages dependent on the old implementation were found incompatible.

To support the community, we aim to review a few community plugins and, where feasible, submit PRs to update them, offering practical examples.

This upgrade effort has been attempted multiple times over the years, but often proved too challenging to complete. We want to acknowledge and thank those who contributed to earlier PRs addressing this task, including @JamesAlias, @73nici, @skurfuerst , @mhsdesign , and @markusguenther 

**Testing Notes**:
Acceptance tests for editable content elements encountered issues with the new CKEditor. The workaround implemented in this PR leverages TestCafe’s proposed solution, but it only works reliably in Chrome. Fortunately, the TestCafe team has already merged a PR that will likely resolve this issue in the future.

Special thanks to everyone involved in this significant effort!

Please see https://discuss.neos.io/t/feedback-needed-in-which-release-to-include-ckeditor-5-update/6707 for context and discussion.

Resolves: https://github.com/neos/neos-ui/issues/2836

PR #3898, opened 2024-12-10
BUGFIX: Render table dropdowns with correct icons

markusguenther:
**What I did**
The icons were broken, and the fallback title was not aligned pretty well. So the SVG markup string will be transformed now to a data-uri and this can be used as an image source.

**How to verify it**
Described in the linked issue.

Resolves: #3897

### DIFF
--- a/packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/TableDropDown.js
+++ b/packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/TableDropDown.js
@@ -42,7 +42,6 @@ export default class TableDropDownButton extends PureComponent {
 
     render() {
         const iconDataUri = svgToDataUri(ckeIcons[this.props.icon]);
-        console.log({iconDataUri});
         return (
             <DropDown
                 padded={false}
--- a/packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/TableDropDown.js
+++ b/packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/TableDropDown.js
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import {$get} from 'plow-js';
 
 import {neos} from '@neos-project/neos-ui-decorators';
+import {svgToDataUri} from '@neos-project/utils-helpers';
 import ckeIcons from './icons';
 
 import style from './TableDropDown.module.css';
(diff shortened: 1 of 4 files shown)


markusguenther:
I already fixed that in the CKEditor Upgrade PR. Did not know that it happened in 8.3.
https://github.com/neos/neos-ui/pull/3883

packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/TableDropDown.js
```diff
@@ -40,12 +41,13 @@ export default class TableDropDownButton extends PureComponent {
...
 
     render() {
+        const iconDataUri = svgToDataUri(ckeIcons[this.props.icon]);
         return (
             <DropDown
                 padded={false}
             >
                 <DropDown.Header title={this.props.i18nRegistry.translate(this.props.tooltip)}>
-                    <img style={{verticalAlign: 'text-top'}} src={ckeIcons[this.props.icon]} alt={this.props.i18nRegistry.translate(this.props.tooltip)} />
+                    <img style={{verticalAlign: 'text-top'}} src={iconDataUri} alt={this.props.i18nRegistry.translate(this.props.tooltip)} />
```
mhsdesign, line 50: öhm do i get that right that we want to have the svgs as data-url and not as raw content string?
That can be done via the build setup. For the neos ui i introduced the convention that files ending with `.dataurl.svg` will get that treatment via the build setup already introduced via:

https://github.com/neos/neos-ui/pull/3836

packages/utils-helpers/src/svgToDataUri.ts
```diff
@@ -0,0 +1,47 @@
+const REGEX = {
```
mhsdesign: lol 😅 i hope you dindt had to write this yourself? :D I think more simple it would be to just use base64 decode and done :D

Edit: see comment above - i believe we dont need this at all hopefully

mhsdesign (changes requested):

packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/TableDropDown.js
markusguenther: The issue here was that the content of `this.props.icon` is just the SVG markup.
And that in a src tag is not working, and I bet that the build pipeline also doesn't handle it.

packages/utils-helpers/src/svgToDataUri.ts
markusguenther: Can discuss other solutions of course :)

markusguenther: I simplified that and now use base64 for the inline dataUri SVG stuff.

mhsdesign:
Okay i gave the build time data url thing way a shot here: https://github.com/neos/neos-ui/pull/3915

The difference is letting the data url be generated via build tooling instead of in the runtime

mhsdesign:
as now https://github.com/neos/neos-ui/pull/3915 is merged

Issue #3899, opened 2024-12-10
BUG: React PropType warning that the canBePasted type is wrong 


markusguenther:
### Description

For the `NodeToolBar`, we evaluate the current number of node context paths in the clipboard. Depending on this count, we proceed to check if there is a node that can be pasted.

When there are no nodes in the clipboard, the function returns a numeric value. This mismatch between a number and a boolean causes the warning.

### Steps to Reproduce

1. Open a neos backend
2. ensure you have no nodes in the clipboard
3. click on a inline editable element

#### Expected behavior

No warning!

#### Actual behavior

<img width="1072" alt="Screenshot 2024-12-10 at 19 31 38" src="https://github.com/user-attachments/assets/1178fc86-5287-487d-8a64-46bcd814aad7">

### Affected Versions

Neos: 8.3

UI: 8.3

PR #3900, opened 2024-12-10
BUGFIX: Prevent proptype warning regarding canBePasted property

markusguenther:
**What I did**

If `clipboardNodesContextPaths.length` is 0, this part will be interpreted as false (because 0 is considered "falsy" in JavaScript). The entire expression will then be 0 without evaluating the second part. This happens due to short-circuit evaluation in JavaScript.

**How I did it**
With this change we ensure that the value is always a boolean value.

Resolves: #3899


mhsdesign (commented):
Thanks makes sense ^^

Sebobo (approved):

PR #3902, opened 2024-12-11
FEATURE: Mark inherited disabled nodes in document and content tree

### LINKED ISSUE #3896: FEATURE: Mark inherited disabled nodes in tree
As we now know which node is disabled (hidden) by one of it's ancestors, we should show that in the tree similar to the disabled node itself - but different.

image

dlubitz:
Fixes #3896 

This marks nodes, which are disabled by inheritence, because a ancestor node is disabled, with a dedicated icon.

image

### DIFF
--- a/Classes/Fusion/Helper/NodeInfoHelper.php
+++ b/Classes/Fusion/Helper/NodeInfoHelper.php
@@ -105,6 +105,7 @@ class NodeInfoHelper implements ProtectedContextAwareInterface
             // TODO: we should export this correctly named, but that needs changes throughout the JS code as well.
             '_hidden' => $node->tags->withoutInherited()->contain(NeosSubtreeTag::disabled()),
             'hiddenInMenu' => $node->getProperty('hiddenInMenu'),
+            '_hiddenByAncestors' => $node->tags->onlyInherited()->contain(NeosSubtreeTag::disabled()),
             '_hiddenInIndex' => $node->getProperty('hiddenInMenu'),
             '_hasTimeableNodeVisibility' =>
                 $node->getProperty('enableAfterDateTime') instanceof \DateTimeInterface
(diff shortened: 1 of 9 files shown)


Classes/Fusion/Helper/NodeInfoHelper.php
```diff
@@ -103,6 +103,7 @@ public function renderNodeWithMinimalPropertiesAndChildrenInformation(
             // ensure _isHidden is sent to hidden nodes are correctly shown in the tree.
             // TODO: we should export this correctly named, but that needs changes throughout the JS code as well.
             '_hidden' => $node->tags->withoutInherited()->contain(SubtreeTag::disabled()),
+            '_hiddenByAncestors' => $node->tags->onlyInherited()->contain(SubtreeTag::disabled()),
             '_hiddenInIndex' => $node->getProperty('hiddenInMenu'),
             '_hasTimeableNodeVisibility' =>
```
mhsdesign, line 108: all this stuff is slightly getting more and more bloated ...  i dont like that :D maybe we should do something similar to what we have done with other data: use bit masks ... this is especially useful to denote that there is likely of all the things only one state that we really want to highlight in the ui.

https://github.com/neos/neos-ui/blob/81d01bed49549acb67c1637fdee99b38140af501/packages/neos-ui-redux-store/src/CR/Workspaces/index.ts#L12-L17

aside from this the chnage looks good :D Code wise it looks fine and should work ... is there anything missing than to disucss if we like to keep introducing data bloat :D ?

dlubitz:
I've rebased this on 9.2 so we can have it there as a feature.

mhsdesign (changes requested):
Hi, i see you have also created https://github.com/neos/neos-ui/pull/4161 but this feature pr is not feature complete without would behave buggy - e.g not update? We cannot have one without the other ...

but speaking about the other ... the change introduces more complexity to our already hopeless tree.

I am restricting myself to adding more features to the tree - for example i wanted to show deleted nodes to allow their publication - and rather started to work on incorporating a new foundation for our tree which is fully server/php backed and should also be fast.

https://github.com/neos/neos-ui/pull/3944

The problem is that it was not a super big priority so far. Our current tree is already not maintainable and fully undertested by E2E tests.

I think as developers we cannot add more and more things this broken tree but must refactor first and consider there the features we would like for a version 2 like showing and updating inherited nodes as well as showing deleted nodes as well as a possible way to paginate.

mhsdesign:
An alternative we could consider is to still have this feature on our current tree implementation but add e2e tests for all scenarios and then later switch out the implementation ... but that still makes our current tree more complex which it can barely handle

dlubitz:
I see your point, but stopping developing at all is also not an option for me. IF there were already hard progress on rewriting thing, ok. But nobody seems to work active on this.

This PR here works stand alone for the current node. BUT the child nodes are not updated immediatelly. Only a manual reload of the trees/full UI shows the hidden children. And yes, this was basically the idea of this change.

PR #3903, opened 2024-12-13
DRAFT: Test CKEditor update with testcafe 3.7.1 RC

markusguenther:
Just a Pipeline test and if we can remove the workarounds with the new RC.


markusguenther:
Ok the testcafe 3.7.1 RC is not resolving our issues with the content editable elements in CKEditor :/

PR #3904, opened 2024-12-20
TASK: Update workspace review URL

pKallert:
**What I did**
With https://github.com/neos/neos-development-collection/pull/5132 the workspace URL is switched from /show to /review. 
We need to update the URL in the Neos UI that links to the workspace detail view. 

**How to verify it**
The "review changes" button works with the new Workspace UX PR. 
Bildschirmfoto 2024-12-20 um 09 25 22


mhsdesign (approved):
;) nice uri building we use here 😅

PR #3907, opened 2025-01-20
BUGFIX: Forgiving `I18nRegistry.translate` for strings with colons and `undefined`

mhsdesign:
The following node type configuration

```yaml
placeholder: 'ClientEval: node.properties.tagName'
```

currently yields an error which is caught by the error boundary for properties:

> Error: TranslationAddress must adhere to format "{packageKey}:{sourceName}:{transUnitId}". Got "ClientEval: node.properties.tagName" instead.

The newly added strictness in https://github.com/neos/neos-ui/pull/3804 introduces this regression here.

Previously the function `getTranslationAddress` would just attempt to spit a string and expect three items to be returned:

https://github.com/neos/neos-ui/blob/2cecfcf136aafcd9e3f72537f513eb81b01e993b/packages/neos-ui-i18n/src/registry/I18nRegistry.js#L7

The deconstruction would already cause an error in PHP but not in JavaScript:

```
const [packageKey, sourceName, id] = 'ClientEval: node.properties.tagName'.split(':')
```

Returns

```
packageKey: "ClientEval"
sourceName: " node.properties.tagName"
id: undefined
```

The previous forgiveness needs to be reintroduced as its easy to create errors with strings containing colons at any position which is definitely likely for placeholders. Imagine: `placeholder: "The title of the blog:"`.

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/packages/neos-ui-i18n/README.md
+++ b/packages/neos-ui-i18n/README.md
@@ -22,7 +22,7 @@ Neos:
             // ...
 ```
 
-At the beginning of the UI bootstrapping process, translations are loaded from an enpoint (see: [`\Neos\Neos\Controller\Backend\BackendController->xliffAsJsonAction()`](https://neos.github.io/neos/9.0/Neos/Neos/Controller/Backend/BackendController.html#method_xliffAsJsonAction)) and are available afterwards via the `translate` function exposed by this package.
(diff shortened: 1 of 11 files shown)


mhsdesign:
This could also crash the whole ui in case a string contains a colon in the help message for example for the neos demo:

https://github.com/neos/Neos.Demo/blob/2f8081c3a39ff3b67e4306ff5891ef3bd8923108/NodeTypes/Content/Carousel/Carousel.yaml#L20

image


And in https://github.com/neos/neos-ui/pull/3907/commits/5520eb488f9267e08c87b1244d88bc456d4a1c10 i fixed that passing no id does not crash as well.

Ill merge this as i could not find a reviewer.

PR #3910, opened 2025-01-23
BUGFIX: Improve conflict resolution edge-cases in sync and publish workflow

### LINKED ISSUE #3908: After resolved conflict, don't attempt to re-synchronise
While testing https://github.com/neos/neos-development-collection/pull/5337 i found out through the e2e tests that the Neos Ui issues an obsolete call to rebase after the conflicts were resolved. Discarding or forcing a rebase _does_ already rebase so a second rebase afterwards is obsolete, and will be handled more strictly with the core adjustments:

Image

The error should be prevented client side by not calling the endpoint in this case.

mhsdesign:
Resolves #3908

fixes edecases during the conflict resolution:
- stopping publish if the attempted changes were dropped after a conflict:  #3908
- fixes 'Discard workspace "User workspace"' as resolution during publishing (previously froze)
- fixes navigating back after selecting 'Discard workspace "User workspace"' in sync mode and selecting another strategy

### DIFF
--- a/Resources/Private/Translations/en/SyncWorkspaceDialog.xlf
+++ b/Resources/Private/Translations/en/SyncWorkspaceDialog.xlf
@@ -54,7 +54,7 @@
         <source>Discard all changes in workspace "{workspaceName}"</source>
       </trans-unit>
       <trans-unit id="resolutionStrategy.DISCARD_ALL.confirmation.message" xml:space="preserve">
-        <source>You are about to discard all changes in workspace "{workspaceName}". This includes all changes on other sites.
+        <source>You are about to discard all {numberOfChanges} change(s) in workspace "{workspaceName}". This includes all changes on other sites.
 
         Do you wish to proceed? Be careful: This cannot be undone!</source>
       </trans-unit>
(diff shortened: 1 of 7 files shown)


mhsdesign:
I also noticed what @pKallert found out, that during publishing the conflict resolution strategy "discard all" does not work. Thanks for the hint how to solve that :D I took the changes you prepared for https://github.com/neos/neos-ui/pull/3893 into this pr in 66c11d6d971f23ab477a1a8ce8b8a0dfdfa7fb40 - i hope thats okay:)

Further after staring a lot onto the code i was able to fix the overlaying modals and the result message in dc1dbd61670b492e798bf4f7e784e151a3a796bd instead of using your approach to add a `SyncingPhase.IDLE` state https://github.com/neos/neos-ui/pull/3893/commits/cec189e04d670a615cae6a37cc5210b952b22fa2 to avoid the overlaying popups.

https://github.com/user-attachments/assets/bacdf436-507c-42c4-b765-978792b712d5

Further as my changes result in the use of the `SyncWorkspaceDialog -> ResolutionStrategyConfirmationDialog -> DiscardAllConfirmationDialog` instead of the `PublishingDialog -> ConfirmationDialog` the "No, cancel" button works correctly and leads back to the selection instead of stopping fully and the visual appearance and the text is slightly different:

| Old | New |
|--------|--------|
| image | image |

PR #3911, opened 2025-01-25
TASK: Show testcafe output in circle ci overview

mhsdesign:
Apparently none of the options for `saucectl run`

--show-console-log
--verbose
--live-logs

do output the actual live console logs from testcafe to the circle ci shell.

--show-console-log seems to output them at the end but only for happy cases.
This pr introduces a separate window where just the output is cleanly shown - hopefully also on error.

image

Further the artifact upload was fixed as the path was wrong.

**What I did**

**How I did it**

**How to verify it**


markusguenther:
So, should we also apply this to older branches? Thanks for digging deeper into this!

markusguenther (approved):

PR #3912, opened 2025-01-26
TASK: Use new `translate()` method if possible

mhsdesign:
Followup to https://github.com/neos/neos-ui/pull/3804 which deprecates the use of the <i18n> component and the i18nRegistry, as this is now a global concept via the `translate` method.

**What I did**

Using various careful regex replacements and manual adjustments, commit by commit the `translate()` method was introduced.

As written here https://github.com/neos/neos-ui/pull/3804#issuecomment-2613992743 the new method does not handle user input gracefully, which is the reason why non static strings are still using the old translation methods to ensure nothing breaks here.

**How I did it**

**How to verify it**

Via the regex

```
translate\('(?!(Neos\.Neos:(Main|Inspector|Modules)|Neos\.Neos\.Ui:(Error|Main|PublishingDialog|SyncWorkspaceDialog)):[a-zA-Z._-]+')
```

All source files can be checked that each call to `translate()` now specifies the translation as fully qualified shorthand string.

Replacing the <i18n> component with a direct call to `translate()` might not always look correct as this leads to a missing `<span />` tag. Those cases have to be identified by eye.

### DIFF
--- a/packages/neos-ui-containers/src/InsertModeSelector/index.js
+++ b/packages/neos-ui-containers/src/InsertModeSelector/index.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 
 import {ButtonGroup, Button, ResourceIcon} from '@neos-project/react-ui-components';
 import {neos} from '@neos-project/neos-ui-decorators';
-import I18n from '@neos-project/neos-ui-i18n';
+import I18n, {translate} from '@neos-project/neos-ui-i18n';
 
 import style from './style.module.css';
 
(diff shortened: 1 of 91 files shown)


markusguenther:
Wow, that’s a lot to take in! I’ll need some time to look into it. Thanks a bunch!

mhsdesign:
Currently the typescript interface of `translate()` demands a second parameter, the fallback. In some instances we do provide a fallback but we only provided a fallback so far half the time approximately ~130 times. The other half we have no fallback.
Now to satisfy typescript i added an empty string for the cases where we dont have anything filled out yet. That will be ~130 needless empty strings added then :) The question is do we want to have fallbacks explicitly for new code - in that case typescript will force us to think about it (or you insert `''` to make it silent, or do we dont think that Neos is navigateable without translation framework and files.

A third solution could be to use translation based on the actual label already what flow supports but what we have not used too much. See `Translator::translateByOriginalLabel()`

Sebobo:
IMHO, the empty string is worse. If the translation fails, you don't see anything and might not even the notice the issue. F.e. when the label is next to an icon in a button. Seems harder to debug this way?

mhsdesign:
Typescript fails here because we require the change https://github.com/neos/neos-ui/pull/4083 upmerged

Issue #3906, opened 2025-01-08
Non-scrollable Neos.Neos/Inspector/Editors/RichTextEditor


auwaerter:
### Description

Scrolling within the `Neos.Neos/Inspector/Editors/RichTextEditor` seems not to be possible. The **Discard** and **Apply** button moves below the visible area (even more when re-entering the editor after an apply).

When i click Apply, re-enter into the editor i also can't scroll down using the mouse. I can only enter and use the keyboard arrows or the page down button.

This bug was supposed to be fixed within https://github.com/neos/neos-ui/issues/3664 but probably came back smh.

### Steps to Reproduce

1. Use Google Chrome v131.0.6778.265
1. Create a new page that includes the 'Super.Vendor:Content.Text' NodeType, which contains the 'Neos.Neos/Inspector/Editors/RichTextEditor' using the above properties.
1. Create 1000 words here: https://www.loremipsum.de/
1. Navigate to it and add Text using "Toggle the editor".
1. Apply (you already notice, that you can't scroll up to edit the content or e.g. select some properties)
1. Enter again, to "Toggle the editor" (you can't scroll down, but you can use the arrow up/down key / page down button)

In the mentioned case, i'm using:

```yaml
'Super.Vendor:Content.Text':
  superTypes:
    'Neos.Neos:Content': true
  ui:
    label: 'i18n'
    icon: file-text
    position: 20
    inspector:
      groups:
        settings:
          label: 'Settings'
  properties:
    'text':
      type: 'string'
      ui:
        label: 'i18n'
        reloadIfChanged: true
        inspector:
          group: 'settings'
          editor: 'Neos.Neos/Inspector/Editors/RichTextEditor'
          editorOptions:
            placeholder: '<p>Text</p>'
            autoparagraph: true
            linking:
              anchor: true
              title: true
              relNofollow: true
              targetBlank: true
              download: true
            formatting:
              strong: true
              em: true
              u: true
              sub: true
              sup: true
              del: true
              p: true
              h1: true
              h2: true
              h3: true
              h4: true
              h5: true
              h6: true
              pre: true
              underline: true
              strikethrough: true
              removeFormat: true
              left: true
              right: true
              center: true
              justify: true
              table: true
              ol: false
              ul: false
              a: true
```

#### Expected behavior

1. Scrollbar appears
3. The **Apply** and **Discard** button is fixed

#### Actual behavior
[...]

PR #3913, opened 2025-01-31
BUGFIX: Non-scrollable Neos.Neos/Inspector/Editors/RichTextEditor

auwaerter:
Fixes the scrolling behavior within Neos.Neos/Inspector/Editors/RichTextEditor

**What I did**

Added a height with 0px value to enable scrolling within `RichTextEditor`.
=> Not sure if there is any better option/way for this fix as i didn't received any feedback in my [issue](https://github.com/neos/neos-ui/issues/3906), but it is working fine for me.

Before:
before

After (with height of 0px):
after

### DIFF
--- a/packages/neos-ui-editors/src/SecondaryEditors/CKEditorWrap/index.module.css
+++ b/packages/neos-ui-editors/src/SecondaryEditors/CKEditorWrap/index.module.css
@@ -15,4 +15,5 @@
     outline: 0;
     padding: 15px;
     overflow: scroll;
+    height: 0px;
 }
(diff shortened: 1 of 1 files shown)


Sebobo (approved):
Thx, don't know exactly why the fix works, but I can verify its functionality with Safari and Vivaldi (Chrome).

packages/neos-ui-editors/src/SecondaryEditors/CKEditorWrap/index.module.css
```diff
@@ -15,4 +15,5 @@
     outline: 0;
     padding: 15px;
     overflow: scroll;
+    height: 0px;
```
jonnitto:
```suggestion
    height: 0;
```

Just a small detail

jonnitto (changes requested):

jonnitto (approved):

Issue #3882, opened 2024-11-09
BUG: Icons to big in conflict dialog apply / okay button


mhsdesign:
have an outdated workspace and synchronise it

is:

<img width="887" alt="image" src="https://github.com/user-attachments/assets/f8cee833-006e-4ca2-aa70-a5b523e2e6dc">

wanted:

<img width="824" alt="image" src="https://github.com/user-attachments/assets/cffdaa51-6aa4-4faf-8b55-e531b9de446f">

Related https://github.com/neos/neos-ui/pull/3861

PR #3801, opened 2024-06-01
TASK: Upgrade to FontAwesome 6.5.2

wiegandj:
Issue: https://github.com/neos/neos-ui/issues/3793

Upgraded FontAwesome in neos-ui to 6.5.2.

Versions were tested in neos-development-collection before.

See upgrade for neos-development-collection:
Issue: https://github.com/neos/neos-development-collection/issues/5107
PR: https://github.com/neos/neos-development-collection/pull/5108

PR #3914, opened 2025-02-04
BUGFIX: Prevent override of height for icons

markusguenther:
The icons have multiple classes. From Font Awesome, we obtain the custom replacement class neos-svg-inline—fa, which sets a height of 1rem. However, we define the selectBoxHeaderWithSearchInput__icon class and set the height to 40px via a variable. The issue arises because the order in which these classes are applied determines the winning height. In Neos 8.4, the 40px height prevails, while in Neos 9.0, the 1rem height takes precedence. Consequently, the icons appear misaligned.

To address this issue, we implement an adjustment that ensures that when an element possesses both of these classes, the height of 40px from our custom class prevails.

Related: #3882

### DIFF
--- a/packages/react-ui-components/src/SelectBox_HeaderWithSearchInput/style.module.css
+++ b/packages/react-ui-components/src/SelectBox_HeaderWithSearchInput/style.module.css
@@ -19,6 +19,14 @@
     }
 }
 
+/*
+    Prevent that the height is defined by neos-svg-inline--fa when it is on the same level
+    like the selectBoxHeaderWithSearchInput__icon
+*/
+:global(.neos-svg-inline--fa).selectBoxHeaderWithSearchInput__icon {
+    height: var(--spacing-GoldenUnit);
+}
+
 .selectBoxHeaderWithSearchInput__icon {
     composes: reset from '../reset.module.css';
     height: var(--spacing-GoldenUnit);


markusguenther:
With the Upgrade in https://github.com/neos/neos-ui/pull/3801 some Icons were misaligned.

**Before:**
Screenshot 2025-02-04 at 18 36 15


**After:**
Screenshot 2025-02-04 at 18 34 34

markusguenther:
The acceptance tests for Firefox seem to be a bit wonky at the moment.

mhsdesign:
Thanks for fixing this

And i tested it and the override works ... thanks! Now im wondering why this change is targeting 8.4 when the bug only appears in 9.0?

> In Neos 8.4, the 40px height prevails, while in Neos 9.0, the 1rem height takes precedence.

In Neos 8.4 the class is indeed ordered differently in the build output:

```
.neos-svg-inline--fa{display:inline-block;font-size:inherit;height:1em;
.mAesQG_selectBoxHeaderWithSearchInput__icon{height:40px;width:1em}
```

Instead of `neos-svg-inline--fa` being last like in 9.0:

```
.mAesQG_selectBoxHeaderWithSearchInput__icon{height:40px;width:1em}
.neos-svg-inline--fa{display:var(--fa-display, inline-block);height:1em;
```

Also i noticed that the "width:1em" part is new with the font awesome upgrade too and wonder how that comes into play.

https://github.com/neos/neos-ui/pull/3801/files#diff-487adff98ba8cb5993634e088dd98a9904b1860ed6c58d4f788f1895822c5793

----------

overall its pretty mean to have the styles somehow wind up in a different order than expected and i also expect that explains: https://github.com/neos/neos-ui/issues/3882 and https://github.com/neos/neos-ui/pull/3861
... maybe we can find out the reason for that?

markusguenther:
Hi @mhsdesign 

the `neos-svg-inline--fa` brings the height of 1rem and depending on the ordering of the classes the height comes from the `neos-svg-inline--fa` or the `selectBoxHeaderWithSearchInput__icon` class. I target 8.4 as the upgrade was intoduced here and when somehow the ordering changes, the bug will be the same.

So it is more prevention to make clear that our class will win :)

pKallert (approved):
Tested it out locally and looks fine

Sebobo (approved):
Cannot reproduce as previously mentioned by Marc, but ok 👍

markusguenther:
> Cannot reproduce as previously mentioned by Marc, but ok 👍

It will fails on Neos 9.0, but as it was introduced here I wanted to be sure that our class rule wins.

markusguenther:
Thanks for testing ❤️

PR #3915, opened 2025-02-05
BUGFIX: Render cke toolbar table dropdowns with data urls

mhsdesign:
Resolves: #3897
Replaces https://github.com/neos/neos-ui/pull/3898 by letting the data url be generated via build tooling instead of runtime

We leverage our custom defined `.dataurl.svg` extension to let the build tooling use the data-url loader instead of inlining the svg contents.

This is a regression from #3836 as it changes the default `.svg` loading to "text" from "dataurl" and provides and alternative new `.dataurl.svg` loading with "dataurl" to handle the previous cases.

**What I did**

**How I did it**

**How to verify it**

before:

image

after:

image


pKallert (approved):
Tried it out locally and works for me

Sebobo (approved):
Works, thx!

Issue #3917, opened 2025-02-06
BUG: Inconsistent document tree behavior when copy/cutting documents


pKallert:
When creating, copying and cutting documents and content there is a very inconsistent behavior of what happens in the content and document tree. 
This leads to confusion for Editors. This is especially bad is when there are references to Documents and the wrong document is edited. 

**Current behavior**

In content tree when copy/pasting/creating new elements => you jump to new element and the element is selected in the content tree.

In document tree when creating a new document => you jump to the new page and the the new page ist selected in the document tree

In document tree when copy/pasting a new document => you stay on the "current" document

**Why is this happening** 

For the content tree, a RenderContentOutOfBand Feedback is added when copy/cutting/creating. When creating new elements, additionally a NodeCreated feedback is added. 
For creating a new document, a NodeCreated Feedback is added. 
For copy/cut, no Feedback is added that leads to a redirect in the Content Tree 

Neos UI: 8.4 
Neos: 8.4

PR #3918, opened 2025-02-06
BUGFIX: add feedback to switch to new document in tree on cut and copy

pKallert:
see #3917 

**What I did**
When using cut/copy in the document tree, the user is not automatically redirected to the "new" page. 

**How I did it**
To keep it consistent with  all other operations in the content/document tree I added feedback in the operations that will lead to the user being on the new page 

**How to verify it**
When using cut/copy/create in both the document and content tree, the user is always redirected to the "new" node.

### DIFF
--- a/Classes/Domain/Model/Changes/AbstractCopy.php
+++ b/Classes/Domain/Model/Changes/AbstractCopy.php
@@ -36,6 +36,7 @@ abstract class AbstractCopy extends AbstractStructuralChange
         $updateNodeInfo->setNode($node);
         $updateNodeInfo->recursive();
         $this->feedbackCollection->add($updateNodeInfo);
+        $this->addNodeCreatedFeedback($this->getSubject());
         parent::finish($node);
     }
 
--- a/Classes/Domain/Model/Changes/AbstractCopy.php
+++ b/Classes/Domain/Model/Changes/AbstractCopy.php
@@ -36,7 +36,7 @@ abstract class AbstractCopy extends AbstractStructuralChange
         $updateNodeInfo->setNode($node);
         $updateNodeInfo->recursive();
         $this->feedbackCollection->add($updateNodeInfo);
-        $this->addNodeCreatedFeedback($this->getSubject());
+        $this->addNodeCreatedFeedback($node);
         parent::finish($node);
     }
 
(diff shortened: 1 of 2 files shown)


Classes/Domain/Model/Changes/AbstractCopy.php
```diff
@@ -36,6 +36,7 @@ protected function finish(NodeInterface $node): void
         $updateNodeInfo->setNode($node);
         $updateNodeInfo->recursive();
         $this->feedbackCollection->add($updateNodeInfo);
+        $this->addNodeCreatedFeedback($this->getSubject());
```
mhsdesign: why do we need this line here? Rest looks understandable:)!

pKallert: This will select the new document as active, basically does the same as Redirect (from a UI behavior standpoint).
Fun fact: This is already included in Neos 9, but there the AbstractCopy does not exist anymore: https://github.com/neos/neos-ui/blob/9.0/Classes/Domain/Model/Changes/CopyBefore.php#L88

mhsdesign:
```suggestion
        $this->addNodeCreatedFeedback($node);
```

okay in my test i had to change this line to use the `$node` as `getSubject` seems to be the old node which was copied which resulted in moving back to the copy source location instead of the target.

Classes/Domain/Model/Changes/AbstractMove.php
```diff
@@ -40,7 +41,11 @@ protected function finish(NodeInterface $node)
             $updateNodePath->setNewContextPath($this->getSubject()->getContextPath());
             $this->feedbackCollection->add($updateNodePath);
         }
-
+        if ($this->getSubject()->getNodeType()->isOfType('Neos.Neos:Document')) {
```
mhsdesign, line 44: this seems to work well. and for content nodes it seems that `RenderContentOutOfBand` takes care of highlighting the node where it was moved to.

pKallert: yes, I was pretty happy to find something that works out of the box since I did not want to trigger addNodeCreatedFeedback on a node move

Sebobo (approved):
Works great, thx!

Sebobo:
@pKallert can you retry the test, I cannot trigger it again.

dlubitz (approved):
LGTM!

dlubitz:
@Sebobo As the e2e tests seem to be broken in general, can we merge this anyways?

Issue #3367, opened 2023-02-15
Cleanup our git repos `.yarn/cache` folder


mhsdesign:
With the upgrade to yarn 3 in https://github.com/neos/neos-ui/pull/3116 we check in all zip files of all dependencies in the `.yarn` folder so you don't need to install anything form a remote. Therefore achieving exactly the same build for every dev.

Now with changes to dependencies like the removal of 1000 unneeded ones (https://github.com/neos/neos-ui/pull/3382) or big upgrades of testcafe, ckeditor and the like there will be a lot of movement in the yarn cache folder of newly added zips and a lot of removed ones.

Our git history holds a lot of unused cached yarn packages, and this will make downloading this repo slower on the long run.

We should somehow clear cache entries, that are not used actively in any branch

Via the command

```
git rev-list --objects --all |                                     
  git cat-file --batch-check='%(objecttype) %(objectname) %(objectsize) %(rest)' |
  sed -n 's/^blob //p' |
  sort --numeric-sort --key=2 |
  cut -c 1-12,41-
```

i could find out that the biggest files indeed are just build related - and sometimes just accidentally checked in (Host.js.map & Host.js) and also as expected a lot of `zip` files say hi.

```
0b836e14902f 961867 yarn.lock
a50b0fdf0599 962266 yarn.lock
2699a267a03b 962566 yarn.lock
242d97157b0f 963709 yarn.lock
35e4b54fbb92 963709 yarn.lock
924350a9bbbf 964496 yarn.lock
bac0ab373d28 965271 yarn.lock
284df683c7b9 973740 .yarn/cache/@storybook-addon-knobs-npm-3.2.16-405c3e747d-e83327c1ce.zip
4c0ca4220a3f 984413 .yarn/cache/core-js-npm-2.6.11-15178ded27-6944011e7a.zip
1aabca37bd3e 1012514 .yarn/cache/@fortawesome-free-brands-svg-icons-npm-5.15.3-ebacfdf66f-03d65f66de.zip
96bfde0df4cb 1041577 .yarn/cache/core-js-pure-npm-3.26.1-d78c7431c6-d88c40e5e2.zip
5080d3cb2b11 1057225 .yarn/cache/@ckeditor-ckeditor5-special-characters-npm-43.3.1-4bd25c2915-ccbfb83d11.zip
[...]

PR #3920, opened 2025-02-11
TASK: Optimize cloning in update-neos-ui-compiled.sh

kdambekalns:
The script clones `git@github.com:neos/neos-ui-compiled.git` – and that is a **massive** repository. Cloning it uses a lot of space (over 1 GB) and takes long, as the Jenkins log shows (timestamps are elapsed time):

```
00:00:50.125 + git clone git@github.com:neos/neos-ui-compiled.git tmp_compiled_pkg
00:00:50.137 Cloning into 'tmp_compiled_pkg'...
00:21:54.271 + cd tmp_compiled_pkg
00:21:54.272 + git checkout origin/8.4
```

That took… more than 20 minutes! Yikes!

I suggest to do this instead:

```
git clone --depth 1 git@github.com:neos/neos-ui-compiled.git --single-branch --branch=${BRANCH} tmp_compiled_pkg
```

Takes about 10 seconds for me, transfers around 5 MB of data, uses 23 MB of disk space and even saves us the `git checkout`, as the branch is already taken care of.

### DIFF
--- a/Build/Jenkins/update-neos-ui-compiled.sh
+++ b/Build/Jenkins/update-neos-ui-compiled.sh
@@ -38,11 +38,7 @@ make clean && make setup
 NEOS_UI_VERSION="${GIT_TAG:-${GIT_BRANCH#*/}-dev}" make build-production
 
 rm -Rf tmp_compiled_pkg
-git clone git@github.com:neos/neos-ui-compiled.git tmp_compiled_pkg
-cd tmp_compiled_pkg
-git checkout "$GIT_BRANCH"
-cd ..
-
+git clone --depth 1 git@github.com:neos/neos-ui-compiled.git --single-branch --branch=${GIT_BRANCH} tmp_compiled_pkg
 
 mkdir -p tmp_compiled_pkg/Resources/Public/Build
 


kdambekalns:
"Fun" fact: All our Neos checkouts on Jenkins (as they prefer source checkouts) have a 1 GB Neos.Neos.Ui.Compiled package… 😑 Too bad we cannot seem to change that, as the dev versions can only be checked out "from cache" via Composer, even overriding the `preferred-install` for that package doesn't change that.

See also https://github.com/composer/composer/issues/10074#issuecomment-2652195920

kdambekalns:
New issue with Composer: https://github.com/composer/composer/issues/12310 🤞

markusguenther (approved):
Thanks for that nice improvement. :)

mhsdesign (approved):
Thank you for noticing this and fixing it ;)

This definitely goes into the direction of https://github.com/neos/neos-ui/issues/3367 -> through our `.yarn` cache folder the repo just grows quite fast ... we might need to promote similar ways to checkout also for developers so they save diskspace. Maybe in case we still want to merge the Neos Ui repo into Neos.Neos we can avoid these heavy artifacts to be merged.

PR #3921, opened 2025-02-12
TASK: Add ability for read-only users to interact with nodes in the preview page and view node data in the inspector

JamesAlias:
**What I did**
Previous Behaviour:
- When a view-only user clicks on a node on the preview page nothing happens.
- The user can only select nodes via the content tree (left sidebar)
- The user can not see any node data in the inspector (right sidebar)

Fixed Behaviour:
- The user can interact with nodes on the preview page (highlighting is visible)
- The user can view - but not edit - node data in the inspector (right sidebar)

**How I did it**
- I let the backend send metadata to the client when the user has permission to read (previously it was only sent, when the user had permission to edit).
- I handle the read-only case in CKEditor to prevent the user from editing and throwing errors that way.

This should also apply to read-only workspaces.

> Note: When I tested editing content via circumventing browser behaviour (disabled fields etc) the server responded with an error. This has not been a very thorough test though.

**How to verify it**
- Create a role (e.g. with the Sandstorm.NeosAcl package) that has permission to view but not to edit/create.
- When browsing content the user should be able to click on nodes in the preview page and see their data in the inspector.

**BEFORE**
Screenshot before change
**AFTER**
Screenshot after change

### DIFF
--- a/Classes/Aspects/AugmentationAspect.php
+++ b/Classes/Aspects/AugmentationAspect.php
@@ -22,6 +22,7 @@ use Neos\Fusion\Service\HtmlAugmenter;
 use Neos\Neos\Domain\Service\ContentContext;
 use Neos\Neos\Ui\Domain\Service\UserLocaleService;
 use Neos\Neos\Ui\Fusion\Helper\NodeInfoHelper;
+use Neos\Neos\Ui\Service\NodePolicyService;
 
 /**
  * - Serialize all nodes related to the currently rendered document
@@ -60,6 +61,12 @@ class AugmentationAspect
      * @var \Neos\Flow\Mvc\Controller\ControllerContext
      */
     protected $controllerContext = null;
+
+    /**
+     * @Flow\Inject
+     * @var NodePolicyService
+     */
+    protected $nodePolicyService;
 
     /**
      * @Flow\Before("method(Neos\Neos\Fusion\ContentElementWrappingImplementation->evaluate())")
@@ -145,7 +152,8 @@ class AugmentationAspect
             return $content;
         }
 
-        if ($this->nodeAuthorizationService->isGrantedToEditNode($node) === false) {
+        // TODO: this is where we could use a feature flag
+        if ($this->nodePolicyService->canReadNode($node) === false) {
             return $content;
         }
 
--- a/Classes/Aspects/AugmentationAspect.php
+++ b/Classes/Aspects/AugmentationAspect.php
@@ -152,7 +152,6 @@ class AugmentationAspect
             return $content;
         }
 
-        // TODO: this is where we could use a feature flag
         if ($this->nodePolicyService->canReadNode($node) === false) {
             return $content;
         }
--- a/Classes/Aspects/AugmentationAspect.php
+++ b/Classes/Aspects/AugmentationAspect.php
@@ -152,7 +152,7 @@ class AugmentationAspect
             return $content;
         }
 
-        if ($this->nodePolicyService->canReadNode($node) === false) {
+        if ($this->nodePolicyService->canReadProperties($node) === false) {
             return $content;
         }
 
--- a/Classes/Service/NodePolicyService.php
+++ b/Classes/Service/NodePolicyService.php
@@ -82,6 +82,7 @@ class NodePolicyService
             'disallowedNodeTypes' => $this->getDisallowedNodeTypes($node),
             'canRemove' => $this->canRemoveNode($node),
             'canEdit' => $this->canEditNode($node),
+            'canView' => $this->canReadNode($node),
             'disallowedProperties' => $this->getDisallowedProperties($node)
         ];
     }
@@ -98,6 +99,22 @@ class NodePolicyService
 
         return $this->privilegeManager->isGranted(
             NodeTreePrivilege::class,
+            new NodePrivilegeSubject($node)
+        );
+    }
+
+    /**
+     * @param NodeInterface $node
+     * @return bool
+     */
+    public function canReadNode(NodeInterface $node): bool
+    {
+        if (!isset(self::getUsedPrivilegeClassNames($this->objectManager)[ReadNodePropertyPrivilege::class])) {
+            return true;
+        }
+
+        return $this->privilegeManager->isGranted(
+            ReadNodePropertyPrivilege::class,
             new NodePrivilegeSubject($node)
         );
     }
--- a/Classes/Service/NodePolicyService.php
+++ b/Classes/Service/NodePolicyService.php
@@ -19,6 +19,7 @@ use Neos\ContentRepository\Security\Authorization\Privilege\Node\EditNodePrivile
 use Neos\ContentRepository\Security\Authorization\Privilege\Node\EditNodePropertyPrivilege;
 use Neos\ContentRepository\Security\Authorization\Privilege\Node\NodePrivilegeSubject;
 use Neos\ContentRepository\Security\Authorization\Privilege\Node\PropertyAwareNodePrivilegeSubject;
+use Neos\ContentRepository\Security\Authorization\Privilege\Node\ReadNodePrivilege;
 use Neos\ContentRepository\Security\Authorization\Privilege\Node\RemoveNodePrivilege;
 use Neos\Flow\Annotations as Flow;
 use Neos\ContentRepository\Domain\Model\NodeInterface;
@@ -82,7 +83,7 @@ class NodePolicyService
             'disallowedNodeTypes' => $this->getDisallowedNodeTypes($node),
             'canRemove' => $this->canRemoveNode($node),
             'canEdit' => $this->canEditNode($node),
-            'canRead' => $this->canReadNode($node),
+            'canReadProperties' => $this->canReadProperties($node),
             'disallowedProperties' => $this->getDisallowedProperties($node)
         ];
     }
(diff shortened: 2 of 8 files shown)


packages/neos-ui-guest-frame/src/initializePropertyDomNode.js
```diff
@@ -4,6 +4,8 @@ import {actions} from '@neos-project/neos-ui-redux-store';
 import {validateElement} from '@neos-project/neos-ui-validators';
 
 import {getGuestFrameWindow, closestContextPathInGuestFrame} from './dom';
+// TODO: this import feels foreign to the rest of neos code, but it feels strange to import the complete selectors object
+import {isWorkspaceReadOnlySelector} from '@neos-project/neos-ui-redux-store/src/CR/Workspaces/selectors';
```
JamesAlias: Consider using

```js
import { selectors } from '@neos-project/neos-ui-redux-store';

// ...

const isWorkspaceReadOnly = selectors.CR.Workspaces.isWorkspaceReadOnlySelector(store.getState());
```

packages/neos-ui-guest-frame/src/initializeGuestFrame.js
```diff
@@ -112,9 +112,11 @@ export default ({globalRegistry, store}) => function * initializeGuestFrame() {
 
     const editPreviewMode = $get(['ui', 'editPreviewMode'], state);
     const editPreviewModes = globalRegistry.get('frontendConfiguration').get('editPreviewModes');
-    const isWorkspaceReadOnly = selectors.CR.Workspaces.isWorkspaceReadOnlySelector(state);
     const currentEditMode = editPreviewModes[editPreviewMode];
-    if (!currentEditMode || !currentEditMode.isEditingMode || isWorkspaceReadOnly) {
+
+    // ReadOnly workspaces are handled by ckEditor directly
+    // TODO: this will break other editors, correct?
```
JamesAlias: This might be true, but afaik there are no other editors used.

Classes/Aspects/AugmentationAspect.php
```diff
@@ -145,7 +152,8 @@ public function editableElementAugmentation(JoinPointInterface $joinPoint)
             return $content;
         }
 
-        if ($this->nodeAuthorizationService->isGrantedToEditNode($node) === false) {
+        // TODO: this is where we could use a feature flag
```
JamesAlias: remove this TODO

composer.json
```diff
@@ -7,7 +7,7 @@
   ],
   "require": {
     "neos/neos": "^8.3.0",
-    "neos/neos-ui-compiled": "self.version"
+    "neos/neos-ui-compiled": "8.4.x-dev"
```
JamesAlias, line 10: **!!!Revert this before merging!!!**

> I had to do this to allow composer to install an existing `neos/neos-ui-compiled` version.

skurfuerst (approved):
I really like this change; I reviewed this in depth.

IMHO, the functional change is a BUGFIX; so no need for a feature flag from my side.

Thanks <3

Classes/Service/NodePolicyService.php
```diff
@@ -82,6 +82,7 @@ public function getNodePolicyInformation(NodeInterface $node): array
             'disallowedNodeTypes' => $this->getDisallowedNodeTypes($node),
             'canRemove' => $this->canRemoveNode($node),
             'canEdit' => $this->canEditNode($node),
+            'canView' => $this->canReadNode($node),
```
Sebobo: Would prefer `canRead` to implicate the relationship to the privilege and other methods.

Sebobo (changes requested):
Thx for the change, this is useful!

Please look at my suggestion and solve the todos.

Classes/Service/NodePolicyService.php
JamesAlias: done

composer.json
Sebobo: You could have probably also did this override from your root composer.json.
I don' remember having this issue.

packages/neos-ui-ckeditor5-bindings/src/ckEditorApi.js
```diff
@@ -58,26 +58,34 @@ export const createEditor = store => async options => {
...
         .then(editor => {
-            const debouncedOnChange = debounce(() => onChange(cleanupContentBeforeCommit(editor.getData())), 1500, {maxWait: 5000});
-            editor.model.document.on('change:data', debouncedOnChange);
-            editor.ui.focusTracker.on('change:isFocused', event => {
-                if (!event.source.isFocused) {
-                    // when another editor is focused commit all possible pending changes
-                    debouncedOnChange.flush();
-                    return
-                }
+            editor.isReadOnly = isReadOnly;
```
mhsdesign, line 61: i think i also attempted to use the `IsReadOnly` flag from ck and that did not work always correctly ... instead in https://github.com/neos/neos-ui/pull/3842 i made it work by just NOT booting the ckeditor when there is no reason too and think we should do the same here? Not booting the ckeditor and still showing an outline or something is imo much more stable.

Sebobo (approved):
Works fine for me now.

Just noticed one editor that doesn't show a proper disabled state: 
CleanShot 2025-02-25 at 08 43 28@2x

Not related to your change though.

Classes/Service/NodePolicyService.php
```diff
@@ -102,6 +103,22 @@ public function isNodeTreePrivilegeGranted(NodeInterface $node): bool
         );
     }
 
+    /**
+     * @param NodeInterface $node
+     * @return bool
+     */
+    public function canReadNode(NodeInterface $node): bool
+    {
+        if (!isset(self::getUsedPrivilegeClassNames($this->objectManager)[ReadNodePropertyPrivilege::class])) {
```
mhsdesign, line 112: The symbol `ReadNodePropertyPrivilege` was not imported and thus this is just a noop and returns always `true`

mhsdesign: The symbol `ReadNodePropertyPrivilege` was not imported and thus this is just a noop and returns always `true`

composer.json
mhsdesign: ☝️

Classes/Service/NodePolicyService.php
mhsdesign: And when i import it i get a an exception

```
ReadNodePropertyPrivilege only support subjects of type PropertyAwareNodePrivilegeSubject or MethodPrivilegeSubject, but we got a subject of type: NodePrivilegeSubject
```

mhsdesign (changes requested):
Okay aside from the comment regarding the ckeditor thing i actually cant get this change to work:

The neos ui doesnt go in restricted editing mode and i can still edit things and get raw errors back

```
Access denied for method Method: Neos\ContentRepository\Domain\Model\Node::setProperty() Evaluated following 1 privilege target(s): "RestrictEditing": ABSTAIN (0 granted, 0 denied, 1 abstained) Authenticated roles: Neos.Flow:Everybody, Neos.Flow:AuthenticatedUser, Neos.Neos:Administrator, Neos.Neos:Editor, Neos.Neos:AbstractEditor, Neos.ContentRepository:Administrator, Neos.ContentRepository:InternalWorkspaceAccess, Neos.Neos:LivePublisher
```

this is the configuration i used

```
privilegeTargets:
  'Neos\ContentRepository\Security\Authorization\Privilege\Node\EditNodePrivilege':
    'RestrictEditing':
      matcher: 'isDescendantNodeOf("/sites/")'

  'Neos\ContentRepository\Security\Authorization\Privilege\Node\ReadNodePropertyPrivilege':
    'RestrictReading':
      matcher: 'isDescendantNodeOf("/sites/")'

roles:
  'Neos.Neos:Editor':
    privileges:
      -
        privilegeTarget: 'RestrictReading'
        permission: GRANT
```

Classes/Service/NodePolicyService.php
JamesAlias: We decided to use the `ReadNodePrivilege`, which actually exists.

Some Context: The user will not see a node he can not read, so this check is obsolete. We still keep it for AOP shenanigans though.

packages/neos-ui-ckeditor5-bindings/src/ckEditorApi.js
JamesAlias: I see your point and feel free to implement this. For me personally I don't have the time and as long as it works like intended I will just keep it like this.

composer.json
JamesAlias: I will revert this right before merging

Classes/Service/NodePolicyService.php
mhsdesign: okay i will have to retest now as we use the `ReadNodePrivilege` now instead of `ReadNodePropertyPrivilege`

JamesAlias: @mhsdesign I will merge this next week on Tuesday or Wednesday.
If you like, you can test this again. 
Thank you for the reviews so far 😊

mhsdesign:
Hi dear Robert.

We just dug out this one ;)

Sebastian and me just re-reviewed and retested this to evaluate if it can make it as a late candidate in to the 8.4 release.

In the demo i used this policy so no one can edit the blog page of the demo

```yaml
privilegeTargets:
  'Neos\ContentRepository\Security\Authorization\Privilege\Node\EditNodePrivilege':
    'RestrictEditing':
      matcher: 'isDescendantNodeOf("/sites/neosdemo/node-iyyapnswso4eu")'
```

On a blog page

image

Typing also still continues to work in the inspector and yields raw errors.

> Access denied for method Method: Neos\ContentRepository\Domain\Model\Node::setProperty() Evaluated following 1 privilege target(s): "RestrictEditing": ABSTAIN (0 granted, 0 denied, 1 abstained) Authenticated roles: Neos.Flow:Everybody, Neos.Flow:AuthenticatedUser, Neos.Neos:Editor, Neos.Neos:AbstractEditor, Neos.ContentRepository:Administrator, Neos.ContentRepository:InternalWorkspaceAccess, Neos.Neos:LivePublisher


Sebastian and me found a place to disable the editor in the `InspectorEditorEnvelope::options()` method by checking if `node.policy.canEdit` is false. This would disable the editors which is i think what this change wants to do?

image


Typing also still continues to work in CKEditor and yields raw errors and needs some love?.

> Access denied for method Method: Neos\ContentRepository\Domain\Model\Node::setProperty() Evaluated following 1 privilege target(s): "RestrictEditing": ABSTAIN (0 granted, 0 denied, 1 abstained) Authenticated roles: Neos.Flow:Everybody, Neos.Flow:AuthenticatedUser, Neos.Neos:Editor, Neos.Neos:AbstractEditor, Neos.ContentRepository:Administrator, Neos.ContentRepository:InternalWorkspaceAccess, Neos.Neos:LivePublisher

-------------

So thank you for your contribution but as the current state seems not to be working as expected which would need additional  work last minute and we would not feel comfortable introducing a change to the neos ui policies that close before the release. Please dont be discouraged!

t-heuser:
I implemented all the requested changes from here and created a new PR as successor of this PR: https://github.com/neos/neos-ui/pull/4152

skurfuerst:
closing this PR (as Timon cannot push to this repo directly).

PR #3922, opened 2025-02-21
WIP: bugfix prevent deletion of subtrees with pending changes

mhsdesign:
**What I did**

**How I did it**

**How to verify it**


mhsdesign:
this solution doesn't work if the node was only temporary moved into this tree part once but doesnt reside here anymore today and never has in the base workspace! So its not a full fix for the problems of https://github.com/neos/neos-development-collection/issues/5459

Instead we will use soft removals: https://github.com/neos/neos-development-collection/issues/5478

PR #3926, opened 2025-03-07
TASK: Auto rebase workspace on backend module opening if necessary

mhsdesign:
Originally there was a `EditorContentStreamZookeeper` which task was it to create the users workspace on login and also always have it rebased on login.
Then it was adjusted to only rebase if really necessary by using the outdated check: https://github.com/neos/neos-development-collection/issues/4728
But because rebases could likely cause exceptions / legacy OUTDATED_CONFLICT state the rebase on login feature was removed: https://github.com/neos/neos-development-collection/commit/dadb444124ac32483f361ace97d065a32ae1ac42
After all the whole service was removed when the workspace creation was moved to the UI's `BackendController` https://github.com/neos/neos-development-collection/commit/7a2e68abaedf25420a88118644e527575381089a 

Now we discussed that as rebasing is a new 9.0 topic we want to avoid the burden for users who only login to see the current state without even having changes, thus we reintroduce the automatic rebase but directly in the `BackendController`, and in difference than before with the get pending changes check which should prevent a rebase exception.

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/Classes/Controller/BackendController.php
+++ b/Classes/Controller/BackendController.php
@@ -13,6 +13,7 @@ namespace Neos\Neos\Ui\Controller;
  */
 
 use Neos\ContentRepository\Core\SharedModel\Node\NodeAddress;
+use Neos\ContentRepository\Core\SharedModel\Workspace\WorkspaceStatus;
 use Neos\ContentRepositoryRegistry\ContentRepositoryRegistry;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\Mvc\Controller\ActionController;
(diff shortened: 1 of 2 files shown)


nezaniel (approved):
looks good

mhsdesign:
thanks for the flag - understandable because now the soft removal garbage collection is also triggered.
i normally dislike feature flags, and hope that with https://github.com/neos/neos-development-collection/issues/5264 we find a good performant solution for all projections.

mhsdesign:
The tests are also broken on 9.0 because softremoval avoids conflicts

Issue #3342, opened 2023-01-25
BUG: right sidebar in Neos UI missing z-index definition


patricekaufmann:
### Description

When the UI switched to css grids in 8.2 the right sidebar switched from `position:fixed` to `position:relative`. It misses a direct or an indirect z-index definition. None of the parent elements has `position: relative` and `z-index` set. This leads to bugs with custom inspector plugins that use z-index for elements that have z-index set with relative positioning:

![image](https://user-images.githubusercontent.com/41791606/214556835-c535e561-c58e-44ab-b811-9267341057ec.png)

### Possible solution

For any parent of the right sidebar, set  `position: relative` and `z-index`  to anything below `z-index` for the dialog container (which is `1`, maybe set it higher?).

### Affected Versions

UI: 8.2

Sebobo:
Thx for reporting, will try to fix that this week.

Sebobo:
Also causes this problem 

![Image](https://github.com/user-attachments/assets/9fb945ff-8124-49e1-8c66-be3152777a9d)

I'll finally fix it now

PR #3929, opened 2025-03-26
BUGFIX: Have base z-index for inspector to make relative z-indices work

Sebobo:
With this change the issue is solved that the help message of properties is in front of the publish dropdown and custom inspector plugins with z-indices caused visual issues with the creation dialog overlay.

Resolves: #3342

Before:

CleanShot 2025-03-26 at 13 14 11

After:

CleanShot 2025-03-26 at 13 14 19@2x


Sebobo:
Not sure right now why the e2e test with the select box fails.

markusguenther (approved):
Looks good to me. Also don`t see why the test should fail.

PR #3930, opened 2025-03-26
BUGFIX: Dont crash if no site exists

mhsdesign:
**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/Classes/Infrastructure/Neos/MenuProvider.php
+++ b/Classes/Infrastructure/Neos/MenuProvider.php
@@ -75,13 +75,13 @@ final class MenuProvider implements MenuProviderInterface
 
         $result = [];
         foreach ($sitesForMenu as $index => $site) {
-            $name = $site['name'];
+            $name = $site['name'] ?? null;
             $name = is_string($name) ? $name : 'N/A';
 
             $uri = $site['uri'];
             $uri = is_string($uri) ? $uri : '#';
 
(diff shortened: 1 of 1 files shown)


Classes/Infrastructure/Neos/MenuProvider.php
```diff
@@ -75,13 +75,13 @@ private function buildChildrenForSites(ControllerContext $controllerContext): ar
 
         $result = [];
         foreach ($sitesForMenu as $index => $site) {
```
Sebobo, line 77: Why does the array even contain an incomplete item?

Sebobo (changes requested):

Sebobo:
@mhsdesign I wanted to reproduce what this fixes, as I didn't unterstand how the array could have a site without the necessary fields. But when I deactivate the last site I get the following exception instead:

```
Current site and content repository could not be extracted from the Request.
The SiteDetectionMiddleware was not able to determine the site!

Exception Code	1699459565
Exception Type	Neos\Neos\FrontendRouting\SiteDetection\SiteDetectionFailedException
Thrown in File	Packages/Neos/Neos.Neos/Classes/FrontendRouting/SiteDetection/SiteDetectionResult.php
Line	68
```

Which is of course also not optimal, so when the site detection fails, we should redirect to $something.

PR #3933, opened 2025-03-28
FEATURE: Add possibility to display warning

pKallert:
**What I did**
When extending the Neos UI it would be great to have the possibility to easily show a warning. 
The use case is: When a user sets a Document to 'hidden', I would like to show a message if the page exists as the target for a redirect. 
In FlashMessages there is a Message::SEVERITY_WARNING that can be set, but in the Neos UI, using the Feedback collections, there is only the possibility to display Success, error, and info(which is logged to console). 

It is of course possible to display a message using neos extensibility but I thought it would be nice to have an easy way in the Neos UI and be consistent with the flashMessages.

**How I did it**
I added a new Feedback class "Warning" that will show up as an orange message in the Neos UI 

**How to verify it**
```
$warning = new Warning();
$warning->setMessage('This page exists as a source in redirects.');
$this->feedbackCollection->add($warning);
```
will show up as warning in the UI: 

Bildschirmfoto 2025-03-28 um 08 59 22

### DIFF
--- a/Classes/Domain/Model/Feedback/Messages/Warning.php
+++ b/Classes/Domain/Model/Feedback/Messages/Warning.php
@@ -1,0 +1,42 @@
+<?php
+namespace Neos\Neos\Ui\Domain\Model\Feedback\Messages;
+
+/*
+ * This file is part of the Neos.Neos.Ui package.
+ *
+ * (c) Contributors of the Neos Project - www.neos.io
+ *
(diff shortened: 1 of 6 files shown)


Sebobo (approved):
Thx, looks good by reading

dlubitz (approved):
Looks good by reading

PR #3934, opened 2025-03-29
TASK: Fix PHP 8.4 deprecations

nezaniel:
... by removing implicit nullability - 8.3 version

### DIFF
--- a/Classes/Aspects/AugmentationAspect.php
+++ b/Classes/Aspects/AugmentationAspect.php
@@ -57,7 +57,7 @@ class AugmentationAspect
      * inside the wrapContentObject() advice. It will not be necessary once we implement a custom content element
      * wrapping implementation.
      *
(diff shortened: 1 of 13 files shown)


dlubitz (approved):

dlubitz:
This now raises deprecations in PHP 8.3 😢 

```
Deprecated: Optional parameter $haystack declared before required parameter $needle is implicitly treated as a required parameter in /var/www/html/Packages/Application/Neos.Neos.Ui/Classes/Domain/Service/NodeTreeBuilder.php on line 227

Deprecated: Optional parameter $node declared before required parameter $controllerContext is implicitly treated as a required parameter in /var/www/html/Packages/Application/Neos.Neos.Ui/Classes/Fusion/Helper/NodeInfoHelper.php on line 409
```

PR #3935, opened 2025-04-01
TASK: Upmerge 8.4 into 9.0

nezaniel:
Important: Review JS changes from 8.4 to 9.0 as they were mostly removed


pKallert (approved):
Tested it out locally and it works again

Issue #3946, opened 2025-04-29
BUG: Cachebuster appends not standard conform parameter


Sebobo:
The hash that is appended to Neos.Ui resource Uris in `StyleAndJavascriptInclusionService->build` only appends a hash without a parameter name. 
This leads at least in the latest Quobyte S3 storage to an error and could also cause issues with other storages if they expect key value pairs.

I was able to resolve that in the affected project by adding a parameter name "bust=" before the hash parameter.

In addition there are 2 bust parameters if one installs `flowpack/cachebuster` f.e..

![Image](https://github.com/user-attachments/assets/1b0147e2-8567-47a1-a6f1-87e6075faf07)

PR #3947, opened 2025-04-29
BUGFIX: Add parameter name for js/css cachebuster

Sebobo:
Resolves: #3946

**What I did**

Give cache buster parameter a name to have a key & value pair in the query string

**How I did it**

Add configurable cache buster parameter name 

**How to verify it**

Inspect the Neos backend and check that all JS and CSS files loaded by the Neos.Ui like plugins etc. have the parameter appended with a name and value.


markusguenther (approved):
Thanks for taking care. It looks good to me by reading.

kitsunet (approved):

PR #3948, opened 2025-05-06
Optional parameter $node declared before required parameter

patriceckhart:
Solves the problem that occurs when, for example, the backend is called for the first time. In the DOM
`Deprecated: Optional parameter $node declared before required parameter controllerContext is implicitly treated as a required parameter in /tmp/Flow/Development/Cache/Code/Flow_Object_Classes/Neos_Neos_Ui_Fusion_Helper_NodelnfoHelper.php on line 409` is rendered.


markusguenther:
Thank you so much for the PR. I released the fix and am sorry again that I did not spot that in version 8.3.12.

https://github.com/neos/neos-ui/releases/tag/8.3.13

PR #3949, opened 2025-05-07
BUGFIX: Solve deprecation warnings in PHP 8.3

dlubitz:
The fix of PHP8.4 deprecation warnings has introduced new deprecation warnings in PHP 8.3.

https://github.com/neos/neos-ui/pull/3934

```
Deprecated: Optional parameter $haystack declared before required parameter $needle is implicitly treated as a required parameter in /var/www/html/Packages/Application/Neos.Neos.Ui/Classes/Domain/Service/NodeTreeBuilder.php on line 227

Deprecated: Optional parameter $node declared before required parameter $controllerContext is implicitly treated as a required parameter in /var/www/html/Packages/Application/Neos.Neos.Ui/Classes/Fusion/Helper/NodeInfoHelper.php on line 409
```

Both parameter are rather nullable than optional.

### DIFF
--- a/Classes/Domain/Service/NodeTreeBuilder.php
+++ b/Classes/Domain/Service/NodeTreeBuilder.php
@@ -224,7 +224,7 @@ class NodeTreeBuilder
         return $result;
     }
 
-    protected function isInRootLine(?NodeInterface $haystack = null, NodeInterface $needle)
+    protected function isInRootLine(?NodeInterface $haystack, NodeInterface $needle)
     {
         if ($haystack === null) {
             return false;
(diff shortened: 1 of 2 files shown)


markusguenther (approved):
Thanks for taking care @dlubitz 💙

markusguenther:
Would prefer this solution instead of https://github.com/neos/neos-ui/pull/3948
But awesome that @patriceckhart reported that issue and worked on his first contribution for the Neos-UI 👏

kitsunet:
Why don't we switch the order?
Like this was probalby what we shoudl have done 10 years ago^^
To not block a ✅

kitsunet (approved):

dlubitz:
> Why don't we switch the order?

That would be BC in any case. And I assume this was never ment to be optional. Just nullable to make it less breaky if the node is empty for whatever reason.

kitsunet:
None of this is API but fair enough

PR #3954, opened 2025-05-17
BUGFIX: Proper error if site node not found in backend

mhsdesign:
It seems the (internal) contract for providing the UI state requires no site or document node, but this is not true (obviously)
The `defaultNodesForBackend` helper will just crash:

> defaultNodesForBackend(): Argument #1 ($site) must be of type Neos\ContentRepository\Core\Projection\ContentGraph\Node, null given

It is invoked here via YAML - hence we did not detect this null pointer bug:

https://github.com/neos/neos-ui/blob/ebed35152bf480329cb54a18ec40d49fc0076673/Configuration/Settings.yaml#L131

Instead now a pretty error is throw which will also highlight probably its cause:

```
Site node "neosdemo" not found in content repository "default" in dimension space point {"language":"en_US","targetGroup":"private"} and visibility constraints "removed"
```

In my case i had a configuration laying around adding a new dimension `targetGroup` for all sites which was un nice which lead to no nodes being found.


dlubitz (approved):
LGTM! If the interface is internal I'm fine with merging this into 9.0 as a bugfix.

JamesAlias (approved):
Thanks! Reviewed and checked with @fdchriswaechter on contribution day

PR #3955, opened 2025-05-20
BUGFIX: Translate selectbox group labels

Sebobo:
**What I did**

Made it possible to have localised group labels for select box values

**How I did it**

Translate group in addition to the already translated label.
This change needs an additional fix in Neos itself to convert i18n values. But with this change shorthand translation strings already work.

**How to verify it**

CleanShot 2025-05-20 at 22 05 04@2x


dlubitz (approved):

dlubitz:
Only the tests need to be adapted.

Sebobo:
Makes no sense to fill the object with empty values, I've adjusted the code to only modify and not add.

markusguenther (approved):
Works fine :)
Thanks for taking care!

PR #3957, opened 2025-06-21
BUGFIX: (CSS): add visual selection of multiple cells in table in ckeditor

dasoe:
**What I did**
Added CSS for selected cells in tables in CSS editor. The CSS is adding blue outlines to selected cells when multiple cells are selected by dragging the mouse.

**How I did it**
Added 2 CSS definitions in  packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/TableStyles.vanilla-css

Screenshot from 2025-06-21 13-25-32

### DIFF
--- a/packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/TableStyles.vanilla-css
+++ b/packages/neos-ui-ckeditor5-bindings/src/EditorToolbar/TableStyles.vanilla-css
@@ -136,6 +136,11 @@
 .ck-editor__editable.ck-blurred .ck-widget.ck-widget_selected .ck-widget__selection-handle,.ck-editor__editable.ck-blurred .ck-widget.ck-widget_selected .ck-widget__selection-handle:hover,.ck-editor__editable.ck-blurred .ck-widget.ck-widget_selected:hover .ck-widget__selection-handle,.ck-editor__editable.ck-blurred .ck-widget.ck-widget_selected:hover .ck-widget__selection-handle:hover{
     background:var(--ck-color-widget-blurred-border)
 }
+.ck-editor__editable.ck-editor__nested-editable.ck-editor__editable_selected,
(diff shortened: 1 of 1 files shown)


markusguenther:
Thanks @dasoe for providing this change, as we have Neos 8.3 still in maintain mode. We would like to have this bugfix also in version 8.3. The process is to fix the issue in the lowest maintained branch, and after the merge, we bring that to the newer branches as well.

Would you like to try to rebase the PR to version 8.3, or should one of us take care?
Thanks in advance.

dasoe:
Thanks for getting back. Will happily do so. [...]

mhsdesign:
fyi that would be

```
git checkout bugfix/fix-ck-editor-css
git reset --hard origin/8.3
git cherry-pick 5a62c11f0f63ab38d9d8b08ac1a15fa219ecd0a4
git push --force
```

dasoe:
Thanks, that's so kind! 
In 8.3, I can not use the fix, as ckeditor seems to not set any classes to the selected cells, except first selected one (in other words mouse dragging does not work for selecting cells). 
So: might get more complicated. Investigating...

markusguenther:
You are Right. In Neos 8.3 we use CKeditor 5 in Version 16 and in Neos 9.0 in Version 44.

So we are already again one Version behind ;)

bwaidelich:
@dasoe Thanks a lot for your contribution.

So this change adds the blue outlines to selected cells in the table within RTE content, right? Shall we update the description to reflect that?

dasoe:
Yes, I've seen that. Already tried upgrading to version 44 for testing but no change.
Just to be sure (sorry for those basic questions): After editing `package.json`, running `composer install` should be sufficient to have NEOS use the new version, right? I also restarted ddev and flushed the caches NEOS and browser to be sure.

Sebobo (approved):
Works great, thx!

PR #3961, opened 2025-06-24
BUGFIX: Skip content childnodes when returning document tree nodes

### LINKED ISSUE #1508: Tree search is dead slow on large projects => rewrite the tree strucutre

Sebobo:
**What I did**

Speed up response time for return nodes for the document tree by a factor of up to 5x when uncollapsing a document or searching. With ~200 documents (e.g. Blog folder with many articles) the following results were achieved:

* Uncollapse blog page with ~200 articles: 
  * Response size is reduced from 162 KB to 132 KB -> -19%
  * Response time is reduced from 2.5s to 0.6s -> -75%
* Search in the document tree for the same blog posts (also ~200 results)
  * Response size is reduced from 193 KB to 163 KB -> -16%
  * Response time is reduced from 2s to 0.5s -> -75%

**How I did it**

* Skip content child nodes if we only request document types depending on the clients requested usage
* Don't recalculate nodetype filter for each rendered node

**How to verify it**

* Tests still work
* Node trees behave the same way except that uncollapsing and search are much faster

Resolves: #1508

### DIFF
--- a/Classes/Controller/BackendServiceController.php
+++ b/Classes/Controller/BackendServiceController.php
@@ -575,7 +575,8 @@ class BackendServiceController extends ActionController
                 $result = $nodeInfoHelper->renderNodes(
                     array_filter($flowQuery->get()),
                     $this->getControllerContext(),
-                    true
+                    true,
+                    ($finisher['payload'] ?? 'ALL') !== 'PAGE_TREE',
                 );
                 break;
             case 'getForTreeWithParents':
@@ -583,7 +584,8 @@ class BackendServiceController extends ActionController
                 $result = $nodeInfoHelper->renderNodesWithParents(
                     array_filter($flowQuery->get()),
                     $this->getControllerContext(),
-                    $nodeTypeFilter
+                    $nodeTypeFilter,
+                    ($finisher['payload']['usage'] ?? 'ALL') !== 'PAGE_TREE',
                 );
                 break;
         }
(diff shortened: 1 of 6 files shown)


Sebobo:
The follow up for this topic would be to get rid of all children in the response and only provide a parent id for building the tree and a `hasChildNodes` flag for the collapsible state check.

Sebobo:
Another thing to check: we use the similar functions to fill `Neos.Neos.Ui.initialState.cr.nodes.byContextPath` when the preview is rendered and I think those nodes are not even used.

pKallert (approved):
Tested it locally and works for me!

Sebobo:
Another issue I detected when analysing the trace, not directly related to this fix, but the slow tree:

Preview uri generation takes a long time (up to 30-40% of the request) when those uris are not in the cache
-> IMO those Uris could be generated much easier in PHP or JS as they only combine the preview base uri with the **(user specific)** node context as argument

ahaeslich (approved):

PR #3966, opened 2025-07-18
BUGFIX: add CKEditor Heading Plugin for p, pre, blockquote

### LINKED ISSUE #3964: BUG: CKEditor doesnt load Heading Plugin for p, blockquote, pre but uses its command

- [x]  have searched the existing issues

### Current Behavior

Configs that do not allow h-tags will not load the Heading-Plugin.
Therefore `<p>, <pre>, <blockquote>` cannot be used in the CKEditor because they rely on the Heading-Plugin.

### Expected Behavior

Configs that do not allow h-tags can still use `<p>, <pre>, <blockquote>`

### Steps To Reproduce

1. Use formattingOptions in a yaml config file, e.g.
```yaml
p: true
blockquote: true
h1: false
h2: false
h3: false
h4: false
h5: false
``` 
2. Change the Block in inline editing:

Image

3. Nothing happens, errors in console

### Environment

- 8.3
- 8.4
- 9.0

laurahaenel:
This is the MR for 9.0

See #3965
Resolves #3964

### DIFF
--- a/packages/neos-ui-ckeditor5-bindings/src/manifest.config.js
+++ b/packages/neos-ui-ckeditor5-bindings/src/manifest.config.js
@@ -133,7 +133,10 @@ export default ckEditorRegistry => {
         || editorOptions?.formatting?.justify
     )));
     config.set('heading', addPlugin(Heading, editorOptions => (
-        editorOptions?.formatting?.h1
+        editorOptions?.formatting?.p
+        || editorOptions?.formatting?.pre
+        || editorOptions?.formatting?.blockquote
+        || editorOptions?.formatting?.h1
         || editorOptions?.formatting?.h2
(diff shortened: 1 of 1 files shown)


Sebobo:
Thx! 

@markusguenther I don't think we need this PR, as I merged #3964  and we still upmerge fixes from 8.3 right?

laurahaenel:
> Thx!
> 
> @markusguenther I don't think we need this PR, as I merged #3964 and we still upmerge fixes from 8.3 right?

I created a new MR for 9.0 because in 9.0 -> Plow.js is removed, dont know how you would have prefered :D

Sebobo:
@laurahaenel ah thank you, of course :)

Sebobo (approved):

PR #3967, opened 2025-07-18
BUGFIX: Adjust dimension space point handling without a node address in backend controller

nezaniel:
**What I did**

The backend controller now respects Neos' default dimension space point if configured.
It also now prevents using an arbitrary root dimension space point that is not even covered by the requesting site, which previously lead to an exception that could hardly be resolved.

**How I did it**

When no node address is sent to the backend controller, e.g. directly after login, this now checks the requested site node aggregate whether 

1. it covers the default dimension space point configured in Neos; If yes, that is used
2. it covers any of the root dimension space points; If yes, an arbitrary one of those is used
3. it covers any dimension space points at all; If yes, an arbitrary one of those is used

**How to verify it**

* configure a default DSP and verify that one is used
* configure no default DSP and verify that the backend still loads

### DIFF
--- a/Classes/Controller/BackendController.php
+++ b/Classes/Controller/BackendController.php
@@ -12,6 +12,7 @@ namespace Neos\Neos\Ui\Controller;
  * source code.
  */
 
+use Neos\ContentRepository\Core\DimensionSpace\DimensionSpacePointSet;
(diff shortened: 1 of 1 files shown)


mhsdesign (approved):
Thanks the strategy is now really sophisticated and should never fail :D
In a plain neos its unlikely that the site node does not cover all dimensions - let alone not the default dsp, let alone no root generalisation, but its nice to be guarded against that.

Issue #3963, opened 2025-07-16
BUG: SelectBox with DataSource that returns entities shows "invalid value" warning


Benjamin-K:
# Current behaviour

When using a property with a selectbox connected to a DataSource, where the DataSource does not return a `string`-value, the UI shows a warning `invalid value: ...`. This is, because in [SelectBoxHelpers.ts Line 33](https://github.com/neos/neos-ui/blob/242e15b3df46326330dead6e86189fa9d256edad/packages/neos-ui-editors/src/Editors/SelectBox/selectBoxHelpers.ts#L33) the value is used as key of the objects. When using objects (sth. like `{ __value: 'persistence-identifier', __type: 'Some\Package\Model' }`) as described [here](https://discuss.neos.io/t/guide-how-to-select-entities-in-the-inspector-neos-2-3-lts/1664/18), the validation fails.

# Expected behaviour

The selectbox editor also accepts values that are not a string value to be able to select complexer values.

# Steps to reproduce

Follow the steps described in the (old) [post in the Neos Forum](https://discuss.neos.io/t/guide-how-to-select-entities-in-the-inspector-neos-2-3-lts/1664/18) and try to select a value.

# Environment

- Flow: 8.3
- Neos: 8.3
- Neos UI: 8.3.14

Benjamin-K:
As a workaround, one can create a new DataType for the required model and only return the persistent identifier of the entity. But if the data source will return maybe return different types, this will not be an option.

PR #3968, opened 2025-07-21
BUGFIX: Allow selecting Flow entites in selectboxeditor without value error

Sebobo:
Resolves: #3963

**How I did it**

Use the entities identity for the value existence check in the preprocessor the same way
as the DataSourceBasedSelectBoxEditor does internally to map the value to an entity.
This is necessary as the value cannot be used as array key.

**How to verify it**

Use example code from
https://gist.github.com/Benjamin-K/cae7debba7349c4e671bec253ee7f00d

### DIFF
--- a/packages/neos-ui-editors/src/Editors/SelectBox/selectBoxHelpers.ts
+++ b/packages/neos-ui-editors/src/Editors/SelectBox/selectBoxHelpers.ts
@@ -1,10 +1,39 @@
 import {I18nRegistry} from '@neos-project/neos-ts-interfaces';
 import {isNil} from '@neos-project/utils-helpers';
+import {
+    createSelectBoxValueStringFromPossiblyStrangeNodePropertyValue
+} from './createSelectBoxValueStringFromPossiblyStrangeNodePropertyValue';
 
-type RawSelectBoxOptions = {value: string, icon?: string; disabled?: boolean; label: string; group?: string;}[]|{[key: string]: {icon?: string; disabled?: boolean; label: string;
-        group?: string;}};
+type SerializedEntity = {
+    __identity: string;
+    [key: string]: string;
+}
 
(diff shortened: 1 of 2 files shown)


packages/neos-ui-editors/src/Editors/SelectBox/selectBoxHelpers.ts
```diff
@@ -1,10 +1,39 @@
...
+            group?: string;
+        } | null | Record<string, any>;
+    };
+
+type SelectBoxOption = {
+    value: string,
+    icon ? : string;
+    disabled ? : boolean;
+    label: string;
+    group ? : string;
```
mhsdesign:
```suggestion
    icon?: string;
    disabled?: boolean;
    label: string;
    group?: string;
```

mhsdesign (commented):
i thought we would be using the helper `createSelectBoxValueStringFromPossiblyStrangeNodePropertyValue` in the `DataSourceBasedSelectBoxEditor.js`? but i assume my change with the invalid label introduces a new layer beforehand where we expect the value to be preprocessed already.

This woul mean that in `DataSourceBasedSelectBoxEditor.js` we can get rid of the usages to the hacky function then?

Sebobo:
> i thought we would be using the helper `createSelectBoxValueStringFromPossiblyStrangeNodePropertyValue` in the `DataSourceBasedSelectBoxEditor.js`? but i assume my change with the invalid label introduces a new layer beforehand where we expect the value to be preprocessed already.
> 
> This woul mean that in `DataSourceBasedSelectBoxEditor.js` we can get rid of the usages to the hacky function then?

Phew. I hope so, but let's create an issue for that and fix this in a separate PR.
I don't want another regression due to this change for now.

mhsdesign (approved):
thanks

PR #3973, opened 2025-07-30

kitsunet:
We introduced the redirect links for performance in older Neos versions to prevent resolving routes for Nodes which was quite expensive. The Neos 9 preview links are easy to resolve and we have no need for the redirect links for previews.
This helps keep preview links consistent and always use the previewAction.

### DIFF
--- a/Classes/Domain/Model/Feedback/Operations/UpdateNodePreviewUrl.php
+++ b/Classes/Domain/Model/Feedback/Operations/UpdateNodePreviewUrl.php
@@ -111,7 +111,7 @@ class UpdateNodePreviewUrl extends AbstractFeedback
             $contextPath = '';
         } else {
             $nodeInfoHelper = new NodeInfoHelper();
-            $newPreviewUrl = $nodeInfoHelper->createRedirectToNode($this->node, $controllerContext->getRequest());
+            $newPreviewUrl = $nodeInfoHelper->previewUri($this->node, $controllerContext->getRequest());
             $contextPath = NodeAddress::fromNode($this->node)->toJson();
         }
         return [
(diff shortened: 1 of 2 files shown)


kitsunet:
This is independent from the Neos change, but makes sense to keep previews consisent and getting rid of cruft. Without this some preview links go to the showAction an behave differently.

Sebobo (approved):

mhsdesign:
Let me conclude some things we discussed:

Performance: Originally preview routes for nodes would be heavily calculated by Flow using actual resource paths like `localhost/live/user-foo/...` this put a heavy burden on the cache. Later only query parameters where used via `appendExceedingArguments`. That still is expensive as the routing cache build an entry each. Now with Neos 9.0 and the new  NodeUriBuilder (https://github.com/neos/neos-development-collection/pull/4892) we only resolve the base preview uri and append the node query parameter via string manipulation. -> This is the final step to make preview uri building as cheap as possible and thus cheaper than building the `redirectToAction` uri.

But aside from performance there is another part to consider. The `redirectToAction` did not open the node in the user workspace as it was provided to the action but it redirects to the base workspace. So an immediate user workspace on live would make the preview button show the live version of the page - which has been the behaviour since ages but is conceptional broken. (Also if the node was disabled it would show the preview uri for live pages because routing would not work) - the old broken behaviour was just reinstated for the Neos 9 branch via https://github.com/neos/neos-ui/pull/3867 but now removed for good. My followup pr https://github.com/neos/neos-ui/pull/3990 cleans up fully the `redirectToAction`

In the future we hope to be able to fully separate preview vs editing in the actions and thus be able to provide the correct `renderingMode` when opening the preview.

paavo:
Discussion about this topic here:
https://neos-project.slack.com/archives/C050C8FEK/p1762254351004399?thread_ts=1761898855.144669&cid=C050C8FEK

PR #3974, opened 2025-07-30

kitsunet:
We introduced the redirect links for performance in older Neos versions to prevent resolving routes for Nodes which was quite expensive. The Neos 9 preview links are easy to resolve and we have no need for the redirect links for previews. This helps keep preview links consistent and always use the previewAction.

These links should always be preview Urls so performance should be fine.

To restore the 8.3 behaviour use

```yaml
Neos.Neos.Ui.nextVersionPreviewBehavior: false
```

### DIFF
--- a/Classes/Controller/BackendServiceController.php
+++ b/Classes/Controller/BackendServiceController.php
@@ -255,7 +255,7 @@ class BackendServiceController extends ActionController
                 $this->publishingService->publishNode($node, $targetWorkspace);
 
                 if ($node->getNodeType()->isAggregate()) {
-                    $updateNodePreviewUrl = new UpdateNodePreviewUrl();
+                    $updateNodePreviewUrl = new UpdateNodePreviewUrl($this->settings['nextVersionPreviewBehavior'] ?? false);
                     $updateNodePreviewUrl->setNode($node);
                     $this->feedbackCollection->add($updateNodePreviewUrl);
                 }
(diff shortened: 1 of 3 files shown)


kitsunet:
Sister change to the 9.0: https://github.com/neos/neos-ui/pull/3973

Sebobo (approved):
Thx for the change, works fine. But shouldn't we target 8.3 instead? 

Because I noticed last week, that the preview uri generation took quite a bit of time when traversing the page tree, as those uris are all the same but their argument user workspace specific due to the context path. For 100 subpages this took > 20% of request time on the initial request. But in a larger system with regular deployments this can be considered a regular occurrence.

In 9.0 we already have an optimisation by @mhsdesign I would like to back port which uses the standard preview uri and just replaces the node argument for each link. Though I think this could be further improved by not going through the resolver for the same route over and over. It would be nice to get all of this into 8.3 still as the tree performance is not good.

So if I make this adjustments, I would change the code changed here due to adding the `previewUri` to the helper.  Or you add the helper in this change, but without the optimisations I mentioned, so that both versions use the same method. Wdyt?

kitsunet:
I thought UI! 8.4 was intended to work with Neos 8.3 just fine? but fine by me

Sebobo:
Will it? Not sure :D but the lowest maintained version for a bugfix is still 8.3

mhsdesign (changes requested):
For the 8.x branch we discussed to keep the current behaviour to use the base workspace by default - a feature flag to align to the now changed 9.0 behaviour might be the best option.

kitsunet:
we now have a feature flag, would be great to get a second review, but I would want to merge early next week. Thanks

mhsdesign:
As @Sebobo and me also introduced feature flags for other behaviours too i changed my opinion and think that it would do good to introduce with the 8.4 release a soft change to set `nextVersionPreviewBehavior: true` by default.

8.4 introduces already a few things by default which can be reverted by setting a flag. I think this would ensure a better migration to 9.0

-> https://github.com/neos/neos-ui/pull/4005

PR #3978, opened 2025-08-06
FEATURE: Show drop targets when dragging text in the inline editor

### LINKED ISSUE #3977: BUG: Drag & drop indicators are not shown in inline editor
CKEditor supports showing indicators when dragging and dropping selections, but the Neos UI is missing the styles for them, even though the features is installed and enabled.

Can be seen when playing with the editor on
https://ckeditor.com/ckeditor-5/demo/feature-rich/

Sebobo:
**What I did**

Show drag & drop indicators in inline editor. Resolves: #3977 

**How I did it**

* Use a custom body collection class that works inside the iframe and reattaches on location changes or reloads
* Added the missing CSS for the indicators from CKEditor to the Neos styles.

**How to verify it**

https://github.com/user-attachments/assets/d576239e-91d9-4a8c-8122-f57fb4f6083f


Sebobo:
TODO: I think we should move the CK BodyCollection into our Neos backend container, so we have everything inside one element and could use layer styles and maybe a web component for these elements in the future.

Sebobo:
Part of #4016

PR #3979, opened 2025-08-07
FEATURE: Drag & drop nodes in content area

### LINKED ISSUE #1147: Drag-n-Drop in content canvas
The simplest implementation could be just to add move handle to the inline toolbar and drop targets to before and after elements in CCs.
Should work exactly as it does in the tree. The smooth scrolling would be done with the plugin that we use in the tree.
We can even make draging work from the tree into inline and vice versa.

@aertmann had an idea back in the ages to show a map of zoomed-out view of all elements on the page on drag. But this would be ofc way harder to implement, and I'm not even sure it would be beneficial, since with a good smooth scroll in place it should be no problem to move elements all over the page.

UX tasks: design the look of the drag handle and of drop targets.

Sebobo:
Resolves: #1147 

**What I did**

An editor can now drag & drop a node with a new button in the node toolbar.
While dragging the node, the UI shows the allowed drop areas on hover and depending on the cursor position over the hovered element. E.g. drag into collection, drag before / after existing element.

**How I did it**

I added a new button to all nodes with the dragging event to start a drag. A global drag component has events for the drag over, drop and drag end events and shows the indicator while dragging and triggers the move event on drop.

**How to verify it**

https://github.com/user-attachments/assets/aaab43b3-bf5f-4c74-9b10-e51e95455d38


Sebobo:
Will be merged as part of #4016

PR #3980, opened 2025-08-12
BUGFIX: Speed up preview uri generation

Sebobo:
**What I did**

Generate a single preview uri for each required content context and site and adjust its argument for each requested node.

This way we only need to resolve / build a single uri instead of one for each loaded page (and workspace) which can get quite slow.

**How I did it**

Backport of the `previewUri` and its usage from Neos 9. 

**How to verify it**

Page tree should still work and selecting a page loads the correct document in the guest frame.
The xhr requests to load the tree should be around 60-80% faster.

### DIFF
--- a/Classes/Fusion/Helper/NodeInfoHelper.php
+++ b/Classes/Fusion/Helper/NodeInfoHelper.php
@@ -11,13 +11,14 @@ namespace Neos\Neos\Ui\Fusion\Helper;
  * source code.
  */
 
+use GuzzleHttp\Psr7\Uri;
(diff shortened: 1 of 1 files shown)


Sebobo:
Note: this is based in #3961 and needs to be rebased after merging

Sebobo:
This change will benefit even more from #3974

ahaeslich (approved):

mhsdesign (approved):
Thanks for porting the 9.0 optimisation

PR #3981, opened 2025-08-12
BUGFIX: Forward defined insert mode in creation workflow when nodetype is set

Sebobo:
When triggering the node creation workflow and providing a nodetype, the mode ends up being not set due to a typo and the resulting action can lead to an invalid creation action as the default mode is `into` and one might not be allowed to create the same node into itself.
With this patch, the mode is used as provided.


ahaeslich (approved):
by 👀

PR #3982, opened 2025-08-13
BUGFIX: Close creation dialog if back is pressed and a nodetype was preselected

Sebobo:
This is a follow up of #3981.
When the node creation workflow is manually triggered and a nodetype was given, the back button doesn't do anything as the workflow goes back into the node selection and then immediately proceeds to the configuration step again as the nodetype is already set.
With this change, "back" closes the dialog for this case.


mhsdesign (approved):
Okay it would be prettier to name the `back` button cancel instead for this case but it seems like a good step forward already

Issue #3983, opened 2025-08-14
BUG: Copying nodes in the document tree renders recursively all child nodes into response


Sebobo:
Copying nodes especially in the page tree can get very slow. We assumed that this happens due to the many operations on the CR, but during some profile runs of copying 6 pages in a customer project I found out that up to 50% of the time is spent on rendering the response which includes a `UpdateNodeInfo` for each page and for the parent page. 
Each updated node info contains recursively all child nodes which causes lots of database queries and serialisation processes even though the information is not required.
This happens as the `AbstractCopy` class enables the recursive generation in its `finish` method.

The page tree only needs the direct child node context paths, which are also included without the recursive flag.
In addition the currently transmitted information about the content child nodes is not even used as navigating to a page anyway reloads the content nodes.

The content tree COULD use the information about nested child nodes when a content element is copied, but some time ago I refactored the tree to lazy load any missing information. Therefore disabled the recursive feature for both trees doesn't degrade the UX in any meaningful way, and the performance is much better.

PR #3984, opened 2025-08-14
BUGFIX: Speed up node tree copy actions

Sebobo:
Don’t include recursive childnode info in copy action feedback which we don’t need for the node trees to work.

Both trees still work as before. Only difference is that uncollapsing the copied nodes might lazy load deeper nested nodes.

Resolves: #3983

### Testcase:

Copying 4 documents with some content in the Neos.Demo

CleanShot 2025-08-14 at 15 46 17@2x

### Response times

With recursive data: 4.4KB
Without recursive data: 2.5 KB

CleanShot 2025-08-14 at 15 38 45@2x

With more nested data:

CleanShot 2025-08-14 at 15 51 46@2x

### Calls with recursive data

CleanShot 2025-08-14 at 15 47 52@2x

### Calls without recursive data

CleanShot 2025-08-14 at 15 48 26@2x

### DIFF
--- a/Classes/Domain/Model/Changes/AbstractCopy.php
+++ b/Classes/Domain/Model/Changes/AbstractCopy.php
@@ -34,7 +34,6 @@ abstract class AbstractCopy extends AbstractStructuralChange
     {
         $updateNodeInfo = new UpdateNodeInfo();
         $updateNodeInfo->setNode($node);
-        $updateNodeInfo->recursive();
         $this->feedbackCollection->add($updateNodeInfo);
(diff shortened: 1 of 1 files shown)


kdambekalns (approved):
Looks mostly harmless… 😎

mhsdesign (approved):
Thanks nice trick. In Neos 9 or later we can now also remove the whole `\Neos\Neos\Ui\Domain\Model\Feedback\Operations\UpdateNodeInfo::recursive` handing and the internal logic ... but that is for later.

PR #3987, opened 2025-08-18
BUGFIX: Adjust hidden in menu state to new CR property

### LINKED ISSUE #3976: Erroneous "hidden in menu" state
When setting the "hidden in menu" flag, the node will not be greyed in the document tree unless I select another document and reload.

Are there still remnants of "_hiddenInIndex" around?

Neos version: 9.0.4
Neos.UI version: 9.0.1

Sebobo:
Resolves: #3976


mhsdesign (approved):
lol

Issue #3988, opened 2025-08-18
BUG: Hidden state special handling doesn't work with inspector triggered change


Sebobo:
Relates to https://github.com/neos/neos-development-collection/issues/5615 and https://github.com/neos/neos-development-collection/pull/5602

PR #3989, opened 2025-08-18
BUGFIX: Update inline hidden state when hidden property is changed via inspector

Sebobo:
This change handles the case that the hidden state is updated in the inspector which didn't update the guest frame.
Also the case is handled that reloadIfChanged is enabled on the hidden property but the ui didn't send a nodeDomAdress and the backend would throw an exception.

Resolves: #3988

### DIFF
--- a/Classes/Domain/Model/Changes/Property.php
+++ b/Classes/Domain/Model/Changes/Property.php
@@ -117,7 +117,7 @@ class Property extends AbstractChange
     /**
      * Get the node dom address
      *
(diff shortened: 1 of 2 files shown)


dlubitz (approved):
Looks good to me! Thank you for taking care.

mhsdesign (approved):
Thanks for the fix. ... also not requiring the dom address for the property might make kinda sense ... but we can always remove that later ^^. `properties._hidden.ui.reloadIfChanged` definetely is not a usecase

PR #3990, opened 2025-08-19
TASK: Followup #3973 (Use preview uris instead of base workspace preview)

mhsdesign:
With https://github.com/neos/neos-ui/pull/3973 we no longer need `createRedirectToNode` and its routing configuration and action controller. This dead internal code has been removed.

**What I did**

**How I did it**

**How to verify it**


Sebobo (approved):
Thanks, looks good by reading.

PR #3991, opened 2025-08-20
BUGFIX: Replace non existing method call in renderNode

Sebobo:
This affects a deprecated method and is a regression due introduced with #3961


mhsdesign (approved):

mhsdesign:
Yup that code is gone either way with Neos 9.0 ^^

PR #3992, opened 2025-08-20
BUGFIX: Speed up node tree copy actions 9.0 port of #3984

mhsdesign:
See https://github.com/neos/neos-ui/pull/3984

In Neos 9 the `$updateNodeInfo->recursive()` was added to `AbstractStructuralChange` and the `AbstractCopy` was removed.

This change now removes the expensive recursive of `UpdateNodeInfo` as we dont need that for the Ui anymore.

9.0 Benchmarks

9 document nodes
Before: 191kb 1,72s
After: 47kb 1,06s

about 6 times 9 document nodes (nested)
Before: 1.18mb  11,47s
After: 294,41kb  6,03s

**What I did**

**How I did it**

**How to verify it**


Sebobo (approved):
Thx

PR #3994, opened 2025-09-05
BUGFIX: Nested content not editable after copy

Sebobo:
This solves a regression introduced with ef118b19d5de7092cd8f3d2f23f55db77bdb98e4 which prevented editing nested copied content after insertion. 

With this change the out-of-band code rendering takes care to include the nested child node data.

### DIFF
--- a/Classes/Domain/Model/Changes/AbstractStructuralChange.php
+++ b/Classes/Domain/Model/Changes/AbstractStructuralChange.php
@@ -178,6 +178,14 @@ abstract class AbstractStructuralChange extends AbstractChange
                 $this->getParentDomAddress()->getFusionPath() &&
                 $this->getParentDomAddress()->getContextPath() === $node->getParent()->getContextPath()
             ) {
+                // Include nested nodedata for with the rendered content as the change usually only contains the updated node
+                if ($node->hasChildNodes()) {
+                    $updateNestedNodeInfo = new UpdateNodeInfo();
+                    $updateNestedNodeInfo->setNode($node);
+                    $updateNestedNodeInfo->recursive();
+                    $this->feedbackCollection->add($updateNestedNodeInfo);
+                }
+
                 $renderContentOutOfBand = new RenderContentOutOfBand();
                 $renderContentOutOfBand->setNode($node);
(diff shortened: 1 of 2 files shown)


dlubitz (approved):

dlubitz:
This works for me! Thank you for this quick fix!
I wonder if we shouldn't cover this scenario with a test?

Classes/Domain/Model/Changes/AbstractStructuralChange.php
```diff
@@ -178,6 +178,14 @@ protected function finish(NodeInterface $node)
                 $this->getParentDomAddress()->getFusionPath() &&
                 $this->getParentDomAddress()->getContextPath() === $node->getParent()->getContextPath()
             ) {
+                // Include nested nodedata for with the rendered content as the change usually only contains the updated node
+                if ($node->hasChildNodes()) {
```
mhsdesign, line 182: why dont we modify `$updateParentNodeInfo->recursive()` here?

Sebobo: That was also one of my first ideas, but the parent is the outer collection with even more nodes not related to what we are changing.

mhsdesign: humm right ... i was just wondering you wrote

> One drawback now is that the copied node itself is contained twice in the response for this special case. But that’s also an issue of our feedback handler system.

where can i see that this is the case?

Sebobo: The AbstractCopy adds another NodeInfo

dlubitz:
Can we somehow get this merged and released soonish?

jonnitto (approved):

mhsdesign:
And yes more importantly we can add a test simply. If was able to reproduce that in the 9.0 branch with `Container_Test` node and a inline editable text node inside ... ill open a Neos 9.0 fix pr of this with a test attached.

mhsdesign:
9.0 pr https://github.com/neos/neos-ui/pull/4000 okay well we had NO copy and paste tests ... thats the first ... and ckeditor is nasty to test ... :)

PR #3999, opened 2025-09-17

kitsunet:
erronously merged into 8.4, but should actually go in here... See https://github.com/neos/neos-ui/pull/3974


kitsunet:
It's unclear to me how the e2e failure should be related to this, especially with the feature flag false

kitsunet:
well, that was broken, I guess this will fix e2e, a bit scary that this is only indirectly caught by some e2e test.

Issue #3972, opened 2025-07-30
Node Actions Toolbars directly around node to reduce travel time


Sebobo:
* Reduce travel between edited content and inlineUI Toolbar, f.e. to add the next image/text/etc element
* Reduce travel between edited text and formatting
  * Move formatting toolbar (CKEditor) closer to the edited content
  * Take inspiration from https://wordpress.org/gutenberg/
 
<img width="1584" height="1520" alt="Image" src="https://github.com/user-attachments/assets/4ae143f1-3053-47ab-aedc-25c7b3d78746" />
<img width="2426" height="434" alt="Image" src="https://github.com/user-attachments/assets/16a0f7ae-31c2-4897-8044-621bdc17e3e5" />

Sebobo:
Related bugs:
* https://github.com/neos/neos-ui/issues/2691
* https://github.com/neos/neos-ui/issues/2692

Related discussions:
* https://discuss.neos.io/t/neos-editing-ux-inspiration/2453

Inspiration:
* https://wordpress.org/gutenberg/

PR #4001, opened 2025-09-25
BUGFIX: Node toolbar position based on actual size

### LINKED ISSUE #2692: InlineToolbar position wrong if content width is very small
### Description

When an inline editable node is selected in the guest frame, the InlineToolbar jumps to the far left, if the selected element has a small width:

Screenshot 2020-03-26 at 10 40 57

### Steps to Reproduce

- Create an inline editable object (for example a small icon in the footer or a logo in a logowall) with a very small css-styled width of 30px
- select it in the neos backend -> the toolbar is moved to the far left
- Change the width of the element to some larger value, lets say 200px
- select it in the neos backend -> the toolbar is positioned correctly

### Further observations

I first thought my websites styling was messing with the css of the inline UI (which had happend before when I set a style on the `body` element). But when inspecting the toolbar, i saw that there is an inline-style set on one of the containers: `left: 0;`.

Screenshot 2020-03-26 at 10 46 08

I'm suspecting a bug in the javascript code that calculates this `left` value.

### Affected Versions

Reproduced with:
- Neos: 4.3.11; UI: 4.0.5
- Neos: 5.1.15; UI: 5.1.14

Sebobo:
**What I did**

Resolved the issue that the toolbar is incorrectly positioned when the assigned element is too small or the toolbar has more items than the default.

Resolves: #2692 

**How I did it**

Via a ResizeObserver the position is updated when the toolbar switches between elements or changes its size due to its content or window limits.

**How to verify it**

Tested with latest Chrome, Safari, Firefox

NodeToolbarPositionFixed

### DIFF
--- a/packages/neos-ui-guest-frame/src/InlineUI/NodeToolbar/index.js
+++ b/packages/neos-ui-guest-frame/src/InlineUI/NodeToolbar/index.js
@@ -39,10 +39,17 @@ export default class NodeToolbar extends PureComponent {
     };
 
     state = {
-        isSticky: false
+        isSticky: false,
+        toolbarPosition: {}
     };
 
     iframeWindow = getGuestFrameWindow();
+
+    prevNodeToolbarRef = null;
+
+    resizeObserver = null;
+
+    nodeToolbarRef = React.createRef();
 
     updateStickyness = () => {
         const nodeElement = findNodeInGuestFrame(this.props.contextPath, this.props.fusionPath);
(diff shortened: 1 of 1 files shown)


Sebobo:
I wasn't able to fix this with fully with CSS anchor positioning as I only got it working with latest Chrome and Safari. Firefox didn't like it, and the polyfill (https://github.com/oddbird/css-anchor-positioning)  is not easy to apply, as we need to apply it whenever the toolbar is rerendered inside the guest frame. But the polyfill will still be needed for #3972.

pKallert:
Had a look at the code and tested it out locally. 
What I noticed when you resize the window, and the Element was already selected, the position is not reevaluated: 
Bildschirmaufnahme 2025-10-31 um 08 03 13(2)

Sebobo:
Hi @pKallert, thx for finding the issue.
I changed it so that the resize observer also checks for the size of the node element. This solved the issue for me. Can you verify this?

pKallert:
Had another look and noticed that on wide screens (or if you zoom out using browser zoom) the toolbar moves to the right side if you use drag to make the window wider: 
Bildschirmfoto 2025-11-20 um 14 56 24

Sebobo:
@pKallert thx! 
I added the guest frame to be also observed. Now it works IMO.

pKallert (approved):
Tested it out and works perfectly now. Thank you for taking care!

mhsdesign:
The introduction of the resize observer here leads to an odd failure in the e2e tests of "`Discarding: create a content node and then discard it`"

```
JavaScript error details:
ResizeObserver loop completed with undelivered notifications.
    No stack trace available

Tests/IntegrationTests/Fixtures/1Dimension/discarding.e2e.js
 > 129 |        .switchToMainWindow();
```

I thought this is a timing problem but adding a one second wait before the end of the test or after the creation of the node and its removal does not fix the test.
Also i could not reproduce the error when manually carrying out the steps.

PR #4002, opened 2025-09-26
TASK: Introduce PSR12 linting via phpcs and autofix files

mhsdesign:
**What I did**

**How I did it**

**How to verify it**


Sebobo:
StyleCI has 2 complaints. Are those from the upmerge or did phpcs ignore those?

mhsdesign:
no this fix would also replace StyleCi and its riddles ;) Ill see if the reasons are legit...

Okay style CI likes no space after an anonymous class `return new class() {}` but phpcs wants it `return new class () {}` so they fight each other ... 

befor continuing with this well need to see what all our code in Neos.Neos and co says as we only use phpcs partly ... but the style CI revolution should be started there as well..

PR #4003, opened 2025-09-29
BUGFIX: Use correct color for timed visibility icon

### LINKED ISSUE #3997: BUG: Tree shows wrong icon when enableAfterDateTime is used and set to a future date
Neos Version: 9.0.1

## Problem:

When i use the enableAfterDateTime prop, and set the date to a future date, i would expect that the Node is marked as hidden until the configurated date arrives. 

Now there is the blue clock icon for activated and not activated pages. I would expect blue for activated and red for not activated.

Image

## Reproduce the problem:

Add activate the prop for a document node and set a date in the future:

```
  properties:
    enableAfterDateTime:
      type: DateTime
      ui:
        label: i18n
        reloadIfChanged: true
        showInCreationDialog: true
        inspector:
          group: general
```

Sebobo:
**What I did**

This fixes a regression introduced with 43efd9c63709ecf87a029d2636b33b0a935d41f3

Resolves: #3997 

**How I did it**

Readded some of the necessary code from 8.3 and adjusted it to the new visibility property names.

**How to verify it**

CleanShot 2025-09-29 at 09 22 23@2x
CleanShot 2025-09-29 at 09 22 12@2x

### DIFF
--- a/packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js
+++ b/packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js
@@ -1,16 +1,15 @@
 import React, {PureComponent} from 'react';
 import PropTypes from 'prop-types';
 import {connect} from 'react-redux';
-
+import moment from 'moment';
+import hashSum from 'hash-sum';
 import flowright from 'lodash.flowright';
+
 import {Tree, Icon} from '@neos-project/react-ui-components';
 import {stripTags, decodeHtml} from '@neos-project/utils-helpers';
-
 import {actions, selectors} from '@neos-project/neos-ui-redux-store';
 import {isNodeCollapsed} from '@neos-project/neos-ui-redux-store/src/CR/Nodes/helpers';
 import {neos} from '@neos-project/neos-ui-decorators';
-
-import hashSum from 'hash-sum';
 import {urlWithParams} from '@neos-project/utils-helpers/src/urlWithParams';
 
 const getContextPath = node => node?.contextPath;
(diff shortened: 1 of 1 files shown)


packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js
```diff
@@ -172,7 +171,13 @@ export default class Node extends PureComponent {
         const hasTimeableNodeVisibility = node?.properties?._hasTimeableNodeVisibility;
 
         if (hasTimeableNodeVisibility) {
-            const circleColor = isDisabled ? 'error' : 'primaryBlue';
```
dlubitz, line 175: The internal logic of timable node visibility has changed. That's why this also has changed. The state is rather depending on the actual visibility state (disabled/enabled) than the "calculated" state based on the datetime.

https://github.com/neos/timeable-node-visibility

dlubitz (changes requested):

packages/neos-ui/src/Containers/LeftSideBar/NodeTree/Node/index.js
Sebobo: But that also means, the editor has to wait for the job to run so they can see the real state. 

What happens if they set hidden manually, the job would then change the value, correct?

dlubitz: Yes, correct. As I wrote in the issue:

> IIRC we discussed a feature for the UI to show a warning if the current state is not correct (e.g node is visible, but the enableAfter is in future etc). But I couldn't find any issue for this.

Sebobo: We could use this change for that, as it would be more or less the same code ?

dlubitz: I guess so. Would be a feature and we would need some idea how we want to display that.

dlubitz: But I think your solution does not cover all scenatios atm. Please check the tests in the `neos/timeable-node-visibility` package: https://github.com/neos/timeable-node-visibility/blob/9.0/Tests/Behavior/Features/HandleExceeded.feature

dlubitz: And the backend implementation to determine the state: https://github.com/neos/timeable-node-visibility/blob/9.0/Classes/Service/TimeableNodeVisibilityService.php#L129-L154

PR #4007, opened 2025-10-11
BUGFIX: Position Dialog Modals centered and adjust width and height

mhsdesign:
The dialogs always leave a spacing of 20vh from the top which wastes spaces and looks unbalanced on laptops as below the dialog ends often without gap.

This changes centers the dialogs now on the vertical axis so they are always in center no matter their height. (see small vs. big creation dialog)

Additionally, the default width of "wide" was a little reduced as select boxes and text-fields in the creation dialog grew super wide - much wider than one is used from them in the inspector. This size adjustment will also make the new link editor appear by default in the correct width.

The dialog additionally grows by default and in "wide" mode to fit its contents. This is required when the new link editor shows the asset selection which takes the full width.

For reference the original Archaeopterix employed a hack to archive this which is now merged into the core behaviour:

https://github.com/sitegeist/Sitegeist.Archaeopteryx/blob/2d53b63b483b608dcc062a23bf2af608f27d5445/Neos.Ui/core/src/presentation/Modal.tsx#L7-L34

## Visual comparison

| Before | After |
|--------|--------|
| image  | image |
| image | image |

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/packages/react-ui-components/src/Dialog/style.module.css
+++ b/packages/react-ui-components/src/Dialog/style.module.css
@@ -18,9 +18,9 @@
 .dialog__contentsPosition {
     composes: reset from '../reset.module.css';
     position: absolute;
-    top: 20vh;
+    top: 50%;
     left: 50%;
-    transform: translateX(-50%) scale(1);
+    transform: translateY(-50%) translateX(-50%) scale(1);
     background: var(--colors-ContrastDarker);
     box-shadow: 0 20px 40px rgba(0, 0, 0, .4);
     border-radius: 0;
@@ -31,17 +31,17 @@
 .dialog__contents {
     composes: reset from '../reset.module.css';
     position: relative;
-    width: calc(100vw - var(--spacing-GoldenUnit) * 2);
-    max-width: calc(var(--spacing-GoldenUnit) * 16);
+    width: auto;
+    max-width: calc(100vw - var(--spacing-GoldenUnit) * 2);
     border: 2px solid var(--colors-ContrastDark);
     transition: var(--transition-Default) ease max-width;
 
     .dialog--wide & {
-        max-width: calc(var(--spacing-GoldenUnit) * 24);
+        min-width: calc(var(--spacing-GoldenUnit) * 18);
 
         @media(max-width: 576px) {
             max-width: 100vw;
-            width: 100vw;
+            min-width: 100vw;
         }
     }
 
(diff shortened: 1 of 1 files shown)


Sebobo:
I think if we already change this as a "Feature" for 9.1 we should use the popover html feature which takes care of all the positioning, z-index and some of the interaction.

mhsdesign:
Hmm i looked into that but firefox doesnt support the hint value yet which we would need to make the dialog stay open or be dismiss-able when clicking outside.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/popover 

I can label this as `TASK` or `BUGFIX` if you prefer;) But i didnt want to hide this commit in the big link editor pr.

Sebobo:
Why can't we use the "auto" value?

jonnitto:
I think, the `<dialog>` element would be even the better element for this. And as a bonus feature compared to a react element, you don't need to port it to the body, you can right use it anywhere in the markup.

mhsdesign:
Hahah i see i got myself into something here:) Im currently head deep in the Ui and attempting to greatly reduce legacy code or make things better structured ... but the whole dialog to `<portal` refactoring is not my battle to fight - not that i dont think its worth it, but it seems our current `Modal` abstraction might not really survive that and thus there are plugins lives on the line etc etc :) If you have a plan how to approach that please do! -> Just push on this pr! Ill also help support:)

Sebobo:
All good. Turn this into a bugfix for 9.0 and we separate the reimplementation.

Sebobo (changes requested):
The animation is now different when switching between creation dialog steps for example. Before the steps transitioned between each other, now the steps are flying from the bottom which is distracting.

Also the dialog is much smaller by default than before which makes it higher. I'm not exactly sure why yet.

mhsdesign:
Fixed the transitions with jon and also ensured that it works on smaller screens:

image


the dialog is not higher than before in absoulte pixels though the aspect ration changed because its smaller so it might seems higher :)

Any further blockers or should we merge that into 9.0?

Sebobo (approved):
Thx, works great.

mhsdesign:
TODO:

- check if the dialog shouldnt be allowed to go higher than 80vh - on small screens important where place is precious -> maybe but for the creation dialog like in https://github.com/neos/neos-ui/pull/4007#issuecomment-3491248983 the place is already well filled and there needs to be a gap atop
- check if the archeopterix (plugin which uses this dialog) needs adjustments and if so release patch version of this restricting to Neos Ui ^= 9.0.x -> as we the archaeopteryx overrides the styles but we merged those herby in the code as the css is now just doubled there are no negative visual side effects as far as i tested with version 2.0 and this modal styling fix

PR #4008, opened 2025-10-13
TASK: use GitHub ci and run behaviour tests

mhsdesign:
refs https://github.com/neos/neos-ui/issues/4026

Streamlines our CI setup as we use github ci also in Flow and in Neos.

Additionally FunctionalTests are now executed and behat behaviour tests (though this pr adds just an empty one to make CI pass)

**What I did**

**How I did it**

**How to verify it**


Sebobo (changes requested):
IMO we should call those two parts "Backend" and "Frontend". Client and Server is misleading.

I saw in the CI output of the unit tests that there are lots of console errors.
Some seem to be warnings, but some don't look "right".

mhsdesign:
> IMO we should call those two parts "Backend" and "Frontend". Client and Server is misleading.

i thought about that but tbh Backend is even more confusing as in Neos the backend is `/neos` and thus the php server backend but also the js frontend -> in the end we have a client -> server relation and as they are separated and require different techstacks they are tested separately first.

> I saw in the CI output of the unit tests that there are lots of console errors.

i take it you mean the javascript unit tests? This is not a new problem they also appear when you run `yarn test` locally.
@markusguenther attempted here https://github.com/neos/neos-ui/pull/3464 to fix the the console errors ... so thats another part but we should probably do it at some point 😅

Sebobo:
My problem is that here on Github it looks confusing:
CleanShot 2025-10-15 at 13 28 51@2x

Most people wouldn't understand what those 2 things mean and probably I will also think in a few months that the CI has a server and client. 
So I would like to have those two visible labels clearing saying what they are for.

skurfuerst:
lgtm, merging!

mhsdesign:
@skurfuerst i dont know why but the e2e testcafe tests still fail when run via github ... something is wrong with the resource loading (the sauce labs video shows that) Apparently the Neos.Demo plugin is loaded but not the NeosUi code itself ... i also validated this by my last commit where i debugged via curl what the NeosUi resources return.

image

-> i might have been a little adventurous to put all these ci refactoring in one pr and then tell bernhard to base his stuff on my experimental ^^

skurfuerst:
Hmmm ok 😁 probably we still should fix it forward? I do not have any experience with that unfortunately...

PR #4010, opened 2025-10-13
TASK: Introduce aliased `useSelector` hook for Neos global State

mhsdesign:
This can be used in functional components to avoid having to use the `connect()` decorator or hoc
    
While a direct import from "react-redux" would work, the `useSelector` would not be aware of its state type. And as the defaultState was removed https://github.com/reduxjs/react-redux/issues/1879 we resort to now cleanly reexporting it.

### example

exactly like the [original `useSelector`](https://react-redux.js.org/api/hooks#useselector), just with typescript support

```tsx
import {useSelector} from '@neos-project/neos-ui-redux-store';

export const Component = () => {
    const workspaceName = useSelector(selectors.CR.Workspaces.personalWorkspaceNameSelector);

    return <>Your workspace: {workspaceName}</>
}
```

Plugin authors can now also directly import `useSelector` from the neos-ui extensibility api and write functional component.

Some savvy developers might have found out that using in `import {useSelector} from 'react-redux';` in plugins works already as react-redux is part of the extensibility api but this can now be adjusted to use `'@neos-project/neos-ui-redux-store'`.

Support with exported types in plugin is not supported.

**What I did**

**How I did it**

**How to verify it**


Sebobo (approved):
Looks good by reading, thx!

PR #4012, opened 2025-10-14
TASK: Introduce global `getConfiguration()` and cleanup configuration provider

mhsdesign:
Depends on https://github.com/neos/neos-ui/pull/4011 and introduces e2e tests for it.

### 1.) Move `editPreviewModes` to actual neos ui "configuration" instead of "frontendConfiguration"

Now that the "configuration" is cleaned up an only contains real configuration - no state, no endpoints - the question is how does it differ to the "frontendConfiguration" and shouldnt the one be part of the other?

From what git tells me it was like this:

The "configuration" in the bootstrap existed but specifically to pipe the "editPreviewModes" to javascript the "Neos.Neos.Ui.frontendConfiguration" settings was created. This was sorely home to the "editPreviewModes" key and its options which were made available from the original location in Neos.Neos.userInterface. To access these in javascript a new "editPreviewModes" registry was created which was filled by the items in frontendConfiguration.editPreviewModes. -> change 2016 https://github.com/neos/neos-ui/pull/425

One year later it was discussed in slack how to make settings overall available from plugins to the ui (https://neos-project.slack.com/archives/C0U0KEGDQ/p1507709786000036)
The "frontendConfiguration" stuck out as option in the settings with the potential to easily do this. The change https://github.com/neos/neos-ui/commit/15afda40abd50c5e918165c8f1f8eea588436040 replaced the dedicated "editPreviewModes" registry with a "frontendConfiguration" where all settings from the "frontendConfiguration" were made available to.

Now today we end up with the configuration and frontendConfiguration where the former is not extensible. As we alias the editPreviewMode configuration in the "frontendConfiguration" and dont define it there and as this value is an integral part or neos it makes now more sense to put it into the "configuration" and thus get rid of the odd RenderingMode helper - in hindsight this is quiet ironic as the frontendConfiguration was initially made for the rendering modes and now we end up almost not using it anymore.

Future of the "frontendConfiguration"

The neos ui only uses this setting for the "hotkeys" but they are odd and broken so actually we dont rely on that much.

For plugins the frontendConfiguration is still the (only) place to add custom values.

Now an alternate future of the "frontendConfiguration" would be to host everything including the editPreviewModes and the nodeTree loading depth but these values are configured in "Neos.Neos.userInterface" and will probably stay there and should not be replaced by an alternative like "Neos.Neos.Ui.frontendConfiguration" (especially as the "editPreviewModes" are so deep integrated in Neos.Neos that they will need to stay configured there)

### 2.) Separate state from configuration -> move the `allowedTargetWorkspaces` into the redux store

previously these options where fully static and part of the "configuration".
But this is misleading, the information depends on other workspaces and thus the database. It is not configuration.

Now they will also be updated when `const {getWorkspaceInfo} = backend.get().endpoints;` is invoked.

During changes only the totalNumberOfChanges are recalculated but not all the allowedTargetWorkspaces

### 3.) Remove `endpoints` from initialData configuration by using `<link` tag and prefetch for node type schema endpoint

this approach was already started for translations via https://github.com/neos/neos-ui/pull/3804 and now the second endpoint is gone as well.

### 4.) E2E tests for extensibility API

Introduces automatic e2e test to ensure the newest adjustments to our extensibility api are tested before we even publish those to npm.

For that the e2e test distribution now ships a plugin which is build using the extensibility api from the project via yarn workspace.

To build them use

```
yarn workspace @neos-project/neos-ui-extensibility run build
yarn workspace @neos-project/neos-ui-test-plugin run build
```

Which builds the extensibility api and then for `Neos.TestNodeTypes` compiles an example dummy plugin.

Testcafe then boots the real Neos Ui with plugin and uses an exposed global variable `neosUiTestPlugin` to interact with the plugin. In a later step the plugin could also be used to test that certain container extension points work or that a custom ck plugin can be written.

```
 - Manifest boot
 - Create and access custom registry
 - Neos Ui Configuration Access
 - Frontend Configuration Access
 √ Extensibility Plugin Api - Check plugin status via developer api (console)
```

**What I did**

**How I did it**

**How to verify it**


mhsdesign:
@Sebobo following (internal) changes have been made

As we do not have a strict guideline as to how a plugin should behave to be supported across patch,minor or major versions its never easy to judge an internal change like this. Many parts of the Neos Ui carry a warning in code with them if someone extends those but thats just warnings. As most packages are exposed via the extensibility api I with my knowledge could build a plugin which could observe every detail what changed here but that is only because of my deep knowledge of the ui. None of these changes should matter to a Neos Ui plugin as they are core details:

- the endpoints `translations` and `nodeTypeSchema` no longer reside under `configuration.endpoints`
- ~the `frontendConfiguration` no longer carries the `editPreviewModes` setting~ and does not use the internal (new) helper `Neos.Ui.RenderingMode.findAllSorted()`
- ~redux sagas no longer receive the `{configuration}` as parameter~
- ~when using the `@neos()` decorator `neos.configuration` is no longer available~
- ~the state information `configuration.allowedTargetWorkspaces` is no longer stored as configuration but in the redux store~

Sebobo (approved):
Looks good, thx

packages/neos-ui/src/index.js
```diff
@@ -75,20 +88,10 @@ async function main() {
     reloadNodes();
 }
 
-function initializeFrontendConfiguration() {
```
Sebobo, line 78: Sadly this breaks 4 of my plugins, as I use `globalRegistry.get('frontendConfiguration')` via the `neos` decorator.

And I'm quite sure, that other plugins use this to.

Sebobo (changes requested):
Like this, it's breaking extensibility without supporting the previous ways of accessing the `frontendConfiguration`.

packages/neos-ui/src/index.js
Sebobo: Reason I did this probably: With the injection I don't need to "create" the plugin with the configuration and a wrapper from the bootstrap, but simply have it available. This was easier code wise.

Sebobo: ah sorry, I access the editPreviewModes and allowedTargetWorkspaces in the commandbar plugin. The frontendConfiguration still seems to work.

mhsdesign: hi yes `globalRegistry.get('frontendConfiguration')` should still continue to work.

The new location of editPreviewModes and allowedTargetWorkspaces is intentional one is now available via `getConfiguration()` and the other now resides in the redux store:

https://github.com/neos/neos-ui/pull/4012#issuecomment-3466687435

how did you access the neos configuration for `allowedTargetWorkspaces`?

I can brew a commit which makes the `editPreviewModes` and `allowedTargetWorkspaces` available as beforehand but at least with 9.1 id like to remove this legacy layer then.

Sebobo: See https://github.com/Sebobo/Shel.Neos.CommandBar/blob/84a15bf1415d35930ee2e6ff6474a58ea394e8d4/packages/ui-plugin/src/CommandBarUiPlugin.tsx#L685
and https://github.com/Sebobo/Shel.Neos.CommandBar/blob/84a15bf1415d35930ee2e6ff6474a58ea394e8d4/packages/ui-plugin/src/CommandBarUiPlugin.tsx#L55

Sebobo: Shouldn't it be possible to merge the data into the frontendConfiguration so we stay compatible in the 9.x world?

mhsdesign: I reverted the removal of `neos.configuration` and added a b/c layer for the two things you used https://github.com/neos/neos-ui/pull/4012/commits/1db58b16899d18be2fa5c9270b62380a1c6f5349

that way the pr is now completely non breaking to plugins even if they did crazy mans stuff like your command bar. Lets cleanup the legacy layer then in a proper minor or even major:)

mhsdesign: Cleanup pr already today so we dont forget that in the furture https://github.com/neos/neos-ui/pull/4029

Sebobo: Thx!

packages/neos-ui-contentrepository/src/registry/initializeNodeTypesRegistry.ts
```diff
@@ -0,0 +1,90 @@
+import {getInlinedDataFromBackend} from '@neos-project/neos-ui-configuration/src/bootstrap';
+import {nodeTypesRegistry} from './NodeTypesRegistry';
+
+const LINK_ID_FOR_SCHEMA_NODE_TYPE_ROUTE = 'neos-ui-uri:/neos/schema/node-type.json';
+
+export const nodeTypes = getInlinedDataFromBackend('nodeTypes') as {
```
Sebobo:
```suggestion
export const nodeTypeGroupsAndRoles = getInlinedDataFromBackend('nodeTypes') as {
```

packages/neos-ui-contentrepository/src/registry/initializeNodeTypesRegistry.ts
```diff
@@ -0,0 +1,90 @@
...
+        nodeTypesRegistry.set(nodeTypeName, {
+            ...nodeTypesSchema.nodeTypes[nodeTypeName],
+            name: nodeTypeName
+        });
+    });
+
+    nodeTypesRegistry.setConstraints(nodeTypesSchema.constraints);
+    nodeTypesRegistry.setInheritanceMap(nodeTypesSchema.inheritanceMap);
+
+    const {groups, roles} = nodeTypes;
```
Sebobo:
```suggestion
    const {groups, roles} = nodeTypeGroupsAndRoles;
```

Sebobo (approved):
Looks good, just a variable name change would be nice.

Sebobo (approved):

Issue #4006, opened 2025-10-09
BUG: Neos UI 8.3.15+ not compatible with PHP 8.0


michael-mueller:
I’m encountering a PHP 8.0 compatibility issue with the neos-ui package starting from version 8.3.15. According to the specs, Neos 8.3 is compatible with PHP 8.0. However, starting with Neos UI 8.3.15, I get fatal errors due to usage of PHP 8.1+ features.

Error Details:
Cannot unpack array with string keys
Exception Code: 0
Exception Type: Error
Neos_Neos_Ui_Fusion_Helper_NodeInfoHelper.php (Line 457)
Original File: Packages/Application/Neos.Neos.Ui/Classes/Fusion/Helper/NodeInfoHelper.php

Related line:
return [...$arguments, 'node' => $node->getContextPath()];

This is not supported in PHP 8.0 – unpacking arrays with string keys is a PHP 8.1 feature.

Sebobo:
Will be solved with the next patch release of 8.3

PR #4013, opened 2025-10-15
BUGFIX: Resolve PHP 8.0 compatbility regression

Sebobo:
Unpacking arrays with keys is not supported yet with the Neos 8.3 minimal PHP version of 8.0.

Resolves: #4006

DO NOT UPMERGE, as this only affects Neos 8.3 and not 8.4 which has PHP 8.2 as minimal version.


dlubitz (approved):

mhsdesign (approved):
good catch :)

PR #4020, opened 2025-10-23
BUGFIX: properly pass node hidden in menu state to UI

### LINKED ISSUE #4009: BUG: initial hiddenInMenus state in document tree incorrect
When initially loading the UI, all document nodes are marked as not hidden in menu, i.e. with a white icon. That includes nodes that _are_ actually hidden in menus. This changes once those are clicked; they will get the correct grey icon then.

Expected behavior: The hidden in menu state should be visible from the start.

Neos Version: 9.0.6
Neos UI Version: 9.0.2

nezaniel:
This adjusts the node info helper to properly pass the node hidden state to the UI

Resolves: #4009 

Further info: see issue

### DIFF
--- a/Classes/Fusion/Helper/NodeInfoHelper.php
+++ b/Classes/Fusion/Helper/NodeInfoHelper.php
@@ -104,7 +104,7 @@ class NodeInfoHelper implements ProtectedContextAwareInterface
             // ensure _isHidden is sent to hidden nodes are correctly shown in the tree.
             // TODO: we should export this correctly named, but that needs changes throughout the JS code as well.
             '_hidden' => $node->tags->withoutInherited()->contain(NeosSubtreeTag::disabled()),
-            '_hiddenInIndex' => $node->getProperty('hiddenInMenu'),
+            'hiddenInMenu' => $node->getProperty('hiddenInMenu'),
             '_hasTimeableNodeVisibility' =>
                 $node->getProperty('enableAfterDateTime') instanceof \DateTimeInterface
                 || $node->getProperty('disableAfterDateTime') instanceof \DateTimeInterface,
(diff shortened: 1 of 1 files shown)


skurfuerst (approved):
makes sense, LGTM by reading. I checked the source code of Neos to see that it now references hiddenInMenu and NOT hiddenInIndex.

Classes/Fusion/Helper/NodeInfoHelper.php
```diff
@@ -104,7 +104,7 @@ public function renderNodeWithMinimalPropertiesAndChildrenInformation(
             // ensure _isHidden is sent to hidden nodes are correctly shown in the tree.
             // TODO: we should export this correctly named, but that needs changes throughout the JS code as well.
             '_hidden' => $node->tags->withoutInherited()->contain(NeosSubtreeTag::disabled()),
-            '_hiddenInIndex' => $node->getProperty('hiddenInMenu'),
+            'hiddenInMenu' => $node->getProperty('hiddenInMenu'),
```
Sebobo, line 107: Hm this was done for compatibility reasons. I guess for now we should keep both names here until we adjust all of them.

Sebobo (changes requested):
Thx! Works, but I think we should keep the old property additionally for compatibility in plugins for now.

Classes/Fusion/Helper/NodeInfoHelper.php
nezaniel: afaik the whole UI was already refactored to use the new name, which is why this no longer worked.

nezaniel: (see https://github.com/neos/neos-ui/pull/3987)

Sebobo: Yes but plugins. Wouldn't hurt much to keep both for now, until we clean up.

nezaniel: ah, of course, /done

Sebobo (approved):
Thx!

PR #4022, opened 2025-10-26
TASK: Fix-up #4008 use GitHub ci + Speedup E2E !!!

mhsdesign:
refs https://github.com/neos/neos-ui/issues/4026

FixUp for #4008

Performance Speedup:
- runs e2e tests in a single action instead of using multiple circle ci workspaces and attach those
- uses good effective composer `Packages` caching
- for sauce labs we no longer bundle `node_modules` and upload that and everything else only the TestFixtures are uploaded and their version of testcafe and test-cafe-react-selectors is used

**What I did**

**How I did it**

**How to verify it**

todo:
- [x] make seb happy with naming of server and client suites
- [x] fix e2e flow server run
- [ ] optional: run e2e with certificate to allow to test copy & paste via navigation api eventually?


.circleci/config.yml
```diff
@@ -0,0 +1,37 @@
+
+#
+# CircleCI is disabled !!!
+#
+# Starting with Neos 9.0 we no longer use CircleCI. This file is to be eventually.
```
skurfuerst:
```suggestion
# Starting with Neos 9.0 we no longer use CircleCI. This file is to be eventually removed.
```

skurfuerst (approved):

Issue #4023, opened 2025-10-27
BUG: Content elements is not shown in UI after page reload


pKallert:
### Steps to reproduce: 

1. In Neos demo open a third level, for example Subpage 1
2. Reload page using browser reload 

### Expected result:

Content tree is shown.

### Actual result: 

Content tree is initially shown, then disapears.

<img width="321" height="542" alt="Image" src="https://github.com/user-attachments/assets/d055da14-5341-4a36-8554-40b15201131a" />

GuangHa:
@Sebobo Do you know when this fix will be released? The last release was on the 26.9.

Sebobo:
@GuangHa I'm not sure. The e2e tests are currently failing and I haven't been able to understand/solve the issue yet that is thrown in the CI pipeline. After spending a day and creating a fix for another issue in the tree I found during investigation, I gave up for now and asked someone else to take a look.

PR #4024, opened 2025-10-27
BUGFIX: Prevent empty content tree when incomplete node data for pagetree is merged

Sebobo:
With #3961 we don’t return content type childnodes in not „fully loaded“ node data for the ui anymore. When the pagetree data is merged after the content tree data (which is fully loaded) this lead to an empty page tree as the childnodes in the store were overriden.

With this change we ignore a partial childnode list when merging if we already have a „fully loaded“ node in the store.

Resolves: #4023

### DIFF
--- a/packages/neos-ui-redux-store/src/CR/Nodes/index.ts
+++ b/packages/neos-ui-redux-store/src/CR/Nodes/index.ts
@@ -7,7 +7,7 @@ import {actionTypes as editPreviewMode, Action as EditPreviewModeAction} from '.
 import * as selectors from './selectors';
 import {calculateNewFocusedNodes, getNodeOrThrow} from './helpers';
 
-import {FusionPath, NodeContextPath, InsertPosition, NodeMap, ClipboardMode, SelectionModeTypes, NodeTypeName} from '@neos-project/neos-ts-interfaces';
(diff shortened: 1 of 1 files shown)


Sebobo:
I'm still testing and verifying if this fix has a further side-effects.

I would have preferred to check the child nodes what role they have and only replace the "document"-type entries. But I'm not sure how I can access the NodeTypesRegistry here.

pKallert:
Thank you for this fix! 
I tested it locally and one side effect I found: Switching quickly between document nodes is no longer possible. 

1.  click on page 1
2. it starts loading
3. quickly click on page 2
4. page 1 is being loaded and not page 2

Sebobo:
Not sure what you mean. Can you make a clip?
And this doesn't happen without this change?

pKallert:
Tried to caputure it with a screen grab, but in the morning the behavior was more consistent on local. Perhaps if the loading times are worse the race condition of loading is worse? 
ezgif-5c16a324e8aa9f


Other than my test in the morning the behavior also now occurs on 8.3 so I guess we are good to go here 😄

Sebobo:
AFAIK we don't fully cancel the previous events or ignore their responses when you quickly click around. But that has been like that for a while now.

pKallert (approved):

pKallert:
Not sure why e2e test for 2 dimensions fails, on local the dimensions look good

markusguenther (approved):
Tested locally and it seam to work fine :)
Thanks for fixing ❤️

PR #4025, opened 2025-10-27
REMOVE ME: Run tests

mhsdesign:
**What I did**

**How I did it**

**How to verify it**


mhsdesign:
shows that `firefox` latests (e.g. 144) is broken thus we use lastest-1 now (143) in ci

PR #4028, opened 2025-11-10
FEATURE: Update ckeditor from v44 to v47

Sebobo:
Notable changes:

* 44.1.0
	* General performance improvements 15-45% lower editor load times
* 44.2.0
	* Emoji plugin
	* Performance improvements when loading and saving data
* [45.0.0](https://ckeditor.com/docs/ckeditor5/latest/updating/guides/update-to-45.html)
	* [Fullscreen feature](https://ckeditor.com/docs/ckeditor5/latest/features/fullscreen.html)
	* [Custom icons](https://ckeditor.com/docs/ckeditor5/latest/getting-started/setup/customizing-icons.html)
* 45.1.0
	* typing improvements especially for Safari browsers
* 45.2.0
	* improved pasting tables from word documents
* [46.0.0](https://ckeditor.com/docs/ckeditor5/latest/updating/guides/update-to-46.html)
        * New way to import -> allows us to fully export the ck api for extensibility
	* Improved "remove formatting"
	* List marker styling
	* Improved markdown conversion
* [47.0.0](https://ckeditor.com/docs/ckeditor5/latest/updating/guides/update-to-47.html)
	* AI AI AI


packages/neos-ui/src/apiExposureMap.js
```diff
@@ -27,95 +27,45 @@ import * as NeosUiRegistry from '@neos-project/neos-ui-registry';
 // It's not safe to just install CKE5 packages from the extension because then then "instanceof" checks will no longer work,
 // which would break CKE5 in some places.
 // Feel free to export more parts as needed.
```
mhsdesign:
```suggestion
```

we export everything so ...

packages/neos-ui/src/apiExposureMap.js
```diff
@@ -130,7 +80,11 @@ export default {
         reduxSagaEffects,
         reselect,
         reactCssThemr,
-        CkEditor5,
+        CkEditor5, // Backwards compatible export
+        'ckeditor5-engine': CkEditor5Engine,
```
mhsdesign: in order for this to work the name of the key is irrelevant but you need a stub file in `neos-ui-extensiblity/src/shims` and also the map.json there needs entries

mhsdesign:
after https://github.com/neos/neos-ui/pull/4012 we also have the capabilities to introduce a rudimentary ck-editor test plugin for the e2e tests ... that way we can test the extensibility api ... we should probably do that then.

packages/neos-ui/src/apiExposureMap.js
Sebobo: That's not fully true, as @ckeditor contains ~70 packages or so and we don't want to include all of them. But the text still needs to be adjusted, correct.

packages/neos-ui/src/apiExposureMap.js
Sebobo: I added and registered the shims

packages/neos-ui/src/apiExposureMap.js
Sebobo: Updated the comment

mhsdesign:
We also expose now all actual ckeditor packages like '@ckeditor/ckeditor5-engine' via extensibility api.

This should allow to write plugins easier and also might allow to bundle ckeditor plugins from npm for the usage inside of neos as they will use the correct ckeditor.

One difference is there though:

plugins could use `import CkEditor5Engine from '@ckeditor/ckeditor5-engine';` which would not be possible via the actual sources and should be `import * as CkEditor5Engine from '@ckeditor/ckeditor5-engine';` but the way our extensibility api works this is not possible. For that we probably need to use real ES modules inside of the browser.

Still we discussed that its good progress that we do export them despite this small left inconsistency.

mhsdesign:
fyi following packages are yet not exposed via the api

```
@ckeditor/ckeditor5-clipboard
@ckeditor/ckeditor5-icons
@ckeditor/ckeditor5-typing
@ckeditor/ckeditor5-undo
@ckeditor/ckeditor5-upload
@ckeditor/ckeditor5-utils
```

i noticed that at trying to bundle the image plugin natively ... but that is probably hacky either way

Issue #4033, opened 2025-11-20
BUG: Document nodes can appear twice after reset filtering


Sebobo:
Due to the optimisation we did in https://github.com/neos/neos-ui/pull/3961 we have a problem when merging new child node data for the document tree. Previously every request returned all children and we didn't have to merge the data. Now we get either the full list, or a filtered list which can lead to duplicated children. We need a proper way to merge new data into the existing list and at the same time keep the correct ordering.

https://github.com/user-attachments/assets/5241c603-b35c-4235-96cd-b91cb77185b8

PR #4034, opened 2025-11-20
BUGFIX: Prevent duplicate documents in page tree after filtering

Sebobo:
Resolves: #4033 by replacing existing document children with new ones sent by the backend instead of merging new and old children.


markusguenther (approved):

markusguenther:
Tested on Neos 8.3 and adapted that to neos 9.1 and tested it there as well.

mhsdesign:
Thank you both for the fix and for the upmerge <3

PR #4036, opened 2025-11-20
TASK: Undo preview url changes

kitsunet:


Sebobo (approved):

mhsdesign (approved):
looks fine thanks :) now we also need to change that flag back to `false` for 8.4

bwaidelich:
> now we also need to change that flag back to false for 8.4

Was that done in the meantime?

PR #3996, opened 2025-09-12
FEATURE: New link editor (Archaeopteryx)

mhsdesign:
The package is herby merged into the Neos Ui from https://github.com/sitegeist/Sitegeist.Archaeopteryx/tree/2.0.1

- [Initial working adaption of Archaeopteryx](https://github.com/neos/neos-ui/commit/a5ef804eab0ebe0332ca37a6919255be13dd332c)
- ...

With these adjustments the copied code already works and offers the new link editor experience though the task is not complete yet:

- [ ] add link editor documentation to neos docs or some place -> document that `anchor: true` is obsolete and that first level starting point is deprecated.
- [x] remove previous ckeditor link editor and inspector link editor
- [x] remove php linting from circle ci in favour of new workflow -> https://github.com/neos/neos-ui/issues/4026
- [x] discuss where to place `Neos\Neos\Ui\LinkEditor\Link` DTO (the thing which is stuffed into the Node) -> https://github.com/neos/neos-development-collection/pull/5682
- [x] fix `LinkToArrayForNeosUiConverter` hack in core -> https://github.com/neos/neos-ui/pull/3723
- [x] remove neos-ui-link-editor-bridge corruption layer
- [x] remove styled-components and use css modules
- [x] remove rx-js and use framework-observeable if possible
- [x] evaluate the need for react-final-form and final-form (all other neos ui parts are coded without)
- [x] remove or combine neos-ui-link-editor-error-handling in favour of our own error handling
- [x] commit all new .yarn dependencies but evaluate the necessity of each
- [x] satisfy js linting and typescript
- [x] optimise bundle size by lazy loading the huge library "libphonenumber-js"
- [x] ~document new plain node-js tests and also see that we only use one version of esbuild for the testing too (evaluate if we can write node-js tests in typescript?)~
- [x] reintroduce `download` attribute support - ui feature since https://github.com/neos/neos-ui/pull/3197
- [x] extract inline styles to css modules
- [x] extract neos globals change and evaluate where to place the global registry and config
  - https://github.com/neos/neos-ui/pull/4010
  - https://github.com/neos/neos-ui/pull/4011
  - https://github.com/neos/neos-ui/pull/4012
- [x] unify icon choices e.g. use cross (x) icon instead of bin like in assets editor
- [x] review how to use the new linktype configuration
- [x] paulas review
- [x] vertical tabs
- [x] do not use `action: ingored` maybe fix flow already and build my and christians idea or use `FlowRoute` annotation?
[...]

PR #4037, opened 2025-11-22
TASK: New link editor final final final polish

mhsdesign:
Just a tiny followup attempting to polish the new link editor (#3996) to absolute shininess ;)

Most importantly we support now `popovers` inside dialogs and thus clickoutside to close for the advanced link options.

image

Getting popovers to work in our dialog was not trivial as `transform translate -50%` trips browsers off (even though the spec allows it). Now we center the dialog via `margin: auto` on all axis. The change https://github.com/neos/neos-ui/pull/4007 is partly readjusted herby as dialog style `wide` does no longer allow to expand to the full width. For more details see commits below.

 - https://github.com/neos/neos-ui/pull/4037/commits/7caa659cfd0f810e76ab1c76f1d93b50ded21e4a (alternative: use flex/grid: https://github.com/neos/neos-ui/pull/4037/commits/75d9daf6dd29361b1839116b502ddc2c0af788ec)
 - https://github.com/neos/neos-ui/pull/4037/commits/2a4802420501b6fe7573d4ec447e86c6f53857c7

Todo:
- [x] solve new mandatory todos from link editor pr (in code)
- [x] fix dialog flex box centering - its now too wide with too much text
- [x] add preview of anchor in node and asset card
- [x] "Force download" and "Herunterladen erzwingen" as label
- [x] Phone validation without leading 0 is wrong label

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/Resources/Private/Translations/de/LinkEditor/Main.xlf
+++ b/Resources/Private/Translations/de/LinkEditor/Main.xlf
@@ -17,6 +17,10 @@
                 <source>Edit Link</source>
                 <target>Link editieren</target>
             </trans-unit>
+            <trans-unit id="inspector.delete" xml:space="preserve" approved="yes">
+                <source>Delete Link</source>
+                <target>Link löschen</target>
+            </trans-unit>
             <trans-unit id="inspector.notfound" xml:space="preserve" approved="yes">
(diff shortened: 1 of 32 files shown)


Sebobo (approved):
LGTM

PR #4039, opened 2025-12-02
TASK: Fix jenkins release build step

Sebobo:
The first build step was replaced with the subpackage build step in 78c9719e5db7cf6124d3709f99ea0839669d461f, but they rely on each other.


kitsunet (approved):
Lets try :)

PR #4041, opened 2025-12-02
BUGFIX: Correctly handle empty children for notInlineEditable overlay

### LINKED ISSUE #4040: Non-Inline-Editable-Overlay not initialized correctly
Nodes without inline-editable properties or editable children should have this overlay to allow clicking them for selection without invoking their content.

In Neos9 the condition for rendering the overlay seems broken:

https://github.com/neos/neos-ui/blob/0dcdbe433826f322cdef3d9958a9b57bbaa42c0d/packages/neos-ui-guest-frame/src/initializeContentDomNode.js#L30-L32

The overlay is only created initialized, if the node is not inline-editable and doesn't have children.
However, the node seems to always receive an empty array of children, which is still truthy and the condition `hasChildren = Boolean(nodes?.[contextPath]?.children)` is always true, so we never get the overlay.

This could be changed to `hasChildren = Boolean(nodes?.[contextPath]?.children?.length)`, but it might be possible that other places also rely on this. I don't have a good grasp about the changes to the node data.

PRGfx:
Correctly handle empty child-node array when checking if a node can get a not-inline-editable overlay

Resolves #4040

### DIFF
--- a/packages/neos-ui-guest-frame/src/initializeContentDomNode.js
+++ b/packages/neos-ui-guest-frame/src/initializeContentDomNode.js
@@ -19,7 +19,7 @@ export default ({store, globalRegistry, nodeTypesRegistry, inlineEditorRegistry}
     }
 
     const isHidden = nodes?.[contextPath]?.properties?._hidden;
-    const hasChildren = Boolean(nodes?.[contextPath]?.children);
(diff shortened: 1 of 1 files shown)


Sebobo (approved):
Thx! I can confirm that this solves the mentioned issue.

PR #4044, opened 2025-12-03
TASK: Fix test-plugin build in Jenkins release

Sebobo:
This changes makes sure, that all normal packages are build first, so that the test-plugin can access their built sources.


kitsunet (approved):
Looks good

PR #4049, opened 2025-12-14
BUGFIX: Exception when using cut and paste node to move node to current position

mhsdesign:
The CR rightfully complains as we kinda do bungus here :)

> Node aggregate "d8a8a5e5-275f-c01b-c6be-f082ef11d0ad" is no sibling of node aggregate "d8a8a5e5-275f-c01b-c6be-f082ef11d0ad" but was expected to be in dimension space point {"language":"en_US"}

While we could `return;` here in the operation the Move operation could still have an effect with multiple dimension at play and depending on the `RelationDistributionStrategy` - so for consistency we still execute the move.

This fix will be especially relevant with drag and drop where the source node is removed optimistically and not reset because of missing error handling

**What I did**

**How I did it**

**How to verify it**


Sebobo (approved):
Thx, sounds reasonable

PR #4050, opened 2025-12-15
FEATURE: Show content tree by default when opening the Neos Ui for the first time

mhsdesign:
The toggle state of the content tree (structure tree) is persisted in the local storage.

We as users known to users open it on new projects out of reflex but new users might not even be aware that it exists.

To ensure new users to neos get familiar with the structure of a site its crucial to see the layout of the content nodes which also sometimes eases operations that are not possible in the inline Ui because of too complex visual layouts.

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/Configuration/Settings.yaml
+++ b/Configuration/Settings.yaml
@@ -175,7 +175,7 @@ Neos:
           leftSideBar:
             isHidden: false
             contentTree:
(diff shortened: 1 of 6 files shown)


Sebobo (approved):
Thx!

PR #4052, opened 2025-12-16
BUGFIX: Properly handle variation of tethered children of root nodes…

nezaniel:
…and their descendants

**What I did**

Tethered nodes cannot be varied (except if they can, see below), so the UI finds the closest un-tethered ancestor to vary instead.
This becomes a problem if said ancestor is a root node, which can happen if one e.g. tethers a site node to the sites root.
The CR already covers this case by making an exception for tethered variation - in exactly that case.
The UI now uses that functionality to avoid unresolvable errors.

**How I did it**

I threw tests at it and checked for root classification

**How to verify it**

Run the tests of try for yourself


mhsdesign (approved):
Did i ever mention i love behat tests? <3 Thank you!!

dlubitz (approved):

PR #4055, opened 2025-12-18
BUGFIX: Prevent clicking structural toolbar to be behind content

### LINKED ISSUE #4054: BUG: Structural Toolbar not always clickable when below content element
Due to an z-index issue on the StructuralToolbar. `calc(infinity)` instead of `auto` should help, but have to verify why it was set to `auto`.

Image

Sebobo:
In the „below“ case it was possible that the structural toolbar was behind the next content element of the currently selected one. With this change we increase the z-index for the standard and hover case, to also allow selecting the toolbar on hover when the formatting toolbar is also at the bottom.

Resolves: #4054

CleanShot 2025-12-18 at 09 30 29@2x


mhsdesign (approved):
so the image shows the current state... in _after_ it would not be shown in front and only on hover? That makes sense

Sebobo:
Yes its behind the ck toolbar until you hover it, but should now always be in front of the next sibling node.

markusguenther (approved):
Works fine also in the customer project :)

Issue #4053, opened 2025-12-17
BUG: 9.1 Inline editing styles can break site styles


Sebobo:
Example: The ck inline editor styles override there padding for the buttons on neos.io.
Also CKEditor applies `text-align:left` which we don't want and should be provided by the CSS of the website itself.

<img width="2438" height="956" alt="Image" src="https://github.com/user-attachments/assets/6d1b6c85-a1f1-4868-b32c-28808f956dde" />

PR #4057, opened 2025-12-19
BUGFIX: Remove conflicting ck inline editor styles

Sebobo:
This change removes several inline styles from the default ck editor theme that may conflict with the style of the edited website.

In the future (with #4056) we want to have a customised theme that doesn’t require a patch so we don’t have issues when updating CKEditor.

Relates: #4053


markusguenther (approved):
Tested in Customer project and it works fine

PR #4058, opened 2025-12-21
BUGFIX: Convert enums to their backing value for the UI

nezaniel:
**What I did**

I tried to edit enums in the UI and found out that they were serialized as an array. This fixes that.

**How I did it**

I added a specific check to the NodePropertyConverterService

**How to verify it**

declare an enum property value using the textfield editor and try editing it


mhsdesign (approved):
yes makes fully sense here ... and thats analog to the native `JsonSerializable` support  #3723

PR #4062, opened 2026-01-07
BUGFIX: Prevent confirmation dialog when switching inspector tabs

### LINKED ISSUE #4061: BUG: Cannot change inspector tab without applying changes
When I try to switch the inspector tab while there are unapplied changes, the apply confirmation dialog is displayed instead.
This impedes editorial workflows.

Sebobo:
The overlay that triggers the dialog was not adjusted for the removed „Current Selection“ element.

Resolves: #4061

CleanShot 2026-01-07 at 10 32 40@2x


mhsdesign (approved):
we have a funny codebase :D

Issue #4064, opened 2026-01-12
BUG: Click behaviour issues in content tree


Sebobo:
* First click only selects element
* Second click scrolls to element, but also sometimes to the previously selected one
* Every click triggers a backend request for new node data.

Sebobo:
The backend request is triggered by a regression in `packages/neos-ui-sagas/src/UI/Inspector/index.js` of 02ef655858912ca12230df9104b5b5d7b79329c2 which incorrectly checks whether a node is fully loaded.

PR #4065, opened 2026-01-12
BUGFIX: Prevent unnecessary reload of selected node

Sebobo:
Resolves regression from typescript replacement of plow.js logic.

Relates: #4064

### DIFF
--- a/packages/neos-ui-sagas/src/UI/Inspector/index.js
+++ b/packages/neos-ui-sagas/src/UI/Inspector/index.js
@@ -41,7 +41,7 @@ export function * inspectorSaga({globalRegistry}) {
                 yield put(actions.UI.Inspector.closeSecondaryInspector());
 
                 const focusedNode = yield select(getFocusedNode);
-                if (focusedNode?.isFullyLoaded) {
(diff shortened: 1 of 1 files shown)


packages/neos-ui-sagas/src/UI/Inspector/index.js
```diff
@@ -41,7 +41,7 @@ export function * inspectorSaga({globalRegistry}) {
                 yield put(actions.UI.Inspector.closeSecondaryInspector());
 
                 const focusedNode = yield select(getFocusedNode);
-                if (focusedNode?.isFullyLoaded) {
+                if (focusedNode && !focusedNode.isFullyLoaded) {
```
mhsdesign, line 44: yes it was `!$get('isFullyLoaded', focusedNode)` when introduced with https://github.com/neos/neos-ui/pull/3364 though when upmerged in the plow-js removal change **I** missed the negation part (`!`).

Originally the code from #3364 would also reload a node if its not at all in the redux store - e.g. `null` - but that was probably not indented and not tested as actual feature.

mhsdesign (approved):
Good catch:)

PR #4066, opened 2026-01-12
BUGFIX: Scroll to focused element in guest frame

### LINKED ISSUE #4064: BUG: Click behaviour issues in content tree

Sebobo:
This solves a regression in Neos 9.1 when the toolbar was refactored to a functional component. The focusedNode in the state was still the previous node in the function instead of the newly selected one. With this change we make sure that the focusedNode matches the updated state with the scroll request.

Resolves: #4064


mhsdesign (approved):
Yes fetching state twice can lead to some funky behaviour:D

PR #4070, opened 2026-01-17
BUGFIX: Refactor `tryFromString` 

robinroloff:
Currently, it is not possible to have a colon in a translation identifier. I don't see a reason for prohibit this, it only confuses people.

With this change, it is now possible to have as many colons as you want after the file XLF file definition.

Example from my use case, which failed:

```
Medienreaktor.Site:Content.Section:properties.renderComponent.values.Medienreaktor.Site:Component.Section.Default
```

Didn't work before, not it works

### DIFF
--- a/packages/neos-ui-i18n/src/model/TranslationAddress.ts
+++ b/packages/neos-ui-i18n/src/model/TranslationAddress.ts
@@ -25,7 +25,9 @@ export class TranslationAddress {
         new TranslationAddress(props.id, props.sourceName, props.packageKey, `${props.packageKey}:${props.sourceName}:${props.id}`);
 
     public static tryFromString = (value: string): TranslationAddress | null => {
-        const [packageKey, sourceName, ...rest] = value.split(TRANSLATION_ADDRESS_SEPARATOR);
+        const split = value.split(TRANSLATION_ADDRESS_SEPARATOR);
+        if(split.length < 3) return null;
+        const [packageKey, sourceName, ...rest] = split;
 
         if (!packageKey || !sourceName || rest.length === 0) {
             return null;
--- a/packages/neos-ui-i18n/src/model/TranslationAddress.ts
+++ b/packages/neos-ui-i18n/src/model/TranslationAddress.ts
@@ -24,16 +24,17 @@ export class TranslationAddress {
     }): TranslationAddress =>
         new TranslationAddress(props.id, props.sourceName, props.packageKey, `${props.packageKey}:${props.sourceName}:${props.id}`);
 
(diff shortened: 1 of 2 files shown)


mhsdesign:
Hi thanks for drafting a pr so quickly to show how it can be solved:)

Now yes starting with Neos 9.0 via #3804 a translation id is now strictly defined by three segments separated by a colon `:`. The first is the package key, the second the xlf file e.g. `Main` and the last is the field within that file.

I was not aware that you can use as the last field name a value with a colon but i see the usecase in your example with a node type name. I assume `renderComponent` is a selectbox or the sort and `values` is the node type?

Anyway Flows `Neos\Flow\I18n\EelHelper\TranslationHelper::I18N_LABEL_ID_PATTERN` also seems to only require at least 3 colons but handles your above translation id also well. Under that premise and that it probably worked in Neos 8.3 Ui as well i agree with this change.

We only need to target Neos 9.0 as its a bugfix and thats when the regression first came in and also a new test should showcase that this is a valid translation id and also to assert that the whole split and join back together thing works as expected:)

robinroloff:
> We only need to target Neos 9.0 as its a bugfix and thats when the regression first came in and also a new test should showcase that this is a valid translation id and also to assert that the whole split and join back together thing works as expected:)

Okay, rebased it.

robinroloff:
Also added tests

packages/neos-ui-i18n/src/model/TranslationAddress.ts
```diff
@@ -24,16 +24,23 @@ export class TranslationAddress {
...
+        if (split.length < 3) {
             return null;
         }
 
-        const [packageKey, sourceName, id] = parts;
+        const [packageKey, sourceName, ...rest] = split;
 
-        return new TranslationAddress(id, sourceName, packageKey, string);
-    }
+        if (!packageKey || !sourceName || rest.length === 0) {
```
mhsdesign: this line is also new logic, i guess just out of habit but yes it makes sense if any of these are empty its not a valid id.
a test to assert that `::` or `a::` `:a:` `::a` are `null` would be cherry on top <3

mhsdesign (approved):
Thanks looks good, lets see if all the test pass CI ... added one last bonus task ^^

EDIT: Linting fails.

markusguenther (approved):
Thanks @robinroloff and @mhsdesign .. it looks good to me.
Should we than also take care for flow to have the same behavoir here.

mhsdesign:
Flows Translation EEL helper already allows this, otherwise i would have been skeptical for implementing something in the Ui which the Flow core doesnt handle

PR #4074, opened 2026-01-29
BUGFIX: clear dataloader debounce timers

### LINKED ISSUE #4047: BUG: Reference editor: Repeated search for same term returns different results
## Current behaviour

* If I repeatedly search for the same term in the reference editor, I potentially get a different result set each time.

## Expected behaviour

* The search for the same term always returns the same result.

## Steps to reproduce

* I configured a reference from one node type to another
* In my video example, I'm searching for the term "Hund". There are exactly two matching items.

https://github.com/user-attachments/assets/6708ee0f-c7e5-46b6-a0e5-6ec55fa29792
 
* I enter "Hund" a first time.
* ✅ Two results are being displayed
* I delete the text and enter "Hund" a second time.
* 🐞 Four results are being displayed
* I delete the text and enter "Hund" a third time.
* 🐞 Even more results are being displayed
* I delete the text and enter "Hund" a fourth time.
* 🐞 Even more results are being displayed.
* I delete the text and enter "Hund" a fifth time.
* ✅ Two results are being displayed

## What happens

* As you can see in the Network tools, for the first search, "Hund" is being sent to the API.
* The second request has "Hun" as a payload
* The third request has "Hu" as a payload
* The fourth request has "H" as a payload
* When I enter "Hund" the fifth time, somehow Neos realized that it already has results for "Hund" in the cache and returns the correct results.

I double-checked this in another installation and with longer words and it's always the same. Each time I manually re-enter the same term, one more character is trimmed from the end of the word until no character is left.

## Environment

* Neos 8.3

PRGfx:
If the search is not cancelled, when a cached search term is requested, the last uncached substring of the search term will be requested and its results returned.

Resolves #4047

### DIFF
--- a/packages/neos-ui/src/manifest.dataloaders.js
+++ b/packages/neos-ui/src/manifest.dataloaders.js
@@ -126,6 +126,9 @@ manifest('main.dataloaders', {}, globalRegistry => {
 
             const cacheKey = makeCacheKey('search', {options, searchTerm});
             if (this._lru().has(cacheKey)) {
+                if (this._debounceTimer) {
+                    window.clearTimeout(this._debounceTimer);
+                }
                 return this._lru().get(cacheKey);
             }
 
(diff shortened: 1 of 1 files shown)


Sebobo (approved):
Thx, makes sense by reading!

mhsdesign (approved):
I tested that we indeed solve https://github.com/neos/neos-ui/issues/4047 fully! Thanks a lot.

lorenzulrich:
Thanks @PRGfx! We had the same problem and are glad it was solved by you!

PR #4077, opened 2026-02-19
TASK: Upgrade CKEditor to version 47.5.0

markusguenther:


Sebobo:
Thx, as some ckeditor minor updates are super breaky or actually offer great fixes or improvements, I would like that we have a short compressed information with the PR what has changed in the specified release(s) which might be relevant for us.

And was there some reason you updated or just general maintenance?

markusguenther:
I just tested something locally and saw that we are having some new releases on the CKE side.
They did not write that it is breaking. But need to summarize the changes. Most was the AI stuff that we are not using at the moment.

Issue #4076, opened 2026-02-09
BUG: Media search box disabled when initially inserting an inline link


nezaniel:
When inserting a new inline link and selecting the media tab, the search box is disabled and cannot be used.
It only becomes available after closing the dialog and re-opening it for editing the link.

![Image](https://github.com/user-attachments/assets/7975dda2-38d7-4260-bae8-6437a56cb3c5)
The disabled state can be seen as a grey background

mhsdesign:
Yes i can confirm this. The `readonly` attribute on the input seems to be the problem. And indeed only if an asset is set and the editor is reopened the search box is enabled.

Issue #4059, opened 2025-12-22
New link editor followup


mhsdesign:
The new link editor was introduced via #3996 and finalised via #4037 - now we still found a few bits and pieces that have to be finial final finalised:

2. During search the tree also searches through content and shows all these contents ... that makes the search already slower and yields a lot of output which is of no use as one cannot link to a main content collection or its children

3. Bottom to top approach for search and for selected node id possibly in-performant as we go upside to find the ancestors and for each level we find the siblings. -> The resulting order is thus also possibly wrong! See test.

Maybe findSubtree needs a nodeId instead of depth to look that deep? And also a search term filter?

4. Use findSubtree instead of multiple findChildren queries to build up initial tree

5. Use new json dimension space point in client instead of legacy 8 format

6. Starting points as node paths are effectively obsolete with nodes carrying no names regularly, promote to use node ids but find a pattern that developers define node ids for project? -> HIghlander nodes.

7.  Write documentation how to write new link type as plugin (figure out how the `Web` tab is by default asked at last with `isSuiteable` as it matches all always?

8. Some projects require to use the FlatNav package and hide nodes from the tree which appear a lot like blog posts or news entries. If the base node type filter excludes them and they are better displayed in another view like FlatNav the link editor should also show them separate:
- this probably requires that FlatNav or the like get supported in the core or that FlatNav ships a link editor plugin
- if this results in two tabs, e.g. two link types the `node://` protocol has no clear tab where editing could start, but that information must be available for the link editor during opening without requires.
  - either we extend the node protocol like `node://news/` or we see that the FlatNav will be part of the current link editor tab and one can switch inside that tab between the document tree and a flat nav

9. extract `NodeTree` building into `Shared` namespace so the upcoming #4021 can us it.

There are some utterly verbose and heavy CR core things like node type multiplication which must definitely reside in the CR core. We should attempt to make the tree builder more lightweight and use rather CR core things than that the Neos Ui introduces duplicated code.

10. Adjust media screen to have less padding in the `media-browser-inspector` class on small screens or when embedded in the link editor (see https://github.com/neos/neos-ui/pull/4081)

PR #4081, opened 2026-02-25
BUGFIX: Dont attempt to manipulate MediaBrowser iframe in LinkEditor which disables search input

mhsdesign:
Subtask of https://github.com/neos/neos-ui/issues/4059
Solves https://github.com/neos/neos-ui/issues/4076

As part of an artefact from the original Archaeopteryx LinkEditor implementation the iframe html was modfied to has less spacing around, no footer (for navigation) and all inputs were attempted to be disabled.

Due to a timing issue this modification to the iframe only takes place for new links. Existing asset links did show the media browser in its original form.

Now the html manipulation is flawed (obviously):
- Because its a hack and behaviour should instead be configured via a query parameter or new route
- Editing of images was attempted to be prohibited, but "Edit", "Delete" was still clickable
- From the "Edit" screen there is no navigation back because we remove the footer
- In the "Edit" screen checkboxes are still clickable because they were not considered
- Valid inputs like the search field do no longer work
- The html manipulation is hard to properly apply on an iframe and did not apply for existing asset links

Instead this change removes the manipulation as it is totally superfluous in the first way considering that in other points the Neos Ui - linke for image selection in the inspector - shows also the full Media Module with all its features. Also the padding is now slightly bigger around the MediaSelection but everything is still well accessible.

| Type | Overview | Details (dot menu -> "Edit") |
|--------|--------|--------|
| Before | Bildschirmfoto 2026-02-25 um 19 15 | Bildschirmfoto 2026-02-25 um 19 16 | 
| After | Bildschirmfoto 2026-02-25 um 19 14 01 | Bildschirmfoto 2026-02-25 um 19 14 |

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/packages/neos-ui-link-editor-core/src/application/LinkTypes/Asset/MediaBrowser.tsx
+++ b/packages/neos-ui-link-editor-core/src/application/LinkTypes/Asset/MediaBrowser.tsx
@@ -10,8 +10,6 @@ interface Props {
 }
 
 export const MediaBrowser: React.FC<Props> = (props) => {
-    const containerRef = React.useRef<HTMLDivElement>(null)
-
     const secondaryEditorsRegistry = getRegistryById('inspector')?.get('secondaryEditors');
 
     const secondaryEditorId = 'Neos.Neos/Inspector/Secondary/Editors/MediaSelectionScreen';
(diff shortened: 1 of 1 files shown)


mhsdesign:
That being said the media screen looks prettier of course with the `.media-browser-inspector` class removed but doing that in any way on this end is wrong and will probably never be cleaned up. Instead we could attempt to make the media browser have less spacing on a small screen. That should do the trick.

PR #4082, opened 2026-02-25
TASK: Expose UI framework functions to plugin extensibility api

mhsdesign:
With https://github.com/neos/neos-ui/pull/3804 a new `translate` method was made available.
This change now exposes this new `translate()` function for plugins while preserving the legacy default export.

Currently, translation is possibly by importing (the default export)

```
import I18n from '@neos-project/neos-ui-i18n';
```

Now by allowing 

```
import {translate} from '@neos-project/neos-ui-i18n';
```

We are in the struggle as the default export is now due to the CJS conversion as `import I18n from` no longer accessible but double nested into `I18n.default`. This is perfectly elaborated here:

https://blog.andrewbran.ch/default-exports-in-commonjs-libraries/

To denote that everything we put into `module.exports` is to be taken as ESM - meaning the `default` key is special - we use the conventional flag `__esModule`.

As we adjust this part in the host Neos Ui code existing plugins will not break and continue to be able to import `<I18n>` from the default export. Simultaneously though we allow the use of `{translate}`

--------

Framework libraries for building the Neos Ui itself (e.g state handling) is now also exposed via the respective import. See the packages readmes for more detail, but be aware that these packages are currently for advanced plugin authors only and still under development.

- `@neos-project/framework-observable`
- `@neos-project/framework-observable-react`
- `@neos-project/framework-promise-react`

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/packages/neos-ui-extensibility/extensibilityMap.json
+++ b/packages/neos-ui-extensibility/extensibilityMap.json
@@ -34,5 +34,8 @@
     "@neos-project/neos-ui-guest-frame": "@neos-project/neos-ui-extensibility/dist/shims/neosProjectPackages/neos-ui-guest-frame",
     "@neos-project/neos-ui-registry": "@neos-project/neos-ui-extensibility/dist/shims/neosProjectPackages/neos-ui-registry",
     "@neos-project/neos-ui-configuration": "@neos-project/neos-ui-extensibility/dist/shims/neosProjectPackages/neos-ui-configuration",
+    "@neos-project/framework-observable": "@neos-project/neos-ui-extensibility/dist/shims/neosProjectPackages/framework-observable",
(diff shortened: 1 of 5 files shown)


mhsdesign:
Tested with a dummy plugin using

```js
import React from 'react';
import I18n, {translate} from '@neos-project/neos-ui-i18n';
import manifest from '@neos-project/neos-ui-extensibility';

manifest('@packagefactory/ckinlineimages-plugin', {}, (globalRegistry) => {
	globalRegistry.get('containers').set('App', () => <div><I18n id="Preserve me" /> <I18n id="Neos.Neos.Ui:Main:below" />{translate("Neos.Neos.Ui:Main:below")}</div>)
})
```

PR #4085, opened 2026-03-04
BUGFIX: Fix esbuild watch mode

JamesAlias:
**What I did**
- fixed watch mode for esbuild which I didn't notice not working 🙈 

**How I did it**

**How to verify it**
- `make setup`
- `make build-watch`
- change a file

### DIFF
--- a/esbuild.js
+++ b/esbuild.js
@@ -93,7 +93,10 @@ const options = {
 }
 
 if (isWatch) {
-    esbuild.context(options).then((ctx) => ctx.watch())
+    (async () => {
+        const ctx = await esbuild.context(options);
+        await ctx.watch();
+    })();
 } else {
     esbuild.build(options).then(result => {
         if (isAnalyze) {


mhsdesign (commented):
interesting. its working for in my cases and also how esbuild promotes the use of it?
so im unsure this feels odd:D but i can test it out. ...Also for dev tasks i always target the lowest version as well.

mhsdesign:
We also promote this pattern in the docs to build plugins ... so id really like to know what is broken^^

See esbuild https://github.com/evanw/esbuild/releases/tag/v0.17.0

JamesAlias:
> interesting. its working for in my cases and also how esbuild promotes the use of it? so im unsure this feels odd:D but i can test it out. ...Also for dev tasks i always target the lowest version as well.

You are completely right. The problem was sitting in front of the computer. idk
Thanks for questioning the PR.

PR #4088, opened 2026-03-04
BUGFIX: Sort NodeType groups in NodeTypeGroupsAndRolesProvider

### LINKED ISSUE #4087: BUG: Sorting groups in NodeType Select Dialog is not working
### Description

Since Neos 9.0 the sorting of the NodeType groups is not working anymore. In pre 9.0 Versions the sorting happend in Fusion, this has been removed and the new `NodeTypeGroupsAndRolesProvider` does not handle the position sorting.

### Steps to Reproduce

1. Create NodeType Setting for group adjustment
2. Change postion of general for instance to end

#### Expected behavior

```
Neos:
  Neos:
    nodeTypes:
      groups:
        textMedia:
          label: "Text & media"
          position: 100
        general:
          label: "Standard"
          position: 200
        structure:
          label: "Structure"
          position: 250
```

With that configuration I would expect to have the `Text & media` at the fist position.

#### Actual behavior

The general is the first and not Text & media.

### Affected Versions

Neos: 9.0

UI: 9.0

markusguenther:
To prevent sorting the groups within the NeosUI on the JavaScript side, the NodeTypeGroupsAndRolesProvider is taking care of it, and so the initialState in the Neos backend will already start with correctly sorted groups.

In earlier versions, this was handled by Fusion.

Fixes: #4087

### DIFF
--- a/Classes/Infrastructure/ContentRepository/NodeTypeGroupsAndRolesProvider.php
+++ b/Classes/Infrastructure/ContentRepository/NodeTypeGroupsAndRolesProvider.php
@@ -16,6 +16,8 @@ namespace Neos\Neos\Ui\Infrastructure\ContentRepository;
 
 use Neos\Flow\Annotations as Flow;
 use Neos\Neos\Ui\Domain\InitialData\NodeTypeGroupsAndRolesProviderInterface;
+use Neos\Utility\Exception\InvalidPositionException;
+use Neos\Utility\PositionalArraySorter;
 
 /**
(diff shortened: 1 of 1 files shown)


Sebobo (approved):
Makes sense since this was moved from Fusion to PHP without the sorting we had in Fusion, thx!

mhsdesign (approved):
yes indeed seems to be caused by https://github.com/neos/neos-ui/pull/3682

Issue #4089, opened 2026-03-04
BUG: Margin between icon and value in select box doesn't work


Sebobo:
See

<img width="680" height="646" alt="Image" src="https://github.com/user-attachments/assets/d6f9dbf6-8a32-45f5-a83e-1155aad0c8b4" />

This can be solved by moving and increasing the margin from the icon itself to the icon wrapper.

PR #4090, opened 2026-03-04
BUGFIX: Icon margin in selectbox

Sebobo:
Adjust margin between icon and value to match the padding of the selectbox

Resolves: #4089

CleanShot 2026-03-04 at 11 58 00@2x


markusguenther (approved):
Thanks for the detail fix :)

PR #4093, opened 2026-03-11
BUGFIX: Fix `search` operation to find nodes by aggregate id

bwaidelich:
Related: https://github.com/neos/neos-development-collection/issues/5756, https://github.com/neos/neos-development-collection/issues/3771, https://github.com/neos/neos-development-collection/pull/1477

### DIFF
--- a/Classes/FlowQueryOperations/SearchOperation.php
+++ b/Classes/FlowQueryOperations/SearchOperation.php
@@ -13,6 +13,7 @@ namespace Neos\Neos\Ui\FlowQueryOperations;
 
 use Neos\ContentRepository\Core\Projection\ContentGraph\Filter\FindDescendantNodesFilter;
 use Neos\ContentRepository\Core\Projection\ContentGraph\Node;
+use Neos\ContentRepository\Core\SharedModel\Node\NodeAggregateId;
 use Neos\ContentRepositoryRegistry\ContentRepositoryRegistry;
 use Neos\Eel\FlowQuery\FlowQuery;
 use Neos\Eel\FlowQuery\Operations\AbstractOperation;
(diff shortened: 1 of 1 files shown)


mhsdesign (approved):

Sebobo:
Thx, but I don't see this fixing the issue with the references editor, which uses the `NodesController`. And that one doesn't use the `SearchOperation`. My tests were also still failing with this patch.

bwaidelich:
> I don't see this fixing the issue with the references editor

uh, I missed that part and fixed it for the node tree searchbox – so we use another mechanism for the references editor?! I'll look into that

> My tests were also still failing with this patch.

@Sebobo Which tests fail for you? The failing e2e tests don't seem to be related

Sebobo:
This manual test :D 
CleanShot 2026-03-12 at 10 18 54@2x

I updated the issue, so it's clearer which part of Neos I meant ;)

Classes/FlowQueryOperations/SearchOperation.php
```diff
@@ -83,6 +84,13 @@ public function evaluate(FlowQuery $flowQuery, array $arguments): void
             $contextNode->aggregateId,
             $filter
         );
+        $nodeAggregateIdOrNull = NodeAggregateId::tryFromString($arguments[0] ?? '');
```
Sebobo, line 87: Does it make sense to run a full `findDescendantNodes` call, in case we have a valid NodeAggregateId?

bwaidelich: A valid NodeAggregateId is every string that matches the regex of `/^([a-z0-9\-]{1,64})$/` – so I'm afraid we can't detect those.
We _could_ search by id first and skip the `findDescendantNodes` call if there was a match, but I think that can be very counter productive if you use speaking ids like `product-123`

Sebobo (approved):
Works, thx!

Issue #4071, opened 2026-01-19
BUG: Content Tree shows duplicated nodes


connykaelinvisol:
Bug in Version 8.4.1: The ContentTree shows duplicated nodes like ContentCollections with identical children. When opening one of the shown ContentCollections both getting opened and show their (identical) content. This goes far more deep if there are again nested children.

First it appears in the main ContentCollection of the page. After page reload and opening the "main" i can see all my Sections double for a tiny fraction of a second, then they disappear. The nodes on lower level stay duplicated all the time.

Issue can be fixed by downgrading to 8.4.0.

<img width="308" height="248" alt="Image" src="https://github.com/user-attachments/assets/56789ec8-e01f-431c-aa80-eac601d28f9e" />

Sebobo:
Thx for the issue. This could already be fixed and just needs a release. Needs to be verified.

connykaelinvisol:
The issue is also in 8.3.17, just to be precise ;-)

Sebobo:
Hi @connykaelinvisol can you check if the problem still persist with the latest ui releases?

connykaelinvisol:
Hi @Sebobo
As far as I can see there is no fix in the 8.3. version, right? Unfortunately I had to de-migrate back to 8.3.16 (as the bug seems to be in 8.3.17 too, as I mentioned before, I chose 8.3.16), so I can't test with a running project. Is it possible to get the fix in 8.3.18?

dmitrijkiltau:
@Sebobo the same happens in 9.1.4, but only with one of our extensions: An interactive map, where the markers can be edited as nodes (they open the node tree when clicking the marker on the map and bidirectionally open the marker popup, when clicking inside the node tree).

At first, I thought that it might be caused by the JS of our extension, but after removing all the JS, it still happens, therefore it must be a neos-ui issue.

It either duplicates itself with its children or even the parent node, depending on how it is being opened (directly in the node tree, or via clicking the inline editor node.

The duplication bug happens in versions `8.4.1`, `9.1.3` and `9.1.4`.

<img width="207" height="374" alt="Image" src="https://github.com/user-attachments/assets/ef3a9c65-8c62-40fc-8837-b51c18ab0b4d" />

Sebobo:
Yes I experience the issue now too in neos.io.

Sebobo:
[...]

PR #4095, opened 2026-03-16
BUGFIX: Prevent duplicate nodes in structure tree

Sebobo:
**What I did**

Prevent nodes from appearing twice in the structure tree under certain circumstances described in #4071.

Resolves: #4071

**How I did it**

1. Always return fully loaded nodes when requested for structure tree. 
2. Deduplicate childnodes when merging partially loaded nodes in redux store. (This is just for safety in case there are more ways to cause this kind of issue, performance wise this should be fine, as we always need fully loaded content nodes anyway and only some requests only loaded partial nodes before).

**How to verify it**

Set structure tree loading depth to 1:

```yaml
Neos:
  Neos:
    userInterface:
      navigateComponent:
        structureTree:
          loadingDepth: 1
```

Have these two node types:

```yaml
'Neos.Demo:Test.Collection':
  superTypes:
    'Neos.Neos:Content': true
    'Neos.Neos:ContentCollection': true
    'Neos.Demo:Constraint.Content.Column': true
    'Neos.Demo:Constraint.Content.Main': true
  ui:
    label: 'Test collection'
    icon: 'atom'
  constraints:
    nodeTypes:
      '*': false
      'Neos.Demo:Content.ItemForTestCollection': true

'Neos.Demo:Content.ItemForTestCollection':
  superTypes:
    'Neos.Neos:Content': true
  ui:
    label: 'Item for test collection'
    icon: 'atom'
```

With the following Fusion:

```
prototype(Neos.Demo:Test.Collection) < prototype(Neos.Neos:ContentComponent) {
    renderer = Neos.Neos:ContentCollectionRenderer
}
prototype(Neos.Demo:Content.ItemForTestCollection) < prototype(Neos.Fusion:Component) {
    renderer = afx`
        <div>I am a test item</div>
    `
}
```

Create a test collection with a test item. Reload the backend and uncollapse the structure tree. The test item should then appear twice.

### DIFF
--- a/Classes/Controller/BackendServiceController.php
+++ b/Classes/Controller/BackendServiceController.php
@@ -572,11 +572,12 @@ class BackendServiceController extends ActionController
                 );
                 break;
             case 'getForTree':
+                $usage = $finisher['payload']['usage'] ?? 'ALL';
                 $result = $nodeInfoHelper->renderNodes(
                     array_filter($flowQuery->get()),
                     $this->getControllerContext(),
-                    true,
-                    ($finisher['payload']['usage'] ?? 'ALL') !== 'PAGE_TREE',
+                    $usage === 'PAGE_TREE', // We only need minimal data for the document tree
+                    $usage !== 'PAGE_TREE', // Child node are only required for the structure tree
                 );
                 break;
             case 'getForTreeWithParents':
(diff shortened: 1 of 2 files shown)


erkenes (approved):
The fix resolves the problem for me in various projects 👍

Kleisli:
We just noticed the same issue today and this also fixes it in our project. Thank you @Sebobo!

Pingu501:
Encountered the same issue today after upgrading to Neos 8.4 and this does fix it!

skurfuerst:
@Sebobo awesome, verified by @Pingu501 and +1 by reading by me.

If you like, I can take care of the upmerge and UI release. This can also go into 9x, right?

All the best,
Sebastian
PS: Do you think the test failure was a random error or related to this problem? I currently simply tried re-running them...

skurfuerst (approved):

Pingu501 (approved):

PR #4099, opened 2026-03-20
BUGFIX: Partial Publishing node type changes did not work if child nodes were removed (because constraints did not match anymore)

skurfuerst:
The core fix is in https://github.com/neos/neos-development-collection/pull/5767, but this change enables it


Classes/Domain/Model/Changes/Property.php
```diff
@@ -290,7 +290,7 @@ private function handleNodeTypeChange(Node $subject): void
                 $subject->workspaceName,
                 $subject->aggregateId,
                 NodeTypeName::fromString($value),
-                NodeAggregateTypeChangeChildConstraintConflictResolutionStrategy::STRATEGY_DELETE
+                NodeAggregateTypeChangeChildConstraintConflictResolutionStrategy::markWithTag(NeosSubtreeTag::removed())
```
mhsdesign, line 293: todo: Update to Neos.Neos change and use class exists check to prevent Neos ui using api which does not exist in neos in the case the ui is only updated. We currently dont enforce a specific version match

skurfuerst: @mhsdesign good idea, will make this more forgiving. Totally forgot about this change 🙈

skurfuerst:
will merge once CI is green

mhsdesign:
Post merge +1 Thanks for adding the adjustments as requested

PR #4103, opened 2026-03-25
TASK: New 9.1 Drag and drop: Let server feedback determine node removal from the dom

mhsdesign:
With the 9.0 fix https://github.com/neos/neos-ui/pull/4018 we can adjust Neos 9.1 where we currently get the inline drag and drop to function by removing the dom node manually. This is no longer necessary and saves us the todo to actually verify that the server can move the node

How to reproduce: Implement a command hook or ACL which the Neos Ui is unaware of.

**What I did**

**How I did it**

**How to verify it**


Sebobo:
As I said in #4018 we have to verify that the feedback is handled fast enough in real environments, to prevent having the node twice for a while.

PR #4107, opened 2026-03-31
BUGFIX: Prevent type error in `Object.keys()` when node type is no longer existant and also its configuration

mhsdesign:
>  TypeError null is not an object (evaluating 'Object.keys(configuration3)')

Called in `Inspector`

Now instead the ui half loads but it doesnt throw cryptic errors and you should see that a node type is missing;)

image

**What I did**

**How I did it**

**How to verify it**


Sebobo (approved):
Amazing, thx!

PR #4113, opened 2026-04-25
FEATURE: Move workspace selector to primary toolbar

### LINKED ISSUE #3684: Improve Workspace Selector Placement for Enhanced User Experience
### Description

Currently, the workspace selector in the Neos User Interface is nested within the publishing dropdown, leading to a less intuitive user experience. Users have expressed difficulty in locating and accessing the workspace selector, particularly those less familiar with Neos.

### Proposed Change:
Move the workspace selector to the top level of the UI to streamline accessibility and improve overall user experience. Placing it prominently will eliminate the need for users to navigate through the publishing dropdown, making it more intuitive for both new and existing Neos users.

Screenshot 2024-01-11 at 20 10 01

### Expected Benefits:

* Simplified access to the workspace selector.
* Improved user navigation and efficiency.
* Enhanced usability for users of varying familiarity with Neos.

### Additional Information:
We've identified an opportunity to enhance the Neos User Interface (UI) by making it more consistent and user-friendly. The proposed changes include restructuring the primary and secondary toolbars, moving the workspace selector to the top level, and incorporating feedback from users, such as Daniel Kestler. So this is just one step to improve the UX of Neos.
Inspiration comes from the discussions in the following tickets.

https://github.com/neos/neos-ui/issues/3247
https://github.com/neos/neos-ui/issues/3803

Sebobo:
With this change the workspace selector is now part of the primary toolbar and not hidden in the publish dropdown anymore.

Resolves: #3684

https://github.com/user-attachments/assets/e3d9d8cc-fbe9-4d90-a29c-84b4c7c314f4

Clean workspace:

CleanShot 2026-06-12 at 09 31 00@2x

CleanShot 2026-06-12 at 09 31 07@2x

Dirty workspace:

CleanShot 2026-06-12 at 09 30 51@2x

With sync button:

CleanShot 2026-06-04 at 15 09 53


packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/index.js
```diff
@@ -211,13 +197,13 @@ export default class PublishDropDown extends PureComponent {
         }
 
         if (isPublishing) {
-            return translate('Neos.Neos:Main:publishTo', 'Publish to {0}', [baseWorkspaceTitle]) + ' ...';
```
markusguenther, line 214: This label can be removed in another PR. Seems that it is not used somewhere else in the Neos-UI or Workspace module.

markusguenther (approved):
Thanks it looks gorgeous :)

packages/neos-ui/src/Containers/PrimaryToolbar/WorkspaceSelector/index.js
```diff
@@ -0,0 +1,89 @@
...
+            {anyWorkspacesAvailable ? (
+                <SelectBox
+                    className={style.selectBox}
+                    options={workspacesOptions}
+                    value={baseWorkspace}
+                    onValueChange={onWorkspaceSelect}
+                    disabled={!changingWorkspaceAllowed}
+                    headerIcon="layer-group"
+                />
+            ) : ''}
```
so-grimm: If an editor is on a shared workspace and then their permissions to use that workspace get taken away, the current selection in the workspace dropdown is empty, which is fine. If however that was the only other workspace the editor had access to and now the editor only has access to the live workspace, the dropdown disappears. The editor still has their base workspace set to the shared workspace and is now unable to switch their base workspace to live and unable to publish.

Image

Sebobo: Yes can you create a new ui issue as we need some initial check in the ui if the current state is invalid. This also affects 9.0 and 9.1 and I'm not even completely sure yet how we solve both cases, as we need to rebase the user workspace forcefully.

Sebobo: This problem is also unrelated to the movement of the workspace selector but we discussed it to also mention it here.

Sebobo: Readonly workspace issue fixed via https://github.com/neos/neos-ui/pull/4134

Sebobo:
Decided to also enable the filter for the workspace selector and improve indication when the current workspace is readonly which works best with #4134 .

mhsdesign:
looks great - i like that the publish button still kinda groups it via the orange border <3

during writing anything this is still too much of a lightshow and looks buggy:

https://github.com/user-attachments/assets/c6efa9c6-b92a-4de4-8e00-45cd4a56303a

i think we should prevent any content shift by:
- not changing the label
- just adding the spinner and then showing the dropdown again
- letting the dropdown not light up as not disabled

mhsdesign:
Then regarding the icon. Neos 9.2 should contain upon a uniform choice for workspaces no? so its either

dropdown

image

(Also the IMO the "review chances" option should always be in the menu and just disabled if no changes instead GONE)

workspace module

image

mhsdesign:
I like the idea of the new searchbox ... but it behaves horribly on safari here in the top :joy:

... im not sure we make ourselves any friends like that:D

https://github.com/user-attachments/assets/217ad6b9-885e-4562-81a5-058c4f03141b

problems in quick testing:
- click does not close the dropdown / to toggle
- typing a search term and removing it fully closes also the dropdown fully instead of showing all options again
- the search box has odd content shift upon opening it
  - and its options are only repositioned once they are hovered via mouse
  - or after some random decision when the term matches and there is only one or two entries left

If the search time matches nothing and the select box is closed by clicking outside it just shows

image

Also the filter icon is weird here? I like to keep the pretty new workspace icon. Maybe a stacked icon but not replace the workspace icon.?

i think the problems are caused as the selectbox is mounted right aligned ... it likes to float right instead ... it doesnt work up there ... we need to use a native new written selectbox using css popover and anchors? wdyt? I think its not worth fixing the broken react repositioning stuff - also it still uses portals and z-indexes ... what can go wrong.

mhsdesign:
In the workspace module or in the code we never refer to "public" or "internal" workspaces ... i just wanted to point this out 

neos ui users are just known by the workspace module that workspaces "exist" - though in that list the "live" workspace does not exist. Also they know they can create shared or private workspaces.

In the code we have these neos classifications

PERSONAL
SHARED
ROOT

and additionally via the workspace permission the feature to have a workspace readonly or not ... but that can for example also be the "live" workspace ... so "live" can be both in the readonly and public group 🤦 

we just need to communicate one thing clearly for the editors ... what is in the code is obviously uninteresting

mhsdesign:
Wenn man die base workspaces zu schnell hintereinander ändert bekommt man gratis irgendwie ne endlos loop hin die den neos server durchgängig mit anfragen zuballert:

image

```
[Wed Jun 10 23:43:23 2026] 127.0.0.1:58727 Accepted
[Wed Jun 10 23:43:23 2026] 127.0.0.1:58727 Closing
[Wed Jun 10 23:43:23 2026] 127.0.0.1:58729 Accepted
[Wed Jun 10 23:43:23 2026] 127.0.0.1:58729 Closing
[Wed Jun 10 23:43:23 2026] 127.0.0.1:58731 Accepted
[Wed Jun 10 23:43:23 2026] 127.0.0.1:58731 Closing
[Wed Jun 10 23:43:23 2026] 127.0.0.1:58733 Accepted
[Wed Jun 10 23:43:23 2026] 127.0.0.1:58733 Closing
[Wed Jun 10 23:43:23 2026] 127.0.0.1:58735 Accepted
[Wed Jun 10 23:43:24 2026] 127.0.0.1:58735 Closing
[Wed Jun 10 23:43:24 2026] 127.0.0.1:58737 Accepted
[Wed Jun 10 23:43:24 2026] 127.0.0.1:58737 Closing
[Wed Jun 10 23:43:24 2026] 127.0.0.1:58739 Accepted
[Wed Jun 10 23:43:24 2026] 127.0.0.1:58739 Closing
[Wed Jun 10 23:43:24 2026] 127.0.0.1:58741 Accepted
[Wed Jun 10 23:43:24 2026] 127.0.0.1:58741 Closing
[Wed Jun 10 23:43:24 2026] 127.0.0.1:58743 Accepted
[Wed Jun 10 23:43:26 2026] 127.0.0.1:58765 Closing
[Wed Jun 10 23:43:26 2026] 127.0.0.1:58767 Accepted
[Wed Jun 10 23:43:26 2026] 127.0.0.1:58767 Closing
[Wed Jun 10 23:43:26 2026] 127.0.0.1:58769 Accepted
[Wed Jun 10 23:43:26 2026] 127.0.0.1:58769 Closing
[Wed Jun 10 23:43:26 2026] 127.0.0.1:58771 Accepted
[Wed Jun 10 23:43:27 2026] 127.0.0.1:58771 Closing
[Wed Jun 10 23:43:27 2026] 127.0.0.1:58773 Accepted
[Wed Jun 10 23:43:27 2026] 127.0.0.1:58773 Closing
[Wed Jun 10 23:43:27 2026] 127.0.0.1:58775 Accepted
[Wed Jun 10 23:43:27 2026] 127.0.0.1:58775 Closing
[Wed Jun 10 23:43:27 2026] 127.0.0.1:58777 Accepted
[Wed Jun 10 23:43:27 2026] 127.0.0.1:58777 Closing
[Wed Jun 10 23:43:27 2026] 127.0.0.1:58779 Accepted
[Wed Jun 10 23:43:27 2026] 127.0.0.1:58779 Closing
[Wed Jun 10 23:43:27 2026] 127.0.0.1:58781 Accepted
[Wed Jun 10 23:43:27 2026] 127.0.0.1:58781 Closing
```

bis der tab geschlossen wird

Sebobo:
> i think we should prevent any content shift by:
> 
> not changing the label
> just adding the spinner and then showing the dropdown again
> letting the dropdown not light up as not disabled

I fixed the "light up" which was a bug that was always there, as the state was calculated incorrectly but never visible before.
The labels and spinner if the publish dropdown I would keep for now and improve afterwards as they are not fully related.
I also adjusted the min-width to prevent most of the jumping in the primary toolbar.

> * click does not close the dropdown / to toggle

Will be fixed with https://github.com/neos/neos-ui/pull/4139

Sebobo:
I think I fixed all important issues. 
* 1 bug and 1 change in the selectbox
* Adjusted the labels how you suggested. 
* Fixed size for the dropdown.
* Icon fix in https://github.com/neos/neos-development-collection/pull/5866.

mhsdesign:
thank you so much <3 especially for removing the content shifts everywhere!!!

- the current selected select box option is twice made darker, once by `istPreviewElement--isDisabled` which sets it to `#999` instead of `#fff` and on top there is an opacity 50%. That now results int this being darker than even the icon and fully unreadable.

expected

image

is

image

- the workspace module icon needs to be adjusted separately: https://github.com/neos/neos-ui/pull/4113#issuecomment-4674674236

- still not sure regarding a clear naming scheme but we can run with yours for now and also adjust them or regroup in the future:) https://github.com/neos/neos-ui/pull/4113#issuecomment-4674826071

- i cannot reproduce https://github.com/neos/neos-ui/pull/4113#issuecomment-4674854003 anymore maybe this was fixed but should probably be solved by adding a blocking dialog and disabling the selectbox in the future

mhsdesign (approved):
Pre-approve but needs the disabled state color fix:)
... and obviously CI ...

https://github.com/neos/neos-ui/pull/4139 needs to be merged and this rebased onto the upmerge for a clean history

Sebobo:
@mhsdesign I fixed and extended the preview badge test and overhauled the design. I updated the screenshots (except the sync button one).

And I added titles!

kdambekalns:
I hopefully fixed the linting error. The e2e test seems to be too slow, is that all there is to the failure?

mhsdesign:
ABSOLUTLELY 

LOVE ❤️ 

IT

image

<3

mhsdesign:
Im brewing together an installable 9.2 beta would be nice to have this also soon included ... https://github.com/neos/Neos.NeosIo/issues/725

I have not tried to run the e2e tests locally ... or debugged this further if they pass on 9.2-dev ...

mhsdesign:
Possible further improvements:

- fix #4148 
- adjust label "Published", especially german "Veröffentlicht" to be different in review workspaces!
- avoid further jumping and allow smaller size and avoid losing current context by adding select box as dropdown on opening


TODO
- add spacing to left so that the other not so important things like language and edit mode are separated

**EDIT:** Issue to discuss followups https://github.com/neos/neos-ui/issues/4153

packages/neos-ui/src/Containers/PrimaryToolbar/WorkspaceSelector/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+    const workspacesOptions = useMemo(() => Object.keys(allowedWorkspaces).map((workspaceName) => {
+        const workspace = allowedWorkspaces[workspaceName];
+        if (!workspace) {
+            return {
+                label: workspaceName,
+                value: workspaceName,
+                group: '',
+                icon: 'x-mark'
+            };
+        }
```
mhsdesign, line 70: these lines are obsolete by definition as we loop over the keys of `allowedWorkspaces` and thus the key exists.

To make the linter happy i usually use Object.entries.map

packages/neos-ui/src/Containers/PrimaryToolbar/WorkspaceSelector/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+        [style['workspaceSelector--isDirty']]: hasUnpublishedNodes,
+        [style['workspaceSelector--isReadOnly']]: isWorkspaceReadOnly
+    });
+
+    const title = changingWorkspaceAllowed ?
+        translate('Neos.Neos.Ui:Main:workspaceSelectorTitle', 'Select target workspace') :
+        translate('Neos.Neos.Ui:Main:workspaceSelectorTitleDisabled', 'Cannot change target workspace while there are unpublished changes');
+
+    return (<div className={classNames} title={title}>
+        {anyWorkspacesAvailable ? (
```
mhsdesign, line 100: this check `Object.keys(allowedWorkspaces).length > 1` cannot happen as we always have the current base workspace i think.
The design even would look broken still as we wrap the outer `div` which claims an empty space but again this case cannot happen. Also a warning would be rather required then?

Maybe you wanted to hide this dropdown if there is no workspace available to switch?

mhsdesign (commented):
During trying to refactor the code to use the native popover api for the dropdown and stuff i noticed two code paths which make no sense as of now

packages/neos-ui/src/Containers/PrimaryToolbar/WorkspaceSelector/index.tsx
Sebobo: Nope, we have cases where the list is empty. Of course this shouldn't happen, but it does due to workspace ACL and one more case I forgot. That's why I said the UI should make a sanity check and redirect on error. Same for invalid dimensions etc.

packages/neos-ui/src/Containers/PrimaryToolbar/WorkspaceSelector/index.tsx
Sebobo: See below

packages/neos-ui/src/Containers/PrimaryToolbar/WorkspaceSelector/index.tsx
mhsdesign: Okay where it was fully empty with your pr or the whole backend crashed because of the PHP in the WorkspaceHelper

image


my change now shows the technical name of the workspace but the dropdown is empty:

image

i think its now good enough for 9.2

mhsdesign: or rather i see that this PR had a bug that this caused all installations with just one workspace e.g. live to totally blackout

image


but thats fixed now #4158 

image

Issue #4112, opened 2026-04-25
EPIC: E2E Test Refactoring


JamesAlias:
I think it's time to ditch TestCafe and move to Playwright.

With Playwright we could also remove the SauceLabs dependency - fixing the "shared API key" problem.

Despite a lot of (relatively easy) work regarding the rewrite from TestCafe to Playwright APIs we would need to rethink the tests that are implemented with the `testcafe-react-selectors`. My guess would be using `data-test-id` if stable selectors are not achievable.

I believe that fixing and improving the test setup will help the team and contributors. 
Especially @mhsdesign and @Sebobo who have to _manually_ test the build before merging/releasing.

mhsdesign:
Thank you for bringing this up as a discussion here;)

A few points from the top of my head:

1. I have thought of using Playwright as well but with the introduction of native CDP in playwright i was consolidated that testcafe is not dead. And CDP promises native automation which is just as good as in playwright. But we dont use that in CI as we dont get chromium to work and firefox does not support that (yet? -> or last time) https://github.com/neos/neos-ui/issues/3591

2. Testcafe is often really slow to merge and release bugfixes that allow to work on newer firefox versions or to set default flags for chromium to ensure the browser does not open any popups like "allow detecting other devices in network"
-> that sucks and we default now to pinning firefox to an old version and during local development i often need to research which flags i have to pass to chromium to behave vanilla.

3. We use SauceLabs at our own choice so we can easily access screen recordings (though _easy_ is an understatement -> you have to find the right url). That comes at the cost that SauceLabs only allows a limited amount of parallel runs which is quickly exceeded during upmerges or heavy work before an release which results in failing CIs. Also there is a second cost: We have to way to share the API key and thus disable E2E tests in fork prs which makes it even less attractive to write tests: https://github.com/neos/neos-ui/pull/4091

4. Currently the tests fail on 8.x on the old CircleCi even-though they pass locally

5. Currently the tests fail on 9.x even-though they pass locally. With https://github.com/neos/neos-ui/issues/4026 i refactored to use Github actions instead of CircleCi and by that removed complexity and speed up the tests a lot.

> Now the e2e testing is decoupled from the other steps and runs in parallel and finishes in under 10 minutes.

That refactoring from last October was already lots of work and frustrating:D And even more frustrating to see now that there is still flakiness - or rather always false positives. Maybe this can be resolved by using another Firefox or Chrome...

[...]

Issue #3380, opened 2023-02-17
Feature: Add `data-testid` to important places for external custom e2e tests


mhsdesign:
E2E testing neos can be hard sometimes and i lately relied much on css classes like:

```
[class^="_secondaryInspector"]
[class^="_markActiveNodeAsFocused--focusedNode"]
[class^="_leftSideBar__bottom"]
```
which is in my case fairly accurate as we dont rely on the exact hash but use starts with (`^⁼`)

but if you watch https://youtu.be/hzJPLKVarMw?t=1717 youll understand that its challenging ^^

with https://github.com/neos/neos-ui/pull/3366 the above will not continue to work but one woul need to use the less specific: contains string `*=` as the css hash will be prepended to the classname

Sebobo:
So you want to stop using ReactSelectors?

mhsdesign:
no but for my own e2e tests (where I cannot use react selector because we dint enable this for production build)

Sebobo:
Ah so the use case is e2e tests for plugins. OK cool!
Maybe you can adjust the title and description 😉

PR #4115, opened 2026-04-28
EPIC: Playwright BDD E2E Test Setup

JamesAlias:
https://github.com/neos/neos-ui/issues/4112

!!!
Right now the SUT installs a specific version of Neos.
This means when I port this to 9.1 branch I need to update the composer.json to also use Neos 9.1.

This is not a final decision, but I am not sure if guessing the Neos version via target branch is stable enough. I mean the "automatic labeling" workflow does it too, I guess?
Up for discussion!
!!!

**What I did**

**How I did it**

**How to verify it**

1. checkout this branch
2. setup Neos Ui (`make setup && make build`)
2. `cd Tests/E2E`
3. `make setup && make test`


JamesAlias:
@Sebobo @mhsdesign 
Please have a look and try the test setup locally.

I am in the process of resolving some cosmetic issues in the Github workflow, but the tests themselves are ready to roll.

mhsdesign:
Thank you so much on working on this ill try to get them running ... i work locally and just mount the Neos Ui testing packages back into my main dev distribution... that should still work with your new system_under_test thing.

My question would be:
- why is there so much low level configuration e.g. some php inis etc that seems utterly hard to maintain in the long run, we dont have these things now ... but maybe they are just hidden in ci
- are these tests considered a full replacement to testcafe?
  - ... if so we should remove testcafe and blend the change together
- sometimes we do odd things during testsetup like modifying the configuration of the system during testing. I think this should probably still work.

ping also @markusguenther

JamesAlias:
> * why is there so much low level configuration e.g. some php inis etc that seems utterly hard to maintain in the long run, we dont have these things now ... but maybe they are just hidden in ci

It should be ok. If not, we can always modify the SUT.
I will remove the PHP ini (needed for vips) and remove all the vips related stuff - the SUT uses gd.

> * are these tests considered a full replacement to testcafe?

Yes. At least for all 9.x branches. I will later look into porting to Neos 8.x.

>   * ... if so we should remove testcafe and blend the change together

I'd say we keep both for a while and if everything is to our satisfaction, we remove TestCafe, CircleCI, SauceLabs from all 9.x branches.

There are, however, tests that are currently deactivated in TestCafe that I did not port yet.
I will look into that after the setup is merged.

> * sometimes we do odd things during testsetup like modifying the configuration of the system during testing. I think this should probably still work.

Seems to work.
We can also run the SUT with different Flow Contexts to enable different Configuration combinations.

> ping also @markusguenther

Yes :) @markusguenther

mhsdesign:
Okay wow i just saw what playwright offers for debugging. Its definitely convincing - way more convincing than tescafe, but maybe we just use it wrong^^

in the artifact upload step following is uploaded
- a simple html file with an embedded react application to view and filter all tests and their execution times
- dedicated tracing to view in the browser
- for each failed step:
  - a video of the failure
  - in a zip, all possible informations. Each HTML snapshot, each server response - everything

The only thing i would also wish for is that the logs and exceptions of the SUT are also provided as download, possibly in the same zip.

So yes id say this is more than sufficient to replace saucelabs. Immediately?!

----------


I do really like the idea that  `createUser` for example directly issues a `flow user:create` command. That makes the test more encapsulated. Also in the future we could use CLI commands to quickly create certain states of nodes that are to be tested. This is easier than the current import export way.

That also means the hack "WriteAdditionalSettingsController" can be exposed as CLI controller instead which is IMO more stable and a better way to interact with the SUT than HTTP. Now the question is, is the docker file system mutable 😅 -> We do not use subcontext's here currently because there would be too many ^^ There are so many configurations we like to tests.

Ill just have to see if i can via ENV variable ensure that my local flow installation is used. I use that heavily when modifying Neos.Neos in sync with the Ui.

-----------

Thank you for removing vips again then, id say this is just additional complexity. And from complexity we have enough already ^^. On that note:

Haha i see you really build us an enterprise Neos ready to ship:D Thanks for doing that and also adding redis but in case it stops working $somehow and is to complex to maintain we can also dial back to file caches right? But for now it works jippy:D

Also as webserver now frankenphp is used in combination with caddyfile im not familiar with those. But i see the need for something the like. Previously we did use `flow server:run` which is limited and does not support HTTPS. Am i correct that using frankenphp is the simples way then to have HTTPS which can enable clipboard testing in the future? Or what are your thoughts @markusguenther @Sebobo in the end we need to maintain and understand this too.

Just for consistency and also again to minimise complexity we should reduce tools if possible or use the ones already used. For example i found that you use a dedicated js package installation for playwright but use npm instead of yarn. Npm is obviously always existing but nevertheless we should decide on one. And that would be yarn still? (open for discussion)

Also the docker image is cached, does the caching respect changes in the SUT Neos packages? E.g. adding a new nodeType or updating the composer json?

As a next step we should also remove references to `[class=*` and use stable test ids as proposed in https://github.com/neos/neos-ui/issues/3380

---------

mhsdesign:
Okay i had a look at frankenphp (again i think its a bit familiar) and it does seem like a great tool especially as its just one binary. So nothing hard to maintain -> even simpler than the previous flow server run which took me quite some time to get functioning ^^

Also as i argued for less complexity, i forgot that we already have the complexity with a docker stack in regard of the `make test-e2e-docker` as this will become obsolete (correct?) This is definitely the right way to proceed as this ensures we use the docker setup actually during testing and developers can be sure it works locally.

JamesAlias:
> The only thing i would also wish for is that the logs and exceptions of the SUT are also provided as download, possibly in the same zip.

Good idea. This is absolutely doable. "Follow up" or "change request"?

> Now the question is, is the docker file system mutable 😅 -> We do not use subcontext's here currently because there would be too many ^^ There are so many configurations we like to tests.

I will share a similar test setup I am working on for our 2FA package, when it's ready. There I test some different Configuration combinations. We can then discuss, if we can and want to do this for the neos tests as well.

> Haha i see you really build us an enterprise Neos ready to ship:D Thanks for doing that and also adding redis but in case it stops working $somehow and is to complex to maintain we can also dial back to file caches right?

Again. We can modify the SUT however we want - as long as it's able to run Neos of course.

> Am i correct that using frankenphp is the simples way then to have HTTPS which can enable clipboard testing in the future?

FrankenPHP uses a Caddy webserver. The, now legacy, way for us at Sandstorm to run Neos has been an Nginx server and FPM. With Caddy we get local HTTPS so yes: Features that require HTTPS are now ready to be tested 🚀 

Caddy is also very low maintenance (experience until the time of writing).

> npm instead of yarn

I can change it to yarn, if this is an issue. I just find myself reaching for the OG more often lately - even more if I do not need features that the likes of yarn or pnpm provide.

> Also the docker image is cached, does the caching respect changes in the SUT Neos packages? E.g. adding a new nodeType or updating the composer json?

My assumption is, that the Docker image will be re-build if dependencies change and the files in `sut_file_system_overrides` _are_ dependencies.

Update: I verified this assumption ✅ 

> As a next step we should also remove references to [class=* and use stable test ids as proposed in https://github.com/neos/neos-ui/issues/3380

Absolutely! It's on the list, but i wanted to try to touch as little as possible outside the test setup for the migration.

Well, _Actually_ I added one (1) testId to show that it works. 😬 

> Also as i argued for less complexity, i forgot that we already have the complexity with a docker stack in regard of the make test-e2e-docker as this will become obsolete (correct?)

Correct.

> This is definitely the right way to proceed as this ensures we use the docker setup actually during testing and developers can be sure it works locally.

Yes! The best part: The SUT can also be used for development - even test driven.

I am in the process of pushing this kind of test setup into our Neos Packages - for me it helps tremendously to have a simple way to run a Neos instance to develop a package without piggy-backing or side-loading it into another Project.

Tests/E2E/steps/hooks.ts
```diff
@@ -0,0 +1,11 @@
...
+import {logout, removeAllUsers} from "../helpers/system";
+
+const {AfterScenario} = createBdd();
+
+// cleanup for each scenario
+AfterScenario(async ({page}) => {
+    await logout(page);
+
+    removeAllUsers();
+});
```
markusguenther, line 11:
```suggestion
AfterScenario(async ({page}) => {
    await logout(page);

    try {
        removeAllUsers();
    } catch (err) {
        console.warn("[AfterScenario] removeAllUsers failed — next scenario may see leftover users:", err);
    }
});
```

Tests/E2E/steps/navigation.steps.ts
```diff
@@ -0,0 +1,27 @@
...
+const {When} = createBdd();
+
+When("I navigate to the {string} page", async ({page}, pageTitle: string) => {
+    const tree = new NeosTree(page);
+    await tree.nodeLabel(pageTitle).click();
+
+    // Wait for the focused tree node to reflect the navigation. Using DOM observation
+    // (no Redux store access): when a tree node becomes the current document, its
+    // ancestor button gets a class containing "isFocused" applied by the tree.
+    await expect(
```
markusguenther, line 14:
```
await expect(
        page.locator('[role="treeitem"] [role="button"][class*="isFocused"]')
            .filter({hasText: pageTitle}),
    ).toBeVisible();
```

Prevent failing tests after navigating to a node...

Sometimes it happens that the Firefox or WebKit fails in the test 'Discard a multi-node move keeps the moved nodes in the tree after a reload' and the node that we want to navigate to is not focused at the test and we have still the previous node focused. This change will take care that we wait until the timeout is reached or the focus is set.

markusguenther:
What should I say! @JamesAlias that is an awesome improvement of the integration tests for the neos ui. I have tested it locally and spotted just a small issue that occurs sometimes when I had run the tests multiple times in a row. 

I had a discussion with @mhsdesign and @Sebobo and we came to the conclusion that we should remove the TestCafe setup and don’t keep both setups. If something will fail, we can revert it anyways. So I just need to thank you for all the work you put into that topic. I needed to create a new PR to be able to write to the branch. Sorry for that.

markusguenther:
Closes in favor of #4129

Issue #4117, opened 2026-05-12
BUG: Dimension switcher label of selected option is trimmed


Sebobo:
The label width is not correct and so the text is replaced with an ellipsis

<img width="1334" height="668" alt="Image" src="https://github.com/user-attachments/assets/1b5ab0b0-7969-4dcb-8489-8676c8a4f903" />

PR #4118, opened 2026-05-12
BUGFIX: Label of current dimension selection was trimmed

Sebobo:
**What I did**

This fixes a regression that was introduced with the bugfix #4090 that was incompatible with the styles introduced with #3680. Luckily it works again by removing most of those styles :D

Resolves: #4117

**How to verify it**

Before:

Image

After:

CleanShot 2026-05-12 at 09 05 32@2x


dlubitz (approved):

markusguenther (approved):
Seems to work again, thanks for the fix

PR #4126, opened 2026-05-31
Obsolete - Part of Neos.Neos - Add custom rendering for hidden document nodes in preview mode

dlubitz:
Adds additional styling for dedicated screen to show hidden document nodes in preview mode

Is required for: https://github.com/neos/neos-development-collection/pull/5843
Relates to https://github.com/neos/neos-development-collection/issues/5724

### DIFF
--- a/Resources/Private/Fusion/Prototypes/NotVisibleDocument.fusion
+++ b/Resources/Private/Fusion/Prototypes/NotVisibleDocument.fusion
@@ -1,0 +1,18 @@
+prototype(Neos.Neos:Page) {
+    head {
+        notVisibleDocumentCSS = Neos.Fusion:Tag {
+            @position = 'after guestFrameApplication'
+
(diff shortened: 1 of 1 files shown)


Sebobo:
As we keep these kind of rendering in Neos.Neos since 9.0. I moved the change to https://github.com/neos/neos-development-collection/pull/5843/changes so we don't have to change both during upmerge.

mhsdesign:
Ah got it - the pr was still linked - now its incorporated for 8.3 in the new Neos.Neos component BackendInformation.

Neos.Neos is also the correct place as the Neos.Ui rather focuses on the JS backend and should not care about fusion. As long as we think its a good idea to render the shortcut preview and also now the hidden nodes via fusion instead of replacing the iframe with a "proper" more capable javascript react rendering for these cases the code rightfully absolutely belongs to Neos.Neos!

Issue #4125, opened 2026-05-22
Logout is very well hidden


kdambekalns:
I just logged in to Neos, did my work, now want to log out again… but… where? Eventually I open the navigation, but still… no?

<img width="499" height="864" alt="Image" src="https://github.com/user-attachments/assets/1b6930c2-259d-4b0e-ba70-e418d0ed3155" />

Ah, a scrollbar! Now we're talking:

<img width="464" height="277" alt="Image" src="https://github.com/user-attachments/assets/5f6f7434-1506-45d9-a80c-d8df50de11bc" />

So, I guess I need to expand that?

<img width="559" height="283" alt="Image" src="https://github.com/user-attachments/assets/b8c368b6-2466-4c08-a3df-394f6467bff7" />

Ah, I need to scroll yet again:

<img width="594" height="352" alt="Image" src="https://github.com/user-attachments/assets/7aaf9a30-58d4-4157-a0f4-5bf72d671a23" />

Relates: #3740 , neos/neos-development-collection#4934

kdambekalns:
If that "user menu button" would be fixed down left and always visible, that would help. Now make it "expand upwards" when opened and the issue is resolved. 😇

Sebobo:
I would make it sticky to the bottom left and expand from below. Should be a nice issue for the contribution day or sprint

kdambekalns:
> I would make it sticky to the bottom left and expand from below. Should be a nice issue for the contribution day or sprint

Exactly, that's what I meant in my comment. 👍

minedesign-jl:
I would get on this

minedesign-jl:
While making the drawer__bottom sticky we* discussed the position and importance of the version panel and decided to remove the from the drawer and instead add a log. 

*@Sebobo @mhsdesign @ahaeslich

ahaeslich:
thx to @minedesign-jl for fixing this in #4127

PR #4127, opened 2026-06-03
BUGFIX: Make user dropdown in left bottom drawer sticky

minedesign-jl:
**What I did**
Made user drop down in left bottom drawer sticky to ensure logout is always visible regardless of number of backend modules in menu.
Version panel was moved from left bottom drawer to a console.log.

**How to verify it**
With your eyes. 
For longish left drawer Logout should always be reachable while menu list is scrollable.
image

While logged into backend, open browser console & check for Neos.Ui log.
image

Resolves: #4125


ahaeslich (approved):
Checked with my own 👀

Sebobo (approved):
That looks friggin' amazing Jasmin! 🤯

Works as expected, thx!

PR #4129, opened 2026-06-04
EPIC: Playwright BDD E2E Test Setup

markusguenther:
> Followup of https://github.com/neos/neos-ui/pull/4115 as I was not able to write to the branch. For comments and discussions take a look at the PR before.

https://github.com/neos/neos-ui/issues/4112

!!!
Right now the SUT installs a specific version of Neos.
This means when I port this to 9.1 branch I need to update the composer.json to also use Neos 9.1.

This is not a final decision, but I am not sure if guessing the Neos version via target branch is stable enough. I mean the "automatic labeling" workflow does it too, I guess?
Up for discussion!
!!!

**What I did**

**How I did it**

**How to verify it**

1. checkout this branch
2. setup Neos Ui (`make setup && make build`)
2. `cd Tests/E2E`
3. `make setup && make test`


markusguenther:
TestCafe has been removed, and the new tests are running fine for Chrome, Firefox, and WebKit browsers.
Bildschirmfoto 2026-06-04 um 16 47 36

PR #4131, opened 2026-06-04
FEATURE: Sync base workspace from UI

### LINKED ISSUE #4130: FEATURE: Allow syncing base workspace from sync button in primary toolbar
With https://github.com/neos/neos-ui/pull/4113 we moved the workspace dropdown out of the publish dropdown.
Not it also makes sense to help editors understand that their base workspace might also be outdated and needs syncing and also allow to do that.

Sebobo:
Resolves: #4130


Sebobo:
@mhsdesign is this still a hot topic after all your changes regarding the hierarchy layers, etc., or should we close it?

PR #4132, opened 2026-06-04
FEATURE: Upgrade to ckeditor 48.2.0

### LINKED ISSUE #4111: FEATURE: Upgrade CKEditor to version 48
Version 48 mainly brings signifacnt changes to inline tables:
https://ckeditor.com/blog/ckeditor-48-0-0-release-highlights/

It also adds much needed accessibility like scope attributes.

Upgrade guide
https://ckeditor.com/docs/ckeditor5/latest/updating/guides/update-to-48.html

Sebobo:
This change updates yarn to simplify our patch workflow and makes it more reliable and that we are up-to-date again. This is necessary to manage the various patches we need for CKEditor.
Due to deprecations the editor creation and the theme inclusion had to be adjusted.

In addition we now offer enabling and disabling the individual CKEditor table features via the nodetype definition:

```yaml
formatting:
  table: true
table:
  tableColumn: true
  tableRow: true
  mergeTableCells: true
  toggleTableCaption: true
  tableProperties: false
  tableCellProperties: false
```

Just enabling the formatting table feature will enable the above defaults like before. The table and cell property editors require to be activated.

CKEditor update 48.2.0 release notes https://ckeditor.com/docs/ckeditor5/latest/updating/guides/update-to-48.html.

Resolves: #4111


packages/neos-ui-ckeditor5-bindings/src/ckEditorApi.js
```diff
@@ -166,7 +166,12 @@ export const createEditor = () => async options => {
     }
 
     return NeosEditor
-        .create(propertyDomNode, ckEditorConfig)
+        .create({
+            root: {
+                element: propertyDomNode,
+            },
+            ...ckEditorConfig
+        })
```
mhsdesign, line 179: `ckEditorConfig` should not be allowed to override `root`

also we need to check that the `placeholder` support still works ... this is now part of `root`.

.yarnrc.yml
```diff
@@ -34,4 +34,16 @@ patchFolder: ./patches
 plugins:
   - ./legacyDependencies.js
 
+supportedArchitectures:
+  os:
+    - darwin
+    - linux
+    - win32
```
mhsdesign: i dont think we need windows support

mhsdesign (approved):
With heavy heart i accept the new bloat of yarn caches...

> seb and me just rediscussed our yarn cache again. At the moment we use yarn 3 now we update to 4 and all npm package zips are redownloaded and have possibly new content and thus claim new storage. We aggreed to move forward with this but once we move the Neos ui into the Neos dev collection or into a Ui collection we should reevaluate the whole yarn cache idea again and most importantly prune all current zips from our history (see git filter repo) - Maybe git LFS is something we should look into in combination with github artefacts?

Issue #4133, opened 2026-06-06
BUG: It's possible to edit in readonly workspaces, even though one cannot publish


Sebobo:
UI shouldn't allow writing to personal workspace when the base workspace is readonly.
Currently always only the personal workspace permissions are checked, but as the personal
workspace is writable for its owner (the current) user, `readonly` defaults to false.

As the CR allows this state, but it is bad for the UX, we need to check both the personal workspace and the base workspace.

PR #4134, opened 2026-06-06
BUGFIX: Check baseworkspace for write permission

Sebobo:
Resolves: #4133

Switching to a (base)workspace in which the user is only a `viewer` will not disable all editing features.
Switching to a (base)workspace with write permissions works as normal.

### DIFF
--- a/packages/neos-ui-redux-store/src/CR/Workspaces/selectors.ts
+++ b/packages/neos-ui-redux-store/src/CR/Workspaces/selectors.ts
@@ -12,7 +12,8 @@ export const baseWorkspaceSelector = (state: GlobalState) => state?.cr?.workspac
 export const allowedTargetWorkspacesSelector = defaultMemoize((state: GlobalState) => state?.cr?.workspaces?.personalWorkspace?.allowedTargetWorkspaces);
 
 export const isWorkspaceReadOnlySelector = (state: GlobalState) => {
-    return state?.cr?.workspaces?.personalWorkspace?.readOnly || false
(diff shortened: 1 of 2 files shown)


packages/neos-ui-redux-store/src/CR/Workspaces/selectors.ts
```diff
@@ -12,7 +12,8 @@ export const baseWorkspaceSelector = (state: GlobalState) => state?.cr?.workspac
 export const allowedTargetWorkspacesSelector = defaultMemoize((state: GlobalState) => state?.cr?.workspaces?.personalWorkspace?.allowedTargetWorkspaces);
 
 export const isWorkspaceReadOnlySelector = (state: GlobalState) => {
-    return state?.cr?.workspaces?.personalWorkspace?.readOnly || false
+    const personalWorkspace = state?.cr?.workspaces?.personalWorkspace;
+    return !personalWorkspace || personalWorkspace.readOnly || personalWorkspace.allowedTargetWorkspaces[personalWorkspace.baseWorkspace]?.readonly;
```
mhsdesign: i wondered why this is not a bugfix for 8.3 but it seems that

```
\Neos\Neos\Ui\Fusion\Helper\WorkspaceHelper::getPersonalWorkspace()
```

changed the behaviour in 9.0 to return always `false` for `readOnly`. Instead we should check here the base workspace.

Sebobo: Ok I adjusted the php part and reverted part of this change.

mhsdesign (approved):
THanks

PR #4136, opened 2026-06-09
BUGFIX: Simplify UpdateWorkspaceInfo

kitsunet:
Refactor `UpdateWorkspaceInfo` to accept the workspace via constructor instead of a setter, eliminating the possibility of using the class without a workspace set.

Extract the user workspace name directly from context paths on publish and discard, avoiding an unnecessary node lookup that could fail when the node no longer exists in context.

Add a guard to prevent accidentally passing a base workspace to `UpdateWorkspaceInfo`.

### DIFF
--- a/Classes/Controller/BackendServiceController.php
+++ b/Classes/Controller/BackendServiceController.php
@@ -15,7 +15,9 @@ namespace Neos\Neos\Ui\Controller;
 
 use Exception;
 use Neos\ContentRepository\Domain\Model\NodeInterface;
+use Neos\ContentRepository\Domain\Model\Workspace;
 use Neos\ContentRepository\Domain\Repository\WorkspaceRepository;
+use Neos\ContentRepository\Domain\Utility\NodePaths;
 use Neos\Eel\FlowQuery\FlowQuery;
 use Neos\Flow\Annotations as Flow;
 use Neos\Flow\I18n\Exception\InvalidLocaleIdentifierException;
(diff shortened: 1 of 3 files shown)


Classes/Domain/Model/Feedback/Operations/UpdateWorkspaceInfo.php
```diff
@@ -38,15 +38,12 @@ class UpdateWorkspaceInfo extends AbstractFeedback
...
-     * Set the workspace
-     *
-     * @param Workspace $workspace
-     * @return void
-     */
-    public function setWorkspace(Workspace $workspace)
+    public function __construct(Workspace $workspace)
     {
         $this->workspace = $workspace;
+        if ($workspace->getBaseWorkspace() === null) {
```
kdambekalns, line 43: I'd move this before the assignment, but only because it seems more logical to reject an invalid value first.

kdambekalns (approved):

Classes/Controller/BackendServiceController.php
```diff
@@ -604,4 +599,31 @@ public function generateUriPathSegmentAction(NodeInterface $contextNode, string
...
+
+    /**
+     * Extract which workspace we are operating on in publish and discard requests. This should be a user workspace so
+     * we limit the results to that. There shouldn't be more than a single result, for now though we do not throw an
+     * exception if that happens but use the first.
+     * Empty string return means no user workspace was found.
+     *
+     * @param array<int,string> $contextPaths
+     */
+    protected function extractUserWorkspaceNameFromContextPaths(array $contextPaths): string
```
mhsdesign, line 611: because its a bugfix its oke i think^^ this will find no application in Neos 9 because we dont have `user-` anymore and also i hope that we dont need it?  Can you please also take care of upmerging this? The workspace in constructor makes also sense for 9.0

kitsunet: exactly!

kitsunet: I will see on upmerge how this is handled in 9, but hopefully better already :D

dlubitz:
FYI: This is a breaking change (for 8.3+) with adding a mandatory parameter to `UpdateWorkspaceInfo` constructor.
https://github.com/neos/neos-ui/commit/dee1c2dc347e55a05f5105c191b0614fc838cdb2#r198391330

mhsdesign:
The whole PHP codebase of the Neos Ui is internal. It was unfortunately never declared before Neos 9.0 and https://github.com/neos/neos-ui/pull/3704

What was your usecase?

dlubitz:
Notify the UI about changes made during translation of nodes.

dlubitz:
Same issue was here: https://github.com/sitegeist/Sitegeist.LostInTranslation/pull/104

kitsunet:
Right, I think that is unfortunate. That said, our rule originally was if it's not marked api, it is not meant for  public consumption and could change, with narrow exceptions due to long time "habits/customs". 

I think as soon as someone starts using stuff externally, it makes sense to request changes in core to make it API or suggest a sensible API. This code was always brittle.

SORRY for breaking <3

Issue #4137, opened 2026-06-10
BUG: UI crashes after clicking after discard back into inline editable text


mhsdesign:
It seems the new 9.1 breadcrumb triggers

```
    const closestDocumentNodeInParentLineIndex = focusedNodeParentLine
        .findIndex((node) => nodeTypesRegistry.hasRole(node.nodeType, 'document'));
```

a beautiful 

<img width="475" height="208" alt="Image" src="https://github.com/user-attachments/assets/0c8b7460-9f16-48cb-ab8b-3ce65332c04d" />

also following things are logged

<img width="232" height="46" alt="Image" src="https://github.com/user-attachments/assets/c7498395-9d13-4e57-88e4-5c983e1e69e9" />

so something is wrong after discarding ...

mhsdesign:
https://github.com/neos/neos-ui/pull/4138 fixes that though there is still a little bit of a weird in between state

https://github.com/user-attachments/assets/c867285d-cde9-4d18-9160-748fe1fc45ac

Sebobo:
I think we need a separate issue for fixing the state itself. The breadcrumb is just displaying what's there.

PR #4138, opened 2026-06-11
BUGFIX: Prevent errors in breadcrumb when a node is inaccessible

Sebobo:
This change makes sure that the breadcrumb can handle temporary state changes during discard better.
If the focused node is invalid, the breadcrumb is not rendered.
If a node in the path is invalid, we now show a question mark.

Resolves: #4137

### DIFF
--- a/packages/neos-ui/src/Containers/Breadcrumb/index.tsx
+++ b/packages/neos-ui/src/Containers/Breadcrumb/index.tsx
@@ -45,20 +45,32 @@ const Breadcrumb: React.FC<{
     }, [focusNode, focusedNode]);
 
     const closestDocumentNodeInParentLineIndex = focusedNodeParentLine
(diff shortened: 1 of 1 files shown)


mhsdesign (approved):
Thanks by reading ... not tested ... but you did? I guess .... then merge<3

Sebobo:
> Thanks by reading ... not tested ... but you did? I guess .... then merge<3

I wasn't really sure how to reproduce the error I just added safeguards, as I suspect the error lying somewhere else. But we anyway have to handle weird node states in all kind of places currently until we refactor the whole thing.

mhsdesign:
ill test it later then

mhsdesign:
before merge! :)

mhsdesign (approved):
Now also tested: 

https://github.com/neos/neos-ui/issues/4137#issuecomment-4680588590

PR #4139, opened 2026-06-11
BUGFIX: Allow toggling selectbox with searchbox and value

Sebobo:
Before this change the selectbox accepted the click event on the filter box and prevent toggling with the value change which was „random“.

Now the filter prevents bubbling of the click and the toggle is not triggered.

### DIFF
--- a/packages/react-ui-components/src/SelectBox/selectBox.js
+++ b/packages/react-ui-components/src/SelectBox/selectBox.js
@@ -326,19 +326,15 @@ export default class SelectBox extends PureComponent {
     }
 
     handleToggleExpanded = () => {
-        // Return earyl if disabled
+        // Return early if disabled
         if (this.props.disabled) {
             return;
         }
 
         let isExpanded;
         if (this.props.displaySearchBox) {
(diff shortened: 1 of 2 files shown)


mhsdesign (approved):
Tested in combination with https://github.com/neos/neos-ui/pull/4113/

mhsdesign:
See video https://github.com/neos/neos-ui/pull/4113#issuecomment-4674747938

what was weird before

mhsdesign:
Unfortunately this causes a regression in all other parts. The select box no longer opens when clicked into

image

expected

image


This causes also the e2e tests to fail. Lets revert @Sebobo? Also we dont need this change anymore with my new workspace selector

Sebobo:
> Unfortunately this causes a regression in all other parts. The select box no longer opens when clicked into
> 
> image
> expected
> 
> image
> This causes also the e2e tests to fail. Lets revert @Sebobo? Also we dont need this change anymore with my new workspace selector

No, let me check and adjust the fix. This is still relevant for plugins and custom usages.

PR #4141, opened 2026-06-14
BUGFIX: Remove extra <ul> from react ui component SelectBox

### LINKED ISSUE #3826: BUG: SelectBox generates invalid markup
There is a least one occasion of the select box generating directly nested `<ul>`-tags, which is invalid markup.

I noticed it in the select box for node tree presets:

Screenshot_2024-07-11_20-00-32 SelectBox vis

Screenshot_2024-07-11_20-00-04 SelectBox markup

There might be other occasions yet.

Affected Versions:
Neos UI: 8.3

minedesign-jl:
Resolves: #3826

**What I did and how**
* selectBox react ui component: fixing invalid double `<ul>` wrap
* editPreviewDropdown: fixing invalid markup - `<div>`is direct child of `<ul>`
* same icon spacing in editPreviewDropdown, dimensionDropdown & userDropdown

**How I did it**
* selectBox: removing `<ul>` from SelectBox.js - it uses drowdown.contents which already returns a `<ul>`
* editPreviewDropdown: wrapping groupHeader & it's options in `<li>`and options in own `<ul>`
* icon spacing: setting spacing to `var(--spacing-Half)`

**How to verify it**
checking for proper list markup for the following dropdowns

* node tree presets
* workspace dropdown
* dimension dropdown
* view/preview dropdown
* node type dropdown (inspector)
* open graph type (inspector)
* twitter card - card type (inspector)
* xml sitemap (inspector)

checking that dropdown ui component still produces valid markup by checking user dropdown.

### DIFF
--- a/packages/neos-ui/src/Containers/PrimaryToolbar/EditPreviewDropDown/index.js
+++ b/packages/neos-ui/src/Containers/PrimaryToolbar/EditPreviewDropDown/index.js
@@ -70,27 +70,31 @@ export default class EditPreviewModeDropDown extends PureComponent {
                         <span className={style.dropDown__currentEditMode}><I18n id={currentEditMode.title}/></span>
                     </DropDown.Header>
                     <DropDown.Contents className={style.dropDown__contents}>
-                        <div className={style.dropDown__groupHeader}>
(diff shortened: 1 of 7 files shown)


mhsdesign:
I would be more confident targeting 9.2 who knows what demons this can have^^ 8.3 is super stable now and soon to be handed of in security only. Wdyt?

PR #4143, opened 2026-06-23
BUGFIX: Icon not part of multiline option grid

Sebobo:
The multiline option layout had a few quirks and bugs and the icon was in the wrong position.
Also added a big of highlighting on hover.

Before:

CleanShot 2026-06-23 at 18 26 08@2x

After:

CleanShot 2026-06-23 at 18 20 19@2x

With hover but no icons:

CleanShot 2026-06-23 at 18 31 50@2x


mhsdesign (approved):
I trust the screenshots <3

Issue #4146, opened 2026-07-01
BUG: Base workspace change does not redirect if on document that does not exist in new base


mhsdesign:
Instead nothing happens.

```
Neos_Neos_Ui_Controller_BackendServiceController.php: assert($documentNodeInstance !== null)

59 assert(false, "assert($documentNodeInstance !== null)")
58 Neos\Neos\Ui\Controller\BackendServiceController_Original::changeBaseWorkspaceAction("live", "{"contentRepositoryId":"default","workspaceName":"…gregateId":"44328ad4-6f5a-4c54-ade4-3302887ee249"}")
57 Neos\Neos\Ui\Controller\BackendServiceController::changeBaseWorkspaceAction("live", "{"contentRepositoryId":"default","workspaceName":"…gregateId":"44328ad4-6f5a-4c54-ade4-3302887ee249"}")
```

Attempting to reload shows a plain 500 because the node does not exist in the user workspace anymore as the workspace was switched. People have to manually fix the url.

```
Neos_Neos_Ui_Controller_BackendController.php: Could not resolve document node for node address {"contentRepositoryId":"default","workspaceName":"marc-henry-schultz-1","dimensionSpacePoint":{"language":"en"},"aggregateId":"44328ad4-6f5a-4c54-ade4-3302887ee249"}
```

PR #4147, opened 2026-07-01
BUGFIX: redirect node to parent if not exists during base workspace change

mhsdesign:
Solves: https://github.com/neos/neos-ui/issues/4146 
Refixes https://github.com/neos/neos-ui/pull/3734 for Neos 9.0

If we find out that the current node does not exist in the new base workspace, we open a subgraph to the old base workspace - which must have the same content as the previously seen by the user as no pending changes are allowed during change - and there find all ancestor nodes. Now we check one by one, starting with the closest to find the next highest common node.

There should be as last resort the site node. The edgecase that a new site only exists in one workspace is not handled.

**What I did**

**How I did it**

**How to verify it**

### DIFF
--- a/Classes/Controller/BackendServiceController.php
+++ b/Classes/Controller/BackendServiceController.php
@@ -16,6 +16,7 @@ namespace Neos\Neos\Ui\Controller;
 
 use Neos\ContentRepository\Core\DimensionSpace\DimensionSpacePoint;
 use Neos\ContentRepository\Core\Feature\WorkspaceRebase\Dto\RebaseErrorHandlingStrategy;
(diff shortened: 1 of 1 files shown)


Sebobo (approved):
Thx, verified and works. But I would have left the comment that explains what is being done.

mhsdesign:
Hmm yeah i should have left a test tbh no one reads comments ... i want tesssstsssstsss

Sebobo:
> Hmm yeah i should have left a test tbh no one reads comments ... i want tesssstsssstsss

I read useful comments, I don't read automatically tests that are in other places ;)

mhsdesign:
true but if you comment out the logic you see which tests fail. If none fail youre out of luck.

P.s. added a nice PR description - feel free to commit that if you like<3

PR #4152, opened 2026-07-08
Bugfix: add ability for view only users to interact with preview page and view data in inspector

t-heuser:
This is the successor of #3921. It contains the bugfix and implements all remarks from the comments of the original PR.

**What I did**
Previous Behaviour:
- When a view-only user clicks on a node on the preview page nothing happens.
- The user can only select nodes via the content tree (left sidebar)
- The user can not see any node data in the inspector (right sidebar)

Fixed Behaviour:
- The user can interact with nodes on the preview page (highlighting is visible)
- The user can view - but not edit - node data in the inspector (right sidebar)

**How I did it**
- Backend: gate rendering on read permission instead of edit permission — added a canReadProperties check (based on ReadNodePrivilege) to NodePolicyService, exposed it in the node policy data, and switched AugmentationAspect to use it so content metadata is still rendered for view-only users. 
- Frontend: don't boot inline editors for content the user can't edit — initializePropertyDomNode.js skips editor initialization when the workspace is read-only or the (lazy-loaded, awaited via store subscription) node policy disallows editing, replacing the old approach of blocking the whole guest frame / read-only CKEditor instances, so view-only users can interact with the preview page normally.                                                                                                                                                               
- Inspector: show node data read-only — tab/property visibility now checks canReadProperties instead of canEdit, and InspectorEditorEnvelope disables editors when the node policy has canEdit: false, letting view-only users see values without editing them.

**How to verify it**
- Create a role (e.g. with the Sandstorm.NeosAcl package) that has permission to view but not to edit/create.
- When browsing content the user should be able to click on nodes in the preview page and see their data in the inspector.

**BEFORE**
Screenshot before change
**AFTER**
Screenshot after change

### DIFF
--- a/Classes/Aspects/AugmentationAspect.php
+++ b/Classes/Aspects/AugmentationAspect.php
@@ -22,6 +22,7 @@ use Neos\Fusion\Service\HtmlAugmenter;
 use Neos\Neos\Domain\Service\ContentContext;
 use Neos\Neos\Ui\Domain\Service\UserLocaleService;
 use Neos\Neos\Ui\Fusion\Helper\NodeInfoHelper;
+use Neos\Neos\Ui\Service\NodePolicyService;
 
 /**
  * - Serialize all nodes related to the currently rendered document
@@ -60,6 +61,12 @@ class AugmentationAspect
      * @var ?\Neos\Flow\Mvc\Controller\ControllerContext
      */
     protected $controllerContext = null;
+
+    /**
+     * @Flow\Inject
+     * @var NodePolicyService
+     */
+    protected $nodePolicyService;
 
     /**
      * @Flow\Before("method(Neos\Neos\Fusion\ContentElementWrappingImplementation->evaluate())")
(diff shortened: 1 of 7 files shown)


mhsdesign:
Hi thanks ill need to find some time to retest this. It seems everything is in one commit so i cannot judge by reading whats new ... but maybe thats not a problem:)

t-heuser:
@mhsdesign Almost everything in the JS files changed, the PHP part stayed the same :)

t-heuser:
@mhsdesign Any updates? :)

mhsdesign:
Nope im still occupied with the 9.2 release. But i wonder, the original pr from robert did not look that bad and i did understood it last time i reviewed ^^ i dont understand the transition from his javascript code changes to this fully new approach - im missing that diff^^ - why _this_ is the successor. Here are also some oddities we have never used before like `store.subscribe()` so thats definitely a sign that something is not really _right_ here.

PR #4154, opened 2026-07-08
TASK: Extract Neos.Ui JS Content Repository Models into separate package (deprecate `neos-ui-ts-interface`)

mhsdesign:
followup to https://github.com/neos/neos-ui/pull/4121

- Avoid deprecated `i18nRegistry` and make it possible to statically infer possible translation keys
- Move PropertyScope to own location and use in EditorEnvelope
- Move all Shared Model things into own package `neos-ui-contentrepository-model` to futher cleanup the to be removed `neos-ui-ts-interfaces` package
  - we are not using the package `neos-ui-contentrepository` as it contains heavy side effects and imports to globalConfig loading which does not work with jest during testing.


packages/neos-ui-editors/src/EditorEnvelope/index.js
Sebobo, line 148: Why not adjust these lines too, when you want to get rid of the registry?
I used it because it was already there and I thought it to be smarter to adjust the whole class in one go instead of mixing it.

mhsdesign: hi thanks for thinking about this:) My goal was/is to adjust all `translate()` calls with a static label which we know exists and works.

We still use the legacy `<I18n>` (renders also a `<span>` and the registry) for the "feature" that we first detect if this could even be a well formatted translation address and else render the input value as fallback.

If we provide not a translation address to the new utility it dies. So we need still a second new utility like `tryTranslate()` which is allowed to be feeded with anything again.

Sebobo: Didn't we discuss already a few times that a fallback option would be helpful, especially when using the translate in plugins?

mhsdesign: yes there is a fallback then i put it wrong: It just dies when you try `translate("not a translation id")` and does not return a fallback or anything because it throws because its not a translation id ... thus `tryTranslate()` which can be used for softer things.

Is also validated by static analysis that `translate()` only allows string literals and not any `string`

mhsdesign:
Can be merged now:)

packages/neos-ui-editors/src/EditorEnvelope/index.js
Sebobo: Not really important for this change, but as a plugin developer I would then always use `tryTranslate` just to be sure. Because depending on the Flow language and fallback configuration it might still fail even though I provided the translations in my plugin. I just had that issue in a customer project.

packages/neos-ui-contentrepository-model/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@neos-project/neos-ui-contentrepository-model",
```
Sebobo, line 2: Can we expose this for plugin development then?
I think the old one was never properly exposed and I had to always copy & paste some times I needed.

Sebobo:
The name of this change should be changed now, this is so much more.

packages/neos-ui-contentrepository-model/package.json
mhsdesign: yes, should (TM) be as simple as removing `private: true`, and then its published. Will do and lets see if it works next beta.

PR #4158, opened 2026-07-10
FEATURE: New rewritten base workspace selector (without select box component)

mhsdesign:
Followup to #4113 

Requires https://github.com/neos/neos-ui/pull/4165

## Before

lost space even if label is small

image

and flickering when opening the dropdown because select box is the full item. Additionally we lose the information which is the current workspace which should always be present:

image

## After

new green bar to indicate no changes

image

flyout to left without any flickering - still shows current base workspace for reference

image

orange indicator and disabled for changes - also trims label

image


Sebobo (dismissed):
As discussed:

- [x] Make all Dropdown Buttons Blue on Hover
- [x] Add search box again to workspace dropdown if more than 10 items
- [x] Keep publish button light green if there are no changes
- [x] Keep text and icon light in workspace dropdown when there are changes
- [x] Adjust hover state of workspace and editpreviewmode dropdown items to match dimensiondropdown

mhsdesign:
Added a search <3 ;)

image

Sebobo:
@mhsdesign thx, works great! Do you want to adjust the hover style of the dimension and editpreviewmode dropdowns to match the workspace dropdown like we discussed?
Also German translation is missing.

mhsdesign:
> Do you want to adjust the hover style of the dimension and editpreviewmode dropdowns to match the workspace dropdown like we discussed?

yes for the edit preview mode selector but the dimension dropdown warrants for a a bigger refactoring in an own PR see Issue tracking that: https://github.com/neos/neos-ui/issues/4159

> Also German translation is missing.

checking which ...

packages/neos-ui-shared-components/src/Util/popoverAutoCloseFromIframe.ts
```diff
@@ -0,0 +1,41 @@
+const closeAllOpenPopovers = () => {
+    // FIXME possibly respect popover="manual" and keep it open?
```
Sebobo: I feel this FIXME needs to be solved before we can merge this.

packages/neos-ui/src/Containers/ContentCanvas/index.js
```diff
@@ -164,6 +165,8 @@ export default class ContentCanvas extends PureComponent {
         this.skipNextLoaderStatusUpdate = true;
         iframeDocument.__isInitialized = true;
         stopLoading();
+        // FIXME dispose ALL added event-listeners on the ContentCanvas, see also below.
```
Sebobo, line 168: Why is this a FIXME?

packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/index.js
```diff
@@ -90,7 +90,7 @@ export default class PublishDropDown extends PureComponent {
         const publishableNodesCount = publishableNodes ? publishableNodes.length : 0;
         return (
             <div id="neos-PublishDropDown" className={style.wrapper}>
-                <AbstractButton
+                <PublishButton
                     id="neos-PublishDropDown-Publish"
                     className={style.publishBtn}
```
Sebobo: Shouldn't this class be moved to the PublishButton component?

packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/index.js
Sebobo: No clue why I called it MainButton back then, when it only returns the label? We should change this.

packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/index.js
```diff
@@ -99,7 +99,7 @@ export default class PublishDropDown extends PureComponent {
                     >
                     {mainButton} {isWorkspaceReadOnly ? (<Icon icon="lock"/>) : ''}
```
Sebobo: How about moving a bit more of this logic to the component. Seems to be specific to it.

packages/neos-ui/src/Containers/PrimaryToolbar/WorkspaceSelector/index.tsx
```diff
@@ -40,82 +42,127 @@ type WorkspaceSelectorProps = {
...
+        enabled={changingWorkspaceAllowed}
+        buttonTitle={title}
+        buttonIcon={<Icon icon="layer-group" padded="right" />}
+        buttonLabel={baseWorkspaceTitle}
+        buttonClassName={dropDownButtonStyles}
+        dropDownClassName={style.dropDownContents}
+        dropDownIconClassName={style.dropDownIcon}
+    >
+        {Object.keys(allowedWorkspaces).length >= 10 ? (
+            <div className={style.searchContainer}>
```
Sebobo, line 107: I dislike that the filter and group label look the same (same bg color), should we try a lighter one for the filter?

Sebobo (changes requested):
Looks and works so much better, I feel the fixme needs to be resolved and some smaller cleanups of weirdness. Thank you!

packages/neos-ui/src/Containers/ContentCanvas/index.js
mhsdesign: because so far we dont care about our memory bloating up possibly or not - no one investigated heap size when switching trough many pages. Fact is the iframe currently gets already a few event listeners attached, and with this pr 2 more which are not cleaned up when the iframe is changed but i dont know if they are auto cleaned up or not

packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/index.js
mhsdesign: i think this comment does not make sense for someone not reading it with 30 degrees outside and just 24 - what do you mean? :D

packages/neos-ui/src/Containers/PrimaryToolbar/WorkspaceSelector/index.tsx
mhsdesign: i knew you would not like it, thats why i added the border already ... yes maybe?

packages/neos-ui-shared-components/src/Util/popoverAutoCloseFromIframe.ts
mhsdesign: have an idea how to fix it properly, first i wanted to use a negative query selector but using `element.popover` via javascript api is the way as it already normalises values to the allowed ones.

Done:) a6d335170bd8cbae20b503e4b04fc842c10a69d7

packages/neos-ui/src/Containers/PrimaryToolbar/PublishDropDown/index.js
mhsdesign: maybe another day this is already getting out of hand:) Also nothing wrong with using composition and i like that. Less complexity in simple components

mhsdesign:
I went to fix all hover things but that got too complex and is now part of this pr https://github.com/neos/neos-ui/pull/4165 when its merged i can rebase and the duplicate commits go away:)

Also i noticed that we moved some translations in #4113 like `published` and `publish` from `Neos.Neos` to this ui but only manually copied en and de. That sucks for the people having done the translations already in all other languages so i added them via my cool CLI command too :)

packages/neos-ui/src/Containers/PrimaryToolbar/WorkspaceSelector/index.tsx
mhsdesign: image

done:)

Sebobo:
Looks great now, one issue left with many items that are being cut off:

image

I think we need to wrap the 3 groups in an auto-scroll container with a max-height.

mhsdesign:
Thanks for your throughout review, indeed i meant to test the case with many entries but forgot:)

Is fixed as per last commit and everything is rebased and ready for merge.

Sebobo (approved):
Works on my machine ™️

Issue #4153, opened 2026-07-08
Additional UX cleanups after new workspace selector


mhsdesign:
https://github.com/neos/neos-ui/pull/4113 implemented a new workspace selector next to the publish dropdown.

That was extensively discussed years ago also here on github: https://github.com/neos/neos-ui/issues/3247

While this issue can be closed now i like to take the remaining suggestions here to refresh the discussion with the new state in mind:

### Inconsistencies discussed in 2023 (Daniel)

Its funny to see so many things of @danielkestler suggestions fixed now, let me quote:

> 1. Edit mode selector (direct edit / raw view etc.) is placed in the primary toolbar. Why? It has no global meaning, it belongs to the actual page preview / rendering (the guest iframe).
> 2. User account selector is in the primary toolbar. That's maybe okay, one could discuss about this. Could also be moved to the toggle menu in the burger menu.
> 3. What has a global meaning then, you may ask? IMO two things:
> 
> * Currently selected site (in a multisite installation), probably with site switcher
> * Currently selected workspace / branch (maybe even more relevant if the ESCR is ready)
> * Currently selected node variant / language
> 
> 4. Panes need a lot of space. Basically, you have:
> 
> * Document Tree
> * Content Tree / Outliner
> * Selected node properties
>   Do we need them all at the same time?
> 
> 5. The toolbar directly above the guest iframe is a bit weird. There are contextual buttons there (formatting of text) which could be inlined to the actual inline editor, and preview settings like fullscreen, open preview link and the keyboard shortcuts (???).
> 
> Maybe we should rethink the interface and make it more consistent, e.g.
> 
> * Primary toolbar to only contain application (or at least "view-wide") buttons and settings
> * Secondary toolbar to contain settings at the document level (like fullscreen, preview link, edit modes, ... selected node?)
> * Floating toolbars that appear in the right context (e.g. formatting buttons directly above selected text)

The user menu (2) Is moved to drawer #3740 
The workspace selector (3) is now in the global space #4113
The CKEditor toolbar (5) is now inline #4016 

So what is left and would make sense to address:

1. Edit mode selector (direct edit / raw view etc.) is placed in the primary toolbar. (see 1 above)
   - Especially as with https://github.com/neos/neos-ui/pull/4104 the label is even more verbose and does no longer work on small screens
2. Show and allow to switch currently selected site (see 3 above) 
3. I still have found no one using the keyboard shortcuts (see 5 above), but they are now part of the primary toolbar,
[...]

PR #4104, opened 2026-03-26
FEATURE: Preview mode (readonly) indicator for Neos UI

mberhorst:
Resolves: #4105 

- Editing Mode Toggle now has a blue indicator when Preview Mode is active
- Also the iframe with the content gets a blue border with a small label at the top telling you that you cannot edit content

CleanShot 2026-03-31 at 11 35 44@2x

PR #4160, opened 2026-07-13
FEATURE: Move EditPreviewDropdown to secondary toolbar

Sebobo:
Overhauls: https://github.com/neos/neos-ui/pull/4104
Related: https://github.com/neos/neos-ui/issues/4153

This way we can use the dropdown directly to show the current mode instead of duplicating it with the PreviewBadge.
Additional advantages: we use the secondary toolbar better and we have much more space in the primary toolbar on small screens.

Edit Mode:

CleanShot 2026-07-13 at 10 47 30@2x

Preview Mode:

CleanShot 2026-07-13 at 10 47 23@2x

Smaller screen:

CleanShot 2026-07-13 at 10 50 55@2x


mhsdesign (approved):
Absolutely peak perfection <3

mhsdesign:
The only cosmetic unpretty part i see is that the label when the button is blue is not aligned in the absolute middle considering the blue 2px border. But rather centered within the button and thus a little higher than centered.

Though that is beyond this PR as it was before also build with the readonly badge and i have no idea how we want to style it otherwise.

mhsdesign:
Ah i see you added a hack to cover the `1px` border. I think thats not necessary without that hack we would not have the problem i described above and also when the preview mode is non-desktop we are decoupled either way and only partially or not at all have blue covering everything:


image

mhsdesign:
Fixed it by removing the global `41px` height for the row and just using `40px` but additionally adding an `1px` top border to the iframe in grey

image

or when readonly that changes to a 5px blue border and replaces the 1px border:

image


opened:

image

mhsdesign:
Adjusting the grid even _fixes_ an existing "issue" no one noticed. The border is set to high in comparison with the document tree:

before

image


after

image

mhsdesign:
btw why do we put it right most and not to the left of the two small buttons?

image

because of muscle memory i had a hard time finding it really fast and it took me 500ms longer - also the position is not absolutely fixed but depending on the width of the edit preview label and thus my brain does not know where to click

PR #4162, opened 2026-07-13
TASK: simplify page tree search and reuse components

mhsdesign:
The new package `@neos-project/neos-ui-shared-components` is meant for

> Internally shared components of the Neos UI, allowed to access the translate() function - not exposed like react-ui-components

We rarely add components to the `react-ui-components` package as they are _more_ public api. But instead we often have components spread around and often not even shared - like with the `Search`

Ill need the `Search` for a separate change but thus prepared here the extraction.

The spacing of the icon in the `Search` component from the link editor #3996 is a little different but it looks good:

image

**What I did**

**How I did it**

**How to verify it**

TODO:
- [x] check if e2e tests still work (at least locally)


Sebobo (approved):
Looks good by reading

PR #4164, opened 2026-07-14
BUGFIX: Only expand selectbox on filter click

Sebobo:
Resolves the regression in #4139


mhsdesign (approved):
Tested and rebased to 9.0

PR #4166, opened 2026-07-14
WIP FEATURE: Sticky formatting toolbar for long editable areas

Sebobo:
**What I did**

Let the CKEditor toolbar become sticky in long editables when scrolling, similar to the NodeToolbar.
This change was made with some agentic help as the positioning is quite complex and has many edge cases.

**How I did it**

Todos:

- [ ] Code still complex due to workarounds for browser errors with anchors and sticky positioning
- [ ] Test with various browsers
- [ ] Test with elements in various positions
- [ ] Test with elements with multiple (styled) editables

**How to verify it**

https://github.com/user-attachments/assets/d098ca40-d439-4050-a511-59f85039353e

### DIFF
--- a/packages/neos-ui-ckeditor5-bindings/src/ckEditorApi.js
+++ b/packages/neos-ui-ckeditor5-bindings/src/ckEditorApi.js
@@ -36,6 +36,7 @@ let currentEditor = null;
 let currentPropertyDomNode = null;
 let editorConfig = {};
 let resizeObserver = null;
+let currentScrollTarget = null;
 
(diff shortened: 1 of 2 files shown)


Sebobo:
This change was initially not so complex, but after some scrolling chrome lost the connection between the visual and the interactive position of the toolbar and the interaction became madness. This is why the position in the sticky case is now calculated via JS instead of CSS. Maybe the browsers fix this until we release 9.3 and we could simplify it again? 
We cannot ignore the position fallbacks as they are important for a toolbar of unknown size and features which was already hard to solve for the top/bottom position.

Issue #3100, opened 2022-04-09
Umbrella issue: improve shortcuts/hotkeys/ux behaviour


mhsdesign:
### Description

i really like shortcodes/keyboard navigation in programs -> as they make you fast. way faster than always using the mouse.
and i think neos cms editors should also profit from a good keyboard ux to improve their workflow.

see https://github.com/neos/neos-ui/issues/3099
see https://github.com/neos/neos-ui/issues/3032
see https://github.com/neos/neos-ui/issues/3033
see https://github.com/neos/neos-ui/issues/3030
see https://github.com/neos/neos-ui/issues/3031

i would like to see the use of more conventional shorcuts.
is there a reason we dont use a `ctrl + key` combination like a 'normal' desktop app? (as far as i know its pretty easy to determine if `ctrl` was pressed ... -> its in the event)

i found these shortcuts quite odd to remember as one is used to: `simpleKeys` or `modifier (alt, ctrl) and simpleKey`
![image](https://user-images.githubusercontent.com/85400359/162564730-a0e42b56-e6f6-4f96-b176-7d86d7776d71.png)

PR #4167, opened 2026-07-14
TASK: deprecate and disable not use-full and broken keyboard shortcuts

mhsdesign:
Related https://github.com/neos/neos-ui/issues/4153
Related https://github.com/neos/neos-ui/issues/3100

We have discussed thousand times that our keyboard shortcuts are too absurd to be useful and that we want to free the space in the ui and reduce complexity in the code for a feature which doesnt really work.

Notice the missing button? Me neither
 
image

`Neos.Neos.Ui.frontendConfiguration.enableLegacyKeyboardShortcuts` allows to re-enable it in case its really used somewhere and might then lead us to not removing it in 9.3

### DIFF
--- a/Configuration/Settings.yaml
+++ b/Configuration/Settings.yaml
@@ -84,7 +84,8 @@ Neos:
       # const {someKey} = getFrontendConfigurationForPackage('Your.Own:Package');
       #
       frontendConfiguration:
-
+        # keyboard shortcuts will be removed in the next version of neos
+        enableLegacyKeyboardShortcuts: false
         # we use the frontendConfiguration for hotkeys here to preserve compatibility. For new settings the neos ui should use the ConfigurationProvider instead.
         hotkeys:
           'UI.RightSideBar.toggle': 't i'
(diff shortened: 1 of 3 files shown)


Sebobo:
I think there a valid reasons to keep some of the shortcut functionality as I also have plugins that register useful shortcuts and it would be awesome if Neos still provides the API to handle them properly from host or guest frame. For example my AddNodeButton, the Commandbar, Terminal, stuff like that.
But I think almost everyone dislikes the random shortcuts solely based on letters.

mhsdesign:
Hmm external plugins huhm maybe we should provide an extension point for these things indeed but it will look fully different (and work next time) i dont think we should continue to provide the current extension point for plugins. The whole things needs new and plugins could at first attach their own event listeners.

Sebobo:
IMO we can only move the button somewhere in 9.2 so it doesn't bother so much anymore, but not throw everything away.
For 9.3/10 I would like to have a RFC with the team.

mhsdesign:
i think a feature flag for 9.2 is also ok ... thats why i proposed this ... we talked about fully removing it already but yes thats definitely too much for 9.2 now

PR #4169, opened 2026-07-22
TASK: Ensure preview and fullscreen buttons are rightmost (move new edit preview mode left)

mhsdesign:
Followup to https://github.com/neos/neos-ui/pull/4160

image

Because of muscle memory it took a bit to find the old preview button - also the position is not absolutely fixed but depending on the width of the edit preview label which also makes it harder to fully understand the layout.

Dektop (previously in corner)

image

Tablet both corners never match

image


Sebobo (approved):
Impressive, thx

mhsdesign:
Witzbold :'D

PR #4170, opened 2026-07-27
TASK: PSR12 & PHP-CS-Fixer

mhsdesign:
Parent meta issue: https://github.com/neos/neos-development-collection/issues/5929

Adjustments are split into two commits to ease review, first repetitive and not worth reading every part

- b0dd9ff177 Optimise imports and correct spacing at begin of PHP files

Second the remaining bits.

- 2c78aaa168 Enforce remaining PSR12

### DIFF
--- a/.github/workflows/server.yml
+++ b/.github/workflows/server.yml
@@ -24,10 +24,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - php-version: '8.2'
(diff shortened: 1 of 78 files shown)


mhsdesign:
As this PR follows exactly the same patterns as Martins changes in Neos and Flow ill just have it merged.

Issue #4079, opened 2026-02-24
Feature Request: Expose User Roles in Inspector ClientEval (Neos 9)


grizooo:
In Neos 9 it is currently not possible to conditionally show or hide Inspector properties based on the roles of the authenticated backend user.

Expressions like:

hidden: 'ClientEval: !Security.hasRole("Neos.Neos:Administrator")'

or

hidden: 'ClientEval: !configuration.currentUser.roles.includes("Neos.Neos:Administrator")'

do not work because the ClientEval context does not expose user roles or a Security helper.

This makes role-based UI logic in NodeTypes impossible without custom Inspector components.

**Proposal:** Expose current user roles (or provide a `Security.hasRole()` helper) in the ClientEval context to enable simple role-based UI conditions.

Sebobo:
We don't really want security defined in the JS level, as we also already have an existing permission check in the backend via `\Neos\Neos\Security\Authorization\ContentRepositoryAuthorizationService::getNodePermissions` which provides the necessary information to the Neos.Ui and mostly worked in Neos <9 but has some missing features in Neos 9, as the security feature implementation project (https://neos-project.slack.com/docs/T04HF7TKV/F09UXK41H7H) is still WIP.

If you need it asap, you could probably create a patch for the method above, to make another privilege check for a custom NodeType privilege and check that with `isPrivilegeTargetGranted` with the nodeType of the node as parameter, and the contentrepository. That allows you to check the nodetype with the nodeTypeManager.

mhsdesign:
Thank you for your feature request idea.

> We don't really want security defined in the JS level,

This is the Mandelorian way:) And as seb outlined we are working to bringing security in the Neos Ui back to old glory so i think we can safely close this feature request if this is okay for you?

PR #4175, opened 2026-08-22
BUGFIX: Do not create inline editors for nodes without edit permission

### LINKED ISSUE #4171: BUG: Neos UI does not reflect node edit permissions (inline editors stay editable)
### Description
When a node is protected by an `EditNodePrivilege` (Neos 9 subtree tags), the backend UI still
initializes an inline editor for its properties. Editors can type, and the change is only
rejected when it is persisted, with a raw `AccessDenied` error. The inspector, on the other
hand, is rendered completely empty instead of read-only.

This is a known consequence of the CR privileges feature — it is documented in the note of
https://github.com/neos/neos-development-collection/pull/5298 — but I could not find a
dedicated issue for it. The closest one is #4110, which covers only the "Create new" button.

### Steps to reproduce
1. Tag a subtree: `TagSubtree::create(...)` (or the CLI proposed in neos/neos-development-collection#5931)
2. Grant a role `EditNodePrivilege` for that tag only, plus `ReadNodePrivilege` for the site
3. Log in with that role and open a document outside the tagged subtree
4. Click into any inline editable text → the editor accepts input
5. Save → `Command "…SetNodeProperties" was denied: No edit permissions for node "…"`

Reproduced on the Neos demo site with a two-target policy (site-wide `demosite` tag +
`blog` tag), an `EditNodePrivilege` granting only the `blog` tag, and a role with
`Neos.Neos:LivePublisher` (without it the personal workspace is reported read-only and
the UI disables editing altogether, hiding the node-level problem behind the workspace-level
one). Measured property-by-property against what `get-additional-node-metadata` reports:

| page | inline editors initialized | server's `canEdit` |
|---|---|---|
| page outside the tagged subtree | 16 of 16 | all 16 `false` |
| page inside the tagged subtree | 6 of 6 | 5× `true`, 1× `false` (unrelated node) |

i.e. on 9.1.6 as released, every property gets an inline editor regardless of the policy.

### Expected behaviour
Content the user may not edit is not offered for editing in the first place.

### Analysis
`getAdditionalNodeMetadataAction()` already returns the correct `policy.canEdit` (computed by
`ContentRepositoryAuthorizationService::getNodePermissions()`), so the information is available.
Two things prevent it from having an effect:

1. `initializePropertyDomNode.js` gates the inline editor only on `ui.inlineEditable` and
   `policy.disallowedProperties` — and the latter is hardcoded to `[]` server side
   (`// not implemented for Neos 9.0`). `policy.canEdit` is never consulted.
2. The policies are loaded **lazily** by the `CR/Policies` saga after the node data has been
   merged, while `initializeGuestFrame` starts initializing the content DOM nodes immediately.
   `initializeContentDomNode` reads the store at call time, so whether the policy is there is a
   race — which makes any gating on its own flaky. (Measured: a gate-only fix without also
   fixing the load order initialized 1, 5, then 1 editors over three identical reloads on the
   same page.)

Additionally the inspector uses `canEdit` as a *visibility* predicate
(`Inspector/index.js`, `TabPanel/index.js`) rather than disabling the editors, which is why it
ends up empty.

### Questions before I open a PR
Since neos/neos-development-collection#5650 lists "remove `ReadNodePrivilege` etc." as an open
point, is a UX fix against the current privilege API welcome now, or would you rather have this
wait for the security concept? I have a fix for (1) + (2) ready (no revival of the removed
`NodePolicyService`, no new endpoint — the policies are simply loaded together with the node
data) and would prepare the inspector part separately.

Happy to also contribute the repro as a Playwright E2E case if that helps.

ru3fu5z:
Resolves #4171

**What I did**

Inline editors were created for every inline editable property, regardless of whether the user is allowed to edit the node. Editors could type into content protected by an `EditNodePrivilege`, and the change was only rejected once it was persisted — with a raw `AccessDenied` exception from the Content Repository. This is the behaviour described in the note of neos/neos-development-collection#5298.

The node policy already carries the required information (`policy.canEdit`, computed server side by `ContentRepositoryAuthorizationService::getNodePermissions()`), but it was never consulted when creating the inline editors — and it was not even available at that point in time.

**How I did it**

Two changes, both in `neos-ui-guest-frame`:

1. `initializePropertyDomNode.js` now respects `policy.canEdit`. The decision moved into a small pure function `isPropertyInlineEditable()` so it can be unit tested. A missing `canEdit` is deliberately treated as "editable", so nothing is locked down if the policy is unavailable.

2. `initializeGuestFrame.js` loads the node policies **together with** the node data (`all([call(loadNodeData, …), call(loadNodePolicies, …)])`) and merges them in a single store update. Previously the policies were fetched lazily by the `CR/Policies` saga *after* the node data had been merged, while the content DOM nodes are initialized immediately afterwards — and `initializeContentDomNode` reads the store at call time. Whether the policy was there was therefore a race.

The second part is what makes the first one reliable. Gating on `canEdit` alone is flaky: measured on the same page over three identical reloads, a gate-only build initialized **1, 5 and 1** editors — i.e. forbidden editors leak in depending on when the policy request returns. With the policies loaded up front the result is deterministic.

Because both requests now run concurrently, this costs no extra latency: in the network timeline the guest frame's `flow-query` and `get-additional-node-metadata` start simultaneously. The merged policies also keep the `CR/Policies` saga from requesting them a second time (its filter skips nodes that already have a policy).

No security logic was added on the client side — the decision stays entirely server side, as requested in #4079. The removed `NodePolicyService` is *not* revived and no new endpoint is introduced.

Regarding the ongoing security concept work: I asked in neos/neos-development-collection#5650 (comment) whether a UX fix against the current privilege API is welcome now; no objection in two weeks, so here it is — happy to rework if the concept lands on a different model.

**How to verify it**

Reproduction on the Neos demo site (a small command controller is needed to apply the subtree tag, since there is no CLI for it yet — see neos/neos-development-collection#5931):

1. `./flow site:importAll --package-key Neos.Demo`
2. Tag a subtree, e.g. the "Blog" document, via `TagSubtree::create(...)`
3. `Policy.yaml` — note the **site-wide** target: for nodes that no privilege target matches at all, Flow defaults to "granted", so without it the restricted role could edit everything and nothing would reproduce. Note also `LivePublisher`: without it the personal workspace is reported read-only and the UI disables editing altogether, which hides the node-level problem behind the workspace-level one.
   ```yaml
   privilegeTargets:
     'Neos\Neos\Security\Authorization\Privilege\EditNodePrivilege':
       'Neos.Demo:EditAllNodes':
         matcher: 'demosite'
       'Neos.Demo:EditBlogNodes':
         matcher: 'blog'
   roles:
     'Neos.Neos:Editor':
       privileges:
         - privilegeTarget: 'Neos.Demo:EditAllNodes'
           permission: GRANT
         - privilegeTarget: 'Neos.Demo:EditBlogNodes'
           permission: GRANT
     'Neos.Demo:BlogEditor':
       parentRoles: ['Neos.Neos:AbstractEditor', 'Neos.Neos:LivePublisher']
       privileges:
         - privilegeTarget: 'Neos.Demo:EditBlogNodes'
           permission: GRANT
   ```
4. Log in with a user that only has `Neos.Demo:BlogEditor` and open a document outside the tagged subtree
5. **Before:** clicking into any text starts an inline editor; saving fails with `Command "…SetNodeProperties" was denied: No edit permissions for node "…"`
   **After:** the text is not editable; content inside the tagged subtree still is

Measured on the demo site with the setup above, comparing every property against what `get-additional-node-metadata` reports for its node:

| page | 9.1.6 as released | with this PR | server verdict | mismatches |
|---|---|---|---|---|
| Features (not permitted) | 16 of 16 editors | **0 of 16** | all 16 `canEdit: false` | 0 |
| Blog posting (permitted) | 6 of 6 editors | **5 of 6** | 5× `true`, 1× `false` (element outside the blog subtree) | 0 |

And in a real 8.3→9.1 customer migration (74 inline editable properties on one page, editor role scoped to a single tagged element):

| Build | inline editors initialized |
|---|---|
| 9.1.6 as released | 54 of 74 (53 of them forbidden for that role) |
| gate only, policies still lazy | 1 / 5 / 1 over three identical reloads (flaky) |
| this PR | 1 / 1 / 1 — exactly the one node the role may edit |

Before/after GIFs recorded on the demo site follow in a comment.

Side note, unrelated to this PR: the deny message names the wrong privilege type — it says `No privilege of type "…\ReadNodePrivilege" matched` while evaluating an **Edit** target (`"Neos.Demo:EditAllNodes": ABSTAIN`). Happy to file that separately.

Tests: 7 new Jest cases in `initializePropertyDomNode.spec.js`; full suite green (151 suites / 826 tests / 52 snapshots), ESLint clean, production build succeeds.

**Known limitations (deliberately out of scope)**

- The inspector still uses `canEdit` as a *visibility* predicate (`Inspector/index.js`, `TabPanel/index.js`), so it renders empty instead of read-only for such nodes. That is a separate change and overlaps with the inspector part of #4152 — happy to prepare it once the direction here is clear.
- `policy.disallowedProperties` is still hardcoded to `[]` server side (`// not implemented for Neos 9.0`), so per-property restrictions remain unavailable. That belongs to the security concept in neos/neos-development-collection#5650.

### DIFF
--- a/packages/neos-ui-guest-frame/src/initializeGuestFrame.js
+++ b/packages/neos-ui-guest-frame/src/initializeGuestFrame.js
@@ -1,4 +1,4 @@
-import {put, select, takeEvery} from 'redux-saga/effects';
+import {all, call, put, select, takeEvery} from 'redux-saga/effects';
 
 import {actions, actionTypes, selectors} from '@neos-project/neos-ui-redux-store';
 import {requestIdleCallback} from '@neos-project/utils-helpers';
@@ -40,6 +40,18 @@ const eventPath = event => {
     return path;
 };
 
+//
+// Load the node data for the given nodes.
+//
+const loadNodeData = (q, nodeContextPaths) =>
+    nodeContextPaths.length > 0 ? q(nodeContextPaths).get() : [];
+
+//
+// Load the policies for the given nodes, telling us what the user is allowed to do with them.
+//
+const loadNodePolicies = (endpoints, nodeContextPaths) =>
+    nodeContextPaths.length > 0 ? endpoints.getAdditionalNodeMetadata(nodeContextPaths) : {};
+
 export default ({globalRegistry, store}) => function * initializeGuestFrame() {
     const nodeTypesRegistry = globalRegistry.get('@neos-project/neos-ui-contentrepository');
     const inlineEditorRegistry = globalRegistry.get('inlineEditors');
@@ -62,22 +74,36 @@ export default ({globalRegistry, store}) => function * initializeGuestFrame() {
     }
 
     // Load all nodedata for nodes in the guest frame and filter duplicates
-    const {q} = yield backend.get();
+    const {q, endpoints} = yield backend.get();
     const nodeContextPathsInGuestFrame = findAllNodesInGuestFrame().map(node => node.getAttribute('data-__neos-node-contextpath'));
 
     // Filter nodes that are already present in the redux store and duplicates
     const nodesByContextPath = store.getState().cr.nodes.byContextPath;
-    const notFullyLoadedNodeContextPaths = [...new Set(nodeContextPathsInGuestFrame)].filter((contextPath) => {
+    const uniqueNodeContextPathsInGuestFrame = [...new Set(nodeContextPathsInGuestFrame)];
+    const notFullyLoadedNodeContextPaths = uniqueNodeContextPathsInGuestFrame.filter((contextPath) => {
         const node = nodesByContextPath[contextPath];
         const nodeIsLoaded = node !== undefined && node.isFullyLoaded;
         return !nodeIsLoaded;
     });
 
-    // Load remaining list of not fully loaded nodes from the backend if there are any
-    const fullyLoadedNodesFromContent = notFullyLoadedNodeContextPaths.length > 0 ? (yield q(notFullyLoadedNodeContextPaths).get()).reduce((nodes, node) => {
+    // The node policies tell us whether the user may edit a node at all. They are usually fetched
+    // lazily (see the CR/Policies saga), which is too late for us: The inline editors are created
+    // right after this saga merged the node data, so we would create editors before knowing whether
+    // the user is allowed to change the content. We therefore load them together with the node data.
+    const nodeContextPathsWithoutPolicy = uniqueNodeContextPathsInGuestFrame.filter(
+        contextPath => !nodesByContextPath[contextPath]?.policy
+    );
+
+    // Load remaining list of not fully loaded nodes and the missing policies from the backend if there are any
+    const [nodesFromContent, additionalNodeMetadata] = yield all([
+        call(loadNodeData, q, notFullyLoadedNodeContextPaths),
+        call(loadNodePolicies, endpoints, nodeContextPathsWithoutPolicy)
+    ]);
+
+    const fullyLoadedNodesFromContent = nodesFromContent.reduce((nodes, node) => {
         nodes[node.contextPath] = node;
         return nodes;
-    }, {}) : {};
+    }, {});
 
     const nodes = Object.assign(
         {},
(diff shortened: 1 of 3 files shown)


ru3fu5z:
neos-ui-edit-permission-before
neos-ui-edit-permission-after

ru3fu5z:
Friendly ping — ready for review: CI green, before/after screencasts above, running in production on our Neos 9.1 sites since August. Given the recent #neos-general discussion about permissions being a 9.x migration blocker: this PR closes one concrete gap in the classic UI (inline editors ignoring `EditNodePrivilege`).

PR #4176, opened 2026-08-25
TASK: Replace removed and deprecated PhpUnit mock APIs in test

kdambekalns:
getMockForAbstractClass() is removed in PHPUnit 12 without replacement; for a plain interface createMock() is the direct equivalent.

The same file still used MockBuilder::setMethods(), which was already removed in PHPUnit 10 - it stayed unnoticed because every test in this class is skipped. Renamed to onlyMethods() so the file does not fatal the day these tests are revived.

### DIFF
--- a/Tests/Functional/Domain/Service/NodePropertyConverterServiceTest.php
+++ b/Tests/Functional/Domain/Service/NodePropertyConverterServiceTest.php
@@ -27,7 +27,7 @@ class NodePropertyConverterServiceTest extends FunctionalTestCase
 
         $nodeType = $this
             ->getMockBuilder(NodeType::class)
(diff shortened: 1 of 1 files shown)


mhsdesign (approved):

mhsdesign:
We are targeting with this batch of changes 9.2 deliberately right because we want to make it PHPUnit 12 compatible ... or Neos 9.3 but definitely not Neos 9.1 or 9.0 right?

Makes fully sense, can be merged directly if you like one approval for this minor change is enough for this in Ui land;)

kdambekalns:
> We are targeting […] 9.2 deliberately […] or Neos 9.3 but definitely not Neos 9.1 or 9.0 right?

Exactly.

Issue #3923, opened 2025-03-04
BUG: Publish all button (site) shows wrong count


mhsdesign:
The publish all / Discard all buttons in the publish dropdown calculate changes incorrectly.

For example in a multisite having one change in "site-two" but no change int "site-one", will still show one change:

<img width="315" alt="Image" src="https://github.com/user-attachments/assets/238108f4-d24b-4793-b5fd-79f70e7888e5" />

This in-turn causes publication to fail, because here we gather the data correctly and find nothing to publish:

<img width="693" alt="Image" src="https://github.com/user-attachments/assets/3342489e-dcb3-4e80-a691-ea869a8fc8b3" />

At cause is the Neos ui `Neos\Neos\Ui\ContentRepository\Service\WorkspaceService` which attempts to replicate the  neos ui publish logic of `WorkspacePublishingService` beforehand.

An expensive fix would be to check while iterating over each change that it was made in the current closest site node.

This is one of the many flaws of the change projection and will hopefully once be fixed correctly.

PR #4178, opened 2026-09-10
BUGFIX: Scope pending changes indicator to the current site

### LINKED ISSUE #4151: BUG: In a multisite-setup (single CR): Publish indicator shows changes of different site which throws exception when published
I have a multisite setup (single CR). When changing content on site A and switching to site B, I see an orange publish indicator. When clicking "publish all" I get an exception:

```
ServerSideError: InvalidArgumentException
#1737448717: The command "PublishIndividualNodesFromWorkspace" for workspace [user] must contain nodes to publish
```

Tested with: neos/neos and neos-ui 9.1.3.
Reproduced by a colleague in another project who contacted me if I had seen this before, so it does not seem to be an issue of my specific setup.

## Steps to reproduce
1. Change content on site A. Results in a regular publish-button

Image

2. Switch to site B. Results in a publish button that shows "something changed on another document node"

Image

3. Use the dropdown arrow of the publish button and press "publish all". Results in a popup "Changes could not be published" with the exception:

```
ServerSideError: InvalidArgumentException
#1737448717: The command "PublishIndividualNodesFromWorkspace" for workspace [user] must contain nodes to publish
```

4. Change something on site B (with the dirty changeset of the other site A still there) and click on dropdown -> "publish all". Result: When a change is made on site B the problem is masked. Results in a working publishing process

Image

After that, however, the "dirty" changeset still hangs around:

Image

robinroloff:
## What I did

Fixes #4151 and #3923.

In a multi-site setup with a single content repository, `WorkspaceService::getPublishableNodeInfo()` returns **all** pending changes of the personal workspace, while "publish all" is scoped to the current site via `WorkspacePublishingService::publishChangesInSite()`.

So editors working in site A see an orange publish indicator for changes that belong to site B, and publishing fails with:

> The command "PublishIndividualNodesFromWorkspace" for workspace `<user>` must contain nodes to publish

`getPublishableNodeInfo()` now takes an optional `SiteNodeName` and skips changes whose closest site node does not match — mirroring the scope that is actually used for publishing. The four places that build the workspace info pass the current site:

| Where | Source of the site |
| --- | --- |
| `Configuration/Settings.yaml` (initial state) | `site.name` from the Fusion context |
| `AbstractChange::updateWorkspaceInfo()` | closest site node of the changed node |
| `BackendServiceController::getWorkspaceInfoAction()` | `SiteDetectionResult` |
| `BackendServiceController` after changing the base workspace | closest site node of the document |

All four matter: if any of them keeps counting workspace-wide, the indicator falls back to the wrong number as soon as that code path runs.

The parameter is optional and defaults to the previous behaviour, so other callers of this `@internal` API are unaffected.

## How to verify it

In a multi-site setup with a single content repository:

1. Change something in site A, do not publish
2. Switch to site B → the indicator no longer shows the change from site A
3. Change something in site B → only that change is counted
4. Publish in site B → succeeds, the change in site A is untouched
5. Switch back to site A → the change shows up again and can be published

Also worth checking that the normal case still works: several documents changed within the *same* site are all counted and published together.

## Notes

As @mhsdesign points out in #3923 and #4151, the change calculation is due for a rewrite (neos/neos-development-collection#5459). This is the pragmatic fix in the meantime — it makes the indicator agree with what publishing actually does, without touching the change projection.

The content repository side is neos/neos-development-collection#5975, which makes publishing an empty changeset a no-op instead of an exception. The two are independent: this PR stops the UI from offering unpublishable changes, that one stops the exception for any other caller.

One trade-off worth flagging: the filter does an additional `findClosestNode()` per pending change. On workspaces with many pending changes this makes building the initial state slightly more expensive.

### DIFF
--- a/Classes/ContentRepository/Service/WorkspaceService.php
+++ b/Classes/ContentRepository/Service/WorkspaceService.php
@@ -12,14 +12,17 @@ namespace Neos\Neos\Ui\ContentRepository\Service;
  * source code.
  */
 
+use Neos\ContentRepository\Core\Projection\ContentGraph\ContentSubgraphInterface;
 use Neos\ContentRepository\Core\Projection\ContentGraph\Filter\FindClosestNodeFilter;
 use Neos\ContentRepository\Core\Projection\ContentGraph\Node;
 use Neos\ContentRepository\Core\Projection\ContentGraph\VisibilityConstraints;
 use Neos\ContentRepository\Core\SharedModel\ContentRepository\ContentRepositoryId;
 use Neos\ContentRepository\Core\SharedModel\Node\NodeAddress;
+use Neos\ContentRepository\Core\SharedModel\Node\NodeAggregateId;
 use Neos\ContentRepository\Core\SharedModel\Workspace\WorkspaceName;
 use Neos\ContentRepositoryRegistry\ContentRepositoryRegistry;
 use Neos\Flow\Annotations as Flow;
+use Neos\Neos\Domain\Model\SiteNodeName;
 use Neos\Neos\Domain\Service\NodeTypeNameFactory;
 use Neos\Neos\Domain\Service\WorkspacePublishingService;
 use Neos\Neos\PendingChangesProjection\Change;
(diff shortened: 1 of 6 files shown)


mhsdesign (changes requested):
hi thank you for caring. We are aware of this bug and its primary cause is that the change projection and change handling is just a flat storage and without doing expensive queries for each entry during call time more or less needless.

As described in https://github.com/neos/neos-development-collection/issues/5459 id rather like to build a proper solution. Currently a workspace with many changes takes already time to answer the question if a node is on the current document which is N+1 and now we add the complexity for the site question as well.

@dlubitz is working on making the queries actually performant but that is for Neos 9.3+ so id say this change should rather target that. What do you think? If we really need this for 9.0 we definitely need to find out how this impacts performance with workspaces with no, small and many changes - especially the UpdateWorkspaceInfor thing.

VolkerLieber:
Hi @mhsdesign, this is currently a breaking bug for our customers, so it would be highly appreciated if this fix could be accepted before the 9.3+ version.

